PR #26, opened 2016-01-26
[wip] Introduce the io.l5d.http namer

olix0r:
This provides an intermediary namer that supports manipulation of names
generated by by the default http identifier.  This namer is intended to replace
the io.buoyant.http namers, which will be removed in a subsequent change.

TODO:
- docs
- end-to-end tests on linkerd-protocol-http?


olix0r:
instead of solving this as a Namer, we should accomplish this with https://github.com/twitter/finagle/pull/472 and/or https://github.com/BuoyantIO/linkerd/issues/120

PR #59, opened 2016-02-01
RFC: implement defaults and validations on top of ObjectMapper-parsed JSON

gtcampbell:
This is a refinement on #19. It incorporates a new library for the Validation type (to allow us to accumulate validation errors rather than short-circuiting). As discussed offline, I will be pushing an alternate branch that removes the dependency in favor of only stdlib-provided classes like Either for comparison, to see how much readability and composability the added dependency buys us.

Some known remaining items:

- [ ] Implement the remainder of the configuration options represented by the `*Initializer` classes (`NamerInitializer`, etc.)
- [ ] Generate `Stack.Params` for configuration of our domain objects.
- [ ] investigate whether we should be more specific about implicits to improve compile times

I welcome any and all feedback on the approach.


gtcampbell:
closing in favor of #113.

PR #88, opened 2016-02-10
introduce marathon namer

siggy:
skeleton for a marathon namer. primarily want feedback on the config structure and the usage of a `label` key.

### DIFF
--- a/examples/marathon.l5d
+++ b/examples/marathon.l5d
@@ -1,0 +1,17 @@
+namers:
+- kind:      io.l5d.experimental.marathon
+  prefix:    /io.l5d.marathon
+  host:      marathon.mesos
+  port:      80
+  uriPrefix: /marathon
+
+routers:
+- protocol: http
+  baseDtab: |
+    /host     => /io.l5d.marathon;
+    /method   => /$/io.buoyant.http.anyMethodPfx/host;
+    /http/1.1 => /method;
+  httpUriInDst: true
+  servers:
+  - port: 4140
+    ip: 0.0.0.0
--- a/linkerd/namer/marathon/src/main/scala/io/l5d/marathon.scala
+++ b/linkerd/namer/marathon/src/main/scala/io/l5d/marathon.scala
@@ -1,0 +1,80 @@
+package io.l5d.experimental
+
+import com.fasterxml.jackson.core.JsonParser
+import com.twitter.conversions.time._
+import com.twitter.finagle.param.Label
+import com.twitter.finagle.{Http, Stack, Path}
+import io.buoyant.linkerd.{NamerInitializer, Parsing}
+import io.buoyant.marathon.v2.{Api, AppIdNamer}
+
+/**
+ * Supports namer configurations in the form:
+ *
+ * <pre>
+ * namers:
+ * - kind:      io.l5d.experimental.marathon
+ *   prefix:    /io.l5d.marathon
+ *   host:      marathon.mesos
+ *   port:      80
+ *   uriPrefix: /marathon
+ * </pre>
+ */
+object marathon {
+
+  /** The mesos master host; default: marathon.mesos */
+  case class Host(host: String)
+  implicit object Host extends Stack.Param[Host] {
+    val default = Host("marathon.mesos")
+    val parser = Parsing.Param.Text("host")(Host(_))
+  }
+
+  /** The mesos master port; default: 80 */
+  case class Port(port: Int)
+  implicit object Port extends Stack.Param[Port] {
+    val default = Port(80)
+    val parser = Parsing.Param.Int("port")(Port(_))
+  }
+
+  /** URI Prefix; default: empty string */
+  case class UriPrefix(uriPrefix: String)
+  implicit object UriPrefix extends Stack.Param[UriPrefix] {
+    val default = UriPrefix("")
+    val parser = Parsing.Param.Text("uriPrefix")(UriPrefix(_))
+  }
+
+  val parser = Parsing.Params(
+    Host.parser,
+    Port.parser,
+    UriPrefix.parser
+  )
+
+  val defaultParams = Stack.Params.empty +
+    NamerInitializer.Prefix(Path.Utf8("io.l5d.marathon"))
+}
+
+/**
+ * Configures a Marathon namer.
+ */
+class marathon(val params: Stack.Params) extends NamerInitializer {
+  def this() = this(marathon.defaultParams)
+  def withParams(ps: Stack.Params) = new marathon(ps)
+
+  def paramKeys = marathon.parser.keys
+  def readParam(k: String, p: JsonParser) =
+    withParams(marathon.parser.read(k, p, params))
+
+  /**
+   * Build a Namer backed by Marathon.
+   */
+  def newNamer() = {
+    val marathon.Host(host) = params[marathon.Host]
+    val marathon.Port(port) = params[marathon.Port]
+    val marathon.UriPrefix(uriPrefix) = params[marathon.UriPrefix]
+    val NamerInitializer.Prefix(path) = params[NamerInitializer.Prefix]
+    val service = Http.client
+      .configured(Label("namer" + path.show))
+      .newService(s"/$$/inet/$host/$port")
+
+    new AppIdNamer(Api(service, host, uriPrefix), prefix, 250.millis)
+  }
+}
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
@@ -1,0 +1,128 @@
+package io.buoyant.marathon.v2
+
+import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
+import com.twitter.finagle.{Service, http}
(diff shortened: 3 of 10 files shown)


examples/marathon.l5d
```diff
@@ -0,0 +1,17 @@
+namers:
+- kind:      io.l5d.experimental.marathon
+  prefix:    /io.l5d.marathon
+  host:      marathon.mesos
+  port:      80
+  uriPrefix: /marathon
```
esbie: I don't understand why the api is prefixed with /marathon, or what I would do to change it away from the default of /marathon. Adding docs to config.md might be good to do as part of this change

siggy: Good point, I'll update config.md with some of this detail.

That API prefix can vary depending on how marathon is configured. 

For example, running locally based on these instructions:
https://mesosphere.github.io/marathon/docs/
... there is no API prefix, you'd do:
```
$ curl localhost:8080/v2/apps
```

Using the default DCOS / AWS setup:
https://mesosphere.com/amazon/
... you'd do:
```
curl $(dcos config show core.dcos_url)/marathon/v2/apps
```

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -0,0 +1,134 @@
...
+ */
+
+trait Api {
+  def getApps(): Future[Api.Apps]
+  def getAddrs(app: String): Future[Set[SocketAddress]]
+}
+
+object Api {
+
+  type Apps = Set[String]
```
esbie: tioli it's more clear to call these AppIds

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -0,0 +1,119 @@
+package io.buoyant.marathon.v2
```
esbie: the other namers don't package underneath the api version (eg v2)... I think the intent there is that AppIdNamer could work under any version of the api that's released

siggy: yeah, i originally wrote it that way. @olix0r mentioned that it's preferable to tie AppIdNamer to a specific API version.

esbie: *shrug* alright, whatever the bossman says then

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -0,0 +1,119 @@
...
+      Closable.make { deadline =>
+        work.raise(Closed)
+        Future.Unit
+      }
+    }
+    Activity(states)
+  }
+
+  private[this] var appMonitors: Map[String, Var[Addr]] = Map.empty
+  private[this] def watchAddr(app: String): Var[Addr] = synchronized {
```
esbie: tioli feels more like a `getMonitoredAddr`

siggy: hmmm, this is a bit weird because the function both retrieves an address, and also initiates an ongoing watch as a side-effect. we could call it `getMonitoredAddrAndStartWatching`, but i'm inclined to just leave it as-is, to at least be explicit about the side-effect part.

esbie:
documentation looks great, thanks for fixing those examples! :put_litter_in_its_place: 
I would explain what `uriPrefix` is and how to use it in the docs. tioli I would make empty string the default, since that's a possible configuration. other than that it looks :star:

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -0,0 +1,119 @@
...
+   *
+   * and attempts to bind an Addr by resolving named endpoint from the
+   * Marathon master.
+   */
+  def lookup(path: Path): Activity[NameTree[Name]] = path match {
+    case Path.Utf8(app, _*) =>
+      val id = prefix ++ Path.Utf8(app)
+      val residual = path.drop(1)
+      // Marathon appends a leading '/' to app names that don't have it.
+      val appId = if (app(0) == '/') {
```
olix0r: prefer `app.startsWith("/")` which can't blow up on index out of bounds

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -0,0 +1,119 @@
...
+        app
+      } else {
+        s"/$app"
+      }
```
olix0r: style nit, prefer:
```scala
val appId =
  if (app startsWith "/") app
  else s"/$app"
```

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
esbie: sure sure, I had the opposite confusion though, where I was concerned that a new--possibly duplicate--watch request was sent on every lookup

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -0,0 +1,119 @@
...
+      log.debug("marathon lookup: %s %s %s", id.show, path.show, appId)
+      appsActivity.map {
+        case apps if apps(appId) =>
+          val addr = watchAddr(appId)
+          log.debug("marathon app found: (%s) appId: %s addr: %s", apps.mkString(", "), appId, addr)
+          NameTree.Leaf(Name.Bound(addr, id, residual))
+        case apps =>
+          log.debug("marathon app not found: (%s) appId: %s", apps.mkString(", "), appId)
```
olix0r: We have a ticket elsewhere for this, but it's worth preempting it here: Let's use Trace.record, which will save us a lot of time in the longterm.  E.g.
```scala
Trace.recordBinary("marathon.appId", appId)
...
Trace.record("marathon.found", true)
...
Trace.record("marathon.found", false)
```

It's probably not useful to log `addr` here (since it changes on its own, and this will just be some memory address here)

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
siggy: ok i get that. updated with a super-duper-explicit `getAndMonitorAddr`.

also your comments made me think of this:
monitor

esbie:
great! :+1:

linkerd/namer/marathon/src/main/scala/io/l5d/marathon.scala
```diff
@@ -0,0 +1,80 @@
...
+    withParams(marathon.parser.read(k, p, params))
+
+  /**
+   * Build a Namer backed by Marathon.
+   */
+  def newNamer() = {
+    val marathon.Host(host) = params[marathon.Host]
+    val marathon.Port(port) = params[marathon.Port]
+    val marathon.UriPrefix(uriPrefix) = params[marathon.UriPrefix]
+    val path = params[NamerInitializer.Prefix].path.show
```
olix0r: tioli, mild pref for:
```scala
 val NamerInitializer.Prefix(path) = params[NamerInitializer.Prefix]
...
   .configured(Label("namer" + path.show))
```

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -0,0 +1,134 @@
...
+  private[this] case class AppNode(
+    id: Option[String],
+    tasks: Option[Seq[TaskNode]]
+  )
+
+  private[this] case class AppsRsp(
+    apps: Option[Seq[AppNode]] = None
+  ) {
+    def toApps(): Try[Api.AppIds] =
+      Try(
```
olix0r: for blocks:
```scala
Try {
  ...
}
```

But does this really need to be a Try at all?

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -0,0 +1,134 @@
...
+        apps match {
+          case Some(apps) => apps.map { app => app.id.getOrElse("") }.toSet
+          case None => Set[String]()
```
olix0r: Prefer `Set.empty[String]`

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -0,0 +1,134 @@
...
+        }
+      )
+  }
+
+  private[this] case class AppRsp(
+    app: Option[AppNode] = None
+  ) {
+    def toAddresses(): Try[Set[SocketAddress]] = Try {
+      app match {
```
olix0r: Similarly, what exceptions are we expecting here that this needs to be a Try?

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
olix0r: thought was: appid is a feature of the v2 api.  it may not exist in v3?

olix0r:
this is looking good! rebase and it's :+1: :100:

esbie:
review-ninja gods demand :star:

siggy:
:star: rebase

PR #111, opened 2016-02-19
Allow multi-segment marathon ids in AppIdNamer

### LINKED ISSUE #110: support grouped app ids in the marathon namer
We currently do not have a good way to map a hostname to a marathon app id containing slashes.

Given an app ID like `testing/hello`, we should provide a way for an http request to have a host header like `hello.testing` to route to this marathon app.

I think that the marathon namer should be changed to admit slashed names, rather than requiring that app ids be a single path element.

esbie:
Updates AppIdNamer to accept any number of path segments that represent a valid marathon id. Removes support for segments to contain `/` characters.

Updated the marathon example to support http identifiers like `/http/1.1/GET/hello.testing/residual`

Fixes #110

### DIFF
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
@@ -21,37 +21,47 @@ class AppIdNamer(
   import AppIdNamer._
 
   private[this] implicit val _timer = timer
+  private[this] case class PathSpan(app: Path, residual: Path)
 
   /**
    * Accepts names in the form:
    *   /<app-id>/residual/path
    *
+   *   or
+   *
+   *   /<group-id>/<app-id>/residual/path
+   *
+   *   etc
+   *
    * and attempts to bind an Addr by resolving named endpoint from the
    * Marathon master.
    */
-  def lookup(path: Path): Activity[NameTree[Name]] = path match {
-    case Path.Utf8(app, _*) =>
-      val id = prefix ++ Path.Utf8(app)
-      val residual = path.drop(1)
-      // Marathon appends a leading '/' to app names that don't have it.
-      val appId =
-        if (app startsWith "/") app
-        else s"/$app"
-      Trace.recordBinary("marathon.appId", appId)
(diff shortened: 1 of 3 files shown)


marathon/src/test/scala/io/buoyant/marathon/v2/AppIdNamerTest.scala
```diff
@@ -48,10 +48,10 @@ class AppIdNamerTest extends FunSuite with Awaits {
     assert(state == Activity.Ok(NameTree.Neg))
   }
 
-  test("Namer handles looking up both appId and /appId") {
+  test("Namer handles looking up /app/id") {
     class TestApi() extends Api {
       def getAppIds(): Future[Api.AppIds] = {
-        Future.value(Set[String]("/foo", "/servicename"))
+        Future.value(Set[String]("/service/name"))
```
olix0r: tiolo: `[String]` isn't necessary here

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -21,37 +21,46 @@ class AppIdNamer(
...
+          case Some(PathSpan(app, residual)) =>
+            val id = prefix ++ app
+
+            Trace.recordBinary("marathon.appId", app.show)
+            Trace.recordBinary("marathon.id", id.show)
+            Trace.recordBinary("marathon.found", app.show)
+
+            val addr = getAndMonitorAddr(app.show)
+            NameTree.Leaf(Name.Bound(addr, id, residual))
+          case None =>
```
olix0r: tioli: I'd put a newline above case None to give it some visual spacing

olix0r:
:star: i like it! just style nits

tioli: it'd be nice to have a slightly more descriptive commit message (a sentence or two), but not a hard requirement

esbie:
@olix0r thanks! better commit message and nits addressed. 

fyi I set up pull approvals to remain around after rebase (so there's no need to re-star now)

klingerf:
Cool -- looks good to me. :star: :musical_keyboard:

klingerf:
:star: :musical_keyboard:

PR #113, opened 2016-02-19
Refactor config parsing to use jackson ObjectMapper

adleong:
This is a major refactor to move away from using the streaming JsonParser and
instead use ObjectMapper.  Case classes define the structure of the configs.

A hierarchy of config objects explicitly define the config structure.  For example, a `LinkerConfig` contains `RouterConfig`s which contain `ServerConfig`s.  Jackson's ObjectMapper is used to parse the config file into these config objects.  Each config object then knows how to reify itself into a linkerd/core object.

The advantage of this over the JsonParser model is that the structure of the config is very explicit and obvious from the code.  Adding or modifying sections or fields of the config becomes much more straightforward.

Some config objects such as `NamerConfig` are abstract traits and specific implementations must be provided by a plugin.  For example, `NamerInitializer`s are loaded via `ServiceLoader`.  The `NamerInitializer`'s only job is to register a subtype of `NamerConfig` with the `ObjectMapper`.  For example, `fs` register `FsConfig` (a subtype of `NamerConfig`) under the id `io.l5d.fs`.  Now, when jackson attempt to deserialize a `NamerConfig`, it will look at the `kind` property and use this to determine which registered subtype should be used.

Work still to be done:
* all unit/integration/e2e tests pass but this could use "real world" testing as well
* jackson's error messages are pretty good, but we should evaluate if they are sufficient in all cases or if we want to provide better error messages in some cases
* currently no error is thrown if multiple subtypes are registered to the same id.  this should generate an error
* re-add JTestNamer to prove that namers can still be created in java

This work could be done in a followup.


linkerd/config/src/main/scala/io/buoyant/linkerd/config/Parser.scala
```diff
@@ -0,0 +1,50 @@
...
+import scala.reflect.{ClassTag, classTag}
+
+abstract class ConfigDeserializer[T: ClassTag] extends StdDeserializer[T](Parser.jClass[T]) {
+  def register(module: SimpleModule): SimpleModule = module.addDeserializer(Parser.jClass[T], this)
+  protected def catchMappingException(ctxt: DeserializationContext)(t: => T): T =
+    try t catch { case arg: IllegalArgumentException => throw ctxt.mappingException(arg.getMessage) }
+}
+
+object Parser {
+  def jClass[T: ClassTag]: Class[T] = classTag[T].runtimeClass.asInstanceOf[Class[T]]
```
gtcampbell: it might be good to add a comment indicating why we need to do this typecasting (i.e. erasure)

linkerd/config/src/main/scala/io/buoyant/linkerd/config/Parser.scala
```diff
@@ -0,0 +1,50 @@
...
+   * We expose this publicly for testing purposes (to allow easy parsing of config subtrees) at the moment.
+   */
+  def objectMapper(config: String): ObjectMapper with ScalaObjectMapper = {
+    val factory = if (peekJsonObject(config)) new JsonFactory() else new YAMLFactory()
+    val customTypes = Seq(
+      new DirectoryDeserializer,
+      new DtabDeserializer,
+      new InetAddressDeserializer,
+      new PathDeserializer,
+      new PortDeserializer
```
gtcampbell: We should switch these to make use of LoadService so custom types can be defined in plugins.

linkerd/config/src/test/scala/io/buoyant/linkerd/config/ParserTest.scala
```diff
@@ -0,0 +1,143 @@
...
+    assert(baseConfig.baseDtab === Some(Dtab.read("/foo => /bar ;")))
+  }
+
+  test("simple http router") {
+    val baseRouter = baseConfig.routers.get.head.asInstanceOf[HttpRouterConfig]
+    assert(baseRouter.httpUriInDst === Some(true))
+    assert(baseRouter.label == None) // label should default to protocol
+
+    /*
+    TODO: assertions on Stack.Params!!
```
gtcampbell: we should probably address this TODO. :)

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/names/WebDelegator.scala
```diff
@@ -168,7 +168,7 @@ private[admin] class WebDelegator(
   def apply(req: Request): Future[Response] = req.method match {
     case Method.Get => (req.params.get("d"), req.params.get("n")) match {
       case (Some(DtabStr(dtab)), Some(PathStr(path))) =>
-        val DstBindingFactory.Namer(namer) = linker.params[DstBindingFactory.Namer]
+        val DstBindingFactory.Namer(namer) = linker.routers.head.params[DstBindingFactory.Namer]
```
esbie: Is this right? Is it ok to resolve the dtab of one router with the name interpreter of another? Are all router nameinterpreters the same?

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -1,175 +1,94 @@
...
-    val namers = NamerInitializers.load()
-    val tls = TlsClientInitializers.load()
-    mk(protos, namers, tls)
+      // At least one router must be specified
+      if (routers.isEmpty) {
+        throw NoRoutersSpecified
+      }
+
+      // Router labels must not conflict
+      routers.groupBy(_.label).collect {
```
gtcampbell: nit: since we're only doing this for the side-effects, I'd prefer a `foreach` to `collect` here.

linkerd/config/src/test/scala/io/buoyant/linkerd/config/ParserTest.scala
klingerf: Oh man looks like the tests in this file don't compile and aren't being run.  Investigating now, but I think this is actually a legacy file and should be removed.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/names/WebDelegator.scala
klingerf: Good question. It looks like all routers pick up the same set of name interpreters, here:

https://github.com/BuoyantIO/linkerd/blob/alex/config-madness/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala#L65

And since we require at least one router per config, then this assumption is probably ok, although I agree that it is confusing.

klingerf:
Ok, have addressed all of the review feedback in the most recent commit on this branch. Please take another look.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/names/WebDelegator.scala
gtcampbell: I also agree that this is confusing. Is there a reason we can't leave a way to access the namer via the linker? (we could even pass the linker to the router rather than the namer itself)

linkerd/config/src/main/scala/io/buoyant/linkerd/config/Parser.scala
```diff
@@ -0,0 +1,44 @@
...
+  private[this] def peekJsonObject(s: String): Boolean =
+    s.dropWhile(_.isWhitespace).startsWith("{")
+
+  /**
+   * Load a Json or Yaml parser, depending on whether the content appears to be Json.
+   * We expose this publicly for testing purposes (to allow easy parsing of config subtrees) at the moment.
+   */
+  def objectMapper(config: String): ObjectMapper with ScalaObjectMapper = {
+    val factory = if (peekJsonObject(config)) new JsonFactory() else new YAMLFactory()
+    val customTypes = LoadService[ConfigDeserializer[Any]]
```
gtcampbell: Q: does this work with `ConfigDeserializer[_]` rather than `ConfigDeserializer[Any]`?

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/names/WebDelegator.scala
esbie: I'd also be down for adding a more straightforward way to access the namers

linkerd/config/src/main/scala/io/buoyant/linkerd/config/types/ThriftProtocolDeserializer.scala
```diff
@@ -0,0 +1,16 @@
...
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.twitter.finagle.thrift.Protocols
+import io.buoyant.linkerd.config.ConfigDeserializer
+import org.apache.thrift.protocol.{TCompactProtocol, TProtocolFactory}
+
+class ThriftProtocolDeserializer extends ConfigDeserializer[TProtocolFactory] {
+  override def deserialize(jp: JsonParser, ctxt: DeserializationContext): TProtocolFactory = {
+    _parseString(jp, ctxt) match {
+      case "binary" => Protocols.binaryFactory()
+      case "compact" => new TCompactProtocol.Factory()
```
gtcampbell: We should probably add an explicit throw as the default case rather than relying on a MatchError here

esbie:
:star: I don't see anything here that can't be addressed in a followup

linkerd/config/src/main/scala/io/buoyant/linkerd/config/types/ThriftProtocolDeserializer.scala
klingerf: Agreed, but as far as I can tell this exception is swallowed anyway by jackson's JsonMappingException, so I though it would be more confusing to provide an exception and not have that be the actual exception that is thrown.  I think we should address more sane config parsing error messages in a separate branch.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/names/WebDelegator.scala
klingerf: Ok, have addressed this in the latest commit.

olix0r:
note: typo in commit messages "Explose" which is a cool word but probably not what you meant ;)

klingerf:
@olix0r d'oh. will squash that commit when this is ready to go.

linkerd/config/src/main/scala/io/buoyant/linkerd/config/Parser.scala
```diff
@@ -0,0 +1,44 @@
...
+   * We expose this publicly for testing purposes (to allow easy parsing of config subtrees) at the moment.
+   */
+  def objectMapper(config: String): ObjectMapper with ScalaObjectMapper = {
+    val factory = if (peekJsonObject(config)) new JsonFactory() else new YAMLFactory()
+    val customTypes = LoadService[ConfigDeserializer[_]]
+      .foldLeft(new SimpleModule("linkerd custom types")) { (module, d) =>
+        {
+          d.register(module)
+        }
+      }
```
olix0r: needless braces?

not a blocker, just noting style issues as i got through this

olix0r:
:star: this looks good!  my main beef is that there are a bunch of super-long lines that should be wrapped at 100 chars.  in no way a blocker

siggy:
:+1: :star:

klingerf:
Have fixed long lines and unused imports in the latest commit.  Running some external validations of this branch now, then will rebase before merging.

klingerf:
:star: <= rebased

gtcampbell:
:star: thanks for @adleong and @klingerf's hard work on this!

PR #114, opened 2016-02-19
Remove deprecated global flags.

esbie:
No need to keep these flags around since nameinitializers serve the
same purpose and our internal apps have migrated away from global
flags.


klingerf:
:star: :horse_racing: Thanks for cleaning this up!

adleong:
:star: :flags:

PR #116, opened 2016-02-19
Benchmarking proved my speculative optimization wrong. Something else…

stevej:
… made things better.

* Added a microbenchmark with jmh for AccessLogger.
* Switched back to shorter and simpler string interpolation because it's totally fine

I threw this into protocol-benchmark because that's what I was testing. We can move this elsewhere but I wanted an example of how to microbenchmark somewhere in our repo for future use.

### DIFF
--- a/linkerd/protocol/benchmark/src/main/scala/io/buoyant/protocol/AccessLoggerBenchmark.scala
+++ b/linkerd/protocol/benchmark/src/main/scala/io/buoyant/protocol/AccessLoggerBenchmark.scala
@@ -1,0 +1,34 @@
+package io.buoyant.linkerd.protocol
+
+import com.twitter.finagle.Service
+import com.twitter.finagle.http._
+import com.twitter.util.Future
+import com.twitter.util.StdBenchAnnotations
+import com.twitter.logging.NullLogger
+import org.openjdk.jmh.annotations._
+import io.buoyant.linkerd.protocol.http.AccessLogger
+import io.buoyant.test.Awaits
+
+// ./sbt 'project linkerd-protocol-benchmark' 'jmh:run -i 20 -prof gc .*AccessLoggerBenchmark.*'
+@State(Scope.Benchmark)
+class AccessLoggerBenchmark extends StdBenchAnnotations with Awaits {
+  val service = AccessLogger(NullLogger) andThen Service.mk[Request, Response] { req =>
+    val rsp = Response()
+    rsp.status = Status.PaymentRequired
+    rsp.contentType = "application/json"
+    rsp.contentLength = 304374
+    Future.value(rsp)
+  }
+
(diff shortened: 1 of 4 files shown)


linkerd/protocol/benchmark/src/main/scala/io/buoyant/protocol/AccessLoggerBenchmark.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  val req = Request()
+  req.method = Method.Head
+  req.uri = "/foo?bar=bah"
+  req.host = "monkeys"
+  req.contentType = "text/plain"
+
+  @Benchmark
+  def benchmark(): Future[Response] = {
+    service(req)
```
adleong: should this await?

stevej: I think that makes sense and have now made that change. I wouldn't expect it to make a difference (and it doesn't seem to) but there's no chance of a timing bug now.

project/LinkerdBuild.scala
```diff
@@ -126,6 +127,13 @@ object LinkerdBuild extends Base {
 
       val all = projectDir("linkerd/protocol")
         .aggregate(http, mux, thrift)
+
+      val benchmark = projectDir("linkerd/protocol/benchmark")
+        .dependsOn(core, Router.http)
```
klingerf: I don't think you import anything from router/http inside the linkerd/protocol/benchmark project -- maybe just remove line 132?

linkerd/protocol/benchmark/src/main/scala/io/buoyant/protocol/AccessLoggerBenchmark.scala
```diff
@@ -0,0 +1,34 @@
...
+import com.twitter.finagle.Service
+import com.twitter.finagle.http._
+import com.twitter.util.Future
+import com.twitter.util.StdBenchAnnotations
+import com.twitter.logging.NullLogger
+import org.openjdk.jmh.annotations._
+import io.buoyant.linkerd.protocol.http.AccessLogger
+import io.buoyant.test.Awaits
+
+// ./sbt 'project linkerd-protocol-benchmark' jmh:run -i 20 -prof gc .*AccessLoggerBenchmark.*
```
klingerf: Can you change this to:

    ./sbt 'project linkerd-protocol-benchmark' 'jmh:run -i 20 -prof gc .*AccessLoggerBenchmark.*'

adleong:
:star: :christmas_tree:

klingerf:
:star: :dango: Thanks for updating -- looks good to me.

PR #126, opened 2016-02-23
Switch to using lodash _.find in admin JS

### LINKED ISSUE #119: JavaScript error in admin ui
It appears there's a JS error in [`routers.js`](https://github.com/BuoyantIO/linkerd/blob/master/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js#L111).

Replacing `servers.find(` with `_.find(servers` fixes the issue.

Much of the admin UI appears broken without this fix.

gtcampbell:
We were using `Array.prototype.find`, which is not fully supported
across browsers.

Fixes #119.


klingerf:
:star: :zap: Thanks for fixing!

rmars:
:+1:

PR #127, opened 2016-02-23
TracerConfig and TracerInitializer

esbie:
I want to use my own tracer for linkerd routers and configure it
similar to how namers are configured.

My proposed solution is to add a top level `tracers` section to
the config. Each tracer requires a kind to be specified. See
TestTracer for an example on how I plan to use it

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
@@ -1,9 +1,10 @@
 package io.buoyant.linkerd
 
-import com.twitter.finagle.Stack
+import com.twitter.finagle.{param, Stack}
 import com.twitter.finagle.Stack.Param
 import com.twitter.finagle.buoyant.DstBindingFactory
 import com.twitter.finagle.naming.{DefaultInterpreter, NameInterpreter}
+import com.twitter.finagle.tracing.{NullTracer, DefaultTracer, BroadcastTracer, Tracer}
 import com.twitter.finagle.util.LoadService
 import io.buoyant.linkerd.config._
 
@@ -14,7 +15,7 @@ trait Linker {
   def routers: Seq[Router]
   def interpreter: NameInterpreter
   def admin: Admin
-
+  def tracer: Tracer
   def configured[T: Stack.Param](t: T): Linker
 }
 
@@ -40,7 +41,8 @@ object Linker {
     val protocols = LoadService[ProtocolInitializer]
     val namers = LoadService[NamerInitializer]
     val clientTls = LoadService[TlsClientInitializer]
-    parse(config, protocols ++ namers ++ clientTls)
+    val tracers = LoadService[TracerInitializer]
+    parse(config, protocols ++ namers ++ clientTls ++ tracers)
   }
 
   def load(config: String): Linker = {
@@ -50,12 +52,24 @@ object Linker {
   case class LinkerConfig(
     namers: Option[Seq[NamerConfig]],
     routers: Seq[RouterConfig],
+    tracers: Option[Seq[TracerConfig]],
     admin: Option[Admin]
   ) {
     def mk: Linker = {
-      val interpreter = namers.map(nameInterpreter)
 
-      val namersParam = Stack.Params.empty
+      val tracerImpls = tracers.map(_.map(_.newTracer()))
+      val tracer: Tracer = tracerImpls match {
+        case Some(Nil) => NullTracer
+        case Some(Seq(tracer)) => tracer
+        case Some(tracers) => BroadcastTracer(tracers)
+        case None => DefaultTracer
+      }
+
+      val namerParams = Stack.Params.empty + param.Tracer(tracer)
+
+      val interpreter = namers.map(nameInterpreter(namerParams))
+
+      val params = namerParams
         .maybeWith(interpreter.map(DstBindingFactory.Namer(_)))
 
       // At least one router must be specified
@@ -69,7 +83,7 @@ object Linker {
           if (rts.size > 1) throw ConflictingLabels(label)
       }
 
-      val routerImpls = routers.map(_.router(namersParam))
+      val routerImpls = routers.map(_.router(params))
 
       // Server sockets must not conflict
       routerImpls.flatMap(_.servers).groupBy(_.addr).collect {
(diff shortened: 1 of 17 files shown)


linkerd/core/src/main/scala/io/buoyant/linkerd/TracerInitializer.scala
```diff
@@ -0,0 +1,18 @@
+package io.buoyant.linkerd
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility
+import com.fasterxml.jackson.annotation.{JsonAutoDetect, JsonIgnore, JsonProperty, JsonTypeInfo}
+import com.twitter.finagle.Path
```
siggy: `JsonProperty` and `Path` unused

siggy:
after i run `./sbt test`, two files get reformatted:

```
diff --git a/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala b/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
index 5201805..76b35f2 100644
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
@@ -86,8 +86,6 @@ object Linker {
           if (svrs.size > 1) throw ConflictingPorts(svrs(0).addr, svrs(1).addr)
       }

-
-
       new Impl(routerImpls, interpreter.getOrElse(DefaultInterpreter), tracer, admin.getOrElse(Admin()))
     }
   }
diff --git a/linkerd/core/src/main/scala/io/buoyant/linkerd/TracerInitializer.scala b/linkerd/core/src/main/scala/io/buoyant/linkerd/TracerInitializer.scala
index 75e6f54..30de583 100644
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/TracerInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/TracerInitializer.scala
@@ -9,8 +9,8 @@ import com.twitter.finagle.tracing.Tracer
 @JsonAutoDetect(fieldVisibility = Visibility.ANY)
 trait TracerConfig {
   /**
-    * Construct a tracer.
-    */
+   * Construct a tracer.
+   */
   @JsonIgnore
   def newTracer(): Tracer
 }
```

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -50,13 +52,20 @@ object Linker {
...
     def mk: Linker = {
       val interpreter = namers.map(nameInterpreter)
 
-      val namersParam = Stack.Params.empty
-        .maybeWith(interpreter.map(DstBindingFactory.Namer(_)))
+      val tracerImpls = tracers.map(_.map(_.newTracer()))
+      val tracer: Tracer = tracerImpls match {
+        case Some(definedTracers) => BroadcastTracer(definedTracers)
+        case None => DefaultTracer
+      }
```
olix0r: tioli:
```scala
case Some(Nil) => NullTracer
case Some(Seq(tracer)) => tracer
case Some(tracers) => BroadcastTracer(tracers)
case None => DefaultTracer
```

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -50,13 +52,20 @@ object Linker {
...
+
+      val params = Stack.Params.empty
+        .maybeWith(interpreter.map(DstBindingFactory.Namer(_))) + param.Tracer(tracer)
```
olix0r: I'm skeptical about this.

Do we want to be able to trace naming-related requests? Or do we want to be able to use namers in tracers.  I don't think both makes sense; and I think it's probably more important to be able to trace naming-related requests...

Need to think about this a little.

linkerd/core/src/test/scala/io/buoyant/linkerd/TestTracer.scala
```diff
@@ -0,0 +1,21 @@
...
+
+class TestTracer extends TracerInitializer {
+  val configClass = Parser.jClass[TestTracerConfig]
+  val configId = "io.buoyant.linkerd.TestTracer"
+}
+
+object TestTracer extends TestTracer
+
+class TestTracerConfig extends TracerConfig {
+  config =>
```
olix0r: what does `config` do here?  omit it?

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
esbie: oof yeah I forgot about wanting to trace namer requests. On the other hand I really want to pass the name interpreter into the tracer so I can use it to resolve the tracer's client's path :|

linkerd/core/src/test/scala/io/buoyant/linkerd/TestTracer.scala
esbie: oh! vestigial

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
olix0r: Per offline discussion: we want to invert this so that the tracer is passed to the namers rather than passing namers to tracers.

esbie:
ptal

adleong:
:star: :pencil2: nice

klingerf:
:star: :night_with_stars: Looks good to me.

siggy:
:star: :+1:

linkerd/namer/k8s/src/main/scala/io/l5d/k8s.scala
```diff
@@ -73,6 +73,7 @@ case class k8sConfig(
     // namer path -- should we just support a `label`?
     val path = prefix.show
     val service = client
+      .withParams(params)
```
olix0r: so, i think this will entirely replace the client's params (and i.e. wipe out the tls config, and any http-specific param settings).

This probably needs to be:
```scala
.transformed { stk => stk.withParams(stk.params ++ params) }
```

linkerd/namer/consul/src/main/scala/io/l5d/consul.scala
```diff
@@ -41,8 +41,9 @@ case class ConsulConfig(
    * Build a Namer backed by Consul.
    */
   @JsonIgnore
-  def newNamer(): CatalogNamer = {
+  def newNamer(params: Stack.Params): CatalogNamer = {
     val service = Http.client
+      .withParams(params)
```
olix0r: probably has to be

```scala
.transformed { stk => stk.withParams(stk.params ++ params) }
```

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -50,12 +52,24 @@ object Linker {
...
-      val namersParam = Stack.Params.empty
+      val tracerImpls = tracers.map(_.map(_.newTracer()))
+      val tracer: Tracer = tracerImpls match {
+        case Some(Nil) => NullTracer
+        case Some(Seq(tracer)) => tracer
+        case Some(tracers) => BroadcastTracer(tracers)
+        case None => DefaultTracer
+      }
+
+      val tracerParams = Stack.Params.empty + param.Tracer(tracer)
```
olix0r: nit: this is more accurately _namerParams_ -- the params for namers.

olix0r:
looking good! i think we have to fix the `withParams` thing, though, or we'll break the kubernetes namer

linkerd/namer/k8s/src/main/scala/io/l5d/k8s.scala
esbie: huh, transform only operates on the stack, so withParams isn't an option here. gonna go with a `.copy()` instead, I think

PR #140, opened 2016-02-26
Annotate traces with router label

klingerf:
In looking at traces emitted from linkerd, it would be helpful to be able to distinguish traces by router in linkerd setups with multiple routers.  Am updating RoutingFactory to add a "router.label" trace annotation prior to routing requests.


esbie:
:+1: thanks for adding!

olix0r:
:star:

PR #141, opened 2016-02-29
Allow NamerInitializer modules to provide NameInterpreters.

olix0r:
In order to outsource name resolution (i.e. to another service), we need to be able to provide a configured NameInterpeter implementation.

Rather than complicate the configuration file to explicitly differentiate NameInterpreter and Namer configurations, they are merged into the `namers` field. If multiple NameInterpeter configurations are specified or if Namers are specified along with a NameInterpreter, an error is thrown.


siggy:
lgtm :star: :+1: 
tioli: update config.md?

klingerf:
:star: :turtle: Looks good. :+1: to @siggy's comment about config.md.

olix0r:
Going to hold off on documenting this until there is a concrete implementation.  config.md should be a reference, and there's nothing to reference at this point

PR #143, opened 2016-03-01
use http for slack.linkerd.io

siggy:


klingerf:
:star: :insecure: :star:

Issue #120, opened 2016-02-22
Pluggable HTTP request identifiers


olix0r:
Currently, a router's `Identifier[Req]` may be configurable, but it is not possible to replace the request identifier entirely.  Supporting a pluggable `Identifier` type, at least for HTTP, would allow users to write arbitrary request identifiers to inspect authentication, application-specific cookies, or even payloads.

These plugins must be protocol-specific and, at least initially, probably only make sense for HTTP.

PR #145, opened 2016-03-04
Make NameInterpreters configurable for each router

olix0r:
Each router should have its own dtab/name interpretation context.  In order to accommodate this, this change does the following:
- Remove the NamingFactoryConfig types introduced in https://github.com/BuoyantIO/linkerd/pull/141.  The `namers` block now only loads `Namer` configurations.
- Each router now has an optional `interpreter` configuration, backed by the `InterpreterInitializer` module type.  Interpreters are built against the list of process-wide `namers`.  A _default_ configuration is used if none is specified, which performs local name resolution.  This change will allow other (e.g. remote) interpreter contexts to be introduced on a per-router basis.
- Furthermore, module kinds had to be globally unique.  This is problematic if we want to use a value like _default_ for an interpreter `kind` as it would require that no other module types have a _default_ kind.  This change detects conflicts within a module type, but not across module types.


linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -21,46 +20,66 @@ trait Linker {
...
+  )
+
+  private[linkerd] def parse(
+    config: String,
+    inits: Initializers = LoadedInitializers
+  ): LinkerConfig = {
+    val mapper = Parser.objectMapper(config)
+    for (kinds <- inits.iter) {
+      ensureUniqueKinds(kinds)
+      kinds.foreach(_.registerSubtypes(mapper))
```
adleong: A bit dissonant to use for and foreach right next to each other.  Maybe use for here to be consistent?

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -101,26 +107,39 @@ trait RouterConfig {
   var baseDtab: Option[Dtab] = None
   var failFast: Option[Boolean] = None
   var timeoutMs: Option[Int] = None
+  var dstPrefix: Option[String] = None
```
adleong: This is unrelated to this review but I just thought of it.  If linker-level dtabs are gone and now dtabs are per-router, is a per-router `dstPrefix` still useful at all?

docs/config.md
```diff
@@ -130,7 +130,10 @@ Each router must be configured as an object with the following params:
   (default is protocol dependent)
 * *failFast* -- If `true`, connection failures are punished more aggressively.
   Should not be used with small destination pools. (default: false)
-* *timeoutMs* -- Per-request timeout in milliseconds. (default: no timeout)
+* *timeoutMs* -- Per-request timeout in milliseconds. (default: no  timeout)
+* *interpreter* -- an interpreter type name, determining what module
```
adleong: This is actually in interpreter section, right?
```
interpreter:
  kind: default
```

olix0r: ah, yep, correct. i'll clarify that

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
olix0r: Good question.

There's no reason that 2 routers couldn't share a dtab (so that they can be controlled centrally), but have distinct entry points...

Really, I think that we should drop `dstPrefix` in favor of `identifier` modules (#120).  Individual Identifiers could chose to expose a `prefix` configuration, or not.

adleong:
:star: I dig this ⛏

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -68,89 +87,45 @@ object Linker {
...
+      }
 
       // Router labels must not conflict
       routers.groupBy(_.label).foreach {
-        case (label, rts) =>
-          if (rts.size > 1) throw ConflictingLabels(label)
+        case (label, rts) if rts.size > 1 =>
+          throw ConflictingLabels(label)
+        case _ =>
       }
```
gtcampbell: tioli: I'm not a huge fan of `case _ =>`, although the only alternative I can think of is to use `collectFirst` instead of `foreach`. something like: 

```scala
routers
  .groupBy(_.label)
  .collectFirst { case (label, rts) if rts.size > 1 => ConflictingLabels(label) }
  .foreach(throw _)
```

gtcampbell:
:star: looks good; needs a rebase of course.

docs/config.md
```diff
@@ -130,7 +130,10 @@ Each router must be configured as an object with the following params:
   (default is protocol dependent)
 * *failFast* -- If `true`, connection failures are punished more aggressively.
   Should not be used with small destination pools. (default: false)
-* *timeoutMs* -- Per-request timeout in milliseconds. (default: no timeout)
+* *timeoutMs* -- Per-request timeout in milliseconds. (default: no  timeout)
```
klingerf: Smallest of nits -- you're introducing an extra space between "no" and "timeout".

esbie:
looks good, but I don't actually see the `interpreter` config section tested anywhere?
Also I see the default interpreterinitializer defined, but that's it. is the plan to not open source the others until later?

olix0r:
other implementations don't _really_ exist in the yet (i've got a branch cobbled together for demo purposes, but nothing ready to share). i expect that these implementations will land in this repo in the next ~month.

olix0r:
status: i've rebased this down to 2 primary commits. it still needs some docs improvements and better testing.  If someone feels like picking this up and taking it over the line... by all means, please do.  Otherwise I'll get to this as time permits.

gtcampbell:
I added some additional tests, etc. Looks like the branch has merge conflicts, so I'll deal with those now.

gtcampbell:
fixed merge conflicts - there were a bunch of changes conflicting  with @adleong's recent configuration branch, so I'd especially like him to look at this to ensure I resolved them in reasonable fashion.

adleong:
LGTM

klingerf:
:star: Looks good to me.

PR #146, opened 2016-03-05
Add config.json admin endpoint

adleong:
Add a config.json admin endpoint which re-serializes the parsed linkerd config.

### DIFF
--- a/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
+++ b/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
@@ -8,8 +8,9 @@ import com.twitter.server.Admin.Path
 import com.twitter.server.handler.{SummaryHandler => TSummaryHandler, _}
 import com.twitter.server.view.{IndexView, TextBlockView}
 import io.buoyant.linkerd.Linker
+import io.buoyant.linkerd.Linker.LinkerConfig
 
-class LinkerdAdmin(app: App, linker: Linker) {
+class LinkerdAdmin(app: App, linker: Linker, config: LinkerConfig) {
 
   private[this] val log = Logger()
 
@@ -48,6 +49,7 @@ class LinkerdAdmin(app: App, linker: Linker) {
     }
 
   private[this] def linkerdAdminRoutes: Seq[(String, Service[Request, Response])] = Seq(
+
     "/" -> new SummaryHandler(linker),
     "/files/" -> (StaticFilter andThen ResourceHandler.fromDirectoryOrJar(
       baseRequestPath = "/files/",
@@ -57,7 +59,8 @@ class LinkerdAdmin(app: App, linker: Linker) {
     "/delegator" -> DelegateHandler.ui(linker),
     "/delegator.json" -> DelegateHandler.api(linker),
     "/routers.json" -> new RouterHandler(linker),
-    "/metrics" -> MetricsHandler
+    "/metrics" -> MetricsHandler,
+    "/config.json" -> new ConfigHandler(config, Linker.configInitializers)
   )
 
   private[this] def metricsRoutes: Seq[(String, Service[Request, Response])] = Seq(
--- a/linkerd/config/src/main/scala/io/buoyant/linkerd/config/Parser.scala
+++ b/linkerd/config/src/main/scala/io/buoyant/linkerd/config/Parser.scala
@@ -1,8 +1,10 @@
 package io.buoyant.linkerd.config
 
+import com.fasterxml.jackson.annotation.JsonInclude.Include
 import com.fasterxml.jackson.core.JsonFactory
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer
 import com.fasterxml.jackson.databind.module.SimpleModule
+import com.fasterxml.jackson.databind.ser.std.StdSerializer
 import com.fasterxml.jackson.databind.{DeserializationContext, ObjectMapper}
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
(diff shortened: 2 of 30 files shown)


esbie:
looking good! Can you add a unit test to verify it has the kind of json output we expect? Also, this probably has a fair amount of overlap with routers.json and could potentially replace it

adleong:
Caution: Rabbit Hole Ahead

After trying for way too long to add a test for this, I've given up.  It seems that in tests, Jackson adds a field for the subtype class name e.g.
```
"protocol":"Fancy"
```
(Note that the class name is potentially different from the property value used to identify it.  Typically we use fully qualified class names as identifiers whereas Jackson will just print the simple class name here.)

The Jackson documentation (https://fasterxml.github.io/jackson-annotations/javadoc/2.4/com/fasterxml/jackson/annotation/JsonTypeInfo.html#property()) states that if the object has a property with the same name, the value from the object should be printed instead.  This works as expected when linkerd is run normally (not in a unit test).  However, in unit tests, jackson will print the class name in addition to the value from the object.  eg
```
{"protocol":"Fancy","servers":[{"port":2}],"protocol":"fancy"]}]}
```
(The uppercase "Fancy" is the config class name, the lowercase "fancy" is the name of the protocol.)  I have no idea why it does this or how to stop it.

linkerd/config/src/main/scala/io/buoyant/linkerd/config/Parser.scala
```diff
@@ -41,6 +54,19 @@ object Parser {
...
+        throw ConflictingSubtypes(cis(0).namedType, cis(1).namedType)
+    }
+    for (ci <- configInitializers) ci.registerSubtypes(mapper)
+
     mapper
   }
+
+  def jsonObjectMapper(
+    configInitializers: Seq[ConfigInitializer]
+  ): ObjectMapper with ScalaObjectMapper = objectMapper("{", configInitializers)
```
gtcampbell: this is clever, but I think it would be more readable to add another method that takes a `JsonFactory` or `YamlFactory` and have both this and `objectMapper` call into that.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -1,5 +1,6 @@
 package io.buoyant.linkerd.admin
 
+import com.fasterxml.jackson.databind.SerializationFeature
```
gtcampbell: Is this needed? I don't see it used in the file.

linkerd/config/src/main/scala/io/buoyant/linkerd/config/types/DtabDeserializer.scala
```diff
@@ -11,3 +11,11 @@ class DtabDeserializer extends ConfigDeserializer[Dtab] {
       Dtab.read(_parseString(parser, ctx))
     }
 }
+
+class DtabSerializer extends ConfigSerializer[Dtab] {
```
gtcampbell: I would prefer splitting the `*Serializer` classes out into their own `.scala` files, to match the Scala Style Guide's [recommendation on file organization](http://docs.scala-lang.org/style/files). Thoughts?

olix0r: these are so small, and closely related. i think it's fine for them to share a home.

adleong: These classes are so small that I find it more readable to put them in the same file.  I don't have a strong opinion about this, though.  Happy to change.

(Precedent: we violate the "one class per file" guideline all over the place)

olix0r: If you want to adhere to the letter of the style guide, I might do `object DtabCodec { class Serializer ... class Deserializer }` ; but I don't feel strongly about that.  I _do_ think it's a lot easier for these things to share a home...

gtcampbell: Mostly, I think it's a little surprising that a file called FooDeserializer.scala would contain both `FooDeserializer` and `FooSerializer`. Like the rest of you, I don't feel that strongly about this either way. :)

gtcampbell:
:star: I like the way this turned out!

linkerd/config/src/main/scala/io/buoyant/linkerd/config/types/DtabDeserializer.scala
adleong: I don't think you can discover inner classes via ServiceLoader :cry: 

Unless someone starts having :muscle: feelings, I'm going leave this alone.

esbie:
lgtm thanks for the test!

linkerd/config/src/main/scala/io/buoyant/linkerd/config/types/ThriftProtocolDeserializer.scala
```diff
@@ -17,3 +17,18 @@ class ThriftProtocolDeserializer extends ConfigDeserializer[TProtocolFactory] {
...
 }
+
+class ThriftProtocolSerializer extends ConfigSerializer[TProtocolFactory] {
+  override def serialize(
+    value: TProtocolFactory,
+    jgen: JsonGenerator,
+    provider: SerializerProvider
+  ): Unit = {
+    val protocol = value match {
+      case _: TBinaryProtocol.Factory => "binary"
```
klingerf: Unfortunately `TBinaryProtocol.Factory` is not the same type that's produced by calling `Protocols.binaryFactory()` above, so this deserializer always returns `thriftProtocol: unknown` when `thriftProtocol: binary` is specified in a config.  How hard would it be to update ConfigHandlerTest to also test this field?

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -57,7 +59,8 @@ class LinkerdAdmin(app: App, linker: Linker) {
     "/delegator" -> DelegateHandler.ui(linker),
     "/delegator.json" -> DelegateHandler.api(linker),
     "/routers.json" -> new RouterHandler(linker),
-    "/metrics" -> MetricsHandler
+    "/metrics" -> MetricsHandler,
+    "/config.json" -> new ConfigHandler(config, Linker.configInitializers)
```
klingerf: How would you feel about supporting "/config.yaml" as well? I'd accept "bad" as a valid answer :smile:

linkerd/config/src/main/scala/io/buoyant/linkerd/config/types/ThriftProtocolDeserializer.scala
adleong: Good catch, thanks.  Fixed and added to the test.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
adleong: bad

klingerf:
:star: :tanabata_tree: Thanks for updating -- I like your fix a lot.

adleong:
:star: rebase

klingerf:
:star:

PR #148, opened 2016-03-08
Add loadBalancer client config.

### LINKED ISSUE #147: configurable load balancing strategies

adleong:
Fixes #147 

Add a loadBalancer section to the client config where a load balancer can be
specified and configured.  The load balancers that are currently supported are
p2c, ewma, aperture, and heap.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -49,6 +49,8 @@ routers:
       kind: io.l5d.clientTls.static
       commonName: foo
       caCertPath: /foo/caCert.pem
+    loadBalancer:
+      kind: ewma
   timeoutMs: 1000
 
 - protocol: thrift
@@ -150,6 +152,8 @@ local IPv4 interfaces.
 <a name="basic-client-params"></a>
 ### Basic client parameters
 
+#### TLS
+
 * *tls* -- The router will make requests using TLS if this parameter is
 provided.  It must be an object containing keys:
   * *kind* -- One of the supported TlsClientInitializer plugins, by
@@ -158,11 +162,11 @@ provided.  It must be an object containing keys:
 
 Current TLS plugins include:
 
-#### io.l5d.clientTls.noValidation
+##### io.l5d.clientTls.noValidation
 
 Skip hostname validation.  This is unsafe.
 
-#### io.l5d.clientTls.static
+##### io.l5d.clientTls.static
 
 Use a single common name for all TLS requests.  This assumes that all servers
 that the router connects to all use the same TLS cert (or all use certs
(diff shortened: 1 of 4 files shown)


linkerd/core/src/test/scala/io/buoyant/linkerd/LoadBalancerTest.scala
```diff
@@ -0,0 +1,27 @@
...
+      |  client:
+      |    loadBalancer:
+      |      kind: $balancer
+      |  servers:
+      |  - {}
+      |""".stripMargin
+
+      val linker = Linker.load(config, Seq(TestProtocol.Plain))
+      val factory = linker.routers.head.params[LoadBalancerFactory.Param]
+      assert(factory.loadBalancerFactory != DefaultBalancerFactory)
```
adleong: I wish we could test more here, but finagle is pretty opaque about the load balancer.

docs/config.md
```diff
@@ -197,6 +201,47 @@ names:
   commonNamePattern: "{host}.buoyant.io"
 ```
 
+#### Load Balancer
+
```
gtcampbell: it might be good to note here what default load balancer is used if this option is unspecified (unless I missed that?)

gtcampbell:
:star: This looks like a useful thing to be able to configure.

siggy:
lgtm :+1: :star:

docs/config.md
adleong: Good point.  The default (p2c) comes from finagle so I'm a bit uneasy documenting something that could change out from under us.  But in this case I think it makes sense.

olix0r: also, per @stevej, i think we'd _like_ to change the default for routers to be EWMA...

adleong:
:star: rebase

docs/config.md
```diff
@@ -197,6 +201,47 @@ names:
...
+* *maxEffort* -- Optional.
+
+##### [ewma][ewma]
+
+ewma supports the following options (see [here][ewma] for option semantics and defaults):
+
+* *maxEffort* -- Optional.
+* *decayTimeMs* -- Optional.
+
```
olix0r: per my understanding, p2c and ewma (and apperature) aren't mutually exclusive. I believe that EWMA is a costing strategy to feed into P2C, and apperature is a windowing strategy taht still uses P2C.  I think that "p2c" should actually be "queuedepth", here -- but I defer to someone who knows more about the details...

adleong: That's right.  What we call ewma here is actually p2c+ewma.  Similarly, what we call aperture is actually aperture+least loaded.  The finagle docs (which we link to) spell this out more explicitly.

PR #155, opened 2016-03-10
Pluggable HTTP request identifiers

### LINKED ISSUE #120: Pluggable HTTP request identifiers

esbie:
Adds a linkerd-identifer-http module, and uses the current http.Identifier for the BasicHttpIdentifierInitializer.

I'm not very happy about adding a finagle http dependency to linkerd-core, but I needed `HttpIdentifierInitializer` in core in order to call `LoadService[HttpIdentifierInitializer]()` ... while at the same time wanting it typed to `Identifier[http.Request]`. Not sure if anyone has suggestions there.

(also...the reverse of echo is ohce)

Fixes #120

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -269,26 +269,14 @@ heap does not support any options.
 <a name="protocol-http"></a>
 ### HTTP/1.1 protocol parameters
 
-HTTP requests are routed by a combination of Host header, method, and URI.
-Specifically, HTTP/1.0 logical names are of the form:
-```
-  dstPrefix / "1.0" / method [/ uri* ]
-```
-and HTTP/1.1 logical names are of the form:
-```
-  dstPrefix / "1.1" / method / host [/ uri* ]
-```
-
-The default _dstPrefix_ is `/http`. In both cases, `uri` is only considered a part
-of the logical name if the config option `httpUriInDst` is true.
-
 Router configuration options include:
 * *httpAccessLog* -- Sets the access log path.  If not specified, no
 access log is written.
-* *httpUriInDst* -- If `true` http paths are appended to destinations. This allows
- path-prefix routing. (default: false)
+* *identifier* -- [Http-specific identifier](#protocol-http-identifiers) (default:
+default)
 
-The default server _port_ is 4140.
+The default _dstPrefix_ is `/http`
+The default server _port_ is 4140
 
 As an example, here's an http router config that routes all `GET` requests to
 8081 and all `POST` requests to 8091
@@ -304,6 +292,33 @@ routers:
   servers:
     port: 5000
 ```
+<a name="protocol-http-identifiers"></a>
+#### HTTP/1.1 Identifiers
+
+Identifiers are objects responsible for creating logical names from an incoming
+request with the following parameters:
+
+* *kind* -- One of the supported identifier plugins, by fully-qualified class
+ name. Current plugins include:
+  * *default*
+* *httpUriInDst* -- If `true` http paths are appended to destinations. This allows
+ path-prefix routing. (default: false)
+* any other identifier-specific parameters
+
+##### default
+
+HTTP requests are routed by a combination of Host header, method, and URI.
+Specifically, HTTP/1.0 logical names are of the form:
+```
+  dstPrefix / "1.0" / method [/ uri* ]
+```
+and HTTP/1.1 logical names are of the form:
+```
+  dstPrefix / "1.1" / method / host [/ uri* ]
+```
+
+In both cases, `uri` is only considered a part
+of the logical name if the config option `httpUriInDst` is true.
 
 <a name="protocol-thrift"></a>
 ### Thrift protocol parameters
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/IdentifierInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/IdentifierInitializer.scala
@@ -1,0 +1,5 @@
+package io.buoyant.linkerd
+
+import io.buoyant.linkerd.config.ConfigInitializer
+
+trait IdentifierInitializer extends ConfigInitializer
--- a/linkerd/identifier/http/src/main/scala/io/l5d/DefaultHttpIdentifierInitializer.scala
+++ b/linkerd/identifier/http/src/main/scala/io/l5d/DefaultHttpIdentifierInitializer.scala
@@ -1,0 +1,26 @@
+package io.l5d.identifier
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.{Dtab, Path}
+import io.buoyant.linkerd.IdentifierInitializer
+import io.buoyant.router.http.DefaultIdentifier
+import io.l5d.HttpIdentifierConfig
+
+class DefaultHttpIdentifierInitializer extends IdentifierInitializer {
+  val configClass = classOf[DefaultHttpIdentifierConfig]
+  override val configId = DefaultHttpIdentifierConfig.kind
+}
+
+object DefaultHttpIdentifierInitializer extends DefaultHttpIdentifierInitializer
+
+object DefaultHttpIdentifierConfig {
+  val kind = "default"
+}
+
+class DefaultHttpIdentifierConfig extends HttpIdentifierConfig {
+  @JsonIgnore
+  override def newIdentifier(
+    prefix: Path,
+    baseDtab: () => Dtab = () => Dtab.base
+  ) = DefaultIdentifier(prefix, httpUriInDst.getOrElse(false), baseDtab)
+}
--- a/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
+++ b/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
@@ -8,6 +8,8 @@ import com.twitter.finagle.stats.NullStatsReceiver
 import com.twitter.finagle.tracing.{Annotation, BufferingTracer, NullTracer}
 import com.twitter.finagle.{Http => FinagleHttp, Status => _, http => _, _}
 import com.twitter.util._
+import io.buoyant.router.Http
+import io.buoyant.router.http.DefaultIdentifier
 import io.buoyant.test.Awaits
 import java.net.InetSocketAddress
 import org.scalatest.FunSuite
@@ -73,7 +75,6 @@ class HttpEndToEndTest extends FunSuite with Awaits {
 routers:
 - protocol: http
   baseDtab: ${dtab.show}
-  httpUriInDst: true
   servers:
   - port: 0
 """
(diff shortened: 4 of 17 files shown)


adleong:
What about renaming `HttpIdentifierInitializer` to just `IdentifierInitializer` and moving it into core?  You would still have `BasicHttpIdentifierInitializer` extend `IdentifierInitializer` and you'd be able to change the load service line to `LoadService[IdentifierInitializer]` and kill the unwanted dependency.

docs/config.md
```diff
@@ -247,26 +247,16 @@ heap does not support any options.
...
-  dstPrefix / "1.1" / method / host [/ uri* ]
-```
-
-The default _dstPrefix_ is `/http`. In both cases, `uri` is only considered a part
-of the logical name if the config option `httpUriInDst` is true.
-
 Router configuration options include:
 * *httpAccessLog* -- Sets the access log path.  If not specified, no
 access log is written.
 * *httpUriInDst* -- If `true` http paths are appended to destinations. This allows
```
adleong: Should this be a `basicHttp` identifier specific option?

router/http/src/main/scala/io/buoyant/router/http/Identifier.scala
```diff
@@ -5,6 +5,14 @@ import com.twitter.finagle.buoyant.Dst
...
 
+object Identifier {
+  def mk(
+    prefix: Path,
+    uris: Boolean = false,
+    baseDtab: () => Dtab = () => Dtab.base
+  ): RoutingFactory.Identifier[http.Request] = Identifier(prefix, uris, baseDtab)
+}
+
 case class Identifier(
```
adleong: Should this move into the identifier/http project?  It's used as a default value for the identifier param so I guess not.

docs/config.md
esbie: I waffled on this a bit, as I assume most every http identifier will have this option. Ultimately I think you're right, it's specific enough to identifiers that I think it makes sense

esbie:
@adleong yeah I think I understand what you're proposing there, there's just an added wrinkle where I want a type-safe way to make sure the identifier can be used in the http router. Users shouldn't be able to use `io.l5d.identifier.basicHttp` in the thrift router etc.

...maybe a moot point since only http accepts identifiers right now? But I think oliver had a similar request in #120 "These plugins must be protocol-specific"

adleong:
`HttpConfig` would still contain a `HttpIdentifierConfig` so you would only be able to specify an http identifier in the http config.  `HttpIdentifierConfig` would still produce an `Identifier[http.Request]` so I think you're still good from a type-safety point of view.

With my suggestion, all `IdentifierInitializer`s would get loaded (potentially for multiple protocols) but they would only be used if there was a config class that included them.

esbie:
ok cool, I'll try it out, thanks :)

olix0r:
note: i would prefer if users don't have to specify fully-qualified paths.  I'd love it if each protocol could have a _default_ identifier, and others could be provided.  Not a hard requirement but _io.l5d.identifier.basicHttp_ is a bit verbose when we're already in the `identifier` field in the `http` protocol.

adleong:
:star: nice! :id:

esbie:
squashed and rebased after latest master conflict

router/http/src/main/scala/io/buoyant/router/Http.scala
```diff
@@ -1,21 +1,20 @@
...
 
   object param {
-
-    /** Whether URI paths should be included in Http router destinations. */
-    case class UriInDst(enabled: Boolean)
-    implicit object UriInDst extends Stack.Param[UriInDst] {
-      val default = UriInDst(enabled = false)
+    case class HttpIdentifier(id: (Path, () => Dtab) => RoutingFactory.Identifier[fhttp.Request])
+    implicit object HttpIdentifier extends Stack.Param[HttpIdentifier] {
+      val default = HttpIdentifier(id = Identifier.mk)
```
olix0r: tioli: omit `id = ` -- unless you think it adds clarity

router/http/src/main/scala/io/buoyant/router/http/Identifier.scala
```diff
@@ -5,6 +5,13 @@ import com.twitter.finagle.buoyant.Dst
...
 import io.buoyant.router.RoutingFactory
 
+object Identifier {
+  def mk(
+    prefix: Path,
+    baseDtab: () => Dtab = () => Dtab.base
+  ): RoutingFactory.Identifier[http.Request] = Identifier(prefix, false, baseDtab)
+}
+
 case class Identifier(
```
olix0r: now that we're allowing multiple Identifier types -- this should probably become BasicIdentifier or DefaultIdentifier or something that distinguishes it from other kinds.

linkerd/identifier/http/src/main/scala/io/l5d/BasicHttpIdentifierInitializer.scala
```diff
@@ -0,0 +1,26 @@
...
+
+class BasicHttpIdentifierInitializer extends IdentifierInitializer {
+  val configClass = classOf[BasicHttpIdentifierConfig]
+  override val configId = BasicHttpIdentifierConfig.kind
+}
+
+object BasicHttpIdentifierInitializer extends BasicHttpIdentifierInitializer
+
+object BasicHttpIdentifierConfig {
+  val kind = "default"
```
olix0r: tioli: I know before I suggested that this should be "default", but since we refer to this everywhere as _basic_, maybe it should just be that.  My main objection was to having a fully qualified classname.

linkerd/core/src/main/scala/io/buoyant/linkerd/IdentifierInitializer.scala
```diff
@@ -0,0 +1,5 @@
+package io.buoyant.linkerd
+
+import io.buoyant.linkerd.config.ConfigInitializer
+
+trait IdentifierInitializer extends ConfigInitializer
```
olix0r: nbd but files should have newlines at the end.  maybe update your ide/editor to do that by default.

project/LinkerdBuild.scala
```diff
@@ -83,6 +83,16 @@ object LinkerdBuild extends Base {
       .dependsOn(core)
       .withTests()
 
+
+    object Identifier {
+      val http = projectDir("linkerd/identifier/http")
+        .dependsOn(core, Router.http)
+        .withTests()
```
olix0r: tioli: it might be nicer to just keep this in the protocol project since it's a default implementation.  _especially_ since the actual identifier implementation is part of the router package and isn't actually modular.  but generally for default implementations i think it's fine to keep it part of the main protocol package.

olix0r:
This is looking good! just a few comments around clarity

linkerd/core/src/main/scala/io/buoyant/linkerd/IdentifierInitializer.scala
esbie: Found it in Intellij:
Settings→Editor→General→Other→Ensure line feed at file end on save.

linkerd/identifier/http/src/main/scala/io/l5d/BasicHttpIdentifierInitializer.scala
esbie: I like default more, so I'll move to that

linkerd/core/src/main/scala/io/buoyant/linkerd/IdentifierInitializer.scala
siggy: for reference, in sublime: `"ensure_newline_at_eof_on_save": true`

project/LinkerdBuild.scala
esbie: What about additional identifiers that aren't the default? I was thinking of writing some identifiers to replace rewriting namers. Would they also go in protocol.http or should they stay here?

router/http/src/main/scala/io/buoyant/router/Http.scala
```diff
@@ -1,21 +1,20 @@
...
 
-import com.twitter.finagle.{Http => FinagleHttp, Server => FinagleServer, http => _, _}
 import com.twitter.finagle.client.StackClient
 import com.twitter.finagle.http.{Request, Response, TlsFilter}
 import com.twitter.finagle.param.ProtocolLibrary
 import com.twitter.finagle.server.StackServer
-import io.buoyant.router.http.{ForwardedFilter, Identifier}
+import com.twitter.finagle.{Http => FinagleHttp, Server => FinagleServer, http => fhttp, _}
+import io.buoyant.router.Http.param.HttpIdentifier
+import io.buoyant.router.http.{ForwardedFilter, DefaultIdentifier$}
```
klingerf: Pretty sure this import should be:

    import io.buoyant.router.http.{ForwardedFilter, DefaultIdentifier}

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -8,6 +8,8 @@ import com.twitter.finagle.stats.NullStatsReceiver
 import com.twitter.finagle.tracing.{Annotation, BufferingTracer, NullTracer}
 import com.twitter.finagle.{Http => FinagleHttp, Status => _, http => _, _}
 import com.twitter.util._
+import io.buoyant.router.Http
+import io.buoyant.router.http.DefaultIdentifier$
```
klingerf: Same issue here?

    import io.buoyant.router.http.DefaultIdentifier

linkerd/identifier/http/src/main/scala/io/l5d/DefaultHttpIdentifierInitializer.scala
```diff
@@ -0,0 +1,26 @@
+package io.l5d.identifier
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.{Dtab, Path}
+import io.buoyant.linkerd.IdentifierInitializer
+import io.buoyant.router.http.{DefaultIdentifier}
```
klingerf: Unneeded braces here.

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
esbie: huh, Intellij's refactoring was not so hot

klingerf:
:star: :stars: Looks good. Just noting that changing the nesting on `httpUriInDst` is a breaking config change and will require a minor version bump when we release.

esbie:
rebased and added to changelog

PR #156, opened 2016-03-10
Add config options for controlling connection pool size.

### LINKED ISSUE #24: configurable client and server socket pools
It's trivial to create a traffic loop:
```sh
while curl -H 'Host: woop' -H 'Dtab-local: /host/woop=>/$/inet/127.1/4140' localhost:4140 ; do :; done
```
:skull: system runs out of file descriptors! :gun: 

We need to consider other means to prevent/avoid loops, but this highlights the need to be able to constrain:
1. the number of sockets that a server may accept concurrently.
2. the number of sockets that a single client (not router) may open.

Finagle has stack parameters to control these.  We need to ensure that there are sane default values, and expose a means to configure these parameters.

adleong:
Add a `maxConcurrentRequests` config option to limit number of concurrent
requests accepted by a server and a `hostConnectionPool` client config section
to control the number of connections maintained to destination hosts.

Fixes #24

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -156,14 +156,28 @@ local IPv4 interfaces.
   It must be an object containing keys:
   * *certPath* -- File path to the TLS certificate file
   * *keyPath* -- File path to the TLS key file
+* *maxConcurrentRequests* -- Optional.  The maximum number of concurrent
+requests the server will accept.  (default: unlimited)
 
 <a name="basic-client-params"></a>
 ### Basic client parameters
 
+* *hostConnectionPool* -- Optional.  Configure the number of connections to
+maintain to each destination host.  It must be an object containing keys:
+  * *minSize* -- Optional. The minimum number of connections to maintain to each
+  host.  (default: 0)
+  * *maxSize* -- Optional.  The maximum number of connections to maintain to
+  each host.  (default: Int.MaxValue)
+  * *idleTimeMs* -- Optional.  The amount of idle time for which a connection is
+  cached in milliseconds.  (default: forever)
+  * *maxWaiters* -- Optional.  The maximum number of connection requests that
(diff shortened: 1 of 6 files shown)


docs/config.md
```diff
@@ -148,10 +148,24 @@ local IPv4 interfaces.
   It must be an object containing keys:
   * *certPath* -- File path to the TLS certificate file
   * *keyPath* -- File path to the TLS key file
+* *maxConcurrentRequests* -- Optional.  The maximum number of concurrent
+requests the server will accept.  (default: unlimited)
 
 <a name="basic-client-params"></a>
 ### Basic client parameters
 
+* *hostConnectionPool* -- Configure the number of connections to maintain to
```
esbie: maybe mention whether the object is optional or required

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -1,13 +1,13 @@
...
+import com.fasterxml.jackson.annotation.{JsonIgnore, JsonProperty, JsonTypeInfo}
+import com.fasterxml.jackson.core.{io => _}
 import com.twitter.conversions.time._
 import com.twitter.finagle._
+import com.twitter.finagle.client.DefaultPool
 import com.twitter.finagle.param.Label
-import com.twitter.finagle.service.{TimeoutFilter, FailFastFactory}
+import com.twitter.finagle.service.{FailFastFactory, TimeoutFilter}
 import com.twitter.util.Closable
-import io.buoyant.router.RoutingFactory
```
esbie: looks like this import is still needed

esbie:
:star: lgtm

gtcampbell:
:star:

PR #157, opened 2016-03-11
Better unknown namer errors

### LINKED ISSUE #58: hard-to-understand error with non-existing namers
If you put a namer in the dtab that linkerd doesn't know about, you get the following behavior:

* The log says something unintelligible:
```
E 0201 20:23:21.968 THREAD19: Exception propagated to the root monitor!
Failure(io.l5d.http.anyMethodPfx, flags=0x100000000)
	with Service -> 0.0.0.0/4140
	at com.twitter.finagle.NoStacktrace(Unknown Source)
```
* The name of the unknown namer is returned as a text string with a 503 error (at least, with HTTP--haven't tried other protocols).

Ideally, the log message would be more informative. For the HTTP error, the 503 part is good, but ideally, the content would be either nothing or a more descriptive error message.

esbie:
The global namer's lookup method thinks anything in the pattern of `/$/foo` is a class ... meaning a mistyped namer can result in a ClassNotFoundException.

Here I'm wrapping the exception to provide additional context

Fixes #58

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
@@ -1,10 +1,8 @@
 package io.buoyant.linkerd
 
-import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility
-import com.fasterxml.jackson.annotation.{JsonAutoDetect, JsonIgnore, JsonTypeInfo}
-import com.twitter.finagle.{Dtab, Name, NameTree, Namer, Path, Stack}
+import com.twitter.finagle._
 import com.twitter.finagle.naming.NameInterpreter
-import com.twitter.util.Activity
+import com.twitter.util.{Var, Activity}
 
 object DefaultInterpreterConfig {
   val kind = "default"
@@ -23,6 +21,10 @@ class DefaultInterpreterInitializer extends InterpreterInitializer {
 }
 
 object DefaultInterpreterInitializer extends DefaultInterpreterInitializer
+
+case class UnknownNamer[Req](path: Path, cause: Throwable)
+  extends Exception(s"Unknown namer: ${cause.getMessage} for path: ${path.show}", cause)
+  with NoStacktrace
 
 /**
  * Namers are provided in preference-order so that first-match wins.
@@ -48,7 +50,12 @@ case class ConfiguredNamersInterpreter(namers: Seq[(Path, Namer)])
       // For now, punt to the default interpreter. Later, we
       // should consider removing the finagle Namer.global, which
       // provides /$/ names.
-      case Nil => Namer.global.lookup(path)
+      case Nil =>
+        Namer.global.lookup(path).transform {
+          case Activity.Failed(e: ClassNotFoundException) =>
+            Activity.exception(UnknownNamer(path, e))
+          case state => Activity(Var.value(state))
+        }
 
       // try to lookup the path with the matching namer, or
       // fallback to the rest of the namers.


linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
```diff
@@ -48,7 +50,14 @@ case class ConfiguredNamersInterpreter(namers: Seq[(Path, Namer)])
       // For now, punt to the default interpreter. Later, we
       // should consider removing the finagle Namer.global, which
       // provides /$/ names.
-      case Nil => Namer.global.lookup(path)
+      case Nil =>
+        Namer.global.lookup(path).transform {
+          case Activity.Failed(e) if e.isInstanceOf[ClassNotFoundException] =>
+            Activity.exception(UnknownNamer(path, e))
+          case Activity.Failed(e) => Activity.exception(e)
```
adleong: could all of these be grouped into a passthrough like 
```
case act => act
```
?

linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
```diff
@@ -48,7 +50,14 @@ case class ConfiguredNamersInterpreter(namers: Seq[(Path, Namer)])
       // For now, punt to the default interpreter. Later, we
       // should consider removing the finagle Namer.global, which
       // provides /$/ names.
-      case Nil => Namer.global.lookup(path)
+      case Nil =>
+        Namer.global.lookup(path).transform {
```
adleong: It is super lame that `Activity` doesn't have a `rescue` method.

linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
esbie: I'm not an activity expert, but transform expects matches on `Activity.State` that map to `Activity` :|

linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
esbie: yah, I wanted to use `.handle{}`, but it expects you to like... actually handle the error, not turn it into another better error

linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
olix0r: Can do:
```scala
case state => Activity(state)
```

linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
```diff
@@ -48,7 +50,14 @@ case class ConfiguredNamersInterpreter(namers: Seq[(Path, Namer)])
...
+          case Activity.Failed(e) if e.isInstanceOf[ClassNotFoundException] =>
```
olix0r: prefer:
```scala
case Activity.Failed(e: ClassNotFoundException) =>
```

olix0r:
looks good! just some small style nits.

linkerd/core/src/main/scala/io/buoyant/linkerd/DefaultInterpreterInitializer.scala
esbie: cool, turns out I was unnecessarily worried about type erasure

siggy:
:star: :+1: coool
```
$ curl http://localhost:4140
Unknown namer: FOOio.buoyant.http.anyMethodPfx for path: /$/FOOio.buoyant.http.anyMethodPfx/host/GET/localhost:4140
```

adleong:
:star: :mahjong:

PR #158, opened 2016-03-11
Allow serversets namer to admit residuals.

### LINKED ISSUE #151: serverset namer does not admit residuals
@lsjostro reported in slack that the io.l5d.serversets namer does not support residual paths.

If this isn't already supported by finagle-serversets, we should work with @edma and @atollena to try to get a namer with proper residual support into finagle.

adleong:
Fixes #151 

Currently, if zookeeper has a serverset at `/foo/bar` and you attempt to bind `/foo/bar/x/y/z`, the binding will fail.  This changes the serversets namer to recursively try shorter and shorter prefixes until a match is found.  In the above example `/foo/bar` would be bound with `/x/y/z` as the residual.

Ideally we could contribute this upstream to finagle-serversets.  Currently we have our own copy of `ServersetNamer` because of the visibility restrictions on the one from finagle.  Those would need to be relaxed before we could use it.

### DIFF
--- a/linkerd/namer/serversets/src/main/scala/io/buoyant/linkerd/namer/serversets/ServersetNamer.scala
+++ b/linkerd/namer/serversets/src/main/scala/io/buoyant/linkerd/namer/serversets/ServersetNamer.scala
@@ -41,7 +41,10 @@ class ServersetNamer(zkHost: String) extends Namer {
     resolve(s"zk2!$hosts!$path!$endpoint")
 
   /** Bind a name. */
-  protected[this] def bind(path: Path): Option[Name.Bound] = {
+  protected[this] def bind(path: Path, residual: Path = Path.empty): Option[Name.Bound] = {
+    // Clients may depend on Name.Bound ids being Paths which resolve
+    // back to the same Name.Bound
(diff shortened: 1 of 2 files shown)


gtcampbell:
:+1: LGTM.

olix0r:
:star: i think we should ship this, but I'd like to understand one thing before doing so (to determine if there needs to be followup work): are watches cached/shared across namer lookups?  Specifically, if we resolve two names:

1. `/foo/bar/bah/baz` => Bound(`/foo/bar`, `/bah/baz`)
2. `/foo/bar/bah` => Bound(`/foo/bar`, `/bah`)

what zk requests are required to satisfy this?  Ultimately, if we have an active watch on /foo/bar, the second lookup should be ~free.

As I said, I'm fine merging this without this optimization, but we should understand what's going on...

adleong:
I'm not totally sure, but I think it will work as you describe.  Zk2Resolver memoizes `(hosts, path, endpoint) => Var[Addr]` so both resolutions should get the same `Var[Addr]`.  This is similar to if you did multiple calls to resolve the same path.

PR #159, opened 2016-03-11
Don't render HostConnectionPool's default value.

adleong:


olix0r:
:star:

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -166,6 +166,7 @@ case class HostConnectionPool(
   idleTimeMs: Option[Int],
   maxWaiters: Option[Int]
 ) {
+  @JsonIgnore
   private[this] val default = DefaultPool.Param.param.default
```
gtcampbell: tioli: another option would be to move this into the companion object, right?

gtcampbell:
:star: This looks fine. If you like, you could move the default val into the companion object as I mentioned.

More general question: can/should we configure Jackson to ignore private properties?

PR #163, opened 2016-03-14
Chart individual clients on admin index page

### LINKED ISSUE #149: graph multiple stats on admin homepage
The admin homepage currently graphs the sum of all requests. It may be more useful to graph each request individually in the same chart.

esbie:
Recent demos suggest that the admin front page is more useful if
clients are individually charted instead of only showing a some of
all requests. This change assigns colors to each client and uses
that color to chart requests.

Please help me choose some not-terrible colors. Fixes #149 

screen shot 2016-03-14 at 4 42 53 pm

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
@@ -88,19 +88,36 @@ var Routers = (function() {
     return server;
   }
 
+  function onAddedClients(handler) {
+    var wrapper = function(events, clients) {
+      handler(clients);
+    }
+    $("body").on("addedClients", wrapper);
+    return wrapper;
+  }
+
   // Updates router clients and metrics from raw-key val metrics.
   function update(routers, metrics) {
     // first, check for new clients and add them
+    var addedClients = [];
+
     _.each(metrics, function(metric, key) {
       var match = key.match(clientRE);
       if (match) {
         var name = match[1], id = match[2],
             router = routers[name];
-        if (router) {
-          router.dstIds[id] = mkDst(name, id);
+        if (router && !router.dstIds[id]) {
+          var addedClient = mkDst(name, id);
+          addedClients.push(addedClient)
+          router.dstIds[id] = addedClient;
         }
       }
     });
+
+    if (addedClients.length)
+      $("body").trigger("addedClients", [addedClients]);
+
+    //TODO: remove any unused clients
 
     // then, attach metrics to each appropriate scope
     var routerNames = Object.keys(routers);
@@ -171,7 +188,27 @@ var Routers = (function() {
       findByMetricKey: function(key) { return findByMetricKey(this.data, key); },
 
       /** Finds a router associated with a scoped metric name. */
-      findMatchingRouter: function(key) { return findMatchingRouter(this.data, key); }
+      findMatchingRouter: function(key) { return findMatchingRouter(this.data, key); },
+
+      /** Add event handler for new clients */
+      onAddedClients: onAddedClients,
+
+      //convenience methods
+      servers: function(routerName) {
+        if (routerName && this.data[routerName]) {
+          return this.data[routerName].servers;
+        } else {
+          return _(this.data).map('servers').flatten().value();
+        }
+      },
+
+      clients: function(routerName) {
+        if (routerName && this.data[routerName]) {
+          return _.values(this.data[routerName].dstIds);
+        } else {
+          return _(this.data).map(function(router) { return _.values(router.dstIds); }).flatten().value();
+        }
+      }
     };
   };
 })();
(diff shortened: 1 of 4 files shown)


rmars:
I've used http://colorbrewer2.org/ in the past :fried_shrimp:

olix0r:
I'd be inclined to use [D3's color scales](https://github.com/mbostock/d3/wiki/Ordinal-Scales); although they also reference ColorBrewer (which I hadn't seen before, but looks like a nice library).

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
```diff
@@ -36,12 +36,31 @@ var Routers = (function() {
...
 
+  var mkColor = function() {
+    var colors = [
+      [152,224,197],
+      [74,216,172],
+      [64,118,196],
+      [83,176,196],
+      [166,220,0],
+      [0,0,43]
+    ];
```
klingerf: I think I'm just missing this in the diff, but it seems like wherever you're using a color value you're always calling `.join(",")` -- why not just making this an array of strings to start with?

esbie: no reason, I thought I needed more flexibility than I did

klingerf:
:star: :tangerine: Looks good. In the past I've used the Graphviz Brewer color schemes, but they're kinda retro... http://www.graphviz.org/doc/info/colors.html#brewer

klingerf:
Great, thanks for updating.

esbie:
ok thanks I chose some brewer colors
screen shot 2016-03-15 at 10 15 54 am

adleong:
:bike: :house: 
Since the clients don't represent some kind of scale or gradient should we avoid a gradient color scheme?  What about a set* scheme?

esbie:
there are greens in that color set too, they just weren't randomly chosen for that screenshot

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
```diff
@@ -36,12 +36,33 @@ var Routers = (function() {
...
+      "8,64,129"
+    ];
+
+    //returns a color based on summed ascii values
+    return function(label) {
+      var idx = _.sumBy(label.split(""), function(char) { return char.charCodeAt(0); });
+      var color = colors[idx % colors.length];
+      return color;
+    }
+  }();
```
siggy: it's nice that this is deterministic, but i'm concerned about cases where there are only two clients and they both end up the same color. we don't have to address this in this PR, but i think it should be addressed.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/utils.js
```diff
@@ -134,20 +134,26 @@ UpdateableChart.prototype.setMetric = function(metric) {
...
-    }
-  );
-
-  this._getMetrics(metrics);
+
+  this.tsMap = {};
+
+  _.each(metrics, function(metric) {
+    this.tsMap[metric.name] = new TimeSeries();
+    this.chart.addTimeSeries(
```
siggy: the `router clients` and related sections auto-update when clients are added/removed. i think the chart should at least add clients as they are created, removal is less important.

esbie: yeah I see your point. I'll implement client adds. It's going to be a bigger change though, since this will be the first client-side component that keeps state (the others don't really keep state, they just purge and re-render on every single update)

esbie:
alright, BigBoard now uses routers.onAddedClients() to listen for clients that are added after initialization

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/summary.js
```diff
@@ -226,21 +240,22 @@ var Interfaces = (function() {
...
-      .value();
+    var namerIfaces = _.map(namers, prepInterface);
 
-    var clients = _(routers)
-      .map(function(router) { return _.values(router.dstIds); })
-      .flatten()
-      .value();
+    $('#client-info').html(template({
+      name:'router clients',
+      interfaces: prepClients(routers.clients)
```
klingerf: When I run the admin server with your changes, the `prepClients` method throws the following exception:

    Uncaught TypeError: Cannot read property 'map' of undefined

I think that means that calling routers.clients here returns undefined, but I don't see anything obvious in that method definition that would be causing this.

olix0r:
Do we care that a hard refresh is required to show stats for new clients?  I think the client box appears, but no traffic in the top graph.  May be a bikeshed for another day...

olix0r:
Also I tested this in Safari and it all seemed to work

esbie:
sorry @klingerf @siggy @rmars for that last-minute refactor I did in the last commit that broke things.
@olix0r it sounds like you may not have pulled the 2nd commit

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
rmars: Yeahhh this happens for the calculator app
screen shot 2016-03-16 at 1 39 38 pm

olix0r:
@esbie ah, yup. sorry about that.  looks good! even in safari.

olix0r:
:star:

rmars:
:star: :rainbow: works for me now!

klingerf:
:star: :stars: Looks great -- thanks for tracking that js issue down.

siggy:
:star: :+1: Looks awesome!

PR #165, opened 2016-03-15
Refactor linkerd package structure

adleong:
* Created namer-core project (io.buoyant.namer) outside of linkerd and moved
namer code out of linkerd-core for eaiser reuse.
* Moved linkerd-config project to a top level project called config (io.buoyant.config) for easier reuse.
* Created a top level project called admin (io.buoyant.admin) and moved non-linkerd specific admin code out of linkerd-admin for easier reuse.


namer/core/src/test/scala/io/buoyant/namer/booNamer.scala
```diff
@@ -0,0 +1,10 @@
+package io.buoyant.namer
+
+import com.twitter.finagle.Path
+
+/**
+ * Created by alex on 15-3-2016.
+ */
```
klingerf: Maybe remove this comment from these 4 test config files?  We don't tend to include it in any of our other files.

namer/serversets/src/test/scala/io/l5d/ServersetNamerTest.scala
```diff
@@ -42,7 +42,7 @@ class ServersetNamerTest extends FunSuite {
   }
 
   def namer(path: String) = new ServersetNamer("host") {
-    val loopback = new InetSocketAddress(InetAddress.getLoopbackAddress, 1)
+    val loopback = Address(new InetSocketAddress(InetAddress.getLoopbackAddress, 1))
```
klingerf: Hmm, any idea why this compilation failure didn't trigger when we upgraded finagle?  This looks like the right fix, but just curious.

project/LinkerdBuild.scala
```diff
@@ -12,10 +12,12 @@ import pl.project13.scala.sbt.JmhPlugin
...
  * - marathon/ -- marathon client
  * - router/ -- finagle router libraries
- * - linkerd/ -- configuration, runtime, and modules
+ * - namer/ -- name resolution
+ * - config/ -- configuration utilities
+ * - linkerd/ -- runtime, and modules
  * - test-util/ -- async test helpers; provided by [[Base]]
  */
-object LinkerdBuild extends Base {
+object LinkerdBuild extends Base { self =>
```
klingerf: Instead of `self`, the convention we've used previously is to just reference `LinkerdBuild` directly.  See lines 83, 91, 95, etc. for examples of that.  I think we should either switch everything in this file to use `self`, or drop this variable and use `LinkerdBuild` throughout.  I'm good with either approach.

project/LinkerdBuild.scala
```diff
@@ -65,15 +67,57 @@ object LinkerdBuild extends Base {
       .aggregate(core, http, mux, thrift)
   }
 
+  val configCore = projectDir("config")
```
klingerf: TIOLI: I have a slight preference for calling this project val `config` instead of `configCore`. `configCore` implies to me that there's a `core` sub-project inside the `config` project, which is not the case.

namer/core/src/test/scala/io/buoyant/namer/booNamer.scala
adleong: Intellijaaaaaaaaaaaaaaaaay

project/LinkerdBuild.scala
adleong: unfortunately `config` is a name conflict with an sbt method.

klingerf: Ah ok, nevermind me.

klingerf:
:star: :racehorse: Looks good to me -- thanks for updating.

olix0r:
:star:

PR #172, opened 2016-03-18
minimize chance of duplicate colors in admin graph

### LINKED ISSUE #170: ensure unique graph colors on admin
https://github.com/BuoyantIO/linkerd/pull/163 introduced multiple metrics on the main admin page. The current implementation ensures color assignments are deterministic, but runs the risk of duplicate colors.

siggy:
fixes #170

screen shot 2016-03-18 at 10 13 32 am

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
@@ -37,22 +37,21 @@ var Routers = (function() {
       serverRE = /^rt\/(.+)\/srv\/([^/]+)\/(\d+)\/requests$/;
 
   var mkColor = function() {
+    var colorIdx = 0;
     var colors = [
-      "224,243,219",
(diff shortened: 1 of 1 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/routers.js
```diff
@@ -36,23 +36,22 @@ var Routers = (function() {
   var clientRE = /^rt\/(.+)\/dst\/id\/(.*)\/requests$/,
       serverRE = /^rt\/(.+)\/srv\/([^/]+)\/(\d+)\/requests$/;
 
+  var colorIdx = 0;
```
esbie: I'd move this inside mkColor
no reason for the rest of the routers code to access it

esbie:
:star: lgtm! don't think it'll work if we ever want server colors, but we can cross that bridge later if we need to

rmars:
:star: :rainbow: :art:

PR #175, opened 2016-03-19
Acceptance test for configuration

esbie:
Adds a script that runs in ci, which loads all possible l5d config
options, has a few requests run through it, then shuts down.

We might consider ditching consul/marathon examples in favor of this file? (just so we don't have so many configs to update when the format changes)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,3 +17,4 @@ test:
   # unfortunately these have to be two separate sbt commands ¯\_(ツ)_/¯
   - ./sbt coverage test && ./sbt coverageAggregate
   - ./sbt e2e:test assembly
+  - ./test-config.sh
--- a/examples/acceptance-test.l5d
+++ b/examples/acceptance-test.l5d
@@ -1,0 +1,138 @@
+# Configuration for all possible linkerd options
+
+admin:
+  port: 9990
+
+namers:
+- kind: io.l5d.fs
+  rootDir: ./examples/io.l5d.fs
+
+- kind: io.l5d.serversets
+  zkAddrs:
+  - host: 127.0.0.1
+    port: 2181
+
+- kind: io.l5d.experimental.consul
+  host: 127.0.0.1
+  port: 2181
+
+- kind: io.l5d.experimental.k8s
+  host: kubernetes.default.cluster.local
+  port: 443
+  tls: true
+  authTokenFile: ./examples/io.l5d.k8s/kube_token
+
+- kind: io.l5d.experimental.marathon
+  prefix: /io.l5d.marathon
+  host: marathon.mesos
+  port: 80
+  uriPrefix: /marathon
+
+routers:
+- protocol: http
+  label: int
+  baseDtab: |
+    /http/1.1 => /$/inet/127.1/9999;
+  dstPrefix: /http
+  failFast: false
+  timeoutMs: 100
+  httpAccessLog: logs/access.log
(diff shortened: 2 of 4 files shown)


examples/acceptance-test.l5d
```diff
@@ -0,0 +1,138 @@
...
+# TODO: test ssl traffic
+- protocol: http
+  label: tls
+  baseDtab: |
+    /host     => /io.l5d.fs;
+    /method   => /$/io.buoyant.http.anyMethodPfx/host;
+    /http/1.1 => /method;
+  client:
+    tls:
+      kind: io.l5d.clientTls.static
```
rmars: should we also test the other tls plugin (io.l5d.clientTls.boundPath)?

test-release.sh
```diff
@@ -0,0 +1,62 @@
...
+    # Report failure, leave tmp dir intact
+    echo "At least one request to linkerd failed."
+    echo " => Ensure that you have a local service running on port 9999"
+    echo " => Inspect logs at ./logs"
+    cleanup
+    exit 1
+  fi
+}
+
+echo "Running..."
```
rmars: tioli - "Running tests..."?  This statement appears between two "Starting http server" statements so looks a little incomplete without an object
```
Starting http server on 9999...
Running...
Serving HTTP on 0.0.0.0 port 9999 ...
```

rmars:
LGTM! yayyy

circle.yml
```diff
@@ -17,3 +17,4 @@ test:
   # unfortunately these have to be two separate sbt commands ¯\_(ツ)_/¯
   - ./sbt coverage test && ./sbt coverageAggregate
   - ./sbt e2e:test assembly
+  - ./test-release.sh
```
klingerf: Thanks for adding this! How would you feel about calling it `test-config.sh` or something similar? That might be more appropriate since we'll now be running it on every commit (hooray!).

test-release.sh
olix0r: I'd just omit this.  We can always run it with bash -x if we want details.

test-release.sh
```diff
@@ -0,0 +1,62 @@
...
+  http_test
+
+  # Stop linkerd
+  curl localhost:9990/admin/shutdown
+
+  if [ $request_failed = true ]; then
+    # Report failure, leave tmp dir intact
+    echo "At least one request to linkerd failed."
+    echo " => Ensure that you have a local service running on port 9999"
+    echo " => Inspect logs at ./logs"
```
olix0r: slightly better to send these to stderr (`>&2`)

circle.yml
olix0r: +1

olix0r:
:star: just nits

test-config.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Start http server 9999
+if [ -f "http.pid" ]; then
+  kill `cat http.pid`
```
klingerf: Mind adding `http.pid` in .gitignore?

klingerf: And `logs/` would probably also be appropriate.

esbie: ok sure

adleong:
:star: :hamburger: nice

klingerf:
:star: :ship: Looks great -- thanks for updating!

PR #176, opened 2016-03-19
Replace routers.json usage with config.json

esbie:
No need to keep around two endpoint that are essentially the same
except config.json provides more info


klingerf:
:star: :shell: Nice cleanup!

rmars:
:star: :cake: yay

PR #177, opened 2016-03-19
Support building against -SNAPSHOT twitter dependencies

olix0r:
Add an sbt setting, developTwitterDeps, that causes all twitter dependencies to
be set to their SNAPSHOT versions.  The project/install-twitter-develop-deps.sh
script fetches and installs SNAPSHOT version of all relevant projects.

Once SNAPSHOT versions are published, linkerd may be tested against these
dependencies with the following:

    ./sbt 'set developTwitterDeps in Global := true' test

circle.yml has been updated to support develop builds (via environment) and to publish test reports.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,21 +17,13 @@ dependencies:
       mkdir -p .gitshas
       export GIT_SHA_DIR=$PWD/.gitshas
       ./project/install-twitter-develop-deps.sh
-      ./sbt 'set developTwitterDeps in Global := true' update
-    else
-      ./sbt update
     fi
+  - ./sbt update
 
 test:
   override:
-  - >
-    if [ -n "${TWITTER_DEVELOP}" ]; then
-      ./sbt 'set developTwitterDeps in Global := true' coverage test
-      ./sbt 'set developTwitterDeps in Global := true' integration:compile e2e:test assembly
-    else
-      ./sbt coverage test
-      ./sbt integration:compile e2e:test assembly
-    fi
+  - ./sbt coverage test
+  - ./sbt integration:compile e2e:test assembly
   - ./test-config.sh
 
   post:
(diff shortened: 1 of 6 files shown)


project/Base.scala
```diff
@@ -159,6 +159,23 @@ class Base extends Build {
...
+          dep.copy(revision = dep.revision+"-SNAPSHOT")
+        } else dep
+      })
+
+    def withTwitterLibs(deps: Seq[ModuleID]): Project =
+      project.settings(libraryDependencies ++= {
+        if (doDevelopTwitterDeps.value) {
+          deps.map { dep => dep.copy(revision = dep.revision+"-SNAPSHOT") }
+        } else deps
+      })
```
klingerf: How about defining `withTwitterLibs` to call out to `withTwitterLib` so that you don't have to do the revision mucking in multiple places?  Something like:

```
def withTwitterLibs(deps: Seq[ModuleID]): Project =
  deps.foldLeft(project) { (_, mod) => withTwitterLib(mod) }
```

esbie:
:star: yay! was able to run this locally no problem

circle.yml
```diff
@@ -10,10 +10,19 @@ dependencies:
...
+    else
+      ./sbt update
+    fi
 
 test:
   override:
   # unfortunately these have to be two separate sbt commands ¯\_(ツ)_/¯
   - ./sbt coverage test && ./sbt coverageAggregate
-  - ./sbt e2e:test assembly
+  - ./sbt integration:compile e2e:test assembly
```
klingerf: I couldn't easily figure this out so will just ask -- do these sbt commands not also require `'set developTwitterDeps in Global := true'` in order to run against the finagle snapshots?

olix0r: yep, good call

klingerf:
:star: Looks good to me -- thanks for making those changes.

PR #179, opened 2016-03-21
Add an enableProbation flag to the load balancer config

### LINKED ISSUE #166: make Probation configurable
Finagle's [LoadBalancerFactory.EnableProbation](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/LoadBalancerFactory.scala#L28) param controls whether endpoints are eagerly evicted from service discovery.  we should add a configuration option to the router client config.

I also think we should enable this by default (i.e. in Router.scala).  It's a much safer default behavior that Twitter has not been able to change for legacy reasons.

rmars:
Allows us to control whether endpoints are eagerly evicted from service discovery

See https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/LoadBalancerFactory.scala#L28

Fixes #166

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -53,6 +53,7 @@ routers:
       caCertPath: /foo/caCert.pem
     loadBalancer:
       kind: ewma
+      enableProbation: false
   timeoutMs: 1000
 
 - protocol: thrift
@@ -230,6 +231,9 @@ names:
 * *loadBalancer* -- Optional.  Specifies a load balancer to use.  It must be an
 object containing keys:
   * *kind* -- One of the supported load balancers.
+  * *enableProbation* -- Optional.  Controls whether endpoints are eagerly evicted from
+    service discovery. (default: true)
+    See Finagle's [LoadBalancerFactory.EnableProbation](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/LoadBalancerFactory.scala#L28)
   * Any options specific to the load balancer.
 
 If unspecified, p2c is used.
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/LoadBalancerConfig.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/LoadBalancerConfig.scala
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonSubTypes.Type
 import com.fasterxml.jackson.annotation.{JsonIgnore, JsonSubTypes, JsonTypeInfo}
 import com.twitter.conversions.time._
 import com.twitter.finagle.Stack
+import com.twitter.finagle.loadbalancer.LoadBalancerFactory.EnableProbation
 import com.twitter.finagle.loadbalancer.{Balancers, LoadBalancerFactory}
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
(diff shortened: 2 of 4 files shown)


docs/config.md
```diff
@@ -230,6 +231,9 @@ names:
 * *loadBalancer* -- Optional.  Specifies a load balancer to use.  It must be an
 object containing keys:
   * *kind* -- One of the supported load balancers.
+  * *probationEnabled* -- Controls whether endpoints are eagerly evicted from
```
adleong: What do you think about calling this just `probation` instead of `probationEnabled`?

docs/config.md
```diff
@@ -230,6 +231,9 @@ names:
...
+    service discovery.
+    See Finagle's [LoadBalancerFactory.EnableProbation](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/loadbalancer/LoadBalancerFactory.scala#L28)
```
adleong: can you specify that this is optional and what the default is

linkerd/core/src/main/scala/io/buoyant/linkerd/LoadBalancerConfig.scala
```diff
@@ -16,8 +17,11 @@ import com.twitter.finagle.loadbalancer.{Balancers, LoadBalancerFactory}
 trait LoadBalancerConfig {
   val factory: LoadBalancerFactory
 
+  val enableProbation: Option[Boolean] = None
+
   @JsonIgnore
-  def clientParams = Stack.Params.empty + LoadBalancerFactory.Param(factory)
+  def clientParams = Stack.Params.empty + LoadBalancerFactory.Param(factory) +
+    LoadBalancerFactory.EnableProbation(enableProbation.getOrElse(true))
```
adleong: TIOLI
If you want to default to whatever the param's default is instead of explicitly defaulting to true, you could do
```
.maybeWith(enableProbation.map(LoadBalancerFactory.EnableProbation(_)))
```

adleong:
Maybe worth adding a test to LoadBalancerTest to test the presence of the param.

docs/config.md
rmars: Sure! It is a long name as is.

docs/config.md
rmars: yep! thanks!!!

docs/config.md
rmars: Though, if we left it, it would match finagle....

adleong: yeah, matching finagle is nice.  but "enabled" on a boolean is pretty dumb and redundant.  *shrug* I don't feel strongly.

adleong: just noticed this says `probationEnabled` but it's `enableProbation` elsewhere

linkerd/core/src/main/scala/io/buoyant/linkerd/LoadBalancerConfig.scala
rmars: I was trying to be explicit about defaulting to true, but I am totally open to changing it.

docs/config.md
rmars: Yeah, I totally agree with you re: boolean and that name... right now I'm still leaning towards matching finagle, lmk if you have a stronger opinion :)

olix0r:
:star:

adleong:
:star: :ship:

PR #180, opened 2016-03-22
Move delegate handler and related code to i.b.admin for easier reuse"

adleong:
Some minor refactoring to make DelegateHandler and related code independent of linkerd.


admin/src/main/scala/io/buoyant/admin/StaticFilter.scala
```diff
@@ -0,0 +1,30 @@
...
+    "js" -> MediaType.Javascript,
+    "json" -> MediaType.Json,
+    "png" -> MediaType.Png,
+    "svg" -> "image/svg+xml",
+    "txt" -> MediaType.Txt
+  )
+
+  def apply(req: Request, svc: Service[Request, Response]) = {
+    svc(req).flatMap { res =>
+      val ext = req.path.split('.').lastOption.getOrElse("")
```
gtcampbell: I don't think this is correct - `"/foo/bar".split(".").lastOption` is `Some("/foo/bar")`. Can you switch this to use `req.fileExtension`?

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -4,12 +4,24 @@ import com.twitter.app.App
...
 
 class LinkerdAdmin(app: App, linker: Linker, config: LinkerConfig) extends Admin(app) {
 
+  private[this] val dtabs: Future[Map[String, Dtab]] =
+    Future.value(
+      linker.routers.map { router =>
+      val RoutingFactory.BaseDtab(dtab) = router.params[RoutingFactory.BaseDtab]
+      router.label -> dtab()
+    }.toMap
+    )
```
gtcampbell: lines 20-22 should be indented one more level, I think

gtcampbell:
LGTM other than my two cleanup comments

olix0r:
:star:

PR #189, opened 2016-03-23
Add a skeleton endpoint to access new dashboard at /dashboard

### LINKED ISSUE #184: [dashboard] set up a skeleton endpoint for dashboard v2
Part of #183 

This should leave the current dashboard intact, but allow us to develop / test the new one in parallel.

Set up a `/dashboardV2` in linkerdAdminRoutes
Code up something similar to SummaryHandler

Link to this page from the "old" dashboard, perhaps with "Try Beta!" or some other wording.  Should be able to return to original dashboard from new dashboard.

rmars:
So we can start work on the new dashboard.

* Added basic js and css files as starting points/entry points - copied over basics from summary.js
* Copied over SummaryHandler as a base for the new dashboard.  
* We don't use Stat/statsHtml right now anywhere, but the designs have those stats at the bottom of the page, so I added it now.  I can remove/simplify it, right now it's wholesale from SummaryHandler.

Fixes #184

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
@@ -1,0 +1,63 @@
+"use strict";
+
+/**
+ * Number of millis to wait between data updates.
+ */
+var UPDATE_INTERVAL = 1000;
+var OVERVIEW_STATS_DESCRIPTION = {
+  stats: [
+    { description: "uptime", dataKey: "jvm/uptime",  elemId: "jvm-uptime", value: "0s" },
+    { description: "thread count", dataKey: "jvm/thread/count", elemId: "jvm-thread-count", value: "0" },
+    { description: "memory used", dataKey: "jvm/mem/current/used", elemId: "jvm-mem-current-used", value: "0MB" },
+    { description: "gc", dataKey: "jvm/gc/msec",  elemId: "jvm-gc-msec", value: "1ms" }
+  ]
+}
+
+$.when(
+  $.get("/files/template/overview_stats.template"),
+  $.get("/admin/metrics.json")
+).done(function(overviewStatsRsp, metricsJson) {
+  appendOverviewSection();
+
+  $(function() {
+    var dashboard = Dashboard();
+    dashboard.start(UPDATE_INTERVAL);
+  });
+
+  function appendOverviewSection() {
+    var overviewTemplate = Handlebars.compile(overviewStatsRsp[0]);
+    var compiledHtml = overviewTemplate(OVERVIEW_STATS_DESCRIPTION);
+
+    var $overviewStats = $('<div />').html(compiledHtml);
+    $overviewStats.appendTo("body");
+  }
+});
+
+var Dashboard = (function() {
+
+  function render(data) {
+    var json = $.parseJSON(data);
+    $(".test-div").text("Fill in content.");
+  }
+
+  /**
+   * Returns a function that may be called to trigger an update.
+   */
+  return function() {
+
+    function update() {
+      $.ajax({
+        url: "/admin/metrics.json",
+        dataType: "text",
+        cache: false,
+        success: function(metrics) {
+          render(metrics);
+        }
+      });
+    }
+
+    return {
+      start: function(interval) { setInterval(update, interval); }
+    };
+  };
+})();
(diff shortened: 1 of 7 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -0,0 +1,45 @@
...
+
+$.when(
+  $.get("/admin/metrics.json")
+).done(function(metricsJson) {
+
+  $(function() {
+    var dashboard = Dashboard();
+    dashboard.start(UPDATE_INTERVAL);
+  });
+});
```
siggy: just confirming that we'll eventually seed the initial page load with this first call to metrics.json, rather than "fill in content" after a second call?

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -0,0 +1,45 @@
...
+   */
+  return function() {
+
+    function update() {
+      $.ajax({
+        url: "/admin/metrics.json",
+        dataType: "text",
+        cache: false,
+        success: function(metrics) {
+          render(metrics);
```
siggy: just curious: why not do ```success: render``` ?

siggy:
:star: :+1: just a couple questions but otherwise a great start!

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
rmars: correct! that's my intent.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
rmars: ah! eventually I wanted it to call various update functions on success, eg calling a render method on various dashboard components.  Also, calling routers.update(metrics); seemed to make more sense from here than from the render() method inside the dashboard.  I could just move those function calls to render() though.

siggy: that makes sense, agree `routers.update(metrics)` wouldn't make sense inside a `render()` method.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
```diff
@@ -0,0 +1,62 @@
...
+        Stat("memory used", "jvm-mem-current-used", "0MB", "", "jvm/mem/current/used"),
+        Stat("gc", "jvm-gc-msec", "1s", "", "jvm/gc/msec")
+      ).map { stat =>
+          s"""
+        <li data-key="${stat.dataKey}">
+          <strong class="stat-label ${stat.style}">${stat.description}</strong>
+          <span id="${stat.elemId}" class="stat">${stat.value}</span>
+        </li>
+        """
+        }.mkString("\n")
```
esbie: Could you move these stats to be rendered on the client using handlebars instead of the server? We want to head towards client-side html rendering for most parts of the admin page

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
```diff
@@ -0,0 +1,62 @@
...
+private object DashboardHandler {
+  private[this] case class Stat(
+    description: String,
+    elemId: String,
+    value: String,
+    style: String,
+    dataKey: String
+  )
+
+  def summaryHtml(routerCount: Int) = {
```
esbie: nit: rename to dashboardHtml?

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
```diff
@@ -0,0 +1,62 @@
...
+        <hr>
+        <div class="row text-center test-div">
+        </div>
+        <hr>
+      """,
+      csses = Seq("dashboard.css"),
+      javaScripts = Seq("dashboard.js")
+    )
+  }
+}
```
esbie: nit: this file could use a return at the end

esbie:
looks good so far!

esbie:
:+1: awesome thank you!

klingerf:
:star: :white_flower: Nice -- looks good to me.

PR #191, opened 2016-03-24
Move ProcInfo to facilitate code sharing

rmars:
Moved the module ProcInfo into its own file so that the code can be reused for the revamped dashboard.

The only thing I've changed in ProcInfo is including refreshUri in the file rather than getting it from the html

Branched off of #184 so will need to rebase off master when that hits.


rmars:
Note: as soon as #189 passes CI I'm going to merge it into master, and then put this back up again with parent master.  (Since I can't change the parent of this review once it's posted).

rmars:
Rebased off master: see #192

PR #196, opened 2016-03-25
Separate delegator UI from loading logic

esbie:
This review sets the stage from using the delegator code in other
pages besides /delegator

It's a basic reshuffling of html into a separate template and js into a separate module


klingerf:
:star: :sheep: Cool -- looks good to me.

rmars:
:star: :ghost:

PR #197, opened 2016-03-26
Add a metrics collector to query for dashboard data

### LINKED ISSUE #195: [dashboard] Write a metrics module to query for all the data we need for the dashboards
We make very many calls to /metrics to populate the dashboards.  It would be nice to reduce the number of calls we make to this backend (we'd reduce our memory usage too!)

The module will request all the data we need and then pass it to the various view modules on the dashboard page.  Each view will just pick out the parts relevant to it.

We can have everyone interested in metrics register themselves with the metrics collector, and every time the metrics collector gets new metrics, it can push these to interested parties.
Another option would be to register listeners and trigger some event when new metrics are fetched.

Part of #183

rmars:
Fixes #195 

  A module to consolidate our backend metric requests. Collects all metrics that
  we need and gets them in two requests - one to metrics.json and one to metrics
  with the desired params.

  Register a listener to receive metric updates. Each listener needs to implement:
    - onMetricsUpdate(data) - to handle new incoming data
      data is of the form:
      {
        general: {}, // data obtained from /metrics.json
        specific: {} // data obtained from /metrics?m=...
      }
    - desiredMetrics() - returns a list of metrics the listener wants

general and specific probably aren't the best names, suggestions welcome!

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
@@ -11,11 +11,13 @@ $.when(
 ).done(function(overviewStatsRsp, metricsJson) {
   appendOverviewSection();
   var procInfo = ProcInfo();
+  var dashboard = Dashboard();
+
+  var metricsListeners = [procInfo, dashboard];
+  var metricsCollector = MetricsCollector(metricsListeners);
 
   $(function() {
-    var dashboard = Dashboard();
-    procInfo.start(UPDATE_INTERVAL);
-    dashboard.start(UPDATE_INTERVAL);
+    metricsCollector.start(UPDATE_INTERVAL);
   });
 
   function appendOverviewSection() {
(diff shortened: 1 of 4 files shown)


esbie:
This looks solid. Since clients don't start sending metrics until they're used, I'm pretty sure we'll need to allow registering/deregistering for metrics updates after initialization. Something like

function registerListener(listenerFn, desiredMetricsArr)
function deregisterListener(listenerFn)

esbie:
:star: cool thanks

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -11,11 +11,13 @@ $.when(
 ).done(function(overviewStatsRsp, metricsJson) {
   appendOverviewSection();
   var procInfo = ProcInfo();
+  var dashboard = Dashboard();
+
+  var metricsListeners = [procInfo, dashboard];
+  var metricsCollector = MetricsCollector(metricsListeners);
```
klingerf: Do you need the metricsListeners var here?  Could it just be `var metricsCollector = MetricsCollector([procInfo, dashboard]);`?

klingerf:
:star: :honey_pot: Cool, I like this approach a lot.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
rmars: Oh sure! I just wanted to be clear what they were, so people adding future modules would know explicitly what those were.  Would you like me to update this to not have the var?

klingerf: That reasoning makes sense -- totally fine to leave as is.

PR #199, opened 2016-03-28
Delegate page for dashboard v2

esbie:
New route  `/dtab/http` vs. the old `/delegator?router=http`. MVP of delegate UI against dark bg (it's still ugly... maybe it'll get some designer time). Fixes a bug with reliance on globally defined templates var.

screen shot 2016-03-28 at 1 21 53 pm

### DIFF
--- a/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DtabHandler.scala
+++ b/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DtabHandler.scala
@@ -1,0 +1,47 @@
+package io.buoyant.linkerd.admin
+
+import com.twitter.finagle.http.{Request, Response, Status}
+import com.twitter.finagle.{Dtab, Service}
+import com.twitter.util.Future
+import io.buoyant.linkerd.admin.names.DelegateApiHandler
+
+private[admin] class DtabHandler(
+  dtabs: () => Future[Map[String, Dtab]]
+) extends Service[Request, Response] {
+
+  import DtabHandler._
+
+  override def apply(req: Request): Future[Response] = dtabs().flatMap { dtabMap =>
+    req.path match {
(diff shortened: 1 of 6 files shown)


adleong:
Is it possible for this to be moved to `io.buoyant.admin` instead of `io.buoyant.linkerd.admin`?  Eventually we'd want to be able to use this in namerd.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DtabHandler.scala
```diff
@@ -0,0 +1,47 @@
...
+
+private[admin] class DtabHandler(
+  dtabs: () => Future[Map[String, Dtab]]
+) extends Service[Request, Response] {
+
+  import DtabHandler._
+
+  override def apply(req: Request): Future[Response] = dtabs().flatMap { dtabMap =>
+    req.path match {
+      case rexp(routerName) => dtabMap.get(routerName) match {
```
adleong: What do you think about calling these "namespace" instead of "routerName"?  In the context of linkerd each router has one namespace so it amounts to the same thing, but using "namespace" probably makes more sense if this is to be reused by namerd.

esbie:
@adleong the v1/v2 banner distinction is in `io.buoyant.linkerd.admin`, so it's not an easy move into `io.buoyant.admin`.  I don't think it's a big issue though, DtabHandler.scala is a short file to cp to namerd

adleong:
Makes sense!

rmars:
I think our error message colours need to be updated too:
screen shot 2016-03-28 at 2 53 25 pm
screen shot 2016-03-28 at 2 53 30 pm

rmars:
:star: :grapes: yay

klingerf:
:star: :rocket: Works for me.

siggy:
:star: :+1: lgtm

PR #200, opened 2016-03-28
boundPath tls matchError bug fix and regression test

esbie:


linkerd/tls/src/main/scala/io/l5d/clientTls/boundPath.scala
```diff
@@ -32,6 +32,7 @@ case class boundPath(caCertPath: Option[String], names: Seq[NameMatcherConfig],
       } match {
         case None if (strict.getOrElse(true)) =>
           throw new MatcherError(s"Unable to match ${path.show} with available names: ${(names.map(_.prefix)).mkString(",")}")
+        case default => default
```
adleong: TIOLI might be cleaner to move this check down to around line 52:

```
if (commonName.isEmpty && strict.getOrElse(true)) throw ...
```

esbie: I tried that in my previous review, but the yield block never executes if commonName == None

olix0r: @adleong we'll never get to 52/53 if this returns None...  We may want to figure out a better way to do this moving forward (so that we're not just throwing an exception during stack construction) but for now I'm happy just to fix this as-is.

olix0r:
:star: thanks, @esbie!

linkerd/tls/src/main/scala/io/l5d/clientTls/boundPath.scala
adleong: Oh.  Good point.

Could do 
```
val commonName = names.map ...
if if (commonName.isEmpty && strict.getOrElse(true)) throw ...
commonName
```
to avoid the awkward match if you wanted.  But this is fine too.

adleong:
:star: :headphones:

linkerd/tls/src/test/scala/io/l5d/clientTls/boundPathTest.scala
```diff
@@ -7,10 +7,21 @@ import io.buoyant.linkerd.TlsClientInitializer
 import org.scalatest.FunSuite
 
 class boundPathTest extends FunSuite {
-  test("sanity") {
+  test("static sanity") {
     static("hello", None).tlsClientPrep
   }
```
klingerf: Hmm, I don't think this test even belongs in this file. It appears to be defined in two places:

```
$ grep -r 'static("hello", None).tlsClientPrep' ./*
./linkerd/tls/src/test/scala/io/l5d/clientTls/boundPathTest.scala:    static("hello", None).tlsClientPrep
./linkerd/tls/src/test/scala/io/l5d/clientTls/staticTest.scala:    static("hello", None).tlsClientPrep
```

Can you just remove it from here?

klingerf: And then you can call the test you're adding "sanity".

klingerf:
:star: :flags: Thanks for updating.

PR #201, opened 2016-03-28
Add test that bound id will bind to itself

adleong:
We always assume that the `id` field of a `Bound.Name` is a `Path` that resolves to the same `Bound.Name`.  
 
* Add a test mixin that adds an assertion to check that this is true
* Add tests that exercise this assertion in the namers that have tests (fs and serversets)

### DIFF
--- a/namer/core/src/test/scala/io/buoyant/namer/NamerTestUtil.scala
+++ b/namer/core/src/test/scala/io/buoyant/namer/NamerTestUtil.scala
@@ -1,0 +1,26 @@
+package io.buoyant.namer
+
+import com.twitter.finagle.{Name, NameTree, Namer, Path}
+import com.twitter.util.{Await, Future}
+import org.scalatest.FunSuite
+
+trait NamerTestUtil { self: FunSuite =>
+
+  protected def assertBoundIdAutobinds(namer: Namer, path: Path, prefix: Path): Unit = {
+    def lookup(p: Path) = {
+      assert(p.startsWith(prefix))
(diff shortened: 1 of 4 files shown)


siggy:
:+1: :star: tests pass over here.

namer/core/src/test/scala/io/buoyant/namer/NamerTestUtil.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.twitter.util.{Await, Future}
+import org.scalatest.FunSuite
+
+trait NamerTestUtil { self: FunSuite =>
+
+  protected def assertBoundIdAutobinds(namer: Namer, path: Path, prefix: Path): Unit = {
+    def lookup(p: Path) = {
+      assert(p.startsWith(prefix))
+      val pathWithoutPrefix = p.drop(prefix.size)
+      Await.result(namer.lookup(pathWithoutPrefix).values.toFuture.flatMap(Future.const))
```
olix0r: would `namer.lookup(pathWithoutPrefix).sample()` work?

olix0r:
:star: small q, but lgtm

namer/core/src/test/scala/io/buoyant/namer/NamerTestUtil.scala
adleong: that would work in all the tests so far, but would fail for any Activity that starts out as pending.

olix0r: orite

PR #204, opened 2016-03-29
Add module to display summary of router stats.

### LINKED ISSUE #187: [dashboard] write a summary component
Part of #183 

Write a summary component containing prominent metrics (requests, connections, success/fail rate)

This is used in a bunch of components (the top header, plus every subsection).  
Maybe have customizable vertical divider between metrics.

rmars:
As esbie mentioned, there's a lot of data manipulation here. I tried to separate it so it would be less confusing.

* added a RouterController which will handle getting data and distributing it to the modules which will do further processing on the data pertinent to them, and then render the result.
* also added a skeletal implementation of RouterServers to demonstrate how I think RouterController could be used.
* I didn't focus too much on the styling/grid layout of this.  I figure we can get the js logic in first, and style later.
* this uses getSelectedRouter() to figure out whether we've selected a router. Until we add the select router dropdown to the new navbar, this defaults to all routers

screen shot 2016-03-29 at 2 45 54 pm

Fixes #187

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/css/dashboard.css
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/css/dashboard.css
@@ -74,3 +74,49 @@ nav ul.navbar-nav {
 .dentry .dentry-part:hover {
   background-color: #222;
 }
+
+/*
+* Router summary
+*/
+
+.router-summary {
+  clear: both;
+}
+
+.router-summary div {
+  padding: 10px;
+}
+
+.router-header {
+  float: left;
+  text-align: left;
+}
+
+.router-data {
+  float: right;
+}
+
+.router-data span {
+  float: left;
+  margin: 0 30px;
+  text-align: right;
+}
+
+
+.router-summary dt {
+  font-size: 14px;
+}
+
+.router-summary dd {
+  font-size: 30px;
+  line-height: 37px;
+}
+
+/*
+* Router server
+*/
+
+.router-server {
+  text-align: left;
+  padding: 0 10px;
+}
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
@@ -6,19 +6,27 @@
 var UPDATE_INTERVAL = 1000;
 
 $.when(
+  $.get("/files/template/router_container.template"),
+  $.get("/files/template/router_summary.template"),
   $.get("/files/template/process_info.template"),
   $.get("/files/template/request_totals.template"),
   $.get("/admin/metrics.json")
-).done(function(overviewStatsRsp, requestTotalsRsp, metricsJson) {
+).done(function(routerContainerRsp, routerSummaryRsp, overviewStatsRsp, requestTotalsRsp, metricsJson) {
   appendOverviewSection();
 
+  var selectedRouter = getSelectedRouter(); // TODO: update this to avoid passing params in urls #198
   var routers = Routers(metricsJson[0]);
+  var routerTemplates = {
+    summary: Handlebars.compile(routerSummaryRsp[0]),
+    container: Handlebars.compile(routerContainerRsp[0])
+  }
 
   var procInfo = ProcInfo();
   var dashboard = Dashboard();
   var requestTotals = RequestTotals($(".request-totals"), Handlebars.compile(requestTotalsRsp[0]), routers);
+  var routerDisplays = RouterController(selectedRouter, routers, routerTemplates, $(".dashboard-container"));
 
-  var metricsListeners = [procInfo, dashboard, requestTotals];
+  var metricsListeners = [procInfo, dashboard, requestTotals, routerDisplays];
   var metricsCollector = MetricsCollector(metricsListeners);
 
 
@@ -36,6 +44,7 @@ $.when(
   }
 
   function getOverviewStatsData() {
+    // todo: move to process_info.js after old dashboard is killed #198
     var buildVersion = $(".server-data").data("linkerd-version");
 
     return {
@@ -53,7 +62,6 @@ $.when(
 var Dashboard = (function() {
 
   function render(data) {
-    $(".test-div").text("Fill in content.");
   }
 
   return function() {
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/process_info.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/process_info.js
@@ -19,7 +19,7 @@ var ProcInfo = (function() {
 
   function render(data) {
     _(data).each(function(obj) {
-      var id = obj.name.replace(/\//g, "-");
+      var id = obj.name.replace(/[/$.]/g, "-");
       var value = pretty(obj.name, obj.value);
       $("#"+id).text(value);
     });
@@ -49,7 +49,7 @@ var ProcInfo = (function() {
     }
 
     return {
-      start: function(interval) { setInterval(update, interval); }, // TODO: remove once linkerd#183 is complete
+      start: function(interval) { setInterval(update, interval); }, // TODO: #198 remove once linkerd#183 is complete
       onMetricsUpdate: function(data) {
         render(data.specific);
       },
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_server.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_server.js
@@ -1,0 +1,28 @@
+var RouterServer = (function() {
+  // Skeletal implementation to demonstrate functionality of RouterController
+
+  function initializeServerContainers(servers, $serverEl) {
+    _.each(servers, function(server) {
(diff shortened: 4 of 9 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -60,3 +69,181 @@ var Dashboard = (function() {
...
+      .filter(isRouter)
+      .groupBy(getRouterName)
+      .map(function(data, router) {
+        var result = { router: router };
+        _.each(data, function(datum) {
+          var keyParts = datum.name.split("/");
+          var metric = keyParts[keyParts.length - 1];
+
+          result[metric] = datum.value;
+          result[metric + "Delta"] = datum.delta;
```
rmars: I think for the summaries we actually wanted to display the delta.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -60,3 +69,181 @@ var Dashboard = (function() {
     };
   };
 })();
+
+var RouterController = (function () {
```
esbie: I think these modules should go in their own files

rmars: Totally agree

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -60,3 +69,181 @@ var Dashboard = (function() {
...
+      containers[router.router] = $routerEl;
+      return containers;
+    }, {});
+
+    _.each(domContainers, function($container, routerName) {
+      $parentContainer.append($container);
+    });
+
+    return domContainers;
+  }
```
esbie: this would be easier to read as a handlebars template eg
```
{{#each router}}
 <div class="router router-{{router}}">
    <div class="summary"></div>
    <div class="server"></div>
</div>
{{/each}}
```

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -60,3 +69,181 @@ var Dashboard = (function() {
...
+
+var RouterController = (function () {
+  // Contains all the router components, e.g. summaries, graphs, etc
+
+  function initializeRouterContainers(selectedRouter, routers, $parentContainer) {
+    var routerData = !selectedRouter ? routers.data : { router: routers.data[selectedRouter] };
+
+    var domContainers = _.reduce(routerData, function(containers, router) {
+      var $routerEl = $("<div />").addClass("router router-" + router.router);
+      $routerEl.append($("<div />").addClass("summary"));
```
esbie: fyi `$("<div class='foo'>")` is totally valid jquery

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -60,3 +69,181 @@ var Dashboard = (function() {
...
+    metricsParams = metricsParams.concat(clientsToMetricParam(clients));
+  }
+
+  function clientsToMetricParam(clients) {
+    return _(clients).map(function(client) {
+      return _.map(summaryKeys, function(key) {
+        return client.prefix + key;
+      });
+    }).flatten().value();
+  }
```
esbie: yup yup very similar to my branch

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -60,3 +69,181 @@ var Dashboard = (function() {
...
+      if (!summaryEls[stats.router]) {
+        var $summary = $("<div />").html(summaryHtml);
+        var $router = $(routerEls[stats.router].find(".summary")[0]);
+
+        summaryEls[stats.router] = $summary;
+        $summary.appendTo($router);
+      } else {
+        summaryEls[stats.router].html(summaryHtml);
+      }
+    });
```
esbie: I think this method could be a lot simpler if RouterSummary was responsible for rendering a single summary instead of a list of summaries. In a single summary scenario, the RouterController parent could pass in a $summaryEl to render the template in & there would be no need for the `summaryEls` mapping

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
esbie: this is my understanding as well

esbie:
This looks great, I learned a lot from reading it! couple suggestions above

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
```diff
@@ -60,3 +69,181 @@ var Dashboard = (function() {
...
+  function getFailureRate(datum) {
+    // there's some discussion as to whether we should include both success
+    // and failure rate.  this is a very sketchy implementation of this until
+    // we decide for sure
+    if (datum.requests === 0) {
+      return "N/A";
+    } else {
+      return (100*datum.failures/datum.requests).toFixed(2) + "%";
+    }
+  }
```
siggy: there are some bits to do similar things with success rates in utils.js. consider re-using or consolidating?

rmars: I didn't want to invest too much time in consolidating this code because I'm 90% sure we're not going to use failure rates... however I wanted failure rates to be in here so we could see how it looked before making a decision.  Happy to consolidate/refactor, lmk what you think.  

I should add a todo to remove this also.

siggy: no no, that's all good. look forward to all the forthcoming code deletion!

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/dashboard.js
rmars: yeah, this simplifies the code a lot! thanks!

rmars:
screen shot 2016-03-30 at 3 36 24 pm

esbie:
:star: looks very nice, good work! :cherry_blossom: :deciduous_tree:

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/css/dashboard.css
```diff
@@ -74,3 +74,53 @@ nav ul.navbar-nav {
...
+  text-align: left;
+}
+
+.router-data {
+  float: right;
+}
+
+.router-data span {
+  float: left;
+  margin: 0px 30px;
```
klingerf: Tiny tiny nit: `margin: 0 30px;` is more consistent with 0-values throughout the rest of this file.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/process_info.js
```diff
@@ -19,7 +19,7 @@ var ProcInfo = (function() {
 
   function render(data) {
     _(data).each(function(obj) {
-      var id = obj.name.replace(/\//g, "-");
+      var id = obj.name.replace(/\/|\$/g, "-");
```
klingerf: Given that we might have more characters that we eventually want to replace here, I think it's more concise to use a bracket expression:

    var id = obj.name.replace(/[/$]/g, "-");

And that said, how about also escaping "."? I know that sometimes causes problems for jquery selectors. 

    var id = obj.name.replace(/[/$.]/g, "-");

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_server.js
```diff
@@ -0,0 +1,28 @@
+var RouterServer = (function() {
+  // Skeletal implementation to demonstrate functionality of RouterController
+
+  function initializeServerContainers(servers, $serverEl) {
```
klingerf: This just stems from me not knowing javascript -- why are some function parameters prefixed with "$" and not others?  `servers` vs `$serverEl` etc.?  Is that to indicate that serverEl is a fully extended object?

klingerf:
:star: :bike: Looks good! My comments were totally optional.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_server.js
rmars: Ah yes! It's a naming convention - if it is a jquery object, I name it $variable.  I don't know if we actually use this convention in this codebase.  Should I change it?

esbie: I like this convention, so I vote we keep it

klingerf: Cool, yeah, totally fine by me. Just wanted to make sure I understood what was up.

PR #211, opened 2016-03-30
Run Validator in circleci

adleong:
Unfortunately, the head version is hardcoded in `validator.sh` and will need to be updated on each release.  Open to suggestions.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -24,7 +24,7 @@ test:
   override:
   - ./sbt coverage test
   - ./sbt integration:compile e2e:test
-  - ./sbt assembly && ./test-config.sh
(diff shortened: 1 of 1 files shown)


klingerf:
:star: :green_heart:

olix0r:
:-1: Unfortunately, I think we are going to have _a lot_ of problems running this in CI.  It isn't really designed to be a good barometer for correctness -- it's prone to plenty of timing/raciness issues that will make this a poor indicator of regressions, especially in a CI environment with variable performance.

olix0r:
Also, I think we should replace the `validator.sh` script with sbt logic that ensures that the artifacts are built, etc...  We should only have scripts that call out to sbt if we _really_ have to.

adleong:
punting for now until we come up with a more reliable version of Validator

PR #215, opened 2016-03-30
Query util for clientside metrics gathering

esbie:
While routers.js does a great job of creating a hierarchical view
of routers, servers, and clients, we're finding an increasing need
to make requests for metrics across these boundaries, which results in
a lot of mapping/filtering that is not always very readable.

This change introduces a module that builds a regex for querying
across a list of metrics. Simplifies request_totals as a result.

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/query.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/query.js
@@ -1,0 +1,112 @@
+/**
+ * Util for building regexes that map over specifically formatted metric names.
+ * `Query.clientQuery().build()` matches all possible client metrics.
+ *  Use defined helpers to make the query more specific.
+ */
+
+var Query = function() {
+  function escape(string) {
+    return string.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
+  }
+
+  function toR(arr) {
+    if (arr === "*" || !arr || arr.length === 0) {
+      return "(.*)";
+    } else {
+      return "(" + _.map(arr, escape).join("|") + ")";
+    }
+  }
+
+  var generalQuery = function() {
+    var query = {
+      routerLabels: [],
+      metricNames: []
+    };
+    query.allRouters = function() {
+      query.routerLabels = "*";
+      return query;
+    }
+    query.withRouter = function(router) {
+      if (_.isArray(query.routerLabels))
+        query.routerLabels.push(router);
+      return query;
+    }
+    query.withRouters = function(routers) {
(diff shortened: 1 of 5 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/request_totals.js
```diff
@@ -4,54 +4,35 @@ var RequestTotals = (function() {
   const SERVER = "server";
   const CLIENT = "client";
```
rmars: We don't need these any more

rmars:
:star: :grapes: :dart: cool!

esbie:
rebased against master

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/query.js
```diff
@@ -0,0 +1,104 @@
...
+      metricNames: []
+    };
+    query.allRouters = function() {
+      query.routerLabels = "*";
+      return query;
+    }
+    query.withRouter = function(router) {
+      query.routerLabels = [router];
+      return query;
+    }
```
klingerf: Would it make sense to append to the routerLabels array here, rather than replace?  I could imagine somebody calling:

    Query.clientQuery().withRouter("foo").withRouter("bar")

And expecting that to match both routers.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/query.js
```diff
@@ -0,0 +1,104 @@
...
+    query.withRouters = function(routers) {
+      query.routerLabels = routers;
+      return query;
+    }
```
klingerf: Similar comment here.

klingerf:
I like this a lot. Just had that one comment about chained calls.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/query.js
esbie: well they would be WRONG!!!

j/k

I guess I can change it, it's not gonna be as pretty though

klingerf:
@esbie Thanks for updating! Am surprised that you need the `_.isArray` check throughout. What if you dropped support for the "*" character, and just reset to empty array in that case?  Then could you assume that `metricNames` and `routerLabels` are always arrays and skip the check? This is a big gray area of js for me, so apologies if I'm on the wrong track.

klingerf:
:star: :seedling: Looks good to me -- thanks for updating.

PR #217, opened 2016-03-31
Add zipkin tracer initializer

klingerf:
Adding a new trace initializer for configuring Finagle's zipkin-tracer via a linkerd config file. Let me know if this is the appropriate place to put this code within the project or if it should go elsewhere.

### DIFF
--- a/linkerd/tracer/zipkin/src/main/scala/io/l5d/zipkin.scala
+++ b/linkerd/tracer/zipkin/src/main/scala/io/l5d/zipkin.scala
@@ -1,0 +1,32 @@
+package io.l5d
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.tracing.{Record, TraceId, Tracer}
+import com.twitter.finagle.zipkin.thrift.{Sampler, ZipkinTracer}
+import io.buoyant.linkerd.{TracerConfig, TracerInitializer}
+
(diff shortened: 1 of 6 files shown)


esbie:
lgtm

linkerd/tracer/zipkin/src/main/scala/io/l5d/zipkin.scala
```diff
@@ -0,0 +1,34 @@
...
+  port: Option[Int],
+  sampleRate: Option[Double]
+) extends TracerConfig {
+
+  @JsonIgnore
+  override def newTracer(): Tracer = new Tracer {
+
+    private[this] def getHost: String = host.getOrElse("localhost")
+    private[this] def getPort: Int = port.getOrElse(9410)
+    private[this] def getSampleRate: Float =
```
olix0r: tioli: i don't think it adds much value to have `def getFoo` given that these just get called once, immediately.  Could even inline them into ZipkinTrace.mk...

nbd one way or the other

olix0r:
:star: lgtm, one tioli

adleong:
:star: :memo: :penguin:

linkerd/tracer/zipkin/src/main/scala/io/l5d/zipkin.scala
klingerf: Good suggestion, have updated and inlined them.

PR #224, opened 2016-04-01
Move metrics collector registration to be private

esbie:
Going forward I'd like each module to set up its own listening, querying, & filtering behavior. In order to achieve this I'm passing `metricsCollector` in the init of each module (previously it was the opposite: modules were passed into the collector).

I also introduce `Query.find` and `Query.filter` to simplify data transformations

Also removed Dashboard as it looked unused

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/metrics_collector.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/metrics_collector.js
@@ -2,49 +2,41 @@
   A module to consolidate our backend metric requests. Collects all metrics that
   we need and gets them in two requests - one to metrics.json and one to metrics
   with the desired params.
-
-  Register a listener to receive metric updates. Each listener needs to implement:
-    - onMetricsUpdate(data) - to handle new incoming data
-      data is of the form:
-      {
-        general: {}, // data obtained from /metrics.json
-        specific: {} // data obtained from /metrics?m=...
-      }
-    - desiredMetrics() - returns a list of metrics the listener wants
 */
 var MetricsCollector = (function() {
   var generalUpdateUri = "/admin/metrics.json";
   var metricsUpdateUri = "/admin/metrics?";
   var listeners = [];
 
-  function url(listeners) {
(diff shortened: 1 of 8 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/metrics_collector.js
```diff
@@ -17,34 +17,25 @@ var MetricsCollector = (function() {
...
-      console.error("This listener needs to implement onMetricsUpdate and desiredMetrics");
-      return false;
-    }
-    return true;
-  }
-
-  function registerListener(listener) {
-    validateListener(listener);
-    listeners.push(listener);
+  function registerListener(handler, metrics) {
```
rmars: Can you update (or delete :smile:) the comments at the top of this file?

rmars:
:star: :ship: :baby_chick: 
Thanks for removing the dashboard code! I was thinking about doing that too.  And the controller code looks much better.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/metrics_collector.js
esbie: omg thank you for reminding me!

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/query.js
```diff
@@ -68,7 +68,7 @@ var Query = function() {
     }
     q.withClients = function(clients) {
       if (_.isArray(query.clientLabels))
-        query.clientLabels.concat(clients);
+        query.clientLabels = query.clientLabels.concat(clients);
       return q;
     }
```
siggy: i don't see `withRouters`, `allMetrics`, `withClient`, `withClients` used. are these coming?

esbie: they're there for library completeness/symmetry, I expect they'll be used at some point

siggy:
:+1: :star:

PR #225, opened 2016-04-01
recursively create missing zk paths on dtab create

siggy:
if the zk pathPrefix does not exist, namerd will fail. this recursively creates pathPrefix when the user attempts to add a new dtab.

considered using `creatingParentsIfNeeded()`, but that requires the curator api:
https://github.com/twitter/finagle/blob/develop/finagle-benchmark/src/main/scala/com/twitter/finagle/serverset2/LocalServerSetService.scala#L101

### DIFF
--- a/namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/ZkDtabStore.scala
+++ b/namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/ZkDtabStore.scala
@@ -44,11 +44,31 @@ class ZkDtabStore(
     children <- zk.getChildrenOf(zkPrefix)
   } yield children.children.toSet
 
+  private[this] def ensurePath(path: String) {
+    val components = path.split('/')
+    if (components.length > 2) {
+      ensurePath(components.dropRight(1).mkString("/"))
+    }
+
+    writerConnect().flatMap { zkw =>
+      zkw.create(
+        path,
+        None,
+        Seq(Data.ACL.AnyoneAllUnsafe),
+        CreateMode.Persistent
+      ).rescue {
+          case KeeperException.NodeExists(_) => Future.Unit
+        }
+    }.unit
+  }
+
   def list(): Future[Set[String]] = actNs.values.toFuture.flatMap(Future.const)
 
   def create(ns: String, dtab: Dtab): Future[Unit] = {
     val path = s"$zkPrefix/$ns"
     log.info(s"Attempting to create dtab at $path")
+
+    ensurePath(zkPrefix)
 
     writerConnect().flatMap { zkw =>
       zkw.create(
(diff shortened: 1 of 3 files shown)


namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/ZkDtabStore.scala
```diff
@@ -44,12 +44,32 @@ class ZkDtabStore(
     children <- zk.getChildrenOf(zkPrefix)
   } yield children.children.toSet
 
+  private[this] def ensurePath(path: String) {
+    val components = path.split('/')
+    if (components.length > 2) {
+      ensurePath(components.dropRight(1).mkString("/"))
+    }
```
esbie: code golf! you could alternatively use finagle paths (tioli)
```
val components = Path.read(path)
if (components.size() > 2) {
  ensurePath(components.take(components.size()-1).show)
}
```

esbie:
:star: makes sense to me

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/ZkDtabStore.scala
```diff
@@ -44,6 +44,22 @@ class ZkDtabStore(
...
+
+    zkw.create(
+      path,
+      None,
+      Seq(Data.ACL.AnyoneAllUnsafe),
+      CreateMode.Persistent
+    ).rescue {
+        case KeeperException.NodeExists(_) => Future.Unit
+      }.unit
```
olix0r: better:
```scala
...unit.handle { case KeeperException.NodeExists(_) => }
```

i'm a little surprised this compiles given that create returns one type and rescue returns another...

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/ZkDtabStore.scala
```diff
@@ -44,6 +44,22 @@ class ZkDtabStore(
...
+  private[this] def ensurePath(zkw: ZooKeeperWriter, path: String) {
+    val components = Path.read(path)
+    if (components.size > 1) {
+      ensurePath(zkw, components.take(components.size - 1).show)
+    }
+
+    zkw.create(
+      path,
+      None,
+      Seq(Data.ACL.AnyoneAllUnsafe),
```
olix0r: not a blocker, this is obviously not production-ready.  Creating files with global-writable is dangerous for any real usage.  We'll have to revisit permissions and privileges before we tell anyone to use this outside of experimental use.  Maybe worth putting a note in the docs to this effect.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/ZkDtabStore.scala
siggy: hmmm, could it be because `ensurePath` doesn't have a return type?

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/ZkDtabStore.scala
```diff
@@ -44,6 +44,20 @@ class ZkDtabStore(
     children <- zk.getChildrenOf(zkPrefix)
   } yield children.children.toSet
 
+  private[this] def ensurePath(zkw: ZooKeeperWriter, path: String) {
```
adleong: I think this should return `Future[Unit]`

siggy:
closing in favor of https://github.com/BuoyantIO/linkerd/pull/271

PR #226, opened 2016-04-01
harmonize linkerd/namerd example usage

siggy:
this allows for commands like `./sbt namerd-examples/basic:run`, similar
to linkerd.


adleong:
:star: nice, thanks for cleaning this up :curry:

PR #229, opened 2016-04-01
Allow PUT without If-Match in control-http API.  Also allow HEAD.

### LINKED ISSUE #218: namerd: PUT with no If-Match
when writing namerd clients, it's convenient to be able to write a dtab without worrying about concurrent writes.  I.e. "just upload this file".

I think we should support PUT without If-Match to enable this.  I know I had a slightly different take on this originally, but I'm starting to think this is acceptable if sufficiently documented (when namerd has api docs)

adleong:
Fixes #218

### DIFF
--- a/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
+++ b/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
@@ -120,91 +120,100 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response] 
   private[this] def getDtab(ns: String): Future[Option[VersionedDtab]] =
     storage.observe(ns).values.toFuture.flatMap(Future.const)
 
-  def apply(req: Request): Future[Response] = req.path match {
-    case DtabUri(None) =>
-      req.method match {
-        case Method.Get =>
-          storage.list().map { namespaces =>
-            val rsp = Response()
-            rsp.contentType = MediaType.Json
-            rsp.content = Json.write(namespaces)
-            rsp
-          }
-
-        case _ => Future.value(Response(Status.NotFound))
-      }
-
-    case DtabUri(Some(ns)) =>
-      req.method match {
-        case Method.Get =>
-          getDtab(ns).map {
-            case Some(dtab) =>
-              val rsp = Response()
-              val (contentType, codec) = DtabCodec.accept(req.accept).getOrElse(DtabCodec.default)
-              rsp.contentType = contentType
-              rsp.headerMap.add("ETag", versionString(dtab.version))
-              rsp.content = codec.write(dtab.dtab)
-              rsp
-
-            case _ => Response(Status.NotFound)
-          }
-
-        case Method.Put =>
-          req.contentType.flatMap(DtabCodec.byContentType) match {
-            case Some(codec) =>
-              codec.read(req.content) match {
-                case Return(dtab) =>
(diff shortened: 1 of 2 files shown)


namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -120,91 +120,100 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
+    }
 
-    // invalid uri
-    case _ => Future.value(Response(Status.NotFound))
-  }
+  private[this] def handleGetDtab(ns: String, req: Request): Future[Response] =
+    getDtab(ns).map {
+      case Some(dtab) =>
+        val rsp = Response()
+        val (contentType, codec) = DtabCodec.accept(req.accept).getOrElse(DtabCodec.default)
```
olix0r: i know you didn't write this initially, but I just realized that we're supposed to return a 406 Not Acceptable if we are given an Accept header we can't satisfy.  I think in general, the accept stuff I did is not particularly RFC compliant, so maybe best to revisit in a followup

olix0r:
:+1: looks good!

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -120,91 +120,100 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
+        rsp.contentType = contentType
+        rsp.headerMap.add("ETag", versionString(dtab.version))
```
gtcampbell: minor: finagle-http appears to have "ETag" as a constant in `Fields`; might as well use that

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -120,91 +120,100 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
+
+      case None => Response(Status.NotFound)
+    }
+
+  private[this] def handlePutDtab(ns: String, req: Request): Future[Response] =
+    req.contentType.flatMap(DtabCodec.byContentType) match {
+      case Some(codec) =>
+        codec.read(req.content) match {
+          case Return(dtab) =>
+            req.headerMap.get("If-Match") match {
```
gtcampbell: can use the constant in finagle-http here as well

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -120,91 +120,100 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
+    }
+
+  private[this] def handleHeadDtab(ns: String, req: Request): Future[Response] =
+    getDtab(ns).map {
+      case Some(dtab) =>
+        val rsp = Response()
+        rsp.headerMap.add("ETag", versionString(dtab.version))
+        rsp
+
+      case None => Response(Status.NotFound)
```
gtcampbell: tioli: get and head have a bunch of shared logic (look up dtab, 404 if not found, add etag to response) that could be refactored into a shared method

gtcampbell:
:+1: this looks fine to me, although it would probably be good to switch to the finagle-http constants as opposed to the hard-coded strings here

PR #228, opened 2016-04-01
Flesh out servers implementation

rmars:
Also add a dashboard-shared.css file for common styles.
Still holding out on styling this too much until all the js is done.
Colors coming in a branch by esbie.
Similar to esbie's banner code.

Fixes #220

screen shot 2016-04-01 at 12 21 25 pm

Issue #183, opened 2016-03-23
Dashboard revamp tracking ticket


rmars:
Revamp the linkerd dashboard page.
This ticket documents the tasks to get the new dashboard up and running (rough/beta version).
Getting ready to release a v1 is tracked in #275 

The plan:
* [x] #184 Set up a skeleton endpoint for dashboard v2.  This will include:
  + handler (scala)
  + blank html page
* [x] #184 Set up base js and css files. This will include:
  + an update function to poll for metrics
  + js app initialization/entrypoint
  + set up css files for later reference
* Data plumbing work:
  * [x] #195 Write a metrics js module to handle getting the data we need
  * [x] #215 Write a query util for clientside metrics gathering
  * [x] #224
* Write js (sub)components for the new dashboard.
    * [x] #193 a banner component that displays an overview
    * [x] graph component that displays realtime results
      * [x] #186 / #227 combined router graph
      * [x] #241 per client graph that displays latencies
    * [x] #185 a (per client) table / lengend displaying latencies
    * [x] #187 a summary component containing prominent metrics (requests, connections, success/fail rate)
    * [x] #220 write a server module (contains server summary)
    * [x] #231 write a client module (contains table and graphs, and also a summary)
* Styling - make the dashboard look like the mocks.
  * [x] #233 set up grid layout
  * [x] #242 colorize the dashboard: the colors in the overview chart must match the client colors
  * [x] #250 apply stylings from the mocks to the dashboard
* Fix bugs
  * [x] add ``failure rate || null`` to prevent NaN%s in the success rates (related: #258)

**Questions**
* What js frameworks do we want to use? backbone? react?
  + let's try to avoid dependencies as much as possible
* What do we want to use for the graphs? d3? smoothie?
  + we decided to use smoothie for now to make use of the existing UpdateableChart, and revisit in the future
* Do we want this page responsive design? The current spec is very specific pixel-wise.
  + Filed #234 

Comments/discussion on plans/technologies welcome.

rmars:
Latest design specs: 
[color_guide.pdf](https://github.com/BuoyantIO/linkerd/files/189807/color_guide.pdf)
[default_view_grid.pdf](https://github.com/BuoyantIO/linkerd/files/189808/default_view_grid.pdf)
[default_view_spec_typography.pdf](https://github.com/BuoyantIO/linkerd/files/189809/default_view_spec_typography.pdf)
[...]

PR #232, opened 2016-04-04
Add clients module to the new dashboard

### LINKED ISSUE #231: Write a client module
(Similar to servers module #228)
Write a client module that displays the list of clients of a router.
Per the design, each client module contains a summary, graph and table.  

There are separate tickets for graph and table, so this can just be the infra/summary part.

Part of #183

rmars:
(code is very similar to Servers module #228, but updates when new clients are added)

No styling has yet been applied.

Fixes #231
Part of #183

screen shot 2016-04-04 at 2 45 23 pm

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
@@ -1,0 +1,105 @@
+var RouterClient = (function() {
+  var template;
+
+  function getMetricDefinition(routerName, clientName) {
+    return [
+      {
+        description: "Requests",
+        metricSuffix: "requests",
+        query: Query.clientQuery().withRouter(routerName).withClient(clientName).withMetric("requests").build()
+      },
+      {
+        description: "Connections",
+        metricSuffix: "connections",
+        query: Query.clientQuery().withRouter(routerName).withClient(clientName).withMetric("connections").build()
+      },
+      {
+        description: "Successes",
+        metricSuffix: "success",
+        query: Query.clientQuery().withRouter(routerName).withClient(clientName).withMetric("success").build(),
+        getRate: function(data) {
+          var successRate = new SuccessRate(data.requests, data.success, data.failures);
+          return successRate.prettyRate();
+        }
+      },
+      {
+        description: "Failures",
+        metricSuffix: "failures",
+        query: Query.clientQuery().withRouter(routerName).withClient(clientName).withMetric("failures").build()
+      }
+    ];
+  }
+
+  function renderClient($container, client, data) {
+    var clientHtml = template({
+      client: client.label,
+      metrics: data
+    });
+    var $clientHtml = ($("<div />").addClass("router-client").html(clientHtml));
+
+    $container.html($clientHtml);
+  }
+
+  function processData(data, metricDefinitions) {
+    var lookup = {}; // track # requests for SuccessRate()
+
+    return _.map(metricDefinitions, function(defn) {
+      var clientData = Query.filter(defn.query, data);
+
+      if (!_.isEmpty(clientData)) {
+        defn.value = clientData[0].delta;
+        lookup[defn.metricSuffix] = defn.value;
+
+        if(_.isFunction(defn.getRate)) {
+          defn.rate = defn.getRate(lookup);
+        }
+      }
+
+      return defn;
+    });
+  }
+
+  return function (metricsCollector, routers, client, $clientEl, routerName, clientTemplate) {
+    template = clientTemplate;
+    var metricDefinitions = getMetricDefinition(routerName, client.label);
+
+    var metricsHandler = function(data) {
+      var filteredData = _.filter(data.specific, function (d) { return d.name.indexOf(routerName) !== -1 });
+      var transformedData = processData(filteredData, metricDefinitions);
+
+      renderClient($clientEl, client, transformedData);
+    }
+
+    var getDesiredMetrics = function(metrics) {
+      return _.map(metricDefinitions, function(d) {
+        return Query.filter(d.query, metrics);
+      });
+    }
+
+    metricsCollector.registerListener(metricsHandler, getDesiredMetrics);
+
+    return {};
+  };
+})();
+
+var RouterClients = (function() {
+  return function (metricsCollector, routers, $clientEl, routerName, clientTemplate) {
+    var clients = routers.clients(routerName);
+    routers.onAddedClients(addClients);
+
+    _.map(clients, initializeClient);
+
+    function initializeClient(client) {
+      var $el = $("<div />").addClass("router-client");
+      $clientEl.append($el);
+      RouterClient(metricsCollector, routers, client, $el, routerName, clientTemplate);
+    }
+
+    function addClients(addedClients) {
+      _.chain(addedClients)
+        .filter(function(client) { return client.router === routerName })
+        .each(initializeClient)
+        .value();
+    }
+  }
+})();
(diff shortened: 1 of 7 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_summary.js
```diff
@@ -64,6 +64,7 @@ var RouterSummary = (function() {
     renderRouterSummary({ router: routerName }, summaryTemplate, routerName, $summaryEl);
 
     metricsCollector.registerListener(function(data) {
+        routers.update(data.general);
```
esbie: Instead of routers being updated by one of many possible modules, could you pass metricsCollector to routers and have it register itself for updates? It's annoying, but in the end cleaner

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/template/router_container.template
```diff
@@ -2,6 +2,7 @@
   <div class="router router-{{this}}" data-router="{{this}}">
     <div class="summary"></div>
     <canvas class="router-graph" height="181"></canvas>
-    <div class='server router-servers clearfix'><div>Servers</div></div>
+    <div class="clients router-clients clearfix"><div>Clients</div></div>
+    <div class="servers router-servers clearfix"><div>Servers</div></div>
```
esbie: I don't think clearfix is necessary anymore (may need the grid branch to land first? not sure)

esbie:
looks good, just that one routers.update thing. There's a lot of duplication here, but I think it's fine to stick with rule of threes and save a possible refactor for later

rmars:
Yeah, agree about the duplication!

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/template/router_container.template
rmars: Ok! Will revisit once that branch lands, and possibly remove in a later branch.

esbie:
:star: :balloon: :balloon:

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -0,0 +1,105 @@
...
+        if(_.isFunction(defn.getRate)) {
+          defn.rate = defn.getRate(lookup);
+        }
+      }
+
+      return defn;
+    });
+  }
+
+  return function (metricsCollector, routers, client, $clientEl, routerName, clientTemplate) {
```
olix0r: noob question: what does the `$` in `$clientEl` indicate?

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_controller.js
```diff
@@ -81,11 +81,13 @@ var RouterController = (function () {
 
     _.each(routerContainerEls, function(container, router) {
       var $summaryEl = $(container.find(".summary")[0]);
-      var $serverEl = $(container.find(".server")[0]);
+      var $serverEl = $(container.find(".servers")[0]);
```
olix0r: nit: should this be `serversEl` (plural) now?  we have `$clientsEl` -- would be nice to keep them consistent

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
rmars: ah! indicates that it's a wrapped jquery object. just a convention.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_controller.js
rmars: yesss I totally just noticed this too :) :(

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
olix0r: oh, that makes sense! thanks

olix0r:
👍  lgtm (legal disclaimer: i don't have an immense amount of context here. but i can follow what this is doing)

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -0,0 +1,105 @@
+var RouterClient = (function() {
+  var template;
+
+  function getMetricDefinition(routerName, clientName) {
+    return [
+      {
+        description: "Requests",
+        metricSuffix: "requests",
+        query: Query.clientQuery().withRouter(routerName).withClient(clientName).withMetric("requests").build()
```
gtcampbell: tioli: I might cut some of the duplication by defining a baseQuery() function, i.e.

```js
function baseQuery(routerName, clientName) {
  return Query.clientQuery().withRouter(routerName).withClient(clientName);
}
```
and then you could do `baseQuery().withMetric("requests").build()` here and below. Actually, I would have expected to be able to make baseQuery a var since often I see methods named `withFoo` make copies rather than mutating. In a follow-up branch, I might suggest either changing those methods to be called `setFoo` or to have them copy the original query, but that's certainly not worth doing here.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -0,0 +1,105 @@
...
+
+    function initializeClient(client) {
+      var $el = $("<div />").addClass("router-client");
+      $clientEl.append($el);
+      RouterClient(metricsCollector, routers, client, $el, routerName, clientTemplate);
+    }
+
+    function addClients(addedClients) {
+      _.chain(addedClients)
+        .filter(function(client) { return client.router == routerName })
```
gtcampbell: Are we trying to follow the "use triple-equals everywhere" convention in this codebase? If so, this should be a triple-equals.

gtcampbell:
:star: this looks good to me; my comments are optional and/or questions.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
rmars: I think we are? I will change it.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
rmars: That seems reasonable.  I'll address in a future branch.

esbie: Disagree that `withFoo` implies copying over mutating, there's plenty of examples the other way, especially when it comes to the builder pattern:
https://www.thekua.com/atwork/2013/04/a-builder-pattern-implementation-in-javascript/
http://brainbaking.com/enhancing-the-builder-pattern-with-closures/

fwiw jQuery is the most well-known builder pattern lib in use for js, and it mutates using `.foo(bar)`

gtcampbell: True, I've seen the jQuery style too in other contexts - it was just the `with` that tripped me up here. Maybe I'm used to the usage in Finagle?

esbie: Yeah, I understand your thought process. Sounds like jQuery style could be a good compromise

Issue #219, opened 2016-03-31
namerd: Implement k8s/3rdparty namerd storage


olix0r:
Extend linkerd/k8s with the required calls to utilize the 3rdparty API.

Implement a KubernetesDtabStore that uses the 3rdparty api.

PR #239, opened 2016-04-05
Dtab delete

### LINKED ISSUE #209: namerd ctl api: DELETE
namerd ctl api should support DELETE operations

adleong:
* Add a `delete` method to the `DtabStore` interface
* Update in-memory and zk stores to implement delete
* Add a DELETE method to the http controller

Fixes #209

### DIFF
--- a/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
+++ b/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
@@ -18,6 +18,12 @@ trait DtabStore {
   def create(ns: String, dtab: Dtab): Future[Unit]
 
   /**
+   * Deletes a dtab.  Returns a DtabNamespaceDoesNotExistException if the
+   * namespace does not exist.
+   */
+  def delete(ns: String): Future[Unit]
+
+  /**
    * Update an existing dtab.  Returns a DtabVersionMismatchException if the
    * supplied version doesn't match the current version.
    */
@@ -38,6 +44,6 @@ object DtabStore {
     extends Exception(s"The dtab namespace $ns already exists")
   class DtabVersionMismatchException
     extends Exception("Could not update dtab: current version does not match provided version")
-  class DtabNamespaceDoesNotExist(ns: String)
+  class DtabNamespaceDoesNotExistException(ns: String)
     extends Exception(s"The dtab namespace $ns does not exist")
 }
--- a/namerd/storage/in-memory/src/main/scala/io/buoyant/namerd/storage/InMemoryDtabStore.scala
+++ b/namerd/storage/in-memory/src/main/scala/io/buoyant/namerd/storage/InMemoryDtabStore.scala
@@ -3,7 +3,7 @@ package io.buoyant.namerd.storage
 import com.twitter.finagle.Dtab
 import com.twitter.io.Buf
 import com.twitter.util._
-import io.buoyant.namerd.DtabStore.{DtabNamespaceDoesNotExist, DtabVersionMismatchException, DtabNamespaceAlreadyExistsException}
+import io.buoyant.namerd.DtabStore.{DtabNamespaceDoesNotExistException, DtabVersionMismatchException, DtabNamespaceAlreadyExistsException}
 import io.buoyant.namerd.{VersionedDtab, DtabStore}
 import java.nio.ByteBuffer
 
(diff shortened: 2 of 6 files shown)


namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
```diff
@@ -18,6 +18,12 @@ trait DtabStore {
   def create(ns: String, dtab: Dtab): Future[Unit]
 
   /**
+   * Deletes a dtab.  Returns a DtabNamespaceDoesNotExistException if the
+   * namespace does not exist.
+   */
+  def delete(ns: String): Future[Unit]
```
gtcampbell: idle question only tangentially related to this review: would it be useful to have the various CRUD operations return Future[VersionedDtab] rather than Future[Unit]? That way, the caller knows the state/version of the object as of its creation/update/deletion, rather than having to call a get() later.

gtcampbell:
:+1: this looks good to me, and should be easy to merge into my DtabStore work on #219.

namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
adleong: There's some discussion of that here: https://github.com/BuoyantIO/linkerd/issues/207.  Ideally yes, though the cost to implementing this depends on the backing store.

gtcampbell: Got it, thanks for the link. FWIW, k8s also gives us that functionality for free.

rmars:
:star: lgtm!

[Approved with PullApprove](https://pullapprove.com/BuoyantIO/linkerd/pull-request/239/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=BuoyantIO/linkerd)

namerd/storage/in-memory/src/main/scala/io/buoyant/namerd/storage/InMemoryDtabStore.scala
```diff
@@ -47,6 +47,19 @@ class InMemoryDtabStore(namespaces: Map[String, Dtab]) extends DtabStore {
...
+    dtabStatesMu.synchronized {
+      state.sample match {
+        case Some(_) =>
+          state.update(None)
+          Future.Unit
+        case None =>
+          Future.exception(new DtabNamespaceDoesNotExistException(ns))
+      }
+    }
+  }
```
olix0r: Should we actually remove it from the state cache after updating it?  Won't it continue to be returned from `list()` otherwise?

namerd/storage/in-memory/src/test/scala/io/buoyant/namerd/storage/InMemoryDtabStoreTest.scala
```diff
@@ -86,4 +86,12 @@ class InMemoryDtabStoreTest extends FunSuite {
       extractDtab(store.observe("test")) == Dtab.read("/hello => /world")
     )
   }
+
+  test("delete deletes dtab") {
+    val store = mkStore
+    val obs = store.observe("test")
+    assert(obs.sample.isDefined)
+    Await.result(store.delete("test"))
+    assert(obs.sample.isEmpty)
```
olix0r: i'd test that `store.list()` doesn't return "test"

namerd/storage/in-memory/src/main/scala/io/buoyant/namerd/storage/InMemoryDtabStore.scala
adleong: I don't think we want to remove this from the cache.  Anyone observing the discarded state var will no longer receive updates (eg if the namespace is recreated).  Instead, I fixed `list` to only list non-empty namespaces.

gtcampbell:
:star:

namerd/storage/in-memory/src/main/scala/io/buoyant/namerd/storage/InMemoryDtabStore.scala
```diff
@@ -31,8 +31,11 @@ class InMemoryDtabStore(namespaces: Map[String, Dtab]) extends DtabStore {
       }
     }
 
-  def list(): Future[Set[String]] =
-    Future.value(dtabStatesMu.synchronized(dtabStates).keySet)
+  def list(): Future[Set[String]] = Future.value {
+    dtabStatesMu.synchronized(dtabStates).filter {
+      case (key, value) => value.sample.isDefined
+    }.keySet
+  }
```
olix0r: tioli:

I generally prefer to avoid nesting -- easier to understand:
```
val v = ..
Future.value(v)
```

might write this as:
```scala
def list(): Future[Set[String]] = {
  val keys = dtabStatesMu.synchronized(dtabStates).collect {
    case (k, v) if v.sample.isDefined => k
  }
  Future.value(keys.toSet)
}
```

olix0r:
⭐

PR #240, opened 2016-04-05
Add a latency metrics table to the client display.

### LINKED ISSUE #185: Write a table component
Part of #183 

This will display latencies, as well as be a legend for the latency graph

rmars:
Get metrics from client.metrics.
Also updated display to new grid.

Fixes #185 
Part of #183

screen shot 2016-04-05 at 4 05 04 pm

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/template/router_client.template
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/template/router_client.template
@@ -2,27 +2,57 @@
   {{client}}
 </div>
 
-<div class="client-metrics">
-  <dl>
-  {{#each metrics}}
-    <div class="col-md-2">
-      <dt>
-        {{description}}
-      </dt>
+<div class="client-medivics row">
+  <div class="col-md-2">
+    <dl>
+      <dt>Requests</dt>
+      <dd>{{requests}}</dd>
+
+      <dt>Successes</dt>
       <dd>
-        <span class="metric-medium">
-          {{#if rate}}
-            {{rate}}
-          {{else}}
-            {{#if value}}
-              {{value}}
-            {{else}}
-              0
-            {{/if}}
-          {{/if}}
-        </span>
+        {{#if successRate}}
+          {{successRate}}
+        {{else}}
+          {{success}}
+        {{/if}}
       </dd>
+    </dl>
+  </div>
+
+  <div class="col-md-2">
+    <dl>
+      <dt>Connections</dt>
+      <dd>{{connections}}</dd>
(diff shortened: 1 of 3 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/template/router_client.template
```diff
@@ -2,27 +2,58 @@
...
+        {{/if}}
       </dd>
     </div>
-  {{/each}}
-  </dl>
+  </div>
+</div>
+
+<div class="col-md-2">
+  <div>Latencies</div>
```
klingerf: Super minor -- I think it might make sense to bold the Latencies heading, so that it's more similar to the Requests, Failures, Connections, etc. heading from above.  Right now it looks like:

screen shot 2016-04-05 at 3 36 03 pm

klingerf:
⭐ 🐢  Just had that trivial nit.

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/template/router_client.template
```diff
@@ -2,27 +2,58 @@
   {{client}}
 </div>
 
-<div class="client-metrics">
-  <dl>
-  {{#each metrics}}
+<div class="client-metrics col-md-4">
```
esbie: hm, `col-md-*` classes really should be inside a `row`.
to avoid all the extra row/col classes, I'd make `.client-metrics` a row, then have 3 `.col-md-2` divs: one with requests/successes, one with connections/failures, and the last one with latencies

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/template/router_client.template
```diff
@@ -2,27 +2,58 @@
...
       </dd>
     </div>
-  {{/each}}
-  </dl>
+  </div>
+</div>
+
+<div class="col-md-2">
+  <div>Latencies</div>
+  <table class="router-latency-table">
```
esbie: These stats _barely_ fit the definition of what I might consider tabular data, so I'll let it slide. Know that generally going forward I'd like to avoid using tables (I honestly don't care for data lists much either).  It makes separation of content & design much easier if they're not used

rmars: Okay! I am not tied at all to using a table - I could change this.

esbie:
:star: :rabbit2: thank you!

klingerf:
:star: :rabbit:

PR #243, opened 2016-04-06
Start setting rpc annotations on traces

### LINKED ISSUE #221: Missing zipkin annotations
Once #217 ships, we'll be able to configure linkerd to export traces to zipkin.  Looking at the trace data in a zipkin-web UI shows that we're clearly missing a few annotations required by the UI. We should figure out what those are and add them to linkerd.

Example:

screen shot 2016-03-31 at 12 48 19 pm

klingerf:
Previously in linkerd we haven't been emitting [Rpc](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/tracing/Tracer.scala#L51) trace annotations, but Zipkin relies on those heavily in its web UI.  I'm adding them where appropriate.  Before this change:

ef01b0be-f73e-11e5-955f-a42dfc7ab1cc-1

With this change:

screen shot 2016-04-05 at 5 49 47 pm

I'm also fixing a small issue where we were duplicating annotations on http spans, resulting in a loss of some data.

Fixes #221. Relates to #140.

### DIFF
--- a/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
+++ b/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
@@ -80,7 +80,9 @@ class RoutingFactory[Req, Rsp](
     override def close(d: Time) = conn.close(d)
 
     def apply(req: Req): Future[Rsp] = {
+      // we treat the router label as the rpc name for this span
       Trace.recordRpc(label)
+      Trace.recordBinary("router.label", label)
       for {
         dst <- getDst(req).rescue {
           case e: Throwable =>
--- a/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
+++ b/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
@@ -80,7 +80,7 @@ class RoutingFactory[Req, Rsp](
     override def close(d: Time) = conn.close(d)
 
     def apply(req: Req): Future[Rsp] = {
-      Trace.recordBinary("router.label", label)
+      Trace.recordRpc(label)
       for {
         dst <- getDst(req).rescue {
           case e: Throwable =>
--- a/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
@@ -1,7 +1,7 @@
 package io.buoyant.router.http
 
 import com.twitter.finagle.{Service, SimpleFilter}
-import com.twitter.finagle.http.{Message, Request, Response}
+import com.twitter.finagle.http.{Request, Response}
 import com.twitter.finagle.tracing.Trace
 
 /**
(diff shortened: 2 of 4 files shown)


router/http/src/test/scala/io/buoyant/router/http/TracingFilterTest.scala
```diff
@@ -27,25 +26,28 @@ class TracingFilterTest extends FunSuite with Awaits {
     req.uri = "/foo?bar=bah"
     req.host = "monkeys"
     req.contentType = "text/plain"
+    req.contentLength = 94114
 
     Trace.letTracer(tracer) {
-      val f = service(req)
+      service(req)
```
adleong: should this be wrapped in an `Await.result`?  Nothing in `service` is currently asynchronous, but this would turn flaky if that ever changes.

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -10,29 +10,30 @@ import com.twitter.finagle.tracing.Trace
 object TracingFilter extends SimpleFilter[Request, Response] {
 
   def apply(req: Request, service: Service[Request, Response]) = {
+    Trace.recordRpc(req.method.toString)
     Trace.recordBinary("http.method", req.method.toString)
```
adleong: these request annotations won't ever conflict with response annotations, but should we namespace them anyway just so that all the request annotations have the same prefix?

router/http/src/test/scala/io/buoyant/router/http/TracingFilterTest.scala
olix0r: preemptive reminder to use `Await.result(f, timeout)` in tests

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
klingerf: I'm pretty ambivalent on this. Finagle's http codec records "http.uri", so I'd like to keep that one for consistency.  I could go either way on the rest of them.  I also noticed that Finagle's annotations tend to be hierarchical, so we should probably use "http.req" and "http.rsp", instead of "req.http" and "rsp.http".

So that said, how do you feel about this as the set of annotations that we record:

* http.uri
* http.req.method
* http.req.host
* http.req.version
* http.req.content-length
* http.req.content-type
* http.req.transfer-encoding
* http.rsp.version
* http.rsp.content-length
* http.rsp.content-type
* http.rsp.transfer-encoding
* http.rsp.status

adleong: that looks good to me

adleong:
:star: nice 🌵

gtcampbell:
:star:

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -80,7 +80,7 @@ class RoutingFactory[Req, Rsp](
     override def close(d: Time) = conn.close(d)
 
     def apply(req: Req): Future[Rsp] = {
-      Trace.recordBinary("router.label", label)
+      Trace.recordRpc(label)
```
olix0r: tioli: it might be nice to emit both so that it's explicit?  This isn't _really_ the RPC name it's just convenient for zipkin's UI, so this is sort of a hack. it would be nice to reliably be able to look at router.label.

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -10,29 +10,31 @@ import com.twitter.finagle.tracing.Trace
...
-      Trace.recordBinary("http.content-length", length)
+      Trace.recordBinary(s"http.$scope.content-length", length)
     }
     for (t <- msg.contentType) {
-      Trace.recordBinary("http.content-type", t)
+      Trace.recordBinary(s"http.$scope.content-type", t)
     }
     for (te <- msg.headerMap.get("transfer-encoding")) {
-      Trace.recordBinary("http.transfer-encoding", te)
+      Trace.recordBinary(s"http.$scope.transfer-encoding", te)
```
olix0r: tioli: not a _huge_ deal but we could probably do this with fewer allocations by having recordReq and recordRsp methods that use val'd strings (aiui this will have to do new string allocations on each message...)

olix0r: yeah, the more i think about it, the more in favor i am of avoiding having to do string interpolation on every trace...  seems wasteful

gtcampbell: same, sorry I didn't catch that in my review.

klingerf: Ok, have updated to avoid string interpolation.

olix0r:
⭐

Issue #242, opened 2016-04-06
Colorize the dashboard


rmars:
colorize the dashboard: 
* each client should have a main color corresponding to its color in the overview graph
* the latencies in the table should be the legend to the latency graph

Part of #183

PR #244, opened 2016-04-06
Add a graph showing per client latencies.

### LINKED ISSUE #241: Add latency graph to clients module
Add a chart to the clients module that displays latencies

Part of #183

rmars:
Currently the colors are randomly assigned, will fix in #242
The display looks quite jumbled at the moment, hoping the color borders and applying padding will help.
CSS definitely needs tweaking, but hoping to address that later.

Fixes #241
Part of #183

screen shot 2016-04-05 at 6 18 26 pm

### DIFF
--- a/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
+++ b/linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
@@ -24,16 +24,15 @@ var RouterClient = (function() {
 
   function getLatencyData(client, latencyKeys) {
     var latencyData = _.pick(client.metrics, latencyKeys);
+    var tableData = {};
+    var chartData = [];
 
-    var tableData = _.mapKeys(latencyData, function(value, key) {
-      return key.split(".")[1];
-    });
-
-    var chartData = _.map(latencyData, function(latency, metricName) {
-      return {
+    _.each(latencyData, function(latencyValue, metricName) {
+      tableData[metricName.split(".")[1]] = latencyValue;
+      chartData.push({
         name: metricName,
-        delta: latency
-      }
+        delta: latencyValue
+      });
     });
 
     return { tableData: tableData, chartData: chartData };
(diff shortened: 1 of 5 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -23,9 +25,18 @@ var RouterClient = (function() {
...
     });
+
+    var chartData = _.map(latencyData, function(latency, metricName) {
+      return {
+        name: metricName,
+        delta: latency
+      }
+    });
+
+    return { tableData: tableData, chartData: chartData };
```
klingerf: You're iterating over latencyData twice here.  It's not as elegant, but you could save yourself some time if you generate both tableData and chartData in the same pass.  Something like:

```
var tableData = {};
var chartData = [];

_.forEach(latencyData, function(latency, metricName) {
  tableData[split(".")[1]] = metricName;
  chartData.push({name: metricName, delta: latency})
}

return { tableData: tableData, chartData: chartData };
```

... but with proper javascript conventions :)

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -42,17 +53,66 @@ var RouterClient = (function() {
...
+    return chart;
+  }
+
+  return function (metricsCollector, routers, client, $metricsEl, routerName, clientTemplate, $chartEl, colors) {
     template = clientTemplate;
-    var metricDefinitions = getMetricDefinition(routerName, client.label);
-    var latencyKeys = Query.filter(/^request_latency_ms\..*/, _.keys(client.metrics));
+    colorLookup = colors;
+    var metricDefinitions = getMetricDefinitions(routerName, client.label);
+    var latencyKeys = Query.filter(/^request_latency_ms\.(max|min|p9990|p99|p95|p50).*/, _.keys(client.metrics));
```
klingerf: I don't think you need the wildcard match after the histogram value here, since those should be the end of the string that you're trying to match against.  As written, this regex will also match on p9999.  It would be better to include a trailing anchor to avoid that.

```
> Query.filter(/^request_latency_ms\.(max|min|p9990|p99|p95|p50).*/, [{name: "request_latency_ms.p9999"}])[0];
<- Object {name: "request_latency_ms.p9999"}
> Query.filter(/^request_latency_ms\.(max|min|p9990|p99|p95|p50)$/, [{name: "request_latency_ms.p9999"}])[0];
<- undefined
```

esbie:
lgtm! router_clients.js is getting kind of big/messy, so consider modularizing it further in upcoming PRs

klingerf:
:star: :flags: Thanks for updating!

PR #245, opened 2016-04-06
Move the CombinedClientGraph into RouterClients

rmars:
In the upcoming colours branch, each RouterClient in RouterClients has a consistent color scheme that matches the main color in the CombinedColorsGraph.

Letting RouterClients manage them makes it much easier/cleaner to handle color updates when new clients are added, otherwise the parent (RouterController) has to go in and update the combined graph as well as the router clients.


rmars:
I think I"m just going to make this part of #242

PR #246, opened 2016-04-06
Tear-down address observations properly

### LINKED ISSUE #237: namerd: resolution failure after deploy
1. namerd deployed
2. app deployed with web service talking to books service
4. traffic is routed properly
5. redeploy web and books
6. namerd observes service updates
7. On subsequent requests
  - web attempts to bind `/http/.1.1/GET/books`
  - web reports ` io.l5d.k8s/prod/incoming/books: name resolution is negative`
  - requests fail with `com.twitter.finagle.NoBrokersAvailableException: No hosts are available for /http/1.1/GET/books`

olix0r:
namerd's thrift interface has no mechanism to release observations. This is
troublesome, because there is no way to obtain a reference to a *new* address
if a service is deleted and re-created.

In order to fix this, we simplify the thrift interface in the following ways:
- Do not cache Addr observations when `bind`-ing a name.
- Bind concrete names passed to the `addr` call if not cached.
- `addr` observations now take NameTree state into account.  When a concrete
  name becomes anything other than a `NameTree.Leaf`, its observation is torn
  down and removed from cache.

Fixes #237

### DIFF
--- a/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
+++ b/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
@@ -350,8 +350,7 @@ class ThriftNamerInterface(
 
         case None =>
           Trace.recordBinary("namerd.srv/addr.cached", false)
-          val (pfx, namer) = namers.find { case (p, _) => id.startsWith(p) }.getOrElse(DefaultNamer)
-          val resolution = namer.bind(NameTree.Leaf(id.drop(pfx.size))).run.flatMap {
+          val resolution = bindAddrId(id).run.flatMap {
             case Activity.Pending => Var.value(Resolution.Resolved(Addr.Pending))
             case Activity.Failed(e) => Var.value(Resolution.Resolved(Addr.Failed(e)))
             case Activity.Ok(tree) => tree match {
@@ -368,6 +367,11 @@ class ThriftNamerInterface(
           obs
       }
     }
+
+  private[this] def bindAddrId(id: Path): Activity[NameTree[Name.Bound]] = {
+    val (pfx, namer) = namers.find { case (p, _) => id.startsWith(p) }.getOrElse(DefaultNamer)
+    namer.bind(NameTree.Leaf(id.drop(pfx.size)))
+  }
 
   private[this] def releaseAddr(id: Path): Unit =
     addrCacheMu.synchronized {
--- a/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
+++ b/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
@@ -2,7 +2,7 @@ package io.buoyant.namerd.iface
 
 import com.twitter.finagle.naming.NameInterpreter
 import com.twitter.finagle.tracing.Trace
-import com.twitter.finagle.{Addr, Address, Dentry, Dtab, Name, NameTree, Path}
+import com.twitter.finagle.{Addr, Address, Dentry, Dtab, Name, Namer, NameTree, Path}
 import com.twitter.io.Buf
 import com.twitter.util._
 import io.buoyant.namerd.iface.{thriftscala => thrift}
@@ -229,7 +229,8 @@ object ThriftNamerInterface {
  * Exposes a polling interface to Namers.
  */
 class ThriftNamerInterface(
-  val namers: Ns => NameInterpreter,
+  interpreters: Ns => NameInterpreter,
+  namers: Map[Path, Namer],
   stamper: ThriftNamerInterface.Stamper,
   retryIn: () => Duration
 ) extends thrift.Namer.FutureIface {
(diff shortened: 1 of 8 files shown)


olix0r:
It occurs to me that the namerd annotations should be thrift-specific since there are other namerd servers.  Let's do that separately.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -212,27 +279,32 @@ class ThriftNamerInterface(
...
+          Trace.recordBinary("namerd.srv/addr.cached", true)
+          obs
+        case None =>
+          Trace.recordBinary("namerd.srv/addr.cached", false)
+          val addr = getBind(ns, Dtab.empty, id).run.flatMap {
+            case Activity.Failed(e) => Var.value(Resolution.Resolved(Addr.Failed(e)))
+            case Activity.Pending => Var.value(Resolution.Resolved(Addr.Pending))
+            case Activity.Ok(NameTree.Leaf(bound)) if bound.id == id =>
+              bound.addr.map(Resolution.Resolved(_))
+            case _ => Var.value(Resolution.Released)
```
adleong: For clarity can we refine this pattern?  This is really `case Activity.Ok(_)`, right?  And moreover, we expect this to be triggered when it goes to `Activity.Ok(NameTree.Neg)`.  Any other value would be a violation of the assumption that the input is already a bound name, I think.

olix0r: This raises an interesting point: are there situations when Activity.Ok(tree) can be satisfied where NameTree is neither Leaf nor Neg? I think it's uncommon but certainly possible given the way this currently works.  And I don't think we can expect users to know to avoid these situations.

For example, given a dtab like:
```
/http/1.1/* => /io.l5d.k8s/prod/http ;
```
a bound name might be `/io.l5d.k8s/prod/http/web`.

What happens if we update the dtab so it looks like:
```
/http/1.1/* => /io.l5d.k8s/prod/http ;
/io.l5d.k8s/prod => /io.l5d.k8s/pre-prod ;
```

Now `/io.l5d.k8s/prod/http/web` binds to an Alt: `/io.l5d.k8s/pre-prod/http/web | /io.l5d.k8s/prod/http/web`? (Is that right? Or does that go infinitely recursive?)

Or more simply, we may find something like:
```
/http/1.1/* => /io.l5d.k8s/prod/http ;
/io.l5d.k8s/prod/http/web => /$/empty ;
```

Perhaps we can avoid this by doing addr lookups directly against a namer instead of through via a dtab?

Or perhaps we can simply forbid dtabs that contain namer prefixes on the LHS? This seems like a generally good policy in any regard.

olix0r:
@adleong https://github.com/BuoyantIO/linkerd/pull/246/commits/bc3061625cb243e1601dda04196b819e61276125 alters addr resolution to skip delegation entirely so that we have a safeguard against user-provided dtabs interfering with addr binding.  This approach isn't set in stone -- let's weigh our options.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -344,27 +309,75 @@ class ThriftNamerInterface(
...
+          Trace.recordBinary("namerd.srv/addr.cached", true)
+          obs
+
+        case None =>
+          Trace.recordBinary("namerd.srv/addr.cached", false)
+          val resolution = namers.find { case (pfx, namer) => id.startsWith(pfx) } match {
+            case None =>
+              // could not find a namer, this addr will never bind
+              Var.value(Resolution.Resolved(Addr.Neg))
+
```
olix0r: This won't work well with `/$` names like `/$/inet`.  Easy fix incoming

adleong:
I think I like this.  Seems like a nice separation of concerns.  Skipping delegation on the addr call makes perfect sense to me.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -329,50 +311,69 @@ class ThriftNamerInterface(
...
+    addrCacheMu.synchronized {
+      addrCache.get(id) match {
+        case Some(obs) =>
+          Trace.recordBinary("namerd.srv/addr.cached", true)
+          obs
+
+        case None =>
+          Trace.recordBinary("namerd.srv/addr.cached", false)
+          val (pfx, namer) = namers.find { case (p, _) => id.startsWith(p) }.getOrElse(DefaultNamer)
+          val resolution = namer.bind(NameTree.Leaf(id.drop(pfx.size))).run.flatMap {
```
adleong: possibly factor some of this out for readability?

adleong: sorry, I should have been more specific.  I meant to possibly factor out the `Activity.State => Var[Resolution]` mapping since it's long and interrupts the flow.

Not a big deal, TIOLI.

adleong:
:star: 👓

esbie:
:star: looks nice, although git's diff really makes this a difficult read :disappointed: . I'd feel more confident if there was a regression test, but don't worry if it's a pita

olix0r:
@esbie good point.  the thrift tests are slightly unwieldy but I'll at least give it a try.  maybe i can make them less unwieldy, even

olix0r:
simplified tests ThriftNamerInterface `addr` tests and added a test that properly reproduces this issue (i.e. fails as expected against master and passes on this branch).

esbie:
:star: :leaves: :fallen_leaf:

PR #247, opened 2016-04-07
Implement namerd storage using the Kubernetes ThirdPartyResource APIs.

### LINKED ISSUE #219: namerd: Implement k8s/3rdparty namerd storage

gtcampbell:
Fixes #219. I have split this into two commits, and could split further or split into multiple pull requests if needed. I have performed a decent amount of manual testing on this, but suggestions on proper automated tests would be appreciated!

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/Api.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Api.scala
@@ -1,0 +1,187 @@
+package io.buoyant.k8s
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.http
+import com.twitter.finagle.http.MediaType
+import com.twitter.io.{Buf, Reader}
+import com.twitter.util._
+
+/**
+ * Contains various classes and methods useful for interacting with the Kubernetes API.
+ */
+object Api {
+  val BufSize = 8 * 1024
+
+  object Closed extends Throwable
+  case class UnexpectedResponse(rsp: http.Response) extends Throwable
+
+  /**
+   * Represents an HTTP 409 Conflict response, returned by the k8s API when attempting to update a
+   * resource with an out-of-date resource version or when attempting to create a resource at an
+   * existing name.
+   */
+  case class Conflict(rsp: http.Response) extends Throwable
+
+  /**
+   * Represents an HTTP 404 Not Found response.
+   */
+  case class NotFound(rsp: http.Response) extends Throwable
+
+  private[k8s] def mkreq(
+    method: http.Method,
+    path: String,
+    content: Option[Buf],
+    optParams: (String, Option[String])*
+  ): http.Request = {
+    val params = optParams collect { case (k, Some(v)) => (k, v) }
+    val req = http.Request(path, params: _*)
+    req.method = method
+    req.contentType = MediaType.Json
+    content.foreach(req.content = _)
+    req
+  }
+
+  private[k8s] def dechunk(reader: Reader, init: Buf = Buf.Empty): Future[Buf] =
+    reader.read(BufSize).flatMap {
+      case Some(chunk) =>
+        dechunk(reader, init.concat(chunk))
+      case None =>
+        Future.value(init)
+    }
+
+  private[k8s] def getContent(msg: http.Message): Future[Buf] =
+    msg.headerMap.get("transfer-encoding") match {
+      case Some("chunked") => dechunk(msg.reader)
+      case _ => Future.value(msg.content)
+    }
+
+  private[k8s] def parse[T: Manifest](rsp: http.Response): Future[T] =
+    getContent(rsp).flatMap { content =>
+      Future.const(Json.read[T](content))
+    }
+}
+
+/**
+ * Generally required as an implicit for list resources. Provides the kubernetes-designated
+ * name for the resource, as well as a means of transforming an individual instance into a
+ * type-specialized [[Watch]].
+ */
+trait ObjectDescriptor[O <: KubeObject, W <: Watch[O]] {
+  /**
+   * @return the URI path segment used for serving lists of resources of type O, e.g. "endpoints",
+   *         "configmaps".
+   */
+  def listName: String
+
+  /**
+   *
+   * @param o a current object instance
+   * @return a Watch (usually Watch.Modified) wrapping `o`
+   */
+  def toWatch(o: O): W
+}
+
+/**
+ * Describes an Object in the Kubernetes API (i.e.
+ * http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_endpoints)
+ */
+trait KubeObject {
+  def apiVersion: Option[String]
+  def metadata: Option[ObjectMeta]
+  def kind: Option[String]
+}
+
+/**
+ * Describes a List of Objects in the Kubernetes API (i.e.
+ * [[http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_endpointslist EndpointsList]])
+ *
+ * @tparam O the type of object contained in the list
+ */
+trait KubeList[O <: KubeObject] {
+  def items: Seq[O]
+  def apiVersion: Option[String]
+  def metadata: Option[ObjectMeta]
+  def kind: Option[String]
+}
+
+/**
+ * See https://github.com/kubernetes/kubernetes/blob/release-1.2/docs/devel/api-conventions.md#metadata
+ * for descriptions of the meanings of the contained fields.
+ */
+case class ObjectMeta(
+  name: Option[String] = None,
+  generateName: Option[String] = None,
+  namespace: Option[String] = None,
+  selfLink: Option[String] = None,
+  uid: Option[String] = None,
+  resourceVersion: Option[String] = None,
+  generation: Option[String] = None,
+  creationTimestamp: Option[String] = None,
+  deletionTimestamp: Option[String] = None,
+  labels: Option[Map[String, String]] = None,
+  annotations: Option[Map[String, String]] = None
+)
+
+case class ObjectReference(
+  kind: Option[String] = None,
+  namespace: Option[String] = None,
+  name: Option[String] = None,
+  uid: Option[String] = None,
+  apiVersion: Option[String] = None,
+  resourceVersion: Option[String] = None,
+  fieldPath: Option[String] = None
+)
+
+/**
+ * An event resulting from a "watch" on the Kubernetes API:
+ * http://kubernetes.io/docs/api-reference/v1/definitions/#_json_watchevent
+ *
+ * Note: Dealing with this class is a little clunky because we haven't been able to get Jackson to
+ * handle the combination of generics and polymorphic inheritance correctly. Thus, you'll need to
+ * create actual subclasses (i.e. FooWatch extends Watch[Foo]) rather than using Watch[Foo]
+ * directly, and set the correct Jackson annotations on those, to ensure correct parsing.
+ */
+trait Watch[O <: KubeObject] {
+  def resourceVersion: Option[String]
+}
+
+object Watch {
+  trait WithObject[O <: KubeObject] extends Watch[O] {
+    def `object`: O
+    @JsonIgnore
+    def resourceVersion = `object`.metadata.flatMap(_.resourceVersion)
+  }
+  trait Added[O <: KubeObject] extends WithObject[O]
+  trait Modified[O <: KubeObject] extends WithObject[O]
+  trait Deleted[O <: KubeObject] extends WithObject[O]
+  trait Error[O <: KubeObject] extends Watch[O] {
+    def status: Status
+    @JsonIgnore
+    def resourceVersion = None
+  }
+}
+
+case class Status(
+  kind: Option[String] = None,
+  apiVersion: Option[String] = None,
+  metadata: Option[ObjectMeta] = None,
+  status: Option[String] = None,
+  message: Option[String] = None,
+  reason: Option[String] = None,
+  details: Option[StatusDetails] = None,
+  code: Option[Int] = None
+)
+
+case class StatusDetails(
+  name: Option[String] = None,
+  kind: Option[String] = None,
+  causes: Option[Seq[StatusCause]] = None,
+  retryAfterSeconds: Option[Int] = None
+)
+
+case class StatusCause(
+  reason: Option[String] = None,
+  message: Option[String] = None,
+  field: Option[String] = None
+)
+
--- a/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
@@ -1,0 +1,58 @@
+package io.buoyant.k8s
+
+import com.twitter.finagle.{Filter, Http, Stack}
+import com.twitter.finagle.http.{Request, Response}
+import io.buoyant.k8s
+import scala.io.Source
+
+/**
+ * Meant to be implemented by configuration classes that need to
+ * produce a k8s API client.
+ */
+trait ClientConfig {
+  import ClientConfig._
+
+  def host: Option[String]
+  def portNum: Option[Int]
+  def tls: Option[Boolean]
+  def tlsWithoutValidation: Option[Boolean]
+  def authTokenFile: Option[String]
+
+  protected def getHost = host.getOrElse(DefaultHost)
+
+  protected def getPort = portNum.getOrElse(DefaultPort)
+
+  protected def dst = s"/$$/inet/$getHost/$getPort"
+
+  private def authFilter = authTokenFile match {
+    case Some(path) =>
+      val token = Source.fromFile(path).mkString
+      if (token.nonEmpty) new AuthFilter(token)
+      else Filter.identity[Request, Response]
+    case None => Filter.identity[Request, Response]
+  }
+
+  protected def mkClient(
+    params: Stack.Params = Stack.Params.empty
+  ) = {
+    val setHost = new SetHostFilter(getHost, getPort)
+    val client = (tls, tlsWithoutValidation) match {
+      case (Some(false), Some(_)) =>
+        log.warning("tlsWithoutValidation is specified, but has no effect as tls is disabled")
+        Http.client
+      case (Some(false), _) => Http.client
+      case (_, Some(true)) => Http.client.withTlsWithoutValidation
+      case _ => Http.client.withTls(setHost.host)
+    }
+    client.withParams(client.params ++ params)
+      .withStreaming(true)
+      .filtered(setHost)
+      .filtered(authFilter)
+  }
+
+}
+
+object ClientConfig {
+  val DefaultHost = "kubernetes.default.svc.cluster.local"
+  val DefaultPort = 443
+}
--- a/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
@@ -54,5 +54,6 @@ trait ClientConfig {
 
 object ClientConfig {
   val DefaultHost = "kubernetes.default.svc.cluster.local"
+  val DefaultNamespace = "default"
   val DefaultPort = 443
 }
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -3,10 +3,12 @@ package io.buoyant.k8s
 import com.twitter.finagle._
 import com.twitter.finagle.tracing.Trace
 import com.twitter.util._
+import io.buoyant.k8s.Api.Closed
+import io.buoyant.k8s.v1.{Endpoints, EndpointsList, EndpointsWatch, NsApi}
 import java.util.concurrent.atomic.AtomicReference
 import scala.collection.mutable
 
-class EndpointsNamer(idPrefix: Path, mkApi: String => v1.NsApi) extends Namer {
+class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
 
   import EndpointsNamer._
 
@@ -83,12 +85,13 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => v1.NsApi) extends Namer {
 
     private[this] def watch(namespace: String, services: NsCache): Closable = {
       val ns = mkApi(namespace)
+      val endpointsApi = ns.endpoints
       val close = new AtomicReference(Closable.nop)
 
       Trace.letClear {
         val init = {
           log.debug("k8s initializing %s", namespace)
-          val endpoints = ns.endpoints()
+          val endpoints = endpointsApi.get()
 
           close.set(Closable.make { _ =>
             endpoints.raise(Closed)
@@ -103,7 +106,7 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => v1.NsApi) extends Namer {
         }
 
         init.foreach { init =>
-          val (updates, closable) = ns.endpoints.watch(
+          val (updates, closable) = endpointsApi.watch(
             resourceVersion = init.metadata.flatMap(_.resourceVersion)
           )
           close.set(closable)
@@ -117,8 +120,6 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => v1.NsApi) extends Namer {
 }
 
 private object EndpointsNamer {
-
-  val log = v1.log
   val PrefixLen = 3
 
   type VarUp[T] = Var[T] with Updatable[T]
@@ -265,11 +266,11 @@ private object EndpointsNamer {
       }
     }
 
-    def update(watch: v1.EndpointsWatch): Unit = watch match {
-      case v1.EndpointsWatch.Error(e) => log.error("k8s watch error: %s", e)
-      case v1.EndpointsWatch.Added(endpoints) => add(endpoints)
-      case v1.EndpointsWatch.Modified(endpoints) => modify(endpoints)
-      case v1.EndpointsWatch.Deleted(endpoints) => delete(endpoints)
+    def update(watch: EndpointsWatch): Unit = watch match {
+      case EndpointsWatch.Error(e) => log.error("k8s watch error: %s", e)
+      case EndpointsWatch.Added(endpoints) => add(endpoints)
+      case EndpointsWatch.Modified(endpoints) => modify(endpoints)
+      case EndpointsWatch.Deleted(endpoints) => delete(endpoints)
     }
 
     private[this] def getName(endpoints: v1.Endpoints) =
--- a/k8s/src/main/scala/io/buoyant/k8s/README.md
+++ b/k8s/src/main/scala/io/buoyant/k8s/README.md
@@ -1,2 +1,1 @@
-This package contains representations of objects returned by multiple versions of the Kubernetes API.
-Version-specific objects should go in sub-packages (see v1.scala).
+
--- a/k8s/src/main/scala/io/buoyant/k8s/README.md
+++ b/k8s/src/main/scala/io/buoyant/k8s/README.md
@@ -1,0 +1,2 @@
+This package contains representations of objects returned by multiple versions of the Kubernetes API.
+Version-specific objects should go in sub-packages (see v1.scala).
--- a/k8s/src/main/scala/io/buoyant/k8s/SerializationModule.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/SerializationModule.scala
@@ -1,0 +1,10 @@
+package io.buoyant.k8s
+
+import com.fasterxml.jackson.databind.module.SimpleModule
+
+/**
+ * Class used to provide LoadService-compatible serialization/deserialization plugins to k8s JSON parsing.
+ */
+abstract class SerializationModule {
+  def module: SimpleModule
+}
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -1,0 +1,133 @@
+package io.buoyant.k8s
+
+import com.twitter.concurrent.AsyncStream
+import com.twitter.finagle.http
+import com.twitter.finagle.http.Response
+import com.twitter.finagle.param.HighResTimer
+import com.twitter.finagle.service.{Backoff, RetryBudget, RetryFilter, RetryPolicy}
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.io.Reader
+import com.twitter.util._
+import com.twitter.util.TimeConversions._
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * An abstract class that encapsulates the ability to Watch a k8s [[Resource]].
+ */
+private[k8s] abstract class Watchable[O <: KubeObject: Manifest, W <: Watch[O]: Manifest]
+  extends Resource {
+  import Watchable._
+
+  protected def backoffs: Stream[Duration]
+  protected def path: String
+  protected def stats: StatsReceiver
+
+  /**
+   * implementing classes should define this method to retrieve the current state
+   * of the resource to pass as Modified results to any watchers.
+   *
+   * @return a Future containing a sequence of Watches and an optional String representing the current resourceVersion
+   */
+  protected def restartWatches(
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None
+  ): Future[(Seq[W], Option[String])]
+
+  protected def infiniteRetryFilter = new RetryFilter[http.Request, http.Response](
+    RetryPolicy.backoff(backoffs) {
+      // We will assume 5xx are retryable, everything else is not for now
+      case (_, Return(rep)) => rep.status.code >= 500 && rep.status.code < 600
+      case (_, Throw(NonFatal(ex))) =>
+        log.error(s"retrying k8s request to $path on error $ex")
+        true
+    },
+    HighResTimer.Default,
+    stats,
+    RetryBudget.Infinite
+  )
+
+  /**
+   * Watch this resource for changes, using a chunked HTTP request.
+   *
+   * TODO: investigate k8s websockets support as an alternative HTTP approach
+   *
+   * @return An asynchronous stream of [[Watch]] objects containing additions, modifications, and
+   *         deletions to the items in the list being watched, and a Closable handle allowing for
+   *         the watch to be terminated.
+   */
+  def watch(
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None,
+    resourceVersion: Option[String] = None
+  ): (AsyncStream[W], Closable) = {
+    val close = new AtomicReference[Closable](Closable.nop)
+
+    // Internal method used to recursively retry watches as needed on failures.
+    def _watch(resourceVersion: Option[String] = None): AsyncStream[W] = {
+      val req = Api.mkreq(http.Method.Get, path, None,
+        "watch" -> Some("true"),
+        "labelSelector" -> labelSelector,
+        "fieldSelector" -> fieldSelector,
+        "resourceVersion" -> resourceVersion)
+      val retryingClient = infiniteRetryFilter andThen client
+      val initialState = retryingClient(req)
+
+      close.set(Closable.make { _ =>
+        log.trace("k8s watch cancelled")
+        initialState.raise(Api.Closed)
+        Future.Unit
+      })
+
+      AsyncStream.fromFuture(initialState).flatMap { rsp =>
+        rsp.status match {
+          // NOTE: 5xx-class statuses will be retried by the infiniteRetryFilter above.
+          case http.Status.Ok =>
+            close.set(Closable.make { _ =>
+              log.debug("k8s watch closed")
+              Future {
+                rsp.reader.discard()
+              } handle {
+                case _: Reader.ReaderDiscarded =>
+              }
+            })
+            val watchStream = Json.readStream[W](rsp.reader, Api.BufSize)
+            watchStream ++ // if the stream ends (k8s will kill connections after ~30m), restart it)
+              AsyncStream.fromFuture {
+                // It's safe to call lastOption here, because we're after the `++` operator, so
+                // this code will not be executed until `watchStream` has completed.
+                val currentVersion = watchStream.lastOption.map(_.flatMap(_.resourceVersion))
+                currentVersion.flatMap {
+                  case Some(version) => Future.value(_watch(Some(version)))
+                  case None =>
+                    // In this case, we want to try loading the initial information instead before watching again.
+                    restartWatches(labelSelector, fieldSelector).map {
+                      case (ws, ver) =>
+                        AsyncStream.fromSeq(ws) ++ _watch(ver)
+                    }
+                }
+              }.flatten
+
+          case status =>
+            close.set(Closable.nop)
+            log.debug(s"k8s failed to watch resource $path: ${status.code} ${status.reason}")
+            val f = Future.exception(Api.UnexpectedResponse(rsp))
+            AsyncStream.fromFuture(f)
+        }
+      }
+    }
+
+    (_watch(resourceVersion), Closable.ref(close))
+  }
+}
+
+object Watchable {
+  def DefaultBackoff = Backoff.exponentialJittered(1.milliseconds, 5.seconds)
+  implicit class RichAsyncStream[T](as: AsyncStream[T]) {
+    /**
+     * Note: forces the stream. For infinite streams, the future never resolves.
+     *
+     * @return the last element in the stream, if any.
+     */
+    def lastOption: Future[Option[T]] = as.toSeq.map(_.lastOption)
+  }
+}
--- a/k8s/src/main/scala/io/buoyant/k8s/resources.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/resources.scala
@@ -1,0 +1,202 @@
+package io.buoyant.k8s
+
+import com.twitter.finagle.service.Backoff
+import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
+import com.twitter.finagle.{Filter, http}
+import com.twitter.util.TimeConversions._
+import com.twitter.util.{Closable, _}
+
+/**
+ * Basic functionality shared by all k8s Resource implementations: they require an HTTP client,
+ * and can be closed.
+ */
+trait Resource extends Closable {
+  def client: Client
+  def close(deadline: Time) = client.close(deadline)
+}
+
+/**
+ * A Version is a top-level API grouping within the K8s API; examples include "/api/v1" and
+ * "/apis/extensions/v1beta1".
+ *
+ * Implementors are likely to want to extend this to provide named convenience methods on top of the
+ * [[Version.listResource]] method and to type-specialize on the resource types supported by that
+ * version.
+ *
+ * @tparam O The parent type for all [Object]s served by this version.
+ */
+private[k8s] trait Version[O <: KubeObject] extends Resource {
+  val path = s"/$group/$version"
+
+  /**
+   * The first portion of the API path. Currently-known groups are "api" (for the core k8s v1 API)
+   * and "apis" (all others, including extensions and third-party resources).
+   */
+  def group: String
+
+  /**
+   * The second portion of the API path. examples include "v1", "extensions/v1beta1", "buoyant.io/v1".
+   */
+  def version: String
+
+  def withNamespace(ns: String) = new NsVersion[O](client, group, version, ns)
+
+  def listResource[T <: O: Manifest, W <: Watch[T]: Manifest, L <: KubeList[T]: Manifest](
+    backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+    stats: StatsReceiver = DefaultStatsReceiver
+  )(implicit od: ObjectDescriptor[T, W]) = new ListResource[T, W, L](client, path, backoffs, stats)
+}
+
+/**
+ * A Version with a namespace applied. This is a Class because we want to provide a default
+ * `namespace` implementation in [[Version.listResource]], although in many cases implementors may wish to
+ * extend this to provide convenience methods wrapping `list`.
+ *
+ * @param client See [[Version.client]]
+ * @param group See [[Version.group]]
+ * @param version See [[Version.version]]
+ * @param ns The namespace
+ * @tparam O The parent type for all [[KubeObject]]s served by this version.
+ */
+private[k8s] class NsVersion[O <: KubeObject](
+  val client: Client,
+  group: String,
+  val version: String,
+  ns: String
+)
+  extends Resource {
+  val path = s"/$group/$version/namespaces/$ns"
+
+  def listResource[T <: O: Manifest, W <: Watch[T]: Manifest, L <: KubeList[T]: Manifest](
+    backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+    stats: StatsReceiver = DefaultStatsReceiver
+  )(implicit od: ObjectDescriptor[T, W]) = new NsListResource[T, W, L](client, path, backoffs, stats)
+}
+
+/**
+ * A third-party version contains ThirdPartyResource-typed objects. See
+ * https://github.com/kubernetes/kubernetes/blob/master/docs/design/extending-api.md for a description of the
+ * design model.
+ *
+ * Implementors are likely to want to extend this to provide convenience methods on top of the `list` method.
+ *
+ * @tparam O The parent type for all [[KubeObject]]s served by this Version.
+ */
+trait ThirdPartyVersion[O <: KubeObject] extends Version[O] {
+  /** domain name, i.e. "buoyant.io" */
+  def owner: String
+
+  /** version within owner domain, i.e. "v1" */
+  def ownerVersion: String
+
+  def version: String = ThirdPartyVersion.version(owner, ownerVersion)
+  def basePath: String = s"/${ThirdPartyVersion.group}/$version"
+  val group = ThirdPartyVersion.group
+  override def withNamespace(ns: String) = new NsThirdPartyVersion[O](client, owner, ownerVersion, ns)
+}
+
+object ThirdPartyVersion {
+  def version(owner: String, ownerVersion: String) = s"$owner/$ownerVersion"
+  val group = "apis"
+}
+
+/**
+ * A namespaced third-party version, i.e. "/apis/buoyant.io/v1/namespaces/ns/".
+ */
+class NsThirdPartyVersion[O <: KubeObject](client: Client, owner: String, ownerVersion: String, ns: String)
+  extends NsVersion[O](client, ThirdPartyVersion.group, ThirdPartyVersion.version(owner, ownerVersion), ns)
+
+/**
+ * Represents the functionality for a  kubernetes API resource serving a list of objects, for example:
+ * `/api/v1/namespaces/{namespace}/endpoints`.
+ */
+private[k8s] class ListResource[O <: KubeObject: Manifest, W <: Watch[O]: Manifest, L <: KubeList[O]: Manifest](
+  val client: Client,
+  basePath: String,
+  protected val backoffs: Stream[Duration] = Watchable.DefaultBackoff,
+  protected val stats: StatsReceiver = DefaultStatsReceiver
+)(implicit od: ObjectDescriptor[O, W]) extends Watchable[O, W] with Resource {
+  val name = implicitly[ObjectDescriptor[O, W]].listName
+  val watchManifest = implicitly[Manifest[W]]
+  val path = s"$basePath/$name"
+
+  def get(
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None,
+    resourceVersion: Option[String] = None,
+    retryIndefinitely: Boolean = false
+  ): Future[L] = {
+    val req = Api.mkreq(http.Method.Get, this.path, None,
+      "labelSelector" -> labelSelector,
+      "fieldSelector" -> fieldSelector,
+      "resourceVersion" -> resourceVersion)
+    val retry = if (retryIndefinitely) infiniteRetryFilter else Filter.identity[http.Request, http.Response]
+    val retryingClient = retry andThen client
+    retryingClient(req).flatMap(Api.parse[L])
+  }
+
+  protected def restartWatches(
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None
+  ): Future[(Seq[W], Option[String])] =
+    get(labelSelector, fieldSelector, None, true).map { list =>
+      (list.items.map(od.toWatch), list.metadata.flatMap(_.resourceVersion))
+    }
+}
+
+/**
+ * Namespaced list resources support more operations than non-namespaced, including POST, DELETE, and
+ * retrieval of individual items in the list. (We don't have a full implementation yet).
+ */
+private[k8s] class NsListResource[O <: KubeObject: Manifest, W <: Watch[O]: Manifest, L <: KubeList[O]: Manifest](
+  client: Client,
+  basePath: String,
+  backoffs: Stream[Duration] = Watchable.DefaultBackoff,
+  stats: StatsReceiver = DefaultStatsReceiver
+)(implicit od: ObjectDescriptor[O, W]) extends ListResource[O, W, L](client, basePath, backoffs, stats) {
+  def named(name: String): NsObjectResource[O, W] =
+    new NsObjectResource[O, W](client, path, name)
+
+  /**
+   * Creates an Object within the represented List via an HTTP POST.
+   *
+   * @param toCreate the object to POST to the Kubernetes API.
+   * @return if successful, the final created object (will include server-generated attributes like resourceVersion, etc).
+   */
+  def post(toCreate: O): Future[O] = {
+    val req = Api.mkreq(http.Method.Post, path, Some(Json.writeBuf[O](toCreate)))
+    client(req).flatMap(Api.parse[O])
+  }
+}
+
+private[k8s] class NsObjectResource[O <: KubeObject: Manifest, W <: Watch[O]: Manifest](
+  val client: Client,
+  listPath: String,
+  objectName: String,
+  protected val backoffs: Stream[Duration] = Watchable.DefaultBackoff,
+  protected val stats: StatsReceiver = DefaultStatsReceiver
+)(implicit od: ObjectDescriptor[O, W]) extends Resource {
+  private[this] val path = s"$listPath/$objectName"
+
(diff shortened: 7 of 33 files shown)


k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -0,0 +1,176 @@
...
+  private[k8s] def mkreq(
+    method: http.Method,
+    path: String,
+    content: Option[Buf],
+    optParams: (String, Option[String])*
+  ): http.Request = {
+    val params = optParams collect { case (k, Some(v)) => (k, v) }
+    val req = http.Request(path, params: _*)
+    req.method = method
+    req.contentType = "application/json"
```
adleong: I think there's a `MediaType.Json` const for this

k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -0,0 +1,176 @@
...
+    content.foreach(req.content = _)
+    req
+  }
+
+  private[k8s] def dechunk(reader: Reader, init: Buf = Buf.Empty): Future[Buf] =
+    reader.read(BufSize) flatMap {
```
adleong: nit: `reader.read(BufSize).flatMap {`

k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+  private[k8s] def getContent(msg: http.Message): Future[Buf] =
+    msg.headerMap.get("transfer-encoding") match {
+      case Some("chunked") => dechunk(msg.reader)
+      case _ => Future.value(msg.content)
+    }
+
+  private[k8s] def parse[T: Manifest](rsp: http.Response): Future[T] =
+    getContent(rsp) flatMap { content =>
```
adleong: nit: `.flatMap`

k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -0,0 +1,176 @@
...
+   * @param o a current object instance
+   * @return a Watch (usually Watch.Modified) wrapping `o`
+   */
+  def toWatch(o: O): W
+}
+
+/**
+ * Describes an object in the Kubernetes API (i.e. http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_endpoints)
+ */
+trait Object {
```
adleong: maybe rename to not conflict with `java.lang.Object`

k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
```diff
@@ -0,0 +1,48 @@
...
+      else Filter.identity[Request, Response]
+    case None => Filter.identity[Request, Response]
+  }
+
+  protected def mkClient(
+    params: Stack.Params = Stack.Params.empty
+  ) = {
+    val setHost = new SetHostFilter(getHost, getPort)
+    val client = (tls, tlsWithoutValidation) match {
+      case (Some(false), _) => Http.client
```
adleong: Should the `(Some(false), Some(_))` case emit some kind of error or warning since it's not really valid?  ie specifying `tlsWithoutValidation` doesn't have meaning if `tls` is off

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
+import com.twitter.io.{Buf, Reader}
+import com.twitter.util.{Closable, _}
+import com.twitter.util.TimeConversions._
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * Basic functionality shared by all k8s Resource implementations: they require an HTTP client, and can be closed.
+ */
+trait Resource extends Closable {
```
adleong: maybe name this file `Resource.scala` to match this trait?  I know there are a bunch of resources in this file, but it might be good to have the file name match at least one of the things defined in it.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+   * (all others, including extensions and third-party resources).
+   */
+  def group: String
+
+  /**
+   * The second portion of the API path. examples include "v1", "extensions/v1beta1", "buoyant.io/v1".
+   */
+  def version: String
+
+  def client: Client
```
adleong: isn't this already defined in Resource?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+
+  def namespace(ns: String) = new NsVersion[O](client, group, version, ns)
```
adleong: should this be called `withNamespace`?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+ * A namespaced third-party version, i.e. "/apis/buoyant.io/v1/namespaces/ns/".
+ */
+class NsThirdPartyVersion[O <: Object](client: Client, owner: String, ownerVersion: String, ns: String)
+  extends NsVersion[O](client, ThirdPartyVersion.group, ThirdPartyVersion.version(owner, ownerVersion), ns)
+
+/**
+ * A trait that encapsulates the ability to Watch a k8s resource.
+ */
+private[k8s] trait Watchable[O <: Object, W <: Watch[O]] extends Resource {
+  implicit protected val watchManifest: Manifest[W]
```
adleong: can you explain what this line is for?  Should the type param of `Watchable` specify that `W <: Watch[O] : Manifest` ?

k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -0,0 +1,176 @@
...
+      Future.const(Json.read[T](content))
+    }
+}
+
+/**
+ * Generally required as an implicit for list resources. Provides the kubernetes-designated
+ * name for the resource, as well as a means of transforming an individual instance into a type-specialized
+ * Watch.
+ */
+trait ObjectDescriptor[O <: Object, W <: Watch[O]] {
```
adleong: I like to avoid `O` as an identifier because it can be confused for `0` although in this case there's probably not any room for confusion

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+  /**
+   * implementing classes should define this method to retrieve the current state
+   * of the resource to pass as Modified results to any watchers.
+   *
+   * @return a Future containing a sequence of Watches and an optional String representing the current resourceVersion
+   */
+  protected def restartWatches(
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None
+  ): Future[(Seq[W], Option[String])]
```
adleong: Can this return type be made into a case class for clarity?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+  /**
+   * Watch this resource for changes, using a chunked HTTP request.
+   *
+   * TODO: investigate k8s websockets support as an alternative HTTP approach
+   */
+  def watch(
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None,
+    resourceVersion: Option[String] = None
+  ): (AsyncStream[W], Closable) = {
```
adleong: can we have a description of what this returns or a more self-descriptive return type?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+    // Internal method used to recursively retry watches as needed on failures.
+    def _watch(resourceVersion: Option[String] = None): AsyncStream[W] = {
+      val req = Api.mkreq(http.Method.Get, path, None,
+        "watch" -> Some("true"),
+        "labelSelector" -> labelSelector,
+        "fieldSelector" -> fieldSelector,
+        "resourceVersion" -> resourceVersion)
+
+      def currentVersion(initial: Option[String], watch: AsyncStream[W]): Future[Option[String]] = {
+        watch.foldLeft(initial) {
```
adleong: this just takes the resourceVersion of the last watch?  maybe define a helper method `def last[A](as: AsyncStream[A]): Future[Option[A]]`

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+          (version: Option[String], watch: W) => watch.resourceVersion
+        }
+      }
+
+      def watchFromResponse(rsp: Response): AsyncStream[W] = rsp.status match {
+        // NOTE: 5xx-class statuses will be retried by the infiniteRetryFilter above.
+        case http.Status.Ok =>
+          close.set(Closable.make { _ =>
+            log.debug("k8s watch closed")
+            Future {
```
adleong: personal preference: I like `Future.value`

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+              rsp.reader.discard()
+            } handle {
+              case _: Reader.ReaderDiscarded =>
```
adleong: does calling `reader.discard()` throw a `ReaderDiscarded`?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: This was moved from elsewhere; I think I'd actually like to change it to:

```scala
watch.foldLeft(initial) {
  (version: Option[String], watch: W) => watch.resourceVersion orElse version
}
```

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: Sadly, you can't use context bounds in traits :(. I can put a comment to that effect.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: I tried to keep the original API from Endpoints intact when I generalized it, and that used `namespace`, but I think I would prefer either `withNamespace` or `namespaced`; happy to change it.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+            }
+          })
+          val theStream = Json.readStream[W](rsp.reader, Api.BufSize)
```
adleong: maybe name this `watchStream` or `watches`?

k8s/src/main/scala/io/buoyant/k8s/Api.scala
olix0r: yeah, this was pretty jarring when I first looked through this.  perhaps KubeObject or something?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: I was following the convention from http://docs.scala-lang.org/style/files.html#multiunit_files, but maybe we should have a conversation about our preferred file naming convention for multi-unit files.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+          theStream ++
+            AsyncStream.fromFuture(
+              currentVersion(resourceVersion, theStream) flatMap {
```
adleong: nit: `.flatMap`

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -55,13 +57,11 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => v1.NsApi) extends Namer {
...
-            case None =>
-              val ns = mkNs(name)
-              activity() = Activity.Ok(snap + (name -> ns))
-              ns
-          }
+          snap.getOrElse(name, {
+            val ns = mkNs(name)
+            activity() = Activity.Ok(snap + (name -> ns))
+            ns
+          })
```
olix0r: I think this is harder to reason about, since we have no guarantees that this closure is called within the context of the synchronized block

k8s/src/main/scala/io/buoyant/k8s/Defaults.scala
```diff
@@ -0,0 +1,9 @@
+package io.buoyant.k8s
+
+/**
+ * Some useful defaults for configuration of k8s API clients.
+ */
+object Defaults {
+  val Host = "kubernetes.default.cluster.local"
+  val Port = 443
+}
```
olix0r: I think this would be clearer in a ClientConfig companion object

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+            ).flatten
+
+        case status =>
+          close.set(Closable.nop)
+          log.debug("k8s failed to watch endpoints: %d %s", status.code, status.reason)
+          val f = Future.exception(Api.UnexpectedResponse(rsp))
+          AsyncStream.fromFuture(f)
+      }
+
+      val retryingClient = infiniteRetryFilter andThen client
```
adleong: nit: avoid infix

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
gtcampbell: Happy to revert, although I'd actually like to make sure I understand your concern a little better too.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+/**
+ * A namespaced third-party version, i.e. "/apis/buoyant.io/v1/namespaces/ns/".
+ */
+class NsThirdPartyVersion[O <: Object](client: Client, owner: String, ownerVersion: String, ns: String)
+  extends NsVersion[O](client, ThirdPartyVersion.group, ThirdPartyVersion.version(owner, ownerVersion), ns)
+
+/**
+ * A trait that encapsulates the ability to Watch a k8s resource.
+ */
+private[k8s] trait Watchable[O <: Object, W <: Watch[O]] extends Resource {
```
adleong: seems like a lot of the core logic lives in here.  would it be worth splitting this into its own file `Watchable.scala`?  It also might help readability if the nested defs in `watch` were top level `private[this] def`s in `Watchable`.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
olix0r: The signature for `Map[K,V].getOrElse` is effectively
```scala
def getOrElse(k: K, f: => V): V
```

In this case we happen to know that f is going to be called immediately, but generally speaking, this is not a good assumption that if i pass a closure to something that it will be called immediately (in the same thread, etc).  when I see a `synchronized` and then passing a closure to function, i need to stop and make sure that everything's safe ("ok, this isn't a Future, or a view, or something like that").  When there's no closure, it's easy for me to reason about without thinking about the specific types at play.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+  }
+}
+
+object Watchable {
+  def DefaultBackoff = Backoff.exponentialJittered(1.milliseconds, 5.seconds)
+}
+
+/**
+ * Represents the functionality for a  kubernetes API resource serving a list of objects, for example:
+ * `/api/v1/namespaces/{namespace}/endpoints`. This is an abstract class because it makes the Manifest implicit
```
adleong: it doesn't look like `ListResource` is actually abstract

k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -0,0 +1,176 @@
...
+  def apiVersion: Option[String]
+  def metadata: Option[ObjectMeta]
+  def kind: Option[String]
+}
+
+/**
+ * Describes a List of Objects in the Kubernetes API (i.e. http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_endpointslist)
+ * @tparam O the type of object contained in the list
+ */
+trait List[O <: Object] {
```
adleong: rename to avoid conflict with `scala.collection.immutable.List`?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+ */
+private[k8s] class ListResource[O <: Object: Manifest, W <: Watch[O]: Manifest, L <: List[O]: Manifest](
+  val client: Client,
+  basePath: String,
+  protected val backoffs: Stream[Duration] = Watchable.DefaultBackoff,
+  protected val stats: StatsReceiver = DefaultStatsReceiver
+)(implicit od: ObjectDescriptor[O, W]) extends Resource with Watchable[O, W] {
+  val name = implicitly[ObjectDescriptor[O, W]].listName
+  val watchManifest = implicitly[Manifest[W]]
+  val path = basePath + "/" + name
```
adleong: tioli: interpolation?

namerd/examples/k8s.yaml
```diff
@@ -0,0 +1,12 @@
+# This file assumes you are running through `kubectl proxy` on localhost:8001.
+storage:
+  kind: io.buoyant.namerd.storage.experimental.k8s
```
olix0r: generally, i have mixed feelings on using class paths for kinds.  I know we've done this for other ones, but it encodes a bunch of redundant information that isn't particularly user-friendly.  We can revisit this outside of this review, though.

k8s/src/main/scala/io/buoyant/k8s/Api.scala
gtcampbell: I waffled on this, and would be happy to discuss further. I tried to name these based on the Kubernetes documentation: https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#types-kinds, which refers to Object and List, and usually our code style is fine with reproducing simple names in multiple packages (i.e. we have a bunch of `Api` classes). I can call these K8sObject and K8sList, but then you end up with the moderately-repetitive `io.buoyant.k8s.K8sObject` as the fully-qualified class name.

namerd/examples/k8s.yaml
gtcampbell: I agree - I was going with the precedent from the other two storage `kind`s  here, which also use class paths.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/resources.scala
```diff
@@ -0,0 +1,29 @@
...
+  override def ownerVersion: String = Api.ownerVersion
+
+  override def namespace(ns: String) = new NsApi(client, ns)
+  implicit val descriptor = DtabDescriptor
+  def dtabs = list[Dtab, DtabWatch, DtabList]()
+}
+
+object Api {
+  // TODO: should this be buoyant.io, linkerd.io, namerd.io...?
+  val owner = "buoyant.io"
```
olix0r: If this ends up encoded in resrouces, i think i'd prefer "l5d.io"?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Create a new dtab.  Returns a DtabNamespaceAlreadyExistsException if a
+   * dtab already exists with that namespace.
+   */
+  def create(ns: String, dtab: FDtab): Future[Unit] = {
+    api.dtabs.post(namedDtab(ns, dtab)).rescue {
+      case io.buoyant.k8s.Api.Conflict(_) => Future.exception(new DtabNamespaceAlreadyExistsException(ns))
+    }.unit
+  }
```
olix0r: nit: omit needless braces

also I think `.unit.rescue` is slightly cleaner than `.rescue { ... }.unit` but not a big deal either way

namerd/core/src/main/scala/io/buoyant/namerd/DtabSerialization.scala
```diff
@@ -0,0 +1,37 @@
+package io.buoyant.namerd
+
+import com.fasterxml.jackson.core.{JsonGenerator, JsonParser}
+import com.fasterxml.jackson.databind.{DeserializationContext, JsonDeserializer, JsonSerializer, SerializerProvider}
+import com.fasterxml.jackson.databind.module.SimpleModule
+import com.twitter.finagle.{NameTree, Path}
+
+object DtabSerialization {
```
adleong: could call this `DtabCodec` maybe?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -0,0 +1,136 @@
...
+   * Update an existing dtab regardless of the current version or create a new
+   * dtab if one doesn't already exist.
+   */
+  def put(ns: String, dtab: FDtab): Future[Unit] = {
+    api.dtabs.named(ns).put(namedDtab(ns, dtab)).rescue {
+      // In the event of a 404 on a PUT, we should POST instead, as the k8s API doesn't appear to support creating on PUT.
+      // Note that this is potentially racy if the namespace is created between the PUT and the POST, but it's likely
+      // safer to surface that error all the way to the user in that case than to attempt to recover beyond this point.
+      case io.buoyant.k8s.Api.NotFound(_) => create(ns, dtab)
+    }.unit
```
olix0r: I This should almost certainly be `.unit.rescue`?  (Slightly surprised this compiles as i'd expect there to be confusion over the return type of rescue?  Or does `named(ns).put()` return a Unit? If so is `.unit` necessary at all?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -0,0 +1,136 @@
...
+ * @param namespace the Kubernetes namespace being read/written by this store (note that this differs from Dtab namespaces)
+ */
+class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
+  extends DtabStore {
+  private[this] val log = Logger()
+
+  // NOTE: we construct two clients, one for the streaming watch requests, one for everything else.
+  // This is due to an as-yet-undetermined issue where the first non-watch request will hang if using
+  // the same client.
+  val api: NsApi = Api(client.newService(dst)).namespace("default")
```
adleong: should `namespace` be used instead of `"default"`?

olix0r:
style: let's cut line lengths somewhere around 100 chars (guideline, not a rule).  There are some quite long lines, particularly comments, that would be nice to shorten.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
gtcampbell: yes, whoops! :( I'll try to capture that in a test.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  /** List all Dtab namespaces */
+  def list(): Future[Set[String]] = act.run.sample match {
+    case Activity.Ok(nses) => Future.value(nses.keySet)
+    case Activity.Pending => for {
+      tryNses: Try[NsMap] <- act.values.toFuture
+      nses <- Future.const(tryNses)
+    } yield nses.keySet
```
olix0r: nit: would be slightly easier to read if the whole `for { }` loop was indented:

e.g
```scala
case Activity.Pending =>
  for {
    ..
  } yield ..
case ...
```

somewhat visually confusing that the yield is at the same indent level as the next case

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -0,0 +1,136 @@
...
+        case DtabWatch.Modified(m) => nsMap + toDtabMap(m)
+        case DtabWatch.Deleted(d) => nsMap - name(d)
+        case DtabWatch.Error(e) =>
+          log.error("k8s watch error: %s", e)
+          nsMap
+      }
+      witness.notify(Return(newState))
+      newState
+    }
+  }
```
olix0r: This block could probably use a brief comment about what it does (in addition to some of the details of its behavior).  As it is, it's hard to contextualize the comment.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
gtcampbell: My understanding (although I am far from an expert in Scala's type inference):
* `named(ns).put()` returns `Future[Dtab]`
* `rescue`'s signature is `rescue[B >: A](pf: PartialFunction[Throwable, Future[B]): Future[B]`
* because nothing in the `rescue` partialFunction here returns anything that alters the type of the `Future`, the result is also `Future[Dtab]`

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  /** List all Dtab namespaces */
+  def list(): Future[Set[String]] = act.run.sample match {
```
adleong: I think this can be `act.values.toFuture.flatMap(Future.const).map(_.ketSet)`

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Update an existing dtab regardless of the current version or create a new
+   * dtab if one doesn't already exist.
+   */
+  def put(ns: String, dtab: FDtab): Future[Unit] = {
+    api.dtabs.named(ns).put(namedDtab(ns, dtab)).rescue {
+      // In the event of a 404 on a PUT, we should POST instead, as the k8s API doesn't appear to support creating on PUT.
+      // Note that this is potentially racy if the namespace is created between the PUT and the POST, but it's likely
+      // safer to surface that error all the way to the user in that case than to attempt to recover beyond this point.
```
adleong: I agree.  The user can choose to retry the PUT and it should work once the state is no longer changing.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
olix0r: IIUC we have no means to tear down the stream or report an error?
I _think_ I'd be inclined to write this block slightly differently:

```scala
private[this] val states = Var.async[Activity.State[NsMap]] { state =>
  val closeRef = new AtomicReference[Closable](Closable.nop)
  val pending = api.dtabs.get(...).respond {
    case Throw(e) =>
      state = Activity.Failed(e)
    case Return(dtabList) =>
      state() = Activity.Ok(dtabList)
      val (stream, close) = watchApi.dtabs.watch(None, None, dtabList.apiVersion)
      closeRef.set(close)
      stream ... {
        state() = Activity.Ok(...)
      }
  }
  Closable.make { t =>
    pending.raise(...)
    Closable.ref(closeRef).close(t)
  }
}

private[this] val act = Activity(states)
```

We may need to hold this activity open, but we should be able to gracefully tear it down. (If DtabStore isn't Closable, that's another thing we should do in a followup).

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
gtcampbell: I mildly prefer `.rescue.unit` to `.unit.rescue` because the former guarantees that the result is `Future[Unit]` whereas the latter could be anything depending on what's in the rescue block, but I also don't think it's a big deal either way.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+  /**
+   * The second portion of the API path. examples include "v1", "extensions/v1beta1", "buoyant.io/v1".
+   */
+  def version: String
+
+  def client: Client
+
+  def namespace(ns: String) = new NsVersion[O](client, group, version, ns)
+  def path = s"/$group/$version"
+  def list[T <: O: Manifest, W <: Watch[T]: Manifest, L <: List[T]: Manifest](
```
olix0r: perhaps `listResource` would be clearer? it was slightly odd to see `dtabs = list[...]`

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+ * A Version with a namespace applied. This is a Class because we want to provide a default `namespace` implementation
+ * in [[Version]], although in many cases implementors may wish to extend this to provide convenience methods wrapping
+ * `list`.
+ *
+ * @param client The client used to make requests against this API.
+ * @param version Full path for this version of the API.
+ * @param ns The namespace
+ * @tparam O The parent type for all [Object]s served by this version.
+ */
+private[k8s] class NsVersion[O <: Object](val client: Client, group: String, val version: String, ns: String) extends Resource {
```
olix0r: nit: i'd break this up into multiple lines e.g.

```scala
private[k8s] class NsVersion[O <: Object](
  val client: Client,
  group: String,
  val version: String, ns: String
) extends Resource {
  ...
}
```

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+  )(implicit od: ObjectDescriptor[T, W]) = new ListResource[T, W, L](client, path, backoffs, stats)
+}
+
+/**
+ * A Version with a namespace applied. This is a Class because we want to provide a default `namespace` implementation
+ * in [[Version]], although in many cases implementors may wish to extend this to provide convenience methods wrapping
+ * `list`.
+ *
+ * @param client The client used to make requests against this API.
+ * @param version Full path for this version of the API.
```
olix0r: should `group` be documented?

k8s/src/main/scala/io/buoyant/k8s/Api.scala
olix0r: I agree that it's a bit jarring seeing List in this context.  ListResource?  KubeList?

k8s/src/main/scala/io/buoyant/k8s/SerializationModule.scala
```diff
@@ -0,0 +1,10 @@
+package io.buoyant.k8s
+
+import com.fasterxml.jackson.databind.module.SimpleModule
+
+/**
+ * Class used to provide LoadService-compatible serialization/deserialization plugins to k8s JSON parsing.
+ */
+abstract class SerializationModule {
```
olix0r: What are the benefits of this being pluggable?  Don't the concrete ThirdParty types require specific serialization support?

gtcampbell: Not at the moment - they're just case classes, which can be handled by Jackson without any special-casing. The reason to make this pluggable is that the `ObjectMapper` is constructed in the `k8s` project, but only `namerd-storage-k8s` needs it to be able to parse `Path` and `NameTree[Path]` instances.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
adleong: Could this be an abstract class?

gtcampbell:
re 100 char cutoff - I had 120 as the setting in my IDE; I'll adjust that.

namerd/core/src/main/scala/io/buoyant/namerd/DtabSerialization.scala
gtcampbell: Hmm, there's already a `DtabCodec` declared in HttpControlService that does something different. Are you okay with the name duplication?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
gtcampbell: Yes, I was wondering what to do with the Closable as well - DtabStore is not Closable at the moment.

namerd/core/src/main/scala/io/buoyant/namerd/DtabSerialization.scala
adleong: Can the two codecs be consolidated?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
olix0r: i think i agree withNamespace is good/better

k8s/src/main/scala/io/buoyant/k8s/resources.scala
olix0r: I agree that generally Future.value is preferable, but here Future.apply is right so that exceptions are raised into Future.exception.

It seems like this block could simply be the following, though:
```scala
try rsp.reader.discard() catch { case _: Reader.ReaderDiscarded }
Future.Unit
```

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
gtcampbell: This ended up being changed to a one-liner as suggested by @adleong elsewhere.

k8s/src/main/scala/io/buoyant/k8s/Api.scala
gtcampbell: Sure, I went with KubeObject.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
olix0r: oh? i thought that because `create()` returns `Future[Unit]`, `.rescue { ... => create() }` would be a problem?  Is it the case that `Unit >: Dtab`?

k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
gtcampbell: (ultimately, this should probably be a sub-item of a `tls` object, but for now I made your suggested change)

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: I made it a trait because it seems like something one might want to mix in to classes that have other superclasses, but you're right that in the current implementation it could be an abstract class.

k8s/src/main/scala/io/buoyant/k8s/Api.scala
gtcampbell: Would you prefer `KO` now that it's a `KubeObject`?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: I'm having a hard time coming up with a good name - suggestions?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: never mind, after our in-person discussion we decided it's correct just to take the last item of the stream.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: In this case, I don't think we want to use that because the block below might throw an exception, which we want to handle in the `handle` clause; `Future.apply` does the right exception handling, but I don't believe `Future.value` does.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: I believe it does so only if the reader has *already* been discarded (which is fine in this case). That said, @olix0r wrote this code originally (I just relocated it) and can correct me if I'm wrong.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
gtcampbell: Thanks, that seems reasonable.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: looks like we're about half/half for infix vs. dots on `andThen`, and finagle uses infix most of the time.

k8s/src/main/scala/io/buoyant/k8s/Api.scala
gtcampbell: I went with `KubeList` and `KubeObject`.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -0,0 +1,308 @@
...
+
+/**
+ * A Version is a top-level API grouping within the K8s API; examples include "/api/v1" and "/apis/extensions/v1beta1".
+ *
+ * Implementors are likely to want to extend this to provide named convenience methods on top of the `list` method and
+ * to type-specialize on the resource types supported by that version.
+ *
+ * @tparam O The parent type for all [Object]s served by this version.
+ */
+private[k8s] trait Version[O <: Object] extends Resource {
```
adleong: we discussed possibly calling this `Space` so as not to be confused with the api version (v1) or the resource version

k8s/src/main/scala/io/buoyant/k8s/resources.scala
adleong: I think the case class name isn't too important.  What's more important is that the fields are named (eg `watches` and `currentVersion`).  What about `Restart` or `RestartResult`?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/resources.scala
gtcampbell: Sure.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
gtcampbell: Sorry, I was looking at the other `rescue` clause when making that comment. *This* one works not because `Unit >: Dtab`, but, I believe, because of value discarding ("If e has some value type and the expected type is Unit, e is converted to the expected type by embedding it in the term { e; () }."), which I agree is a little weird. I'll switch to .unit.rescue.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
adleong: this took a bit to wrap my head around so it might be worth calling out in a comment or something like that:

currentVersion forces the stream so it will not be satisfied for an in progress response.  This would be a problem because theStream is a streaming response, except for the fact that since this is all on the right-hand-side of `theStream ++`.  That means that for this to be evaluated, `theStream` must be already completed.

This code seems a bit brittle because it relies on this unstated requirement that `currentVersion` be only called with a completed stream.  I don't have a specific suggestion, but at the very least we should be careful.

adleong:
I just finished my second pass.  It makes a lot more sense now that I have the appropriate context, thanks!  Overall I think this looks solid.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
gtcampbell: so I'm trying to figure out the right name scheme. Right now, the Kubernetes API for `apis` returns an `ApiGroupList`, with items like:
```
{
      "name": "buoyant.io",
      "versions": [
        {
          "groupVersion": "buoyant.io/v1",
          "version": "v1"
        }
      ],
    },
```

and "/api" returns an `ApiVersionList`, of which the only element is "v1".

So: it sounds like the closest nomenclature to what we want is actually "GroupVersion"? Or we could introduce another layer in our API, so you'd have something like Group("buoyant").version("v1").namespace("ns")...?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -0,0 +1,122 @@
...
+   *         the watch to be terminated.
+   */
+  def watch(
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None,
+    resourceVersion: Option[String] = None
+  ): (AsyncStream[W], Closable) = {
+    val close = new AtomicReference[Closable](Closable.nop)
+
+    // Internal method used to recursively retry watches as needed on failures.
```
adleong: personal preference: it would be more readable to me if these internal methods were instead `private[this]` methods on `Watchable`

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -0,0 +1,122 @@
...
+    def _watch(resourceVersion: Option[String] = None): AsyncStream[W] = {
+      val req = Api.mkreq(http.Method.Get, path, None,
+        "watch" -> Some("true"),
+        "labelSelector" -> labelSelector,
+        "fieldSelector" -> fieldSelector,
+        "resourceVersion" -> resourceVersion)
+
+      def watchFromResponse(rsp: Response): AsyncStream[W] = rsp.status match {
```
adleong: this is an internal method inside an internal method 😢

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -0,0 +1,122 @@
...
+            Future {
+              rsp.reader.discard()
+            } handle {
+              case _: Reader.ReaderDiscarded =>
+            }
+          })
+          val watchStream = Json.readStream[W](rsp.reader, Api.BufSize)
+          watchStream ++
+            AsyncStream.fromFuture(
+              watchStream.toSeq.map(_.lastOption.flatMap(_.resourceVersion)).flatMap {
```
adleong: might factor out a helper method `def last[T](as: AsyncStream[T]): Option[T]` with a comment here about why it's safe to force the stream

adleong:
:star: 🎱 I like it!

gtcampbell:
(rebased and resolved conflicts, including adding a delete() method)

olix0r:
To recap IRL conversation: this is looking good. But we should give this a bit of a burn-in in a test cluster.

gtcampbell:
rebased/resolved conflicts again.

k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+    client.withParams(client.params ++ params)
+      .withStreaming(true)
+      .filtered(setHost)
+      .filtered(authFilter)
+  }
+
+}
+
+object ClientConfig {
+  val DefaultHost = "kubernetes.default.svc.cluster.local"
```
olix0r: Codacy Issue found: [Field name 'DefaultHost' does not match the regular expression '^[a-z][A-Za-z0-9]*$'](https://www.codacy.com/app/ver/linkerd/file/2415518678/issues/source?bid=3236810&fileBranchId=3236828#l56)

k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+  val DefaultNamespace = "default"
```
olix0r: Codacy Issue found: [Field name 'DefaultNamespace' does not match the regular expression '^[a-z][A-Za-z0-9]*$'](https://www.codacy.com/app/ver/linkerd/file/2415518678/issues/source?bid=3236810&fileBranchId=3236828#l57)

k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+import com.twitter.finagle.http.{Request, Response}
+import io.buoyant.k8s
+import scala.io.Source
+
+/**
+ * Meant to be implemented by configuration classes that need to
+ * produce a k8s API client.
+ */
+trait ClientConfig {
+  import ClientConfig._
```
olix0r: Codacy Issue found: [Put imports in the beginning of the package](https://www.codacy.com/app/ver/linkerd/file/2415518678/issues/source?bid=3236810&fileBranchId=3236828#l13)

k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+      .filtered(setHost)
+      .filtered(authFilter)
+  }
+
+}
+
+object ClientConfig {
+  val DefaultHost = "kubernetes.default.svc.cluster.local"
+  val DefaultNamespace = "default"
+  val DefaultPort = 443
```
olix0r: Codacy Issue found: [Field name 'DefaultPort' does not match the regular expression '^[a-z][A-Za-z0-9]*$'](https://www.codacy.com/app/ver/linkerd/file/2415518678/issues/source?bid=3236810&fileBranchId=3236828#l58)

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/resources.scala
```diff
@@ -0,0 +1,28 @@
...
+  def dtabs = listResource[Dtab, DtabWatch, DtabList]()
+}
+
+object Api {
+  // TODO: should this be buoyant.io, linkerd.io, namerd.io...?
+  val Owner = "l5d.io"
+
+  // TODO: what versioning scheme do we want to use here? I assume we don't want
+  //   to match linkerd/namerd versions, as this schema is unlikely to change with similar regularity.
+  val OwnerVersion = "v1"
```
olix0r: Codacy Issue found: [Field name 'OwnerVersion' does not match the regular expression '^[a-z][A-Za-z0-9]*$'](https://www.codacy.com/app/ver/linkerd/file/2415518826/issues/source?bid=3236810&fileBranchId=3236828#l21)

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/resources.scala
```diff
@@ -0,0 +1,28 @@
...
+  override def ownerVersion: String = Api.OwnerVersion
+
+  override def withNamespace(ns: String) = new NsApi(client, ns)
+  implicit val descriptor = DtabDescriptor
+  def dtabs = listResource[Dtab, DtabWatch, DtabList]()
+}
+
+object Api {
+  // TODO: should this be buoyant.io, linkerd.io, namerd.io...?
+  val Owner = "l5d.io"
```
olix0r: Codacy Issue found: [Field name 'Owner' does not match the regular expression '^[a-z][A-Za-z0-9]*$'](https://www.codacy.com/app/ver/linkerd/file/2415518826/issues/source?bid=3236810&fileBranchId=3236828#l17)

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/DtabWatch.scala
```diff
@@ -0,0 +1,23 @@
...
+  new JsonSubTypes.Type(value = classOf[DtabDeleted], name = "DELETED"),
+  new JsonSubTypes.Type(value = classOf[DtabError], name = "ERROR")
+))
+sealed trait DtabWatch extends Watch[Dtab]
+
+// NOTE: These are not (as would be usual practice) inside a DtabWatch companion object,
+// because doing so leads to intermittent compilation failures from
+// https://issues.scala-lang.org/browse/SI-7551.
+case class DtabAdded(`object`: Dtab) extends DtabWatch with Watch.Added[Dtab]
+case class DtabModified(`object`: Dtab) extends DtabWatch with Watch.Modified[Dtab]
```
olix0r: Codacy Issue found: [`object` is a reserved keyword don't use it as a name](https://www.codacy.com/app/ver/linkerd/file/2415518825/issues/source?bid=3236810&fileBranchId=3236828#l19)

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/DtabWatch.scala
```diff
@@ -0,0 +1,23 @@
...
+  new JsonSubTypes.Type(value = classOf[DtabModified], name = "MODIFIED"),
+  new JsonSubTypes.Type(value = classOf[DtabDeleted], name = "DELETED"),
+  new JsonSubTypes.Type(value = classOf[DtabError], name = "ERROR")
+))
+sealed trait DtabWatch extends Watch[Dtab]
+
+// NOTE: These are not (as would be usual practice) inside a DtabWatch companion object,
+// because doing so leads to intermittent compilation failures from
+// https://issues.scala-lang.org/browse/SI-7551.
+case class DtabAdded(`object`: Dtab) extends DtabWatch with Watch.Added[Dtab]
```
olix0r: Codacy Issue found: [`object` is a reserved keyword don't use it as a name](https://www.codacy.com/app/ver/linkerd/file/2415518825/issues/source?bid=3236810&fileBranchId=3236828#l18)

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/DtabWatch.scala
olix0r: Codacy Issue found: [Field name '`object`' does not match the regular expression '^[a-z][A-Za-z0-9]*$'](https://www.codacy.com/app/ver/linkerd/file/2415518825/issues/source?bid=3236810&fileBranchId=3236828#l19)

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/kubernetes/DtabWatch.scala
```diff
@@ -0,0 +1,23 @@
...
+case class DtabModified(`object`: Dtab) extends DtabWatch with Watch.Modified[Dtab]
+case class DtabDeleted(`object`: Dtab) extends DtabWatch with Watch.Deleted[Dtab]
```
olix0r: Codacy Issue found: [`object` is a reserved keyword don't use it as a name](https://www.codacy.com/app/ver/linkerd/file/2415518825/issues/source?bid=3236810&fileBranchId=3236828#l20)

k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -0,0 +1,187 @@
+package io.buoyant.k8s
```
olix0r: thoughts about scoping this under a v1 package?  Since the API is versioned, and all of this is (i think?) specific to the v1 api, should we try to reflect that in the packaging?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -0,0 +1,133 @@
...
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None
+  ): Future[(Seq[W], Option[String])]
+
+  protected def infiniteRetryFilter = new RetryFilter[http.Request, http.Response](
+    RetryPolicy.backoff(backoffs) {
+      // We will assume 5xx are retryable, everything else is not for now
+      case (_, Return(rep)) => rep.status.code >= 500 && rep.status.code < 600
+      case (_, Throw(NonFatal(ex))) =>
+        log.error(s"retrying k8s endpoints request on error $ex")
```
olix0r: i think this log message is no longer correct? this may not be an endpoints request

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -0,0 +1,133 @@
...
+                    restartWatches(labelSelector, fieldSelector).map {
+                      case (ws, ver) =>
+                        AsyncStream.fromSeq(ws) ++ _watch(ver)
+                    }
+                }
+              }.flatten
+
+          case status =>
+            close.set(Closable.nop)
+            log.debug("k8s failed to watch endpoints: %d %s", status.code, status.reason)
```
olix0r: endpoints-specific log message

k8s/src/main/scala/io/buoyant/k8s/Api.scala
gtcampbell: good question: I was trying to have things match the URL structure. The third-party namespaces are outside the core of the v1 API (i.e. they start with `apis/owner/owner-specified-version` rather than `api/v1`), so I'm not sure how they map into the version of the API as a whole. How would we want to differentiate between the core v1 API and things like `apis/extensions/v1beta1`, etc.?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
gtcampbell: Good catch, will fix these.

k8s/src/main/scala/io/buoyant/k8s/Api.scala
olix0r: Hm... this is probably more trouble than it's worth (and if there really is a v2 api we can just create a k8s-v2 package...)

olix0r:
lgtm modulo those nits!

olix0r:
:star:

k8s/src/main/scala/io/buoyant/k8s/README.md
```diff
@@ -0,0 +1,2 @@
+This package contains representations of objects returned by multiple versions of the Kubernetes API.
+Version-specific objects should go in sub-packages (see v1.scala).
```
klingerf: This file feels a bit buried to me -- we haven't previously included readme files nested this deeply within the project.  Maybe this information would be more visible as a scaladoc comment in k8s/src/main/scala/io/buoyant/k8s/Api.scala?

olix0r: ahh, yeah good catch.  should be in `k8s/` or a comment on `k8s/src/main/scala/io/buoyant/k8s/package.scala`

k8s/src/test/scala/io/buoyant/k8s/ThirdPartyResourceTest.scala
```diff
@@ -0,0 +1,102 @@
...
+ */
+class ThirdPartyResourceTest extends FunSuite with Awaits {
+  import ThirdPartyResourceTest._
+
+  object Fixtures {
+    val bookList = Buf.Utf8("""{"kind":"BookList","apiVersion":"buoyant.io/vTest","metadata":{"selfLink":"/apis/buoyant.io/vTest/namespaces/test/books"},"items":[{"metadata":{"name":"programming-in-scala","selfLink":"/apis/buoyant.io/vTest/namespaces/test/books/programming-in-scala"},"title":"Programming in Scala","author":"Martin Odersky"}]} """)
+  }
+
+  test("namespaced: get books list") {
+    import Fixtures._
```
klingerf: Why not define booklist within the context of ThirdPartyResourceTest, instead of putting it into a separate object that needs to be imported?

```
class ThirdPartyResourceTest extends FunSuite with Awaits {
  import ThirdPartyResourceTest._

  val bookList = Buf.Utf8("""{"kind":"BookList"...

  test("namespaced: get books list") {
    val service = Service.mk[Request, Response] { req =>
```

namerd/docs/config.md
```diff
@@ -47,6 +47,23 @@ Stores the dtab in memory.  Not suitable for production use.
 
 * *namespaces* -- Optional.  A map of namespaces to corresponding dtabs.
 
+### K8sDtabStore
+
+`io.buoyant.namerd.storage.experimental.k8s`
```
klingerf: For consistency with the zookeeper store description in this document, I think this should be written as:

```
### K8sDtabStore
*experimental*
  
`io.buoyant.namerd.storage.experimental.k8s`
```

namerd/examples/3rdparty.yaml
```diff
@@ -0,0 +1,7 @@
+metadata:
+  name: d-tab.l5d.io # the hyphen is required by the Kubernetes API. This will be converted to the CamelCase name "DTab".
+apiVersion: extensions/v1beta1
+kind: ThirdPartyResource
+description: stores dtabs used by Buoyant's `namerd` service
+versions:
+  - name: v1
```
klingerf: Since this isn't a namerd config, I think it's a bit confusing to put it in the `namerd/examples` directory.  For instance, we can't run `./sbt namerd-examples/3rdparty:run`.  How about moving it to a namerd/examples/k8s subdirectory, or something similar?

klingerf:
⭐ 🌱  Looks good to me -- I just left some optional organizational comments, but didn't see any issues with the code.

k8s/src/test/scala/io/buoyant/k8s/ThirdPartyResourceTest.scala
gtcampbell: I had it that way earlier in this branch's lifecycle, but I found it easier to read this way since the single test was a bit buried among all the setup classes otherwise.

olix0r:
:star: :w00t:

klingerf:
⭐ Thanks for making those updates.

PR #248, opened 2016-04-07
Add per client assigned colors to the dashboard graphs

### LINKED ISSUE #242: Colorize the dashboard

rmars:
I also moved the CombinedClientGraph into RouterClients for easier control of colours.
The colours go as follows:
Each new client of a router gets the next colour off the preferred colour list.  The latency graph for that client then gets the shades of this colour.

Adding new clients is a little messy, because we have to assign a colour to the new client, and both the CombinedClientGraph and the RouterClient need to know which colour that is.  So the parent (RouterClients) calls an updateColors(newClientToColorMapping) with the new mapping of which clients are assigned which colours.  For now, I regenerate the color mapping for /all/ clients when new clients are added, as this was the most straightforward.  Can definitely change it though. 

Both RouterClient and RouterClients are getting long.  In another review I will split out the latency graph from RouterClient, and move RouterClients to its own file.

Questions: should we be getting latency metrics from `metrics.json` rather than from `/metrics`?

Fixes #242
screen shot 2016-04-07 at 11 03 41 am

screen shot 2016-04-07 at 11 40 28 am

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/css/dashboard.css
+++ b/admin/src/main/resources/io/buoyant/admin/css/dashboard.css
@@ -133,4 +133,22 @@ nav ul.navbar-nav {
 
 .router-servers-title {
   border-bottom: 1px solid grey;
+  font-size: 18px;
+}
+
+/**
+* Router clients
+*/
+
+.router-clients {
+  border-bottom: 1px solid grey;
+  margin-bottom: 30px;
+}
+
+.router-graph {
+  margin-bottom: 30px;
+}
+
+.chart-container {
+  padding-top: 40px;
 }
--- a/admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
@@ -4,11 +4,12 @@ var CombinedClientGraph = (function() {
   }
 
   return function(metricsCollector, routerName, $root, colors) {
+    var clientColors = colors;
     var query = Query.clientQuery().withRouter(routerName).withMetric("requests").build();
 
-    var timeseriesParams = function() {
+    function timeseriesParams(name) {
       return {
-        strokeStyle: colors[colorIdx++ % colors.length],
+        strokeStyle: clientColors[name.match(Query.clientQuery().build())[2]].color,
         lineWidth: 2
       };
     };
@@ -37,14 +38,16 @@ var CombinedClientGraph = (function() {
     );
 
     var desiredMetrics = _.map(Query.filter(query, metricsCollector.getCurrentMetrics()), clientToMetric);
-    var colorIdx = 0;
-
     chart.setMetrics(desiredMetrics, timeseriesParams, true);
 
     metricsCollector.registerListener(function(data) {
       var filteredData = Query.filter(query, data.specific);
       chart.updateMetrics(filteredData);
     }, function(metrics) { return Query.filter(query, metrics); });
-    return {};
+    return {
+      updateColors: function(newColors) {
+        clientColors = newColors;
+      }
+    };
   };
 })();
--- a/admin/src/main/resources/io/buoyant/admin/js/dashboard.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/dashboard.js
@@ -9,17 +9,19 @@ $.when(
   $.get("/files/template/router_container.template"),
   $.get("/files/template/router_server.template"),
   $.get("/files/template/router_client.template"),
+  $.get("/files/template/router_client_container.template"),
   $.get("/files/template/router_summary.template"),
   $.get("/files/template/process_info.template"),
   $.get("/files/template/request_totals.template"),
   $.get("/admin/metrics.json")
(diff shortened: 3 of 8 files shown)


linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -122,25 +136,57 @@ var RouterClient = (function() {
...
+    _.map(routers.clients(routerName), initializeClient);
 
     function initializeClient(client) {
-      var $metrics = $("<div />").addClass("col-md-6").appendTo($clientEl);
-      var $chart = $("<div />").addClass("col-md-6").appendTo($clientEl);
+      var colorsForClient = clientToColor[client.prefix].colorFamily;
+
+      var $container = $("<div />").css("border-top", "2px solid " + colorsForClient.neutral);
+      var $metrics = $("<div />").addClass("col-md-6").appendTo($container);
+      var $chart = $("<div />").addClass("chart-container col-md-6").appendTo($container);
```
esbie: please add these as part of a template

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -122,25 +136,57 @@ var RouterClient = (function() {
...
   };
 })();
 
 var RouterClients = (function() {
+  function assignColorsToClients(colors, clients) {
+    var colorIdx = 0;
+
+    return _.reduce(clients, function(clientMapping, client) {
+      clientMapping[client.prefix] = colors[colorIdx++ % colors.length];
+      return clientMapping;
```
esbie: I'd prefer clientMapping to be by `client.label` not by prefix

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
```diff
@@ -52,15 +65,15 @@ var RouterClient = (function() {
     return summary;
   }
 
-  function timeseriesParams() {
+  function timeseriesParams(name) {
     return {
-      strokeStyle: colorLookup[colorIdx++ % colorLookup.length], //TODO: #242 per client color scheme
+      strokeStyle: chartLegend[name.replace("request_latency_ms.", "")],
```
esbie: I prefer that we don't try parsing metrics in multiple places. Consider something like:
`strokeStyle: chartLegend[name.match(Query.clientQuery().build())[1]]`

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/combined_client_graph.js
```diff
@@ -4,11 +4,12 @@ var CombinedClientGraph = (function() {
...
 
   return function(metricsCollector, routerName, $root, colors) {
+    var clientColors = colors;
     var query = Query.clientQuery().withRouter(routerName).withMetric("requests").build();
 
-    var timeseriesParams = function() {
+    function timeseriesParams(name) {
       return {
-        strokeStyle: colors[colorIdx++ % colors.length],
+        strokeStyle: clientColors[name.replace("requests", "")].color,
```
esbie: I prefer that we don't try parsing metrics in multiple places. Consider something like:
strokeStyle: clientColors[name.match(Query.clientQuery().build())[1]].color

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_controller.js
```diff
@@ -5,63 +5,77 @@ var RouterController = (function () {
...
 
+  var colorOrder = _.map(baseColorOrder, function(colorName) {
+    var color;
+    var colorParts = colorName.split(".");
+
+    if (colorParts.length > 1) {
+      color = colors[colorParts[0]][colorParts[1]];
+    } else {
+      color = colors[colorParts[0]].neutral;
+    }
```
esbie: simpler as:
`var color = _.property(colorName)(color);`
would have to add .neutrals to the baseColorOrder

esbie:
coming along really nicely :)

linkerd/admin/src/main/resources/io/buoyant/linkerd/admin/js/router_clients.js
rmars: I'm getting the metrics here from client.metrics (from routers.js), which just returns "request_latency_ms.max: 0" etc...  So the client name isn't in name in this case.  

I'm not actually sure whether I should be using client.metrics from routers or if I should be getting latency from metrics.json.

esbie: hm ok fair enough, let's revisit later

esbie:
:star: :stars: thank you for updating!

rmars:
I rebased on top of esbie's branch that moves all the resources.
Also fixed bug where client legends were the wrong colour.

screen shot 2016-04-08 at 4 26 21 pm

olix0r:
:star: for what my review is worth, this looks very nice

PR #252, opened 2016-04-08
Introduce DtabStore.Proxy type to support store-agnostic validation

olix0r:
We'll need to be perform write validation in an interface- and store-agnostic
way.

In order to support this, a DtabStore.Proxy type can be used to wrap DtabStores
with behavior. A special DtabStore.Validator type is introduced to provide a
basis for write validation.

Furthermore, type aliases are introduced to make the interface somewhat purer -- the interface should talk about Namespaces and Versions not Strings and Bufs.

### DIFF
--- a/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
+++ b/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
@@ -4,18 +4,19 @@ import com.twitter.finagle.Dtab
 import com.twitter.io.Buf
 import com.twitter.util.{Activity, Future}
 
-case class VersionedDtab(dtab: Dtab, version: Buf)
+case class VersionedDtab(dtab: Dtab, version: DtabStore.Version)
 
 trait DtabStore {
+  import DtabStore.Namespace
 
   /** List all namespaces */
-  def list(): Future[Set[String]]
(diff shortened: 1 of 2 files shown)


namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
```diff
@@ -4,40 +4,71 @@ import com.twitter.finagle.Dtab
...
+  def put(ns: Namespace, dtab: Dtab): Future[Unit]
 
   /** Watch a dtab and it's version. */
-  def observe(ns: String): Activity[Option[VersionedDtab]]
+  def observe(ns: Namespace): Activity[Option[VersionedDtab]]
 }
 
 object DtabStore {
-  class DtabNamespaceAlreadyExistsException(ns: String)
+  type Namespace = String
```
adleong: unify this with the type alias in the `io.buoyant.namerd` package object?

adleong:
nice

gtcampbell:
LGTM pending @adleong's note about the type alias duplication between `io.buoyant.namerd.DtabStore.Namespace` and `io.buoyant.namerd.Ns`.

adleong:
:star: 🐙

olix0r:
will merge once 0.3.1 is tagged

PR #255, opened 2016-04-08
Only require one shipit for work in linkerd/admin

rmars:
Pullapprove documentation here:
http://docs.pullapprove.com/yaml/reviewers/

I /think/ this is what we want (?)

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -3,14 +3,30 @@ approve_regex: '^(Approved|lgtm|LGTM|:shipit:|:star:|:\+1:|:ship:)'
 reject_regex: ^Rejected
 reset_on_push: false
 reviewers:
-  members:
-  - adleong
-  - esbie
-  - gtcampbell
-  - klingerf
-  - olix0r
-  - rmars
-  - siggy
-  - stevej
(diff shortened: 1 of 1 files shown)


adleong:
:star: 🎣

.pullapprove.yml
```diff
@@ -3,14 +3,30 @@ approve_regex: '^(Approved|lgtm|LGTM|:shipit:|:star:|:\+1:|:ship:)'
...
+    - adleong
+    - esbie
+    - gtcampbell
+    - klingerf
+    - olix0r
+    - rmars
+    - siggy
+    - stevej
+    name: linkerd-admin
+    required: 1
```
adleong: this is crazy, but what happens if you put `-1` here?

rmars: Then pullapprove requires approval from /all/ members :(

`A value of -1 means that approval is required from all members.`

rmars:
There's no way to apply exclusions currently.  There is an open issue in pullapprove:
https://github.com/pullapprove/support/issues/52

rmars:
Won't fix for now...

PR #256, opened 2016-04-08
Create our own version of ZkSession with a more flexible API

adleong:
The API exposed by `com.twitter.finagle.serverset2.ZkSession` is inflexible and doesn't suit our needs very well.  We create our own version of this class with a more flexible API which has

- a single RW zookeeper client which automatically reconnects on session expiry
- direct access to that client
- the `watchedOperation` method allows us to transform the zookeeper api into our concurrency primitives (eg `Activity`)


namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkSession.scala
```diff
@@ -0,0 +1,186 @@
...
+import com.twitter.finagle.serverset2.RetryStream
+import com.twitter.finagle.serverset2.client._
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.logging.Logger
+import com.twitter.util._
+
+/**
+  * ZkSession wraps a ZooKeeper client which automatically reconnects upon session expiry.
+  * Cribbed from https://github.com/twitter/finagle/blob/develop/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/ZkSession.scala
+  */
```
gtcampbell: It might be good to add some comments noting which portions of the code are fully copied from the Finagle implementation and which are specific to this one.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
```diff
@@ -0,0 +1,127 @@
...
+
+    zkSession.zk.create(
+      path,
+      Some(Buf.Utf8(dtab.show)),
+      Seq(Data.ACL.AnyoneAllUnsafe),
+      CreateMode.Persistent
+    ).rescue {
+        case KeeperException.NodeExists(_) =>
+          Future.exception(new DtabNamespaceAlreadyExistsException(ns))
+      }.unit
```
gtcampbell: In @olix0r's review of #247, he mentioned that he prefers `.unit.rescue` over `.rescue.unit`.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
```diff
@@ -0,0 +1,127 @@
+package com.twitter.finagle.serverset2.buoyant
```
gtcampbell: It would be nice to add a comment indicating why this needs to be in the `com.twitter.finagle` package (I assume because we make use of some package-private code?)

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkSession.scala
```diff
@@ -0,0 +1,186 @@
...
+    logger.info(s"Closing zk session $sessionId")
+    zk.close()
+    val newClient = clientBuilder()
+    logger.info(s"Starting new zk session ${sessionId(newClient)}")
+
+    // Upon initial connection, send auth info, then update `u`
+    newClient.state.changes.filter {
+      _ == WatchState.SessionState(SessionState.SyncConnected)
+    }.toFuture.unit.before {
+      // TODO: send auth info
```
gtcampbell: link to relevant issue here?

gtcampbell:
:star: with the caveat that I'm not that familiar with the Zookeeper client. That said, the additions and subtractions from Finagle's ZKSession make sense to me.

olix0r:
For what it's worth -- and we don't have to do this in this branch -- I'd prefer that the ZkDtabStore itself exist in the io.buoyant.namerd package.  We should use the com.twitter.finagle.serverset2.buoyant package to expose the raw utilities that we need to implement a DtabStore.  This is more aspirational than a hard requirement...

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
```diff
@@ -0,0 +1,127 @@
...
+        case Some(Buf.Utf8(s)) =>
+          Activity.value(Some(VersionedDtab(Dtab.read(s), versionBuf(data.stat.version))))
+        case None =>
+          Activity.exception(new IllegalStateException(s"Empty node: $path"))
+      }
+    }.handle {
+      case NoNode(_) => None
+    }
+  }
+
```
olix0r: Not a blocker, but this isn't great for something like namerd.  It means that once a node is observed, it will be observed forever (even if there are no observers).  If for instance, many short-lived dtabs are created and deleted, this will manifest as a memory leak.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
```diff
@@ -0,0 +1,127 @@
...
+  def observe(ns: String): Activity[Option[VersionedDtab]] = dtabWatchOp(s"$zkPrefix/$ns")
+
+  override def put(ns: String, dtab: Dtab): Future[Unit] = {
+    val path = s"$zkPrefix/$ns"
```
olix0r: not a blocker, but we probably need to think more about validation.  Can zkPrefix end in a slash?  Can ns start with a slash?

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkSession.scala
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * The returned Activity is asynchronous: watches aren't reissued
+   * when the Activity is no longer observed.
+   */
+  def watchedOperation[T](go: => Future[Watched[T]]): Activity[T] =
+    Activity(Var.async[Activity.State[T]](Activity.Pending) { u =>
+      @volatile var closed = false
+
+      def loop(): Future[Unit] = {
+        if (!closed) safeRetry(go) respond {
```
olix0r: style: i assume this is copied, but prefer `safeRetry(go).respond`

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkSession.scala
```diff
@@ -0,0 +1,186 @@
...
+   * time the watch fires, applying safe retries when possible.
+   *
+   * The returned Activity is asynchronous: watches aren't reissued
+   * when the Activity is no longer observed.
+   */
+  def watchedOperation[T](go: => Future[Watched[T]]): Activity[T] =
+    Activity(Var.async[Activity.State[T]](Activity.Pending) { u =>
+      @volatile var closed = false
+
+      def loop(): Future[Unit] = {
```
olix0r: It's not clear to me why this returns Future[Unit] and not just Unit.  Perhaps just to satisfy retryWithDelay's signature?

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
```diff
@@ -0,0 +1,127 @@
...
+ * encoded, in the ZooKeeper node /dtabs/<ns>.
+ */
+class ZkDtabStore(
+  hosts: String,
+  zkPrefix: String,
+  sessionTimeout: Option[Duration]
+) extends DtabStore {
+
+  private[this] val log = Logger()
+  private[this] val retryStream = RetryStream()
```
olix0r: what does this mean, practically? Should this be configurable?  Should this be an exponentialJitter backoff?

olix0r:
:star: looks like a good step in the right direction! no blockers -- i think any of my comments could be addressed in a followup branch

adleong:
Thanks for the feedback!  Would you mind creating issues for the items you want addressed in followups?

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
adleong: Doesn't really matter either way in this case.  This ordering makes sense to me because `zk.create` is the fallible operation and I like to have the rescue directly on the operation that it rescues from.  The `.unit` call is just to discard the return value since it's not needed to satisfy the method signature so it makes sense to me to have it last.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
adleong: yeah, pretty much everything in c.t.f.serverset2 is package private, including the ClientBuilder and exceptions.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkSession.scala
adleong: Yes, I think that's the reason.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/buoyant/ZkDtabStore.scala
adleong: The default stream is `Backoff.decorrelatedJittered(10.milliseconds, 10.seconds)`.  Would exponentialJitter be better?  We could also make the parameters configurable.

PR #258, opened 2016-04-09
linkerd-admin: Calculate success rate over success + failure

olix0r:
`requests` is measured in a different place from `success` and `failures`. `requests` counts the number of requests sent, whereas `success` and `failures` measure _responses received_.  Because of this, `success` can easily be greater than `requests`, which leads to odd success rate measurements.

This is easily fixed by calculating total requests as `success + failure`.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/router_server.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/router_server.js
@@ -18,7 +18,7 @@ var RouterServer = (function() {
         metricSuffix: "success",
         query: Query.serverQuery().withRouter(routerName).withServer(serverName).withMetric("success").build(),
         getRate: function(data) {
-          var successRate = new SuccessRate(data.requests, data.success, data.failures);
+          var successRate = new SuccessRate(data.success || 0, data.failures || 0);
           return successRate.prettyRate();
         }
       },
--- a/admin/src/main/resources/io/buoyant/admin/js/utils.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/utils.js
@@ -53,8 +53,8 @@ BytesToStringConverter.prototype.convert = function(b) {
  * Success Rate Utils
  */
 
-function SuccessRate(requests, success, failures) {
-  this.requests = requests;
+function SuccessRate(success, failures) {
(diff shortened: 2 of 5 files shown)


admin/src/main/resources/io/buoyant/admin/js/utils.js
```diff
@@ -53,8 +53,8 @@ BytesToStringConverter.prototype.convert = function(b) {
  * Success Rate Utils
  */
 
-function SuccessRate(requests, success, failures) {
-  this.requests = requests;
+function SuccessRate(success, failures) {
+  this.requests = success + failures;
```
siggy: now that requests is deterministic based on success+failure, the `requests` element of this data structure is redundant information. consider removing and allowing the caller to perform the addition?

rmars: true, though I think naming it "requests" makes it a little clearer to read

adleong:
[Q] just curious, how could you receive more responses than requests sent?

adleong:
:star: 📈

admin/src/main/resources/io/buoyant/admin/js/router_server.js
```diff
@@ -18,7 +18,7 @@ var RouterServer = (function() {
         metricSuffix: "success",
         query: Query.serverQuery().withRouter(routerName).withServer(serverName).withMetric("success").build(),
         getRate: function(data) {
-          var successRate = new SuccessRate(data.requests, data.success, data.failures);
+          var successRate = new SuccessRate(data.success, data.failures);
```
rmars: data.failures isn't always present in the returned metrics (if there haven't yet been any failures) so sometimes the dashboard will display NaN% where it should display N/A.  We should (I can do this later) do some checks on data.failures.

rmars:
:star: :abcd:

olix0r:
@adleong in a let's say we have two time windows: t0 and t1:
- In t0, the process sends 1000 request and receives 500 responses.
- In t1, the process sends 100 requests, and receives 600 responses.

(Yes, this is a contrived example, but there's no reason this can't happen).

adleong:
oh right, windows.

PR #260, opened 2016-04-10
Set default JVM_OPTIONS for linkerd and namerd

olix0r:
When JVM_OPTIONS is not specified at runtime, we provide default JVM_OPTIONS.
This should improve default behavior.

cc @stevej

### DIFF
--- a/project/LinkerdBuild.scala
+++ b/project/LinkerdBuild.scala
@@ -160,10 +160,43 @@ object LinkerdBuild extends Base {
       .dependsOn(core, admin, configCore)
       .withBuildProperties()
 
+    /**
+     * An assembly-running script that adds the namerd plugin directory
+     * to the classpath if it exists.
+     */
+    val execScript =
+      """|#!/bin/sh
+         |
+         |jars="$0"
+         |if [ -n "$NAMERD_HOME" ] && [ -d $NAMERD_HOME/plugins ]; then
+         |  for jar in $NAMERD_HOME/plugins/*.jar ; do
+         |    jars="$jars:$jar"
+         |  done
+         |fi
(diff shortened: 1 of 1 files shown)


project/LinkerdBuild.scala
```diff
@@ -323,9 +339,43 @@ object LinkerdBuild extends Base {
...
+         |   -XX:+AggressiveOpts                   \
+         |   -XX:+UseConcMarkSweepGC               \
+         |   -XX:+CMSParallelRemarkEnabled         \
+         |   -XX:+CMSClassUnloadingEnabled         \
+         |   -XX:+ScavengeBeforeFullGC             \
+         |   -XX:+CMSScavengeBeforeRemark          \
+         |   -XX:+UseCMSInitiatingOccupancyOnly    \
+         |   -XX:CMSInitiatingOccupancyFraction=70 \
+         |   -XX:ReservedCodeCacheSize=32m         \
+         |   -XX:+CMSClassUnloadingEnabled         \
```
olix0r: redundant

siggy: are you saying this is repeated in the namerd section and should be DRY'd up? if so, i agree. if not, i think it should.

stevej: Oh, I think Oliver means that +CMSClassUnloadingEnabled is on by default in Java 8. I missed that.

olix0r: oh, i'm saying that ` -XX:+CMSClassUnloadingEnabled` is specified twice ;)

I opted to declare flags separately for namerd and linkerd because they probably _should_ diverge...

Also, we may not want to set -Xms/mx here -- or if we do we  should support easy override?

siggy:
:star: :+1:

stevej:
:star: :+1:

PR #266, opened 2016-04-11
Update CHANGES.md

adleong:


CHANGES.md
```diff
@@ -1,3 +1,10 @@
+## 0.3.1
+
+* Added beta version of linkerd dashboard version 2.0 :chart_with_upwards_trend:
```
klingerf: How would you feel about calling out the `<admin-port>/dashboard` URL explicitly as part of this bullet point?

klingerf:
:star: :hammer: Thanks for putting this together -- just one optional comment.

CHANGES.md
```diff
@@ -1,3 +1,11 @@
+## 0.3.1
```
olix0r: generally, we shouldn't actually define a version until we're doing the release (so it should just say e.g. `## 0.X.X` -- it's clearer that it isn't released yet so that when someone else goes here they know not to create e.g. 0.3.2.  In this case it's fine, since we're prepping for the release (though you may want to include a placeholder marker for the next set of changes above this.

CHANGES.md
```diff
@@ -1,3 +1,11 @@
...
+
+* Added beta version of linkerd dashboard version 2.0.  Try it out at
+`/dashboard` on the linkerd admin site. :chart_with_upwards_trend:
+* Added zipkin trace initializer for configuring Finagle's zipkin-tracer
+* namerd's HTTP dtab API now supports the HEAD and DELETE methods
+* Tear-down address observations in namerd properly if a service is deleted
```
olix0r: tioli: "properly" feels awkward in this sentence

CHANGES.md
```diff
@@ -1,3 +1,11 @@
+## 0.3.1
+
+* Added beta version of linkerd dashboard version 2.0.  Try it out at
+`/dashboard` on the linkerd admin site. :chart_with_upwards_trend:
+* Added zipkin trace initializer for configuring Finagle's zipkin-tracer
```
olix0r: Prefer the present tense "Add .." to "added"

olix0r:
⭐ this fine, just some language nits

wmorgan:
Apologies for the bikeshed, but I think it's really valuable when items in the changelog describe not just what changed, but what the end user gets as a result of the changes.

E.g. "Added zipkin trace initializer for configuring Finagle's zipkin-tracer. Now you can automatically export Zipkin traces across all your services without having to write any code!" (Or whatever. I have no idea what that feature actually enables.)

klingerf:
For the zipkin bullet point, how about:

* Support Zipkin tracer configuration via config file, to enable automatic export of tracing data from linkerd to a Zipkin collector.

klingerf:
:star: Thanks for updating.

PR #267, opened 2016-04-11
Add toFuture method to Activity using implicit value class.

adleong:
Converting `Activity` to `Future` is something we do in a number of places and it's pretty verbose.


esbie:
:star: yay thank you :)

olix0r:
⭐

PR #269, opened 2016-04-11
Fix tls integration test.

adleong:
One of our TLS integration tests uses a mix of TLS and non-TLS backends.  This test fails without turning strict mode off.


esbie:
:star: good catch

olix0r:
⭐ -- can/should this test run in CI?

adleong:
Seems like the "integration" tests don't run in CI.  I guess because it might be flaky?  Not sure.

PR #274, opened 2016-04-12
Change default k8s master host

klingerf:
The default k8s config namer host value was using a deprecated kube-dns format that was removed in kubernetes 1.2.  Updating our default and documentation accordingly.  More info here:

https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns#backwards-compatibility


rmars:
:star: :bookmark_tabs:

esbie:
:star: :wind_chime: :dash:

PR #276, opened 2016-04-12
Add logging when requests to the k8s API fail

### LINKED ISSUE #272: k8s namer should increment a count and/or log when a request to the k8s API fails
In the case of a failed request to the k8s API, we infinitely retry because this connection to the k8s API is necessary for the correct operation of the k8s namer.  However, if the request fails for a non-transient reason (such as the k8s API can't be reached) then these requests will never succeed.  

To assist with debugging this kind of scenario, we should increment a counter and/or log some information every time a request to the k8s API fails.

adleong:
Fixes #272

### DIFF
--- a/linkerd/examples/k8s.yaml
+++ b/linkerd/examples/k8s.yaml
@@ -1,0 +1,19 @@
+admin:
+  port: 9990
+
+namers:
+- kind: io.l5d.experimental.k8s
+  host: localhost
+  port: 443
+  tls: true
+  tlsWithoutValidation: true
+  authTokenFile: linkerd/examples/io.l5d.k8s/kube_token
+
+routers:
+- protocol: http
+  baseDtab: |
+    /http/1.1/GET => /io.l5d.k8s/foo/http;
+  label: int
+  servers:
+  - port: 4140
+    ip: 0.0.0.0
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -95,7 +95,11 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => v1.NsApi) extends Namer {
             Future.Unit
           })
 
-          endpoints.foreach(services.initialize)
+          endpoints.onSuccess { list =>
+            services.initialize(list)
+          }.onFailure { e =>
+            log.error(e, "k8s failed to list endpoints")
+          }
         }
 
         init.foreach { init =>


linkerd/examples/k8s.yaml
```diff
@@ -0,0 +1,19 @@
+admin:
+  port: 9990
+
+namers:
+- kind: io.l5d.experimental.k8s
+  host: localhost
+  port: 443
+  tls: true
+  tlsWithoutValidation: true
+  authTokenFile: linkerd/examples/io.l5d.k8s/kube_token
```
esbie: any reason to have this file outside of acceptance-test.yaml?
if it's specific to a local kube setup, maybe document that

esbie:
lgtm just 1q

linkerd/examples/k8s.yaml
adleong: I found this was the easiest/fastest way to spin up a linkerd that used the k8s namer

`./sbt linkerd-examples/k8s:run`

I figured this might be useful to others as well.  I'm not super attached to it, though.

rmars:
:star: lgtm!

linkerd/examples/k8s.yaml
esbie: nothing inherently wrong with that
It's not tested though (acceptance-test.yaml is), so it could come back to bite us as these examples fall out of date

k8s/src/main/scala/io/buoyant/k8s/v1.scala
```diff
@@ -96,7 +96,7 @@ package object v1 {
       HighResTimer.Default,
       stats,
       RetryBudget.Infinite
-    )
+    ) andThen new LogFailureFilter(log, stats)
```
olix0r: totally minor tioli: i'd be inclined to separate this from infiniteRetryFilter and have i.e. `val logFailureFilter = ...` and then layer it in below as `retry and then logFailureFilter andThen client`

(Errors should still be logged, etc when retries aren't enabled).

linkerd/examples/k8s.yaml
adleong: That's a good point.  I don't think any of the examples other than acceptance-test.yaml are tested.  I created https://github.com/BuoyantIO/linkerd/issues/280 to track.

k8s/src/main/scala/io/buoyant/k8s/v1.scala
adleong: My thinking was that without the infinite retry filter catching all failures, a failure should bubble up and blow up or log at a higher level.  But probably there's no logging at the higher level either so I think you're probably right we should log failure here regardless of if we are swallowing them or not.

adleong: Actually I just noticed that there's already some logging in the `infiniteRetryFilter`'s predicate.  I removed `LogFailureFilter` entirely.

klingerf:
⭐ 🍴 Much better!

olix0r:
⭐

PR #285, opened 2016-04-13
Add a timeout to path binding

### LINKED ISSUE #273: linkerd should have a timeout for name resolution
If the Activity returned from `namer.bind` stays in the pending state forever, a linkerd request will hang forever.  We should add a timeout so that the request eventually fails.

adleong:
Add a configurable timeout to binding a path.  Requests will now no longer hang
forever if there is an error during binding and will instead fail with an error
message.

Fixes #273

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.{JsonIgnore, JsonProperty, JsonTypeInfo}
 import com.fasterxml.jackson.core.{io => _}
 import com.twitter.conversions.time._
 import com.twitter.finagle._
+import com.twitter.finagle.buoyant.DstBindingFactory
 import com.twitter.finagle.client.DefaultPool
 import com.twitter.finagle.service.{FailFastFactory, TimeoutFilter}
 import com.twitter.util.Closable
@@ -122,13 +123,21 @@ trait RouterConfig {
   def interpreter: InterpreterConfig =
     _interpreter.getOrElse(defaultInterpreter)
 
+  @JsonProperty("bindingTimeoutMs")
+  var _bindingTimeoutMs: Option[Int] = None
+
+  @JsonIgnore
+  def bindingTimeout = _bindingTimeoutMs.map(_.millis).getOrElse(10.seconds)
+
   @JsonIgnore
   def routerParams = Stack.Params.empty
     .maybeWith(baseDtab.map(dtab => RoutingFactory.BaseDtab(() => dtab)))
     .maybeWith(failFast.map(FailFastFactory.FailFast(_)))
     .maybeWith(timeoutMs.map(timeout => TimeoutFilter.Param(timeout.millis)))
     .maybeWith(dstPrefix.map(pfx => RoutingFactory.DstPrefix(Path.read(pfx))))
-    .maybeWith(client.map(_.clientParams)) + param.Label(label)
+    .maybeWith(client.map(_.clientParams)) +
+    param.Label(label) +
+    DstBindingFactory.BindingTimeout(bindingTimeout)
 
   @JsonIgnore
   def router(params: Stack.Params): Router = {
--- a/router/core/src/main/scala/com/twitter/finagle/buoyant/DstBindingFactory.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/buoyant/DstBindingFactory.scala
@@ -1,8 +1,9 @@
 package com.twitter.finagle.buoyant
 
-import com.twitter.finagle.{NoStacktrace => _, _}
 import com.twitter.finagle.naming._
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.finagle.{NoStacktrace => _, _}
 import com.twitter.logging.Logger
 import com.twitter.util._
 import java.util.concurrent.atomic.AtomicReference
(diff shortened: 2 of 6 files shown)


siggy:
:star: :+1: tioli update config.md

klingerf:
:star: :seedling: Looks good to me, and I think 10 seconds is a reasonable default.

router/core/src/main/scala/com/twitter/finagle/buoyant/DstBindingFactory.scala
```diff
@@ -126,12 +133,16 @@ object DstBindingFactory {
...
+    bindingTimeout: BindingTimeout = BindingTimeout.default
+  )(implicit timer: Timer = DefaultTimer.twitter) extends DstBindingFactory[Req, Rsp] {
     private[this]type Cache[Key] = XXX_ServiceFactoryCache[Key, Req, Rsp]
 
     def apply(dst: Dst, conn: ClientConnection): Future[Service[Req, Rsp]] = dst match {
-      case path: Dst.Path => pathCache(path, conn)
+      case path: Dst.Path => pathCache(path, conn).raiseWithin(
+        bindingTimeout.timeout,
+        new TimeoutException(s"Binding ${path.path.show} exceeded ${bindingTimeout.timeout}")
+      )
```
olix0r: should we special case the max timeout so we don't always set timeouts?

i.e.
```scala
val f = pathCache(path, conn)
bindingTimeout match {
  case BindingTimeout(Duration.Top) => f
  case BindingTimeout(timeout) => f.raiseWithin(timeout, exception)
}
```

I think we could also val this exception -- it doesn't need to be instantiated on each request but can just be a private on the class.

olix0r:
I think that this is good safeguard, but doesn't solve the underlying problem.  If a namer's backend isn't available, it shouldn't hang indefinitely -- it should return a NameTree.Neg so that another namer may be tried.

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -122,13 +123,21 @@ trait RouterConfig {
   def interpreter: InterpreterConfig =
     _interpreter.getOrElse(defaultInterpreter)
 
+  @JsonProperty("bindingTimeoutMs")
+  var _bindingTimeoutMs: Option[Int] = None
+
+  @JsonIgnore
+  def bindingTimeout = _bindingTimeoutMs.map(_.millis).getOrElse(10.seconds)
```
olix0r: should this default to `timeoutMs`?  It doesn't make sense for this to be higher than the request timeout.

router/core/src/main/scala/com/twitter/finagle/buoyant/DstBindingFactory.scala
adleong: `Future` already special cases this for us:

```
  def raiseWithin(timer: Timer, timeout: Duration, exc: Throwable): Future[A] = {
    if (timeout == Duration.Top || isDefined)
      return this
```

The exception message contains the path so it can't be val'd on the class.  I expected that the exception would be pass-by-name, but it looks like it's not :(  
Another option would be to make it extend `NoStackTrace` to somewhat reduce allocation.  Or remove the path from the message.

router/core/src/main/scala/com/twitter/finagle/buoyant/DstBindingFactory.scala
```diff
@@ -126,12 +133,16 @@ object DstBindingFactory {
...
+    bindingTimeout: BindingTimeout = BindingTimeout.default
+  )(implicit timer: Timer = DefaultTimer.twitter) extends DstBindingFactory[Req, Rsp] {
     private[this]type Cache[Key] = XXX_ServiceFactoryCache[Key, Req, Rsp]
 
     def apply(dst: Dst, conn: ClientConnection): Future[Service[Req, Rsp]] = dst match {
-      case path: Dst.Path => pathCache(path, conn)
+      case path: Dst.Path => pathCache(path, conn).raiseWithin(
+        bindingTimeout.timeout,
+        new TimeoutException(s"Binding ${path.path.show} exceeded ${bindingTimeout.timeout}") with NoStacktrace
+      )
```
olix0r: Yeah, my concern is that we're creating a new throwable on every lookup regardless of whether a timeout is set or the exception is actually thrown.  Perhaps use [RequestTimeoutException](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/Exceptions.scala#L108) which is mixes in TimeoutException with NoStacktrace?

router/core/src/main/scala/com/twitter/finagle/buoyant/DstBindingFactory.scala
```diff
@@ -126,12 +132,16 @@ object DstBindingFactory {
...
+    bindingTimeout: BindingTimeout = BindingTimeout.default
+  )(implicit timer: Timer = DefaultTimer.twitter) extends DstBindingFactory[Req, Rsp] {
     private[this]type Cache[Key] = XXX_ServiceFactoryCache[Key, Req, Rsp]
 
     def apply(dst: Dst, conn: ClientConnection): Future[Service[Req, Rsp]] = dst match {
-      case path: Dst.Path => pathCache(path, conn)
+      case path: Dst.Path => pathCache(path, conn).raiseWithin(
+        bindingTimeout.timeout,
+        new RequestTimeoutException(bindingTimeout.timeout, s"Binding ${path.path.show} exceeded timeout")
+      )
```
olix0r: style (tioli)

i prefer
```scala
  case path: Dst.Path =>
    val exc = new RequestTimeout..
    pathCache(path, conn).raiseWithin(bindingTimeout.timeout, exc)
```

olix0r:
:star:

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
adleong: I reverted this because I think it's confusing behavior.  Path binding happens outside of the request path so the request timeout doesn't apply to it.

PR #286, opened 2016-04-13
Add auth and acl support to ZkDtabStore

### LINKED ISSUE #249: send auth info in ZkDtabStore
Zk supports sending auth data.  We should configurably send something???

And ACLs should be configurable.  We need to think of a holistic solution for auth/ACL.

adleong:
Add a authInfo section to the zk dtab storage config where ZooKeeper
authenticaiton can be configured.  Add a acls section as well to configure
how ACLs are set on created znodes.

Fixes #249

### DIFF
--- a/namerd/examples/zk.yaml
+++ b/namerd/examples/zk.yaml
@@ -3,6 +3,13 @@ storage:
   pathPrefix: /dtabs
   hosts:
   - localhost:2181
+  authInfo:
+    scheme: digest
+    auth: user_123:password_123
+  acls:
+  - scheme: auth
+    id: ""
+    perms: crwda 
 interfaces:
 - kind: thriftNameInterpreter
 - kind: httpController
--- a/namerd/storage/zk/src/main/scala/io/buoyant/namerd/storage/experimental/ZkDtabStoreInitializer.scala
+++ b/namerd/storage/zk/src/main/scala/io/buoyant/namerd/storage/experimental/ZkDtabStoreInitializer.scala
@@ -12,7 +12,9 @@ class ZkDtabStoreInitializer extends DtabStoreInitializer {
 case class zk(
   hosts: Seq[String],
   pathPrefix: Option[String],
-  sessionTimeoutMs: Option[Int]
+  sessionTimeoutMs: Option[Int],
+  authInfo: Option[AuthInfo],
+  acls: Option[Seq[Acl]]
 ) extends DtabStoreConfig {
 
   @JsonIgnore val sessionTimeout = sessionTimeoutMs.map(_.millis)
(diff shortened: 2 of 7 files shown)


namerd/storage/zk/src/main/scala/io/buoyant/namerd/storage/experimental/ZkDtabStoreInitializer.scala
```diff
@@ -21,6 +23,20 @@ case class zk(
   override def mkDtabStore: DtabStore = new ZkDtabStore(
     hosts.mkString(","),
     pathPrefix.getOrElse("/dtabs"),
-    sessionTimeout
+    sessionTimeout,
+    authInfo,
+    acls.getOrElse(Seq(Acl.AnyoneUnsafe))
   )
```
siggy: maybe default to most restrictive?

namerd/examples/zk.yaml
```diff
@@ -3,6 +3,13 @@ storage:
   pathPrefix: /dtabs
   hosts:
   - localhost:2181
+  authInfo:
+    scheme: digest
+    auth: user_123:password_123
```
siggy: i guess we want to support this use case, but i imagine folks may not be comfortable putting their password in the clear in our config files. a better option would be reading an env var, not sure the best way to accomplish.

namerd/storage/zk/src/main/scala/io/buoyant/namerd/storage/experimental/ZkDtabStoreInitializer.scala
siggy: per offline discussion, onboard with this for now.

namerd/examples/zk.yaml
adleong: per offline discussion: digest auth is fundamentally insecure because it involves sending the password in cleartext: https://zookeeper.apache.org/doc/r3.1.2/zookeeperProgrammers.html#sc_BuiltinACLSchemes

But digest is probably the easiest to show off in an example.

siggy:
:star: :+1:

namerd/storage/zk/src/main/scala/io/buoyant/namerd/storage/experimental/ZkDtabStoreInitializer.scala
```diff
@@ -21,6 +23,20 @@ case class zk(
...
+
+case class AuthInfo(scheme: String, auth: String)
+
+case class Acl(scheme: String, id: String, perms: String) {
+  if (!Acl.validPerms(perms))
+    throw new IllegalArgumentException(s"$perms is not a valid permission string")
+}
+
+object Acl {
+  val validPerms = "[crwda]+".r.pattern.asPredicate.test _
```
gtcampbell: nit: this allows strings with duplicates that I assume are invalid, like "ccccc".

gtcampbell:
:+1:

PR #288, opened 2016-04-14
Add Interpreter into all build target.

adleong:


esbie:
:star: good catch

rmars:
:star: :ice_cream:

PR #290, opened 2016-04-14
Add the http Host header to access logs

benley:
This helps with interpreting logs when routing to backend services by hostname.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/AccessLogger.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/AccessLogger.scala
@@ -14,13 +14,14 @@ case class AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
     val user = "-"
     val referer = reqHeaders.getOrElse("Referer", "-")
     val userAgent = reqHeaders.getOrElse("User-Agent", "-")
+    var hostHeader = reqHeaders.getOrElse("Host", "-")
     val reqResource = s"${req.method.toString.toUpperCase} ${req.uri} ${req.version}"
 
     svc(req).onSuccess { rsp =>
       val statusCode = rsp.statusCode
       val responseBytes = rsp.contentLength.map(_.toString).getOrElse("-")
       val requestEndTime = new TimeFormat("dd/MM/yyyy:HH:mm:ss Z").format(Time.now)
-      log.info(s"""$remoteHost $identd $user [$requestEndTime] "$reqResource" $statusCode $responseBytes "$referer" "$userAgent"""")
+      log.info(s"""$hostHeader $remoteHost $identd $user [$requestEndTime] "$reqResource" $statusCode $responseBytes "$referer" "$userAgent"""")
     }
   }
 }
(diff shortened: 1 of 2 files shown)


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/AccessLogger.scala
```diff
@@ -14,13 +14,14 @@ case class AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
...
     val referer = reqHeaders.getOrElse("Referer", "-")
     val userAgent = reqHeaders.getOrElse("User-Agent", "-")
+    var hostHeader = reqHeaders.getOrElse("Host", "-")
     val reqResource = s"${req.method.toString.toUpperCase} ${req.uri} ${req.version}"
 
     svc(req).onSuccess { rsp =>
       val statusCode = rsp.statusCode
       val responseBytes = rsp.contentLength.map(_.toString).getOrElse("-")
       val requestEndTime = new TimeFormat("dd/MM/yyyy:HH:mm:ss Z").format(Time.now)
-      log.info(s"""$remoteHost $identd $user [$requestEndTime] "$reqResource" $statusCode $responseBytes "$referer" "$userAgent"""")
```
gtcampbell: We're currently logging in the [Combined Log Format](http://httpd.apache.org/docs/2.4/logs.html#combined), which extends the [Common Log Format](http://httpd.apache.org/docs/2.4/logs.html#common) by adding the referer and user-agent headers at the end. I *think* it would be better to add the host header at the end of the log string rather than at the beginning, as that would allow many tools that interact with the Common or Combined formats to continue working (assuming they ignore any additional fields at the end of the string, etc.). Thoughts?

benley: I'm neutral on this one - if putting it at the start violates the standards then I suppose let's not do that :-)

benley: shall I update the patch to put it at the end?

angrylogic:
FWIW this patch follows a suggested convention from the apache httpd: https://httpd.apache.org/docs/2.4/logs.html#virtualhost

There are existing tools like split-logfile which should be able to parse this format:
https://httpd.apache.org/docs/2.4/programs/split-logfile.html

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/AccessLogger.scala
angrylogic: FWIW this patch follows a suggested convention from the apache httpd: https://httpd.apache.org/docs/2.4/logs.html#virtualhost

There are existing tools like split-logfile which should be able to parse this format:
https://httpd.apache.org/docs/2.4/programs/split-logfile.html

gtcampbell: That seems reasonable (although of course the Host header is not a precise match for Apache vhosts). Concern withdrawn. :)

gtcampbell:
:+1: This looks fine to me, thanks for the discussion.

stevej:
👍 LGTM

olix0r:
lgtm

olix0r:
I think we still need you to sign our [Contributor License Agreement](https://buoyant.io/cla/) (as descrived in [CONTRIBUTING](https://github.com/BuoyantIO/linkerd/blob/master/CONTRIBUTING.md)) before we can merge this.  #lawyers

benley:
I did sign the CLA in "Company" mode - but I can sign as an individual too. Done.

olix0r:
@benley so you have, i was looking at the wrong spreadsheet.  Thanks!

PR #291, opened 2016-04-14
Added bind and addr http endpoints.

### LINKED ISSUE #223: namerd: support bind, addr, and delegator endpoints in httpController
It would be nice if the ctl interface could do name and address resolution, perhaps with streaming updates.

I'm envisioning being able to do something like
```
$ namectl bind default /http/1.1/GET/service
/io.l5d.k8s/default/http/service
```
Or
```
$ namectl bind --watch default /http/1.1/GET/service
/io.l5d.k8s/default/http/service
1 * /io.l5d.k8s/default/http/service-v2 & 5 * /io.l5d.k8s/default/http/service
/io.l5d.k8s/default/http/service-v2
```

adleong:
Add `/api/1/bind/<namespace>/<path>` endpoint to the HTTP control interface.  This endpoint returns a streaming response of bound paths for the given name.

Add `/api/1/addr/<namespace>/<path>` endpoint to the HTTP control interface.  This endpoint returns a streaming response of addresses for the given bound path.

Fixes #223


namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -106,14 +108,38 @@ object HttpControlService {
...
   }
 
+  trait NsPathUri {
+    val prefix: String
+
+    def unapply(path: String): Option[(Ns, Path)] = {
+      if (!path.startsWith(prefix)) return None
+      val segments = path.stripPrefix(prefix).split("/")
+      if (segments.size < 2) return None
+      Some((segments.head, Path.Utf8(segments.drop(1): _*)))
```
gtcampbell: tioli: I might avoid the early `return None`s and  write this instead as:

```scala
if (path.startsWith(prefix)) {
  path.stripPrefix(prefix).split("/").toSeq match {
    case head +: tail if tail.nonEmpty => Some(head, Path.Utf8(tail: _*))
    case _ => None
  }
} else {
  None
}
```

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/StreamingResponse.scala
```diff
@@ -0,0 +1,34 @@
...
+class StreamingResponse(writer: Writer) {
+  var closable: Closable = null
+  def close(): Unit = { closable.close() }
+  def start[T](data: Var[T])(toBuf: T => Buf): Unit = {
+    println("Starting streaming response " + this)
+    StreamingResponse.register(this)
+    closable = data.changes.respond { t =>
+      println(s"Streaming response $this got $t")
+      writer.write(toBuf(t)).onFailure { e =>
+        println(s"closing streaming response $this due to $e")
```
gtcampbell: some extraneous printlns here

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftInterpreterInterfaceConfig.scala
```diff
@@ -19,7 +19,7 @@ case class ThriftInterpreterInterfaceConfig(
   protected def defaultAddr = ThriftInterpreterInterfaceConfig.defaultAddr
 
   @JsonIgnore
-  def mk(interpeters: Ns => NameInterpreter, namers: Map[Path, Namer]): Servable = {
+  def mk(interpeters: Ns => NameInterpreter, namers: Map[Path, Namer], store: DtabStore): Servable = {
```
gtcampbell: Not your change, but it would be good to correct the spelling of "interpreters" here.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -225,4 +257,101 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
+  private[this] def streamingResp[T](values: Event[T], render: (T, Closable) => Buf): Future[Response] = {
+    val resp = Response()
+    resp.setChunked(true)
+    val writer = resp.writer
+    var closable: Closable = null
+    var writeFuture: Future[Unit] = Future.Unit
+    closable = values.respond { t =>
+      writeFuture = writeFuture.before {
+        writer.write(render(t, closable)).onFailure { _ =>
+          closable.close()
```
gtcampbell: I find this code a little difficult to understand, so maybe some comments would be in order. Are we guaranteed that `closable` will be non-`null` by the time this line of code is called? Why is writeFuture needed?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -225,4 +257,101 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
+    bindingCacheMu.synchronized {
+      val key = (ns, path)
+      bindingCache.get(key) match {
+        case Some(act) => act
+        case None =>
+          val act = namers(ns).bind(Dtab.empty, path)
+          bindingCache += (key -> act)
+          act
+      }
+    }
```
gtcampbell: I think this (and similar code elsewhere) could be simplified using a `java.util.concurrent.ConcurrentHashMap` (or alternately one of the Guava cache implementations, but I think `ConcurrentHashMap` gives us everything we want). The Java8 version of CHM includes a `computeIfAbsent` method that implements exactly the logic we have here, without the need for any extra `synchronized` blocks.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/StreamingResponse.scala
```diff
@@ -0,0 +1,34 @@
...
+  private[this] val AllMu = new {}
+  private[this] var All: Set[StreamingResponse] = Set.empty
+
+  def register(sr: StreamingResponse): Unit = AllMu.synchronized {
+    All += sr
+  }
+
+  def unregister(sr: StreamingResponse): Unit = AllMu.synchronized {
+    All -= sr
+  }
```
gtcampbell: This could also make use of `ConcurrentHashMap`:
```scala
object StreamingResponse {
  private[this] val set = ConcurrentHashMap.newKeySet[StreamingResponse]
  def register(sr: StreamingResponse): Unit = set.add(sr)
  def unregister(sr: StreamingResponse): Unit = set.remove(sr)
}
```

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
gtcampbell: To wit: I think you could do something like (assuming we can take on a dependency on `scala-java8-compat`; otherwise it would be slightly more verbose)

```scala
private[this] val bindingCache =
  new ConcurrentHashMap[(String, Path), Activity[NameTree[Name.Bound]]]
  
private[this] def getBind(ns: String, path: Path): Activity[NameTree[Name.Bound]] = {
  val fn: ((String, Path)) => Activity[NameTree[Name.Bound]] =
      (ns: String, path: Path) => namers(ns).bind(Dtab.empty, path)
  bindingCache.computeIfAbsent((ns, path), fn.asJava)
}
```

gtcampbell:
On the `ConcurrentHashMap`-related comments: I see we use a similar pattern in a number of other places, so if you'd prefer, we can address those all at once outside the context of this branch.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
adleong: Yeah, it's confusing.  It would be helpful to me to talk through it in person to help get my thoughts straight before taking a stab at commenting it.

olix0r: I think at the very least we should check that closable is not null.  Furthermore, because this is going to cross thread boundaries, i think closable has to be a @volatile....

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -133,6 +163,12 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
       handlePostDtab(ns, req)
     case (DtabUri(Some(ns)), Method.Delete) =>
       handleDeleteDtab(ns)
+    case (BindUri(ns, path), Method.Get) =>
+      handleGetBind(ns, path)
+    case (AddrUri(ns, path), Method.Get) =>
+      handleGetAddr(ns, path)
+    case (DelegateUri(ns, path), Method.Get) =>
+      handleGetDelegate(ns, path)
```
olix0r: can we add tests?

olix0r: would be nice to support polling requests as an alternative to streaming responses.  perhaps require `?stream=true`?

adleong: https://github.com/BuoyantIO/linkerd/issues/310

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
adleong: https://github.com/BuoyantIO/linkerd/issues/309

namerd/iface/control-http/src/test/scala/io/buoyant/namerd/iface/HttpControlServiceStreamingTest.scala
```diff
@@ -0,0 +1,73 @@
...
+import com.twitter.finagle.http.Request
+import com.twitter.finagle._
+import com.twitter.finagle.naming.NameInterpreter
+import com.twitter.io.{Reader, Buf}
+import com.twitter.util.{Return, Var, Activity}
+import io.buoyant.namerd.NullDtabStore
+import io.buoyant.test.Awaits
+import org.scalatest.FunSuite
+
+class HttpControlServiceStreamingTest extends FunSuite with Awaits {
```
olix0r: i'd prefer that this just be part of the HttpControlServiceTest so that all of the endpoints are enumerated in once place, and then test names should be in the same form...

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -106,14 +108,42 @@ object HttpControlService {
...
+  }
+
+  object AddrUri extends NsPathUri {
+    val prefix = "/api/1/addr/"
+  }
+
+  object DelegateUri extends NsPathUri {
+    val prefix = "/api/1/delegate/"
+  }
+
```
olix0r: I suspect that it's a fragile assumption that namespaces can't have slashes...  It would probably be more robust to have the paths be read from a  query param?

olix0r: To put a finer point on it: I expect that we're moving to a model where namespaces are paths.  We can reduce some api churn by anticipating that

adleong: good point

adleong: we already read namespaces from the URI eg `/api/1/dtabs/<ns>`.  Should we frontload the API churn for the sake of consistency and change these to `/api/1/dtabs?namespace=default` now?

Alternatively, we could have a transition period where we support both...

gtcampbell: hmm - we currently also use the namespace as the name for the object we store in Kubernetes in my thirdparty branch, and that's restricted to disallow most special characters other than `-`. Should that be amended to use something else as well?

olix0r: @adleong i think we should consider namespaces as resource trees.  we can create a dtab now (with some backends, at least).  e.g
```
:; namerctl dtab create other/ns - <<EOF
/srv       => /io.l5d.k8s/default/http ;
/host      => /srv ;
/method    => /$/io.buoyant.http.anyMethodPfx/host ;
/http/1.1  => /method ;
EOF
Created other/ns
:; namerctl dtab list
default
other/ns
```

The problem is when we're mixing Dtab Namespaces and Name.Path -- I'd opt to treat the Name.Path as an argument to the Dtab.

@gtcampbell we'll probably need a way of encoding slashes in the k8s store... (maybe just b64 the name and call it a day)

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -99,40 +99,39 @@ object HttpControlService {
...
-    def unapply(uri: String): Option[(Ns, Path)] = {
-      if (uri.startsWith(prefix)) {
-        uri.stripPrefix(prefix).split("/").toSeq match {
-          case ns +: path if path.nonEmpty => Some((ns, Path.Utf8(path: _*)))
-          case _ => None
-        }
+    def unapply(request: Request): Option[(Ns, Path)] = {
+      if (request.path.startsWith(prefix)) {
+        val ns = request.path.stripPrefix(prefix)
+        val path = Path.read(request.getParam("path"))
```
olix0r: this will blow up if the param is not a valid path -- does that cause a 400? (test please?)

olix0r:
⭐  very nice! i definitely think we should add non-streaming variants of these apis, but i'm fine if that happens in another branch

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -200,4 +245,111 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
+        case Some(act) => act
+        case None =>
+          val act = namers(ns).bind(Dtab.empty, path)
+          bindingCache += (key -> act)
+          act
+      }
+    }
+
+  private[this] val renderTryTree =
+    (tryTree: Try[NameTree[Name.Bound]], closable: Closable) => tryTree match {
```
gtcampbell: q: Why is this a val rather than a def?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -200,4 +245,111 @@ class HttpControlService(storage: DtabStore) extends Service[Request, Response]
...
       case Throw(_) => Future.value(Response(Status.InternalServerError))
     }
+
+  private[this] def streamingResp[T](values: Event[T], render: (T, Closable) => Buf): Future[Response] = {
+    val resp = Response()
+    resp.setChunked(true)
+    val writer = resp.writer
+    // closable is a handle to the values observation so that we can close the observation when the
+    // streaming connection is terminated
+    @volatile var closable: Closable = null
```
gtcampbell: q: Could this be `Closable.nop` to avoid the need for the null-check below?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -70,47 +70,92 @@ object HttpControlService {
...
+        }
+        Some(ns, path)
+      } else {
+        None
+      }
+    }
+  }
+
+  object BindUri extends NsPathUri {
+    val prefix = "/api/1/bind"
```
gtcampbell: tioli, tiny: might be good to put `"/api/1"` as a base API URL in `NsPathUri` and construct this and friends via string interpolation, rather than have it repeated for each prefix.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlServiceConfig.scala
```diff
@@ -1,12 +1,19 @@
...
+import com.twitter.finagle.{Namer, Path, Http, ListeningServer}
 import io.buoyant.namerd._
 import java.net.{InetAddress, InetSocketAddress}
 
-class HttpControlServiceConfig extends InterfaceConfig {
-  // note that `namers` is currently ignored here, but we may later wish to expose them via this interface
-  def mk(store: DtabStore, namers: Map[Path, Namer]) = HttpControlServable(addr, store)
+class HttpControlServiceConfig extends InterpreterInterfaceConfig {
+  @JsonIgnore override protected def mk(
+    delegate: (Ns) => NameInterpreter,
```
gtcampbell: parens unnecessary here

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlServiceConfig.scala
```diff
@@ -15,9 +22,13 @@ object HttpControlServiceConfig {
   val defaultAddr = new InetSocketAddress(InetAddress.getLoopbackAddress, 4180)
 }
 
-case class HttpControlServable(addr: InetSocketAddress, store: DtabStore) extends Servable {
+case class HttpControlServable(
+  addr: InetSocketAddress,
+  store: DtabStore,
+  namers: Ns => NameInterpreter
```
gtcampbell: tioli: naming-wise, it's a little odd that you call this `namers` in this constructor but `delegate` above in `HttpControlServiceConfig`, whereas `namers` is used for `Map[Path, Namer]` there. Not a huge deal, of course.

gtcampbell:
⭐  My comments are all minor/non-blocking.

PR #293, opened 2016-04-15
Log which tracer we're loading, replace util-codec Base64

stevej:
Apache's Base64 codec was the top allocator of byte[] based on profiling from load tests as the buffer size was wrong for our use case. Because it's not overridable, I replaced our use of it with the Java 8-only Base64. If we're not ok with asking users for Java 8, then we shouldn't merge this.


gtcampbell:
:star: I believe we already require JDK8, although others can correct me if I'm mistaken.

rmars:
:star: we do advise java 8 in our instructions (https://linkerd.io/doc/getting-started/index.html)

stevej:
Java 8 required! MERGING

PR #298, opened 2016-04-16
Remove sources copied from finagle

olix0r:
We had copied two files from finagle because they were inaccessible.  These
files have since been marked `private[finagle]`, so we can now use the original
implementations from com.twitter.finagle.buoyant.


gtcampbell:
:+1: :raised_hands:

adleong:
:star: 🎸

PR #300, opened 2016-04-17
Limit the namerd mux client to a single endpoint

### LINKED ISSUE #295: limit namerd thrift client to communicate with a single instance
A namerd thrift client will not behave well when it communicates with multiple namerd instances to observe a single name.  This is because stamps are not alligned across instances, so if i get back stamp 32 from one service and then poll again with stamp 32, this stamp is not guaranteed to be relevant on any other instance.  Rather than introduce additional coordination, I suggest that we just limit the connection pool to 1 connection.

Really, we'll need to move to a streaming API, but I'd like to wait to do that with gRPC

olix0r:
The thrift protocol implemented by namerd is designed for subsequent requests
to be sent to different nodes, as there are no guarantees that the 'stamp'
values used are shared across nodes. This watching behavior should later be
moved to a streaming interface. In the meantime, we use the Aperture load
balancer with an unlimited load so that only a single instance will be
considered. When a namerd instance is unavailable, another address in the pool
may be considered.

Fixes #295

### DIFF
--- a/interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
+++ b/interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore
 import com.twitter.conversions.time._
 import com.twitter.finagle.Stack.Transformer
 import com.twitter.finagle._
+import com.twitter.finagle.loadbalancer.Balancers
 import com.twitter.finagle.naming.NameInterpreter
 import com.twitter.finagle.param.HighResTimer
 import com.twitter.finagle.service._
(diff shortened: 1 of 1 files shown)


adleong:
Looks good.  Two questions:

1. How do we test this?
2. What happens if the namerd instance we're talking to goes down and we're forced to talk to a new one?

olix0r:
@adleong 
1. an integration test with multiple namerd ports over a network -- i can try to put that together
2. this should be okay.  when a connection is lost, the node should be marked down, and the client should retry on another endpoint.  at this point, the stamps likely won't match, and but once it receives the current stamps from the remote, it will resume watching properly.

adleong:
I guess the pathological case is where you reconnect to another namerd and the stamps happen to be the same even though the value is different.  In this case, the client would be stuck with a stale value until the next time it's updated.  Maybe we don't need to worry about this for now...

olix0r:
@adleong i think that's right. if we wanted to be rigorous, we could move retries up out of the client and wipe the stamp on error before retrying... i think this is all justifying moving to a streaming api sooner than later.

adleong:
This is less important because we plan to switch to a gRPC API.  However, should we still do this anyway?

siggy:
@adleong @olix0r: do we still want to do this? moving #295 to triage until we decide.

adleong:
Closed in favor of `io.l5d.mesh`

Issue #302, opened 2016-04-18
etcd namer


olix0r:

olix0r:
We have the ability to do this (we have an etcd client), but we don't have any authoritative service registry scheme for etcd. We should open tickets for each of these schemes, if they exist.

Issue #301, opened 2016-04-18
etcd namerd storage


olix0r:

olix0r:
https://github.com/BuoyantIO/linkerd/pull/362

PR #303, opened 2016-04-18
Introduce an etcd client

olix0r:
io.buoyant.etcd provides a client library over etcd's HTTP api. This will facilitate #301 and #302.


olix0r:
This code is actually some of Buoyant's oldest code, so please do give this a thorough review (esp for style).

etcd/src/integration/scala/io/buoyant/etcd/EtcdIntegrationTest.scala
```diff
@@ -0,0 +1,243 @@
...
+    if (process != null) {
+      process.destroy
+    }
+  }
+
+  def serverName =
+    Option(System.getProperty("ETCD_NAME")) match {
+      case None | Some("") => "/$/inet/127.1/4001"
+      case Some(name) => name
+    }
```
olix0r: probably should be able to configure the port that etcd listens on to avoid conflicts...

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
```diff
@@ -0,0 +1,243 @@
...
+
+/*
+ * TODO: use residual path on client to set key prefix.
+ */
+
+/*
+ * Wire-format representations of NodeOp and Node.
+ */
+
+case class Version(etcdserver: String, etcdcluster: String)
```
adleong: A small fraction of this file is actually Etcd.  Can we break this file up?

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
```diff
@@ -0,0 +1,243 @@
...
+        case Status.Ok =>
+          Future.const(readJson[Version](rsp.content))
+
+        case status =>
+          Future.exception(UnexpectedResponse(req.method, req.uri, Nil, status, EtcdState(0)))
+      }
+    }
+  }
+
+  def key(k: Path): Key = new Key(k, client)
```
adleong: Putting all the interesting methods (get, set, etc.) on Key was surprising to me.  I would have found it more intuitive if these methods were on Etcd and took a Key as a parameter.  Can you talk about the motivation for this structure?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,210 @@
...
+import com.twitter.finagle.http._
+import com.twitter.io.Buf
+import com.twitter.util._
+import scala.collection.JavaConverters._
+
+class Key(key: Path, client: Service[Request, Response]) {
+
+  import Etcd._
+
+  def path: Path = key
```
adleong: Using Finagle Paths as urls seems like a weird coupling.  Why not just `Seq[String]`?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,210 @@
...
+  private[this] val noParams = Seq.empty[(String, String)]
+
+  /** Get a key */
+  def get(
+    recursive: Boolean = false,
+    wait: Boolean = false,
+    waitIndex: Option[Long] = None,
+    quorum: Boolean = false
+  ): Future[NodeOp] = {
+    var params = noParams
```
adleong: suggestion:
```
Map(quarum -> "quarum", recursive -> "recursive", ...).collect {
  case (true, param) => "param" -> "true"
}
```

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,210 @@
...
+  import Etcd._
+
+  def path: Path = key
+
+  def key(path: Path): Key = new Key(key ++ path, client)
+  def key(name: String): Key = key(Path.read(name))
+
+  private[this] val noParams = Seq.empty[(String, String)]
+
+  /** Get a key */
```
adleong: Should aim for a consistent level of documentation on these operations.  Some have a lot, some have a little, some have none, some are scaladoc style, some aren't.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,210 @@
...
+   *
+   * If the `prevExist` flag is set to true, the node operation will fail if the node does not
+   * already exist.
+   */
+  def set(
+    value: Option[Buf],
+    ttl: Option[Duration] = None,
+    prevExist: Boolean = false
+  ): Future[NodeOp] = {
+    var params = noParams
```
adleong: Can eliminate this var, similar to above.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,210 @@
...
+    for (ttl <- ttl) {
+      val v = if (ttl == Duration.Zero) "" else ttl.inSeconds.toString
+      params = params :+ "ttl" -> v
+      if (prevExist) {
+        params = params :+ "prevExist" -> "true"
+      }
+    }
+
+    val req = mkReq(keysPrefixPath ++ key, Method.Put, params)
+    client(req) flatMap (toNodeOp(req, _, key, params))
```
adleong: avoid infix notation (here and elsewhere)

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,210 @@
...
+    recursive: Boolean = false,
+    backoff: Stream[Duration] = Stream.empty
+  ): Event[Try[NodeOp]] = new Event[Try[NodeOp]] {
+    private[this] val origBackoff = backoff
+
+    def register(witness: Witness[Try[NodeOp]]) = {
+      @volatile var closing = false
+      @volatile var currentOp: Future[NodeOp] = Future.never
+
+      def loop(idx: Option[Long], backoff: Stream[Duration]): Unit =
```
adleong: I think this starts a polling loop for every witness registered to the event.  What about creating a `Var.async` instead and then returning `that.changes`?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,210 @@
...
+      Closable.make { _ =>
+        closing = true
+        currentOp.raise(new FutureCancelledException)
+        Future.Unit
+      }
+    }
+
+  }
+
+  def watch(backoff: Stream[Duration] = Stream.empty): Activity[NodeOp] =
```
adleong: do we need both `watch` and `events`?  I think we should pick whether we'd prefer to expose an `Activity` or an `Event` and just do one.

etcd/src/main/scala/io/buoyant/etcd/NodeOp.scala
```diff
@@ -0,0 +1,53 @@
+package io.buoyant.etcd
+
+import com.twitter.util._
+
+sealed class Action(val name: String) {
```
adleong: should this be abstract?

etcd/src/main/scala/io/buoyant/etcd/NodeOp.scala
```diff
@@ -0,0 +1,53 @@
...
+  )
+
+  val ByName: Map[String, Action] =
+    All.map(a => a.name -> a).toMap
+
+  case class Invalid(name: String)
+    extends Exception(s"Invalid action: $name")
+
+  def apply(name: String): Try[Action] =
+    ByName.get(name).map(Return(_)) getOrElse Throw(Invalid(name))
```
adleong: suggestion: `Try.orThrow(ByName.get(name))(() => Invalid(name))`

etcd/src/main/scala/io/buoyant/etcd/NodeOp.scala
```diff
@@ -0,0 +1,53 @@
...
+}
+
+case class EtcdState(
```
adleong: could use some docs here, I'm not sure what index represents

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
```diff
@@ -0,0 +1,243 @@
...
+    val Buf.ByteArray.Owned(bytes, begin, end) = Buf.ByteArray.coerce(buf)
+    Try(mapper.readValue[T](bytes, begin, end - begin))
+  }
+
+  private[etcd] def toNodeOp(
+    req: Request,
+    rsp: Response,
+    key: Path,
+    params: Seq[(String, String)] = Seq.empty
+  ): Future[NodeOp] = {
```
adleong: could this return Try[NodeOp] instead of Future[NodeOp]?

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
olix0r: This is similar to how we designed the util-zk api at twitter.  It allows some nice separation of concerns (key selection vs actions).

For instance, if we want to "chroot" a module (i.e. lock into a prefix), we don't have to rely on the module to manage this itself (and vigilantly apply path prefixes).  We can just do:
```scala
val root = etcd.key("/some/root")
```

and then if we path the module `root`, all of its operations will occur under /some/root.

This could all use some better documentation/explanation, but I like this pattern.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: The library deals in finagle primitives (services, futures, etc), so I'm not sure how severe of a coupling this is.  I suppose this puts some additional requirements on permitted characters.  This all came from a Namer implementation, so it was expected it would receive paths.  I think we'd want to maintain support for passing paths to it, though it may be better to support Seq[String] as the underlying primitive.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: After going back on forth on this for a while, I think it would be needlessly restrictive to drop support for events---it exposes etcd's API with minimal interference.  I suspect that it makes sense to provide helpers over this API to model updating states of a single node, but I've removed the watch helper for now until we have a concrete use case for it (to determine whether it makes sense -- I suspect we care more about `def dataWatch: Activity[Option[Buf]]` and `def dirWatch(recursive: Boolean): Activity[Option[Dir]]`)

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: I've documented this, but I think this is correct given the semantics of the API

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: In this case, i think it's probably more readable than the alternative?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,278 @@
...
+
+  def key(path: Path): Key = new Key(key ++ path, client)
+  def key(name: String): Key = key(Path.read(name))
+
+  private[this] lazy val uriPath = keysPrefixPath ++ key
+
+  private[this] def boolParam(cond: Boolean) = if (cond) Some("true") else None
+  private[this] def valueParams(v: Option[Buf]): Seq[(String, String)] = v match {
+    case Some(Buf.Utf8(v)) => Seq("value" -> v)
+    case _ => Seq("dir" -> "true")
```
adleong: This case applies when v is defined but is not Utf8 encoded.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,278 @@
...
+  def set(
+    value: Option[Buf],
+    ttl: Option[Duration] = None,
+    prevExist: Boolean = false
+  ): Future[NodeOp] = {
+    var params = valueParams(value)
+    for (ttl <- ttl) {
+      val v = if (ttl == Duration.Zero) "" else ttl.inSeconds.toString
+      params = params :+ "ttl" -> v
+      if (prevExist) {
```
adleong: is it intentional that prevExist only gets set if ttl is defined?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,278 @@
...
+   * node after some time period (only second-granularity is supported
+   * by etcd).
+   */
+  def create(
+    value: Option[Buf],
+    ttl: Option[Duration] = None
+  ): Future[NodeOp] = {
+    val vp = valueParams(value)
+    val tp = ttl.map("ttl" -> _.inSeconds.toString)
+    val params = vp ++ Seq(tp).flatten
```
adleong: this can just be `vp ++ tp`

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,278 @@
...
+  import Etcd._
+
+  def path: Path = key
+
+  def key(path: Path): Key = new Key(key ++ path, client)
+  def key(name: String): Key = key(Path.read(name))
+
+  private[this] lazy val uriPath = keysPrefixPath ++ key
+
+  private[this] def boolParam(cond: Boolean) = if (cond) Some("true") else None
```
adleong: these are each used in one method respectively.  I might move them inside or nearby those methods for readability.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,278 @@
...
+   * If `recursive` is true, the key's subtree is observed for
+   * changes.
+   *
+   * When an unexpected error is encountered communicating with the
+   * API, the failure is published on the Event and the `backoff`
+   * stream is used to compute the time to wait before retrying. If
+   * the `backoff` stream is exhausted or a fatal error is
+   * encountered, it is reported and polling stops.
+   *
+   * The Event is not reference-counted, so each observer initiates
```
adleong: Thanks for spelling this out.  Makes sense now. 👍

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,278 @@
...
+    recursive: Boolean = false,
+    backoff: Stream[Duration] = Stream.empty
+  ): Event[Try[NodeOp]] = new Event[Try[NodeOp]] {
+    private[this] val origBackoff = backoff
+
+    def register(witness: Witness[Try[NodeOp]]) = {
+      @volatile var closed = false
+      @volatile var currentOp: Future[NodeOp] = Future.never
+
+      def loop(idx: Option[Long], backoff: Stream[Duration]): Unit =
```
adleong: TIOLI: This could return `Future[Unit]` and use `flatMap` instead of `respond`.  That way you would't need `currentOp` and could just raise on the total `Future[Unit]` returned from the original invocation of `loop`.

etcd/src/main/scala/io/buoyant/etcd/Node.scala
```diff
@@ -0,0 +1,107 @@
...
+    modifiedIndex: Long,
+    createdIndex: Long,
+    lease: Option[Lease] = None,
+    value: Buf = Buf.Empty
+  ) extends Node
+
+  /**
+   * Representation of a Node, as returned from etcd.
+   */
+  private[etcd] case class Repr(
```
adleong: I'd prefer spelling out Representation in full

etcd/src/main/scala/io/buoyant/etcd/Node.scala
```diff
@@ -0,0 +1,107 @@
...
+
+    def toLease: Try[Option[Lease]] =
+      for {
+        expiration <- Try {
+          expiration.map { e =>
+            val ms = ISODateTimeFormat.dateTime.parseDateTime(e).getMillis
+            Time.fromMilliseconds(ms)
+          }
+        }
+        ttl <- Try(ttl.map(_.toInt.seconds))
```
adleong: already an int

etcd/src/main/scala/io/buoyant/etcd/NodeOp.scala
```diff
@@ -0,0 +1,111 @@
...
+                  case None => Return(None)
+                  case Some(prev) => prev.toNode.map(Some(_))
+                }
+                prev.map(NodeOp(action, node, etcd, _))
+            }
+        }
+      }
+  }
+
+  private[etcd] object Repr {
```
adleong: I prefer Representation

etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
```diff
@@ -0,0 +1,444 @@
+package io.buoyant.etcd
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.http._
```
adleong: I think the Version imported here conflicts with `i.b.etcd.Version`.  Also I think some of these imports are unused.

project/LinkerdBuild.scala
```diff
@@ -252,7 +244,7 @@ object LinkerdBuild extends Base {
     )
 
     val all = projectDir("namerd")
-      .aggregate(core, Storage.all, Iface.all, main)
+      .aggregate(core, dcosBootstrap, Storage.all, Iface.all, main)
```
adleong: unrelated?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,278 @@
...
+    for (i <- prevIndex) {
+      params = params :+ "prevIndex" -> i.toString
+    }
+    for (Buf.Utf8(v) <- prevValue) {
+      params = params :+ "prevValue" -> v
+    }
+    if (!prevExist) {
+      params = params :+ "prevExist" -> "false"
+    }
+    val Buf.Utf8(v) = value
```
adleong: I think this blows up with a MatchError if value is not Utf8 encoded.  Maybe worth checking first or at least documenting?

adleong:
:star: Thanks for updating.  Just nits remain.  I like the way this looks.  🌾

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: dirty secret: Buf.Utf8 always matches

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: TIL, yes

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: i have some PTSD about this since chained futures once had a subtle reference leak... i'll try to convince myself it's safe to do this.

project/LinkerdBuild.scala
olix0r: yeah, we should fix this outside of this review

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: yeah, this always matches, just sometimes with funny characters.  turns out it's really hard to test whether a byte array contains a utf8 string

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: incomplete sentence in "An Event that reflects"...

etcd/src/integration/scala/io/buoyant/etcd/EtcdIntegrationTest.scala
```diff
@@ -0,0 +1,194 @@
...
+import java.util.UUID
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.fixture.FunSuite
+import scala.sys.process.{Process, ProcessLogger}
+import scala.util.Random
+
+/**
+ * Etcd client integration tests.
+ *
+ * Boots an etcd instance on localhost:4001.
```
siggy: change this to random or hardcode 4001 below.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+  private val Params = Seq.empty[(String, String)]
+
+  private def trueParam(name: String, cond: Boolean): Option[(String, String)] =
+    if (cond) Some(name -> "true") else None
+
+  private def falseParam(name: String, cond: Boolean): Option[(String, String)] =
+    if (!cond) Some(name -> "false") else None
+
+  private[this] val dirParam = Some("dir" -> "true")
+  private def dirOrValueParam(v: Option[Buf]): Option[(String, String)] = v match {
```
adleong: I think this return type can just be `(String, String)`

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+          get(recursive, wait = idx.isDefined, waitIndex = idx).transform {
+            case note@Return(op) =>
+              witness.notify(note)
+              loop(Some(getIndex(op.node) + 1), origBackoff)
+
+            case note@Throw(ApiError(ApiError.KeyNotFound, _, _, idx)) =>
+              witness.notify(note)
+              loop(Some(idx + 1), origBackoff)
+
+            case note@Throw(ApiError(ApiError.EventIndexCleared, _, _, idx)) =>
```
adleong: should the witness be updated here?  If not, `note` is unused.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+                  witness.notify(note)
+                  loop(None, backoff)
+
+                case _ =>
+                  witness.notify(Throw(BackoffsExhausted(key, e)))
+                  Future.Unit
+              }
+
+            case note@Throw(_) =>
+              witness.notify(note)
```
adleong: should we log the fatal error?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+    recursive: Boolean = false,
+    backoff: Stream[Duration] = Stream.empty
+  ): Event[Try[NodeOp]] = new Event[Try[NodeOp]] {
+    private[this] val origBackoff = backoff
+
+    def register(witness: Witness[Try[NodeOp]]) = {
+      @volatile var closed = false
+
+      def loop(idx: Option[Long], backoff: Stream[Duration]): Future[Unit] =
+        if (!closed) {
```
adleong: do we need this? Or is raising sufficient to stop things?

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
```diff
@@ -0,0 +1,115 @@
...
+case class Version(etcdserver: String, etcdcluster: String)
+
+case class UnexpectedResponse(
+  method: Method,
+  uri: String,
+  params: Seq[(String, String)],
+  status: Status,
+  state: Etcd.State
+) extends Exception({
+  val ps = params.map { case (k, v) => s"$k -> $v" }.mkString("(", "), (", ")")
```
adleong: TIOLI:

```
params.map { case (k, v) => s"($k -> $v)" }.mkString(", ")
```

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: raising is advisory and isn't guaranteed to be delivered.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
olix0r: thought was that the consumer of this can log... (we don't have a logger in this module as it is)

etcd/src/main/scala/io/buoyant/etcd/ApiError.scala
```diff
@@ -0,0 +1,64 @@
+package io.buoyant.etcd
+
+case class ApiError(
+  errorCode: Int,
+  message: String,
+  cause: String,
+  index: Int
+) extends Exception(message)
+
+// see: https://github.com/coreos/etcd/blob/master/Documentation/errorcode.md
```
gtcampbell: This appears to be a broken link.

etcd/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+The `io.buoyant.etcd` package contains a generically useful Etcd
+client built on finagle-http. The client supports a variety of
+asynchronous `key` operations on etcd.  It goes beyond the basic etcd
+operations to provide a `watch` utility that models a key's (or
+tree's) state as an `com.twitter.util.Activity`.
+
+### TODO ###
+
+- `compareAndDelete` isn't yet implemented.  minor oversight.
```
gtcampbell: tioli: link to the [API documentation for compareAndDelete](https://coreos.com/etcd/docs/latest/api.html#atomic-compare-and-delete) here?

etcd/README.md
```diff
@@ -0,0 +1,17 @@
...
+- Use residual paths on the client to instrument chrooted key
+operations
```
gtcampbell: I'm not sure I understand what this means.

olix0r: clients can be constructed with names that have residual paths.  I.e. i can create a client with `/$/inet/127.1/4001` or `/$/inet/127.1/4001/some/prefix`.  These both bind clients on 127.1:4001.  The latter is bound with a residual path, `/some/prefix`.  It would be cool to use this as a prefix for all etcd operations.  I don't have a concrete use case for this right now, though.

etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
```diff
@@ -0,0 +1,441 @@
...
+      Some((recursive, idx))
+    }
+  }
+
+  test("Etcd.version") {
+    val uri = Path.Utf8("version")
+    val version = Version("a.b.c.d", "a")
+    val etcd = mkEtcd() { case (Method.Get, `uri`, _) => version }
+    val v = etcd.version()
+    assert(Await.result(v, 250.millis) == version)
```
klingerf: Would it be worth mixing Awaits into this test class? 250ms is the default, so you could just write this as `assert(await(v) == version)`.

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
```diff
@@ -0,0 +1,115 @@
...
+
+case class Version(etcdserver: String, etcdcluster: String)
+
+case class UnexpectedResponse(
+  method: Method,
+  uri: String,
+  params: Seq[(String, String)],
+  status: Status,
+  state: Etcd.State
+) extends Exception({
```
gtcampbell: tioli: is there something more specific than `Exception` we can extend here?

project/LinkerdBuild.scala
```diff
@@ -529,6 +521,7 @@ object LinkerdBuild extends Base {
 
   // Unified documentation via the sbt-unidoc plugin
   val all = project("all", file("."))
+    .settings(unidocSettings)
     .aggregate(
       admin,
       configCore,
```
klingerf: Should the aggregate list for the all project include etcd? It seems to include everything else defined in this file. And I don't think we'll run tests for etcd by default if it's omitted.

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
```diff
@@ -0,0 +1,115 @@
...
+  val ps = params.map { case (k, v) => s"($k -> $v)" }.mkString(", ")
+  s"""$method $uri [$ps] $status"""
+})
+
+object Etcd {
+
+  private[etcd] val keysPrefixPath = Path.Utf8("v2", "keys")
+  private[etcd] val versionPath = Path.Utf8("version")
```
gtcampbell: I will echo @adleong's comment on an earlier revision of this branch: it strikes me as weird to be using a Finagle `Path` to represent a URI, both because it's surprising semantically and because they have different formats (although in this case of course they map to the same string).

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
olix0r: ideas?

project/LinkerdBuild.scala
olix0r: oops, lost in a rebase.

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
```diff
@@ -0,0 +1,115 @@
...
+ * The client service is responsible for setting the Host: header, etc...
+ */
+class Etcd(client: Service[Request, Response]) extends Closable {
+
+  import Etcd._
+
+  /** Fetch etcd server version info */
+  def version(): Future[Version] = {
+    val req = mkReq(versionPath)
+    req.headerMap("accept") = MediaType.Json
```
gtcampbell: tioli: `request` exposes a `accept_=` method

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+
+  private val Params = Seq.empty[(String, String)]
+
+  private def trueParam(name: String, cond: Boolean): Option[(String, String)] =
+    if (cond) Some(name -> "true") else None
+
+  private def falseParam(name: String, cond: Boolean): Option[(String, String)] =
+    if (!cond) Some(name -> "false") else None
+
+  private[this] val dirParam = ("dir" -> "true")
```
gtcampbell: tioli: don't need parens here

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+  private def dirOrValueParam(v: Option[Buf]): (String, String) = v match {
+    case None => dirParam
+    case Some(buf) =>
+      val Buf.Utf8(v) = buf
+      ("value" -> v)
```
gtcampbell: (or here)

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  private def ttlParam(ttl: Option[Duration]): Option[(String, String)] = ttl.map {
+    case Duration.Zero => "ttl" -> ""
+    case ttl => "ttl" -> ttl.inSeconds.toString
```
gtcampbell: tioli: I prefer to avoid shadowing, so maybe it would be good to use a name other than "ttl" here?

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+    waitIndex: Option[Long] = None,
+    quorum: Boolean = false
+  ): Future[NodeOp] = {
+    val params = Params ++
+      trueParam("recursive", recursive) ++
+      trueParam("quorum", quorum) ++
+      trueParam("wait", wait) ++
+      waitIndex.map("waitIndex" -> _.toString)
+    val req = mkReq(uriPath, params = params)
+    req.headerMap("accept") = MediaType.Json
```
gtcampbell: could also use `req.accept =` here

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
olix0r: i think it's likely we should change this, but i think we can do so in a follow-up.  (It would be nice to continue to accept Paths but it shouldn't be the only way to refer to keys...)

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+    client(req).flatMap { rsp => Future.const(NodeOp.mk(req, rsp, key, params)) }
```
gtcampbell: tioli: this line is repeated a bunch throughout this file and could be extracted to a private helper method. not a huge deal since it's a one-liner.

etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -0,0 +1,282 @@
...
+   * the `backoff` stream is exhausted or a fatal error is
+   * encountered, it is reported and polling stops.
+   *
+   * The Event is not reference-counted, so each observer initiates
+   * its own polling loop. This ensures that the initial state of a
+   * tree is reported properly.
+   */
+  def events(
+    recursive: Boolean = false,
+    backoff: Stream[Duration] = Stream.empty
```
gtcampbell: should we default to something sensible like `Backoff.exponentialJittered`?

etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
```diff
@@ -0,0 +1,441 @@
...
+import com.twitter.conversions.time._
+import com.twitter.finagle.{Filter, Path, Service}
+import com.twitter.finagle.buoyant.FormParams
+import com.twitter.finagle.http.{Message, Method, Request, Response, Status}
+import com.twitter.io.Buf
+import com.twitter.util.{Events => _, _}
+import io.buoyant.test.Events
+import org.scalatest._
+
+class KeyTest extends FunSuite with ParallelTestExecution {
```
gtcampbell: I appreciate the comprehensiveness of this test!

etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+  private[this] def addState(etcd: Etcd.State, msg: Message): Unit = {
+    msg.headerMap("X-Etcd-Index") = etcd.index.toString
+    msg.headerMap("X-Etcd-Cluster-Id") = etcd.clusterId
+  }
+
+  private[this] object Recursive {
+    def unapply(params: Params): Option[Boolean] =
+      Some(getParam(params, "recursive").exists(_ == "true"))
```
gtcampbell: tioli, code-golf: `contains("true")` is slightly fewer characters (and also slightly more efficient, not that matters here in a test)

adleong: on the other hand, `exists` is more type-safe than `contains`  :bike: 🏠

gtcampbell: true, but not if you're using the non-type-safe `==` inside it :)

etcd/src/main/scala/io/buoyant/etcd/Etcd.scala
gtcampbell: I was thinking of having a parent trait or abstract class for etcd exceptions, but maybe that's not especially meaningful.

etcd/README.md
gtcampbell: Got it - I think I was thrown by the word "instrument".

gtcampbell:
:star: Looking back at my comments, they're all non-blocking style stuff.

PR #305, opened 2016-04-18
Simplify address observation

adleong:
Since address observations are a `flatMap` of `bind` with the `Var[Addr]`, they
should always be correct, even if the `NameTree` becomes `Neg` and then comes back
with a new `Var[Addr]` (as in the case where a concrete name is deleted and then
recreated).  This means we can simplify the code by not evicting cached
observations when the `NameTree` becomes `Neg`.

### DIFF
--- a/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
+++ b/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
@@ -4,6 +4,7 @@ import com.twitter.finagle.naming.NameInterpreter
 import com.twitter.finagle.tracing.Trace
 import com.twitter.finagle.{Addr, Address, Dentry, Dtab, Name, Namer, NameTree, Path}
 import com.twitter.io.Buf
+import com.twitter.logging.Logger
 import com.twitter.util._
 import io.buoyant.namerd.iface.{thriftscala => thrift}
 import io.buoyant.namerd.Ns
@@ -229,6 +230,8 @@ class ThriftNamerInterface(
 ) extends thrift.Namer.FutureIface {
   import ThriftNamerInterface._
 
+  private[this] val log = Logger.get(getClass.getName)
+
   /*
    * We keep a cache of observations.  Each observer keeps track of
    * the most recently observed state, and provides a Future that will
@@ -262,15 +265,15 @@ class ThriftNamerInterface(
 
         val bindingObserver = observeBind(ns, dtab, path)
         bindingObserver(reqStamp).transform {
-          case Throw(e) =>
-            Trace.recordBinary("namerd.srv/bind.fail", e.toString)
-            val failure = thrift.BindFailure(e.getMessage, retryIn().inSeconds, ref, ns)
-            Future.exception(failure)
-
           case Return((TStamp(tstamp), nameTree)) =>
             Trace.recordBinary("namerd.srv/bind.tree", nameTree.show)
             val (root, nodes, _) = mkTree(nameTree)
             Future.value(thrift.Bound(tstamp, thrift.BoundTree(root, nodes), ns))
+
+          case Throw(e) =>
+            Trace.recordBinary("namerd.srv/bind.fail", e.toString)
+            val failure = thrift.BindFailure(e.getMessage, retryIn().inSeconds, ref, ns)
+            Future.exception(failure)
         }
     }
   }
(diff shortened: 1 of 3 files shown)


namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -100,14 +100,16 @@ object ThriftNamerInterface {
...
-      val promise = synchronized {
-        val previous = pending
-        current = Some(obs)
-        pending = new Promise[Observation]
-        previous
+      if (!current.exists(_.value == v)) {
+        val obs = Observation(nextStamp(), v)
+        val promise = synchronized {
+          val previous = pending
+          current = Some(obs)
```
olix0r: this is doing a read and then a write on current, which I believe needs to be synchronized.  Two threads could read `current`, and evaluate to false, and then both of them would create a new Observation, and an arbitrary one would win, causing setValue to be called twice, which throws an exception.

olix0r: Hmm.. maybe i'm misreading this.  Still the, read/write strikes me as racey.  I think we need to at least mark `current` as `@volatile` if it's being accessed outside of synchronized.  We should think about how this handles with concurrent updates, though...

adleong: I see what you mean.  I don't immediately see how this could go wrong, but that doesn't mean that it can't.  To be safe I've expanded the synchronized block to the whole method.  In reality there should only be one updater so I don't think we'll really need to worry about contention.  Better thread-safe than sorry.

olix0r: yeah, i think the real issue is that i _think_ it would have to be volatile if outside of the synchronized block to ensure visibility? i'm not a jmm lawyer...

olix0r:
lgtm -- how do we test this? (seems like we'll want to vet this against the new EndToEndTest...)

namerd/iface/interpreter-thrift/src/test/scala/io/buoyant/namerd/iface/ThriftNamerEndToEndTest.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    val serverAddr1 = Var[Addr](Addr.Pending)
+    serverState() = Activity.Ok(NameTree.Leaf(Name.Bound(serverAddr1, id)))
+    eventually { assert(clientState == serverState.sample()) }
+    val Activity.Ok(NameTree.Leaf(bound1)) = clientState
+    assert(bound1.id == id)
+
+    @volatile var clientAddr1: Addr = Addr.Pending
+    bound1.addr.changes.respond(clientAddr1 = _)
+
```
adleong: I removed an assert here where we were checking that `clientAddr0` was `Pending`.  I think that once the `NameTree` changes to no longer include the old Var `clientAdd0` then we can no longer make any assumptions about it.  

In practice we should never just sample an `Activity[NameTree[_]]` and watch a `Var` we find inside.  Instead we should always `flatMap` the `Activity[NameTree[_]]` to the `Var[Addr]` to get an observation that will always be correct.

olix0r: i think that was intended to be an assert on `clientAddr1` and not `clientAddr0`...

adleong: That makes sense.  With `clientAddr1` this test fails.  That's because we don't have an explicit pending state: when an `Activity` is pending, we simply don't send any updates.  This means that if the previous `Var[Addr]` becomes `Neg`, we'll keep that value until the new `Var` takes a non-pending value.  Let me know if you disagree, but I think this behavior is okay.

olix0r: although, it's probably okay to assert that clientAddr1 is Neg... sup to you.

gtcampbell:
⭐  (although I am still not a true master of `Activity` and `Var`, this all seems reasonable to me).

PR #315, opened 2016-04-19
Don't use Activity for a simple cache

adleong:
The EndpointsNamer keeps a cache of NsCache objects keyed by name.  We can keep
this cache in a simple Map instead of an Activity to simplify the code and make
it easier to reason about.


olix0r:
:star: good catch

esbie:
⭐  😍 
is it possible to do the same with SvcCache ports?

adleong:
Good question.  I don't think so because ports is exposed and used as an Activity elsewhere.  I'll see what else can be done to simplify, though.

olix0r:
The list of namespaces is updated by a watch. It's a simple cache of namespace observations.  Once we're observing a namespace, the list of services and ports can change dynamically, and so (i think) we need to be able to reflect those updates...  (I'm not saying it can't be simplified, just the motivation is that this is something which can be updated by the watch).

PR #316, opened 2016-04-19
Simplify the NsCache interface

adleong:
Simplify the NsCache interface by making the state Var private and exposing
an Activity instead.


k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -227,19 +227,22 @@ private object EndpointsNamer {
...
   }
 
-  class NsCache(name: String, activity: ActUp[Map[String, SvcCache]]) {
+  class NsCache(name: String) {
 
-    def services: Var[Activity.State[Map[String, SvcCache]]] = activity
+    private[this] val state = Var[Activity.State[Map[String, SvcCache]]](Activity.Pending)
+    private[this] val activity = Activity(state)
+
+    def services: Activity[Map[String, SvcCache]] = activity
```
olix0r: tioli: this could just be `val services = Activity(state)`. `activity` isn't used otherwise.

PR #317, opened 2016-04-20
fix io.l5d.serverset name ids

olix0r:
The ServersetNamer had a hardcoded idPrefix, `/$/io.l5d.serversets`.  This meant that config-specified prefixes were ignored.  Worse yet, the prefix included /$/, which caused resolutions of the id to be searched on the classpath (which failed, since io.l5d.serversets is a NamerInitializer and not a Namer).

Passing the idPrefix into the ServersetNamer constructor resolves this.

This did not cause problems when linkerd resolved serverset names, but made it impossible for serverset names to be resolved through namerd (since the id was not resolvable).

In doing this, we improved error logging in namerd's thrift interface.


klingerf:
⭐ 🐒

adleong:
:star: 🍷

lsjostro:
:+1:

PR #319, opened 2016-04-20
Add streaming and non-streaming versions of namerd http API endpoints

### LINKED ISSUE #309: Namerd's http interfaces should each support both streaming and polling
We should refactor our polling/stamping code so that we can have a principled way of doing streaming and polling on each API endpoint.

adleong:
The following APIs now support the `?watch=true` parameter which requests a
streaming response that streams updates to the requested resource.
* /api/1/dtabs
* /api/1/dtabs/<ns>
* /api/1/bind/<ns>
* /api/1/addr/<ns>
* /api/1/delegate/<ns>

All of these APIs now return a non-streaming response with the current value
of the resource if the `watch` parameter is not set.

To accomplish this, the `DtabStore.list` return type was changed from `Future` to `Activity`.
A bug in the URL parsing of the bind, addr, and delegate APIs was also fixed.

Fixes #309


namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -284,44 +311,40 @@ class HttpControlService(storage: DtabStore, namers: Ns => NameInterpreter)
...
+        rsp.content = Buf.Utf8(tree.show + "\n")
+        rsp
+      }
+    }
 
   private[this] val addrCacheMu = new {}
-  private[this] var addrCache: Map[(String, Path), Var[Addr]] = Map.empty
-  private[this] def getAddr(ns: String, path: Path): Var[Addr] = addrCacheMu.synchronized {
+  private[this] var addrCache: Map[(String, Path), Activity[Addr]] = Map.empty
+  private[this] def getAddr(ns: String, path: Path): Activity[Addr] = addrCacheMu.synchronized {
```
adleong: The change in the return type is worth explaining.

Addr has a `Addr.Pending` value.  Previously, we flatMapped the `Activity[NameTree]` directly to a `Var[Addr]` by mapping `Activity.Pending` to `Addr.Pending`.  This leads to some questions about how this pending state should be displayed in non-streaming responses.  Typically, we use `Activity.toFuture` to get the first non-pending value from an Activity but that doesn't work here because `Addr.Pending` is, in some sense, a non-pending value.

Therefore, I changed this to flatMap the `Activity[NameTree]` into an `Activity[Addr]` so that it made sense to respond with the first non-pending value.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -160,26 +162,73 @@ class HttpControlService(storage: DtabStore, namers: Ns => NameInterpreter)
...
+        writeFuture = writeFuture.before {
+          val buf = render(t)
+          if (buf == Buf.Empty)
+            Future.Unit
+          else
+            writer.write(buf).onFailure { _ => closable.close() }
+        }
+      case Throw(e) => closable.close()
     }
+    Future.value(resp)
```
olix0r: just occurred to me that we should possibly attempt to close the stream if the future is cancelled?  I.e. a user hits ctrl-c on the request, we can close the stream preemptively (i.e. before trying to write on it).  This is an optimization so not really a blocker

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -160,26 +162,73 @@ class HttpControlService(storage: DtabStore, namers: Ns => NameInterpreter)
...
+  }
 
-  private[this] def handleGetDtab(ns: String, req: Request): Future[Response] =
-    getDtab(ns).map {
-      case Some(dtab) =>
+  private[this] def streaming(req: Request): Boolean = req.getBooleanParam("watch")
```
olix0r: maybe clearer as isStreaming

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -160,26 +162,73 @@ class HttpControlService(storage: DtabStore, namers: Ns => NameInterpreter)
...
+    closable = values.respond {
+      case Return(t) =>
+        writeFuture = writeFuture.before {
+          val buf = render(t)
+          if (buf == Buf.Empty)
+            Future.Unit
+          else
+            writer.write(buf).onFailure { _ => closable.close() }
+        }
+      case Throw(e) => closable.close()
```
olix0r: should we report the error on the stream?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
adleong: How can we detect this?  The `Future[Response]` is already complete by the time we're streaming data.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
adleong: Good question.  I'm not sure what the best behavior is.  Print the error and keep streaming?  Print the error and close the connection?  Just close it?  Suggestions appreciated.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
olix0r: yeah, you're right. i was confused about how interrupt handlers worked.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -170,26 +171,75 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
...
+      case Return(t) =>
+        writeFuture = writeFuture.before {
+          val buf = render(t)
+          if (buf == Buf.Empty)
+            Future.Unit
+          else
+            writer.write(buf).onFailure { _ => closable.close() }
+        }
+      case Throw(e) =>
+        val _ = writer.write(Buf.Utf8(e.getMessage).concat(newline)).before(writer.close())
```
olix0r: Should we make render implement `Try[T] => Buf`?  For json streams, in particular, it seems sketchy to return a raw string in a sequence of json objects.  It would be nice if it could be encoded as json...

adleong: In general, that seems reasonable.  In our case we don't stream back any json and each render method would end up having the same handling for Throw (ie just print the exception message).  I think we can maybe defer that until we are returning content types other than plain text.

olix0r: 👍

olix0r:
:star:

siggy:
confirming that `/api/1/delegate/` does not support a `watch` param? should it?

adleong:
It doesn't, currently.  It could, with some refactoring, if you think that would be useful.

siggy:
let's leave it out for now, if we need it, we can revisit.

olix0r:
what changes on `/api/1/delegate`?  The dtab itself can't change...  I suppose we should reflect updates to service discovery?  (It does sound useful, but I'm good with deferring it)

siggy:
note to change the commit message to not include `/api/1/delegate`

siggy:
i noticed `/api/1/dtabs` streams the same response on a deterministic frequency:
```
$ curl :4180/api/1/dtabs?watch=true
["pandoracorn"]
["pandoracorn"]
["pandoracorn"]
```
...is this due to `InMemoryDtabStore` `loop()` ? if so, should we modify the underlying implementation to only stream out changes?

adleong:
@siggy that's correct.  We could implement de-dupping, but I'd prefer to do that in a follow-up.

siggy:
⭐ 👍

PR #321, opened 2016-04-20
Simplify the SvcCache interface

adleong:
Simplify the SvcCache interface by making a private state Var and exposing
ports as an Activity.


olix0r:
:star:

PR #322, opened 2016-04-21
Retry the initial request to the k8s endpoint API if it fails

### LINKED ISSUE #289: Improve failure handling in k8s namer
In the event that the k8s namer can't talk to the k8s API, it should return Neg instead of hanging forever.

adleong:
Problem:
If the initial request to the k8s endpoint API failed, no watch is created and the NsCache is never populated.

Solution:
Infinitely retry the initial request to the k8s endpoint API.

Fixes #289

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -3,9 +3,7 @@ package io.buoyant.k8s
 import com.twitter.finagle._
 import com.twitter.finagle.tracing.Trace
 import com.twitter.util._
-import io.buoyant.k8s.Api.Closed
-import io.buoyant.k8s.v1.{Endpoints, EndpointsList, EndpointsWatch, NsApi}
-import java.util.concurrent.atomic.AtomicReference
+import io.buoyant.k8s.v1.{EndpointsWatch, NsApi}
 import scala.collection.mutable
 
 class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
@@ -23,7 +21,9 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
     case id@Path.Utf8(nsName, portName, serviceName) =>
       val residual = path.drop(PrefixLen)
       log.debug("k8s lookup: %s %s", id.show, path.show)
-      Activity(Ns.get(nsName).services).flatMap { services =>
+      Activity.future(Ns.get(nsName)).flatMap { ns =>
+        Activity(ns.services)
+      }.flatMap { services =>
         log.debug("k8s ns %s initial state: %s", nsName, services.keys.mkString(", "))
         services.get(serviceName) match {
           case None =>
@@ -53,20 +53,23 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
   private[this] object Ns {
     private[this] var caches: Map[String, NsCache] = Map.empty
 
-    def get(name: String): NsCache = synchronized {
+    def get(name: String): Future[NsCache] = synchronized {
       caches.get(name) match {
-        case Some(ns) => ns
+        case Some(ns) =>
+          Future.value(ns)
         case None =>
-          val ns = mkNs(name)
-          caches += (name -> ns)
-          ns
+          mkNs(name).onSuccess { ns =>
+            caches += (name -> ns)
+          }
       }
     }
 
-    private[this] def mkNs(name: String): NsCache = {
+    private[this] def mkNs(name: String): Future[NsCache] = {
       val nsCache = new NsCache(name, Var(Activity.Pending))
-      _watches = _watches + (name -> watch(name, nsCache))
-      nsCache
+      watch(name, nsCache).map { closable =>
+        _watches += (name -> closable)
+        nsCache
+      }
     }
 
     // XXX once a namespace is watched, it is watched forever.  also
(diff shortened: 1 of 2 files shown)


olix0r:
What's the change in behavior? (This is probably where the Problem/Solution style commit message would be helpful ;)

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -53,59 +53,48 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
...
+        case Some(ns) =>
+          Future.value(ns)
         case None =>
-          val ns = mkNs(name)
-          caches += (name -> ns)
-          ns
+          mkNs(name).onSuccess { ns =>
+            caches += (name -> ns)
+          }
       }
```
olix0r: This is racey:

T0: caches.get(name) => None; mkNs(name)
T1: caches.get(name) => None; mkNs(name)

Furthermore, updates to the cache are done from another thread and are not synchronized.

olix0r:
I'm pretty sure that you'll have to update the cache immediately.  It may be workable to store an ADT in the cache instead -- something like:
```scala
sealed trait NsState
object NsState {
  case class Pending(future: Future[Ns]) extends NsState
  case class Cached(ns: Ns) extends NsState
  case class Failed(exception: Throwable, retry: Future[Ns]) extends NsState
}
```

adleong:
Good points.  This requires some more thought.  I'll try to write a test to capture exactly the problem that I'm solving.  Will also update the description accordingly.

k8s/src/test/scala/io/buoyant/k8s/EndpointsNamerTest.scala
```diff
@@ -86,14 +111,13 @@ class EndpointsNamerTest extends FunSuite with Awaits {
         fail(s"unexpected request: $req")
     }
     val api = v1.Api(service)
-    val namer = new EndpointsNamer(Path.read("/test"), api.withNamespace(_))
+    val namer = new EndpointsNamer(Path.read("/test"), api.withNamespace)
 
     @volatile var state: Activity.State[NameTree[Name]] = Activity.Pending
-    val obs = namer.lookup(Path.read("/srv/thrift/sessions")).states respond { s =>
+    val _ = namer.lookup(Path.read("/srv/thrift/sessions")).states respond { s =>
```
klingerf: Question: why do you have to assign this to a val?  It seems like you could just write:

```
namer.lookup(Path.read("/srv/thrift/sessions")).states.respond(state = _)
```

But I think I'm probably misunderstanding what we're trying to do here.  Also, can you make a similar change to this one on line 49 of this file?

adleong: Good question.  Discarding values is a red-flag and usually not intentional.  I want to eventually turn on the compile option -Ywarn-value-discard which will cause a compiler error for discarded values.  In this case, though, we really do want to discard the return value from `respond`.  We can assign it to an unnamed variable to explicitly signal that discarding is intended here.

I'll make the same change on line 49.

klingerf: Ok great, thank for the explanation.  I saw that you were following the same convention in #323 but didn't have the context to understand why.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -52,60 +57,75 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
...
+      * becomes an Activity.Failed and the Future is retried with the given backoff schedule.
+      * Therefore, the legal state transitions are:
+      *
+      * Pending -> Ok
+      * Pending -> Failed
+      * Failed -> Failed
+      * Failed -> Ok
+      */
+    private[this] def retryToActivity[T](go: => Future[T]): Activity[T] = {
+      val state = Var(Activity.Pending: Activity.State[T])
```
olix0r: style: prefer Var[Activity.State[T]](Activity.Pending) (this is done throughout the codebase already)

k8s/src/test/scala/io/buoyant/k8s/EndpointsNamerTest.scala
olix0r: Ah, ok.  We should discuss introducing new stylistic checks.  Throughout our codebase (and certainly finagle), discarded values are used idiomatically.  I'm not opposed to it, but i"m not sure how much value this actually adds if we are diligent about annotating return types

olix0r: nit: `states.respond`

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -52,60 +57,75 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
...
+      }
+    }
 
     def get(name: String): NsCache = synchronized {
       caches.get(name) match {
         case Some(ns) => ns
         case None =>
-          val ns = mkNs(name)
+          val ns = new NsCache(name)
+          val closable = retryToActivity(watch(name, ns))
```
olix0r: might be slightly clearer as `retryToActivity { watch(name, ns) }` to make it obvious that this is a closure

olix0r:
I'll need to give this another read through to get my head around it... it looks promising but I don't quite understand it yet.

k8s/src/test/scala/io/buoyant/k8s/EndpointsNamerTest.scala
adleong: being explicit about discarding values seems like good style to me, but you're right, we should discuss it.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -52,60 +57,75 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
...
-            resourceVersion = init.metadata.flatMap(_.resourceVersion)
+            resourceVersion = list.metadata.flatMap(_.resourceVersion)
           )
-          close.set(closable)
-          updates.foreach(services.update)
+          // fire-and-forget this traversal over an AsyncStream that updates the services state
+          val _ = updates.foreach(services.update)
+          closable
+        }.onFailure { e =>
+          log.error(e, "k8s failed to list endpoints")
```
olix0r: tioli: The prior `AtomicReference` approach before made it possible to cancel the request if its closed before the request is satsified. Practically, it probably doesn't matter, but it seems more correct to do so....

olix0r:
:star:

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -52,60 +57,75 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
...
+     * original future is satisfied.  When the Future is successful, the Activity becomes
+     * an Activity.Ok with a fixed value from the Future.  If the Future fails, the Activity
+     * becomes an Activity.Failed and the Future is retried with the given backoff schedule.
+     * Therefore, the legal state transitions are:
+     *
+     * Pending -> Ok
+     * Pending -> Failed
+     * Failed -> Failed
+     * Failed -> Ok
+     */
```
siggy: tioli:
```
  * Pending-->Ok
  * |         ^
  * v        /
  * Failed--`
  * `-^
```

siggy:
⭐ lgtm

PR #324, opened 2016-04-21
Simplify the Port interface.

adleong:
I kept getting confused by the VarUp and ActUp type aliases.  This change allows us to get rid of them and hopefully improves readability.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -4,7 +4,7 @@ import com.twitter.finagle._
 import com.twitter.finagle.tracing.Trace
 import com.twitter.util._
 import io.buoyant.k8s.Api.Closed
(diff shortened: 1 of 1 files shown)


k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -113,11 +113,9 @@ class EndpointsNamer(idPrefix: Path, mkApi: String => NsApi) extends Namer {
 private object EndpointsNamer {
   val PrefixLen = 3
 
-  type VarUp[T] = Var[T] with Updatable[T]
-  type ActUp[T] = VarUp[Activity.State[T]]
```
klingerf: These type aliases also exist in CatalogNamer.scala -- do you think it's worth trying to remove them from there as well?

adleong: Probably.  I plan to take a pass through the consul namer as well.

klingerf:
⭐ 🐒

olix0r:
:star:

PR #325, opened 2016-04-22
modify namerd /delegate endpoint to accept a dtab

siggy:
also includes readme cleanup, and a refactor to share code between these
three endpoints:
admin /delegator.json
namerd /delegate/namespace
namerd /delegate?dtab=

@adleong: I'd appreciate your eyes on the changes in `HttpControlService.scala`, as it may affect work you are doing in that area.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,28 @@ centrally managing routing policy and fronting service discovery.
 * [linkerd](linkerd/README.md)
 * [namerd](namerd/README.md)
 
+## Quickstart ##
+
+### Boot linkerd ###
+
+```
+$ ./sbt linkerd-examples/http:run # build and run linkerd
+$ open http://localhost:9990      # open linkerd's admin interface in a browser
+```
+
+More details at [linkerd's quickstart](linkerd/README.md#quickstart)
+
+
+### Boot namerd ###
+
+```
+$ ./sbt namerd-examples/basic:run # build and run namerd
+$ open http://localhost:9991      # open namerd's admin interface in a browser
+$ curl :4180/api/1/dtabs          # test namerd's http interface
+```
+
+More details at [namerd's quickstart](namerd/README.md#quickstart)
+
 ## Working in this repository ##
 
 [sbt][sbt] is used to build and test linkerd. Developers should not
@@ -42,57 +64,10 @@ $ ./sbt
 >
 ```
 
-The sbt project consists of many sub-projects:
+The sbt project consists of many sub-projects. To list all projects run:
 
 ```
 > projects
-[info] In file:.../linkerd/
-[info]     admin
-[info]   * all
-[info]     config
-[info]     consul
-[info]     interpreter-namerd
-[info]     k8s
-[info]     linkerd
-[info]     linkerd-admin
-[info]     linkerd-core
-[info]     linkerd-examples
-[info]     linkerd-identifier
-[info]     linkerd-identifier-http
-[info]     linkerd-main
-[info]     linkerd-protocol
-[info]     linkerd-protocol-benchmark
-[info]     linkerd-protocol-http
-[info]     linkerd-protocol-mux
-[info]     linkerd-protocol-thrift
-[info]     linkerd-tls
-[info]     marathon
-[info]     namer
-[info]     namer-consul
-[info]     namer-core
-[info]     namer-fs
-[info]     namer-k8s
-[info]     namer-marathon
-[info]     namer-serversets
-[info]     namerd
-[info]     namerd-core
-[info]     namerd-examples
-[info]     namerd-iface
-[info]     namerd-iface-control-http
-[info]     namerd-iface-interpreter-thrift
-[info]     namerd-iface-interpreter-thrift-idl
-[info]     namerd-main
-[info]     namerd-storage
-[info]     namerd-storage-in-memory
-[info]     namerd-storage-zk
-[info]     router
-[info]     router-core
-[info]     router-http
-[info]     router-mux
-[info]     router-thrift
-[info]     router-thrift-idl
-[info]     test-util
-[info]     validator
 ```
 
 These projects are configured in
--- a/admin/src/main/resources/io/buoyant/admin/js/delegator.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/delegator.js
@@ -29,7 +29,7 @@ var Delegator = (function() {
       var path = $('#path-input').val();
       window.location.hash = encodeURIComponent(path);
       var request = $.get(
-        "/delegator.json?" + $.param({ n: path, d: dtabViewer.dtabStr() }),
+        "/delegator.json?" + $.param({ path: path, dtab: dtabViewer.dtabStr() }),
         renderAll.bind(this));
       request.fail(function( jqXHR ) {
         $(".error-modal").html(templates.errorModal(jqXHR.statusText));
--- a/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.module.SimpleModule
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
 import com.twitter.finagle.http._
+import com.twitter.finagle.naming.NameInterpreter
 import com.twitter.finagle.{Address => FAddress, Addr => FAddr, Path, Status => _, _}
 import com.twitter.io.Buf
 import com.twitter.util._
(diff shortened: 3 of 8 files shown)


namerd/README.md
```diff
@@ -39,3 +39,47 @@ or
...
+$ export NAMERD_URL=http://localhost:4180
+```
+
+Get all dtab namespaces:
+
+```
+$ curl $NAMERD_URL/api/1/dtabs
+```
+
+Get a dtab namespace:
```
rmars: tioli: it might be clearer to say:
Get a dtab for the `default` namespace
since default is a common word...

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -158,29 +159,29 @@ object DelegateApiHandler {
     def writeBuf[T](t: T): Buf = Buf.ByteArray.Owned(mapper.writeValueAsBytes(t))
   }
 
+  def getDelegateRsp(dtab: String, path: String, interpreter: NameInterpreter): Future[Response] =
```
esbie: we should probably move this file from admin to iface if we're going to use it in HttpControlService

esbie:
👍  🐫

namerd/README.md
```diff
@@ -39,3 +39,47 @@ or
...
+
+```
+$ curl -v $NAMERD_URL/api/1/delegate/baz?path=/foo
+```
+
+Evaluate a path against an arbitrary dtab:
+
+```
+$ curl -v "$NAMERD_URL/api/1/delegate?dtab=/foo=>/bar&path=/foo"
+```
```
olix0r: This is okay for now, but we're going to need real API docs...

README.md
```diff
@@ -27,6 +27,23 @@ centrally managing routing policy and fronting service discovery.
...
+$ open http://localhost:9990
+```
+
+### Boot namerd ###
+
+```
+$ ./sbt namerd-examples/basic:run
+$ open http://localhost:9991
+$ curl :4180/api/1/dtabs
+```
```
olix0r: Is this helpful? If this is for the uninitiated, there's no explanation of what's going on.  If this is for the initiated... is this helpful?

admin/src/main/resources/io/buoyant/admin/js/delegator.js
```diff
@@ -29,7 +29,7 @@ var Delegator = (function() {
       var path = $('#path-input').val();
       window.location.hash = encodeURIComponent(path);
       var request = $.get(
-        "/delegator.json?" + $.param({ n: path, d: dtabViewer.dtabStr() }),
+        "/delegator.json?" + $.param({ path: path, dtab: dtabViewer.dtabStr() }),
```
olix0r: pretty sure we're going to have to move this to a POST api in the near future...  When dtabs have comments, etc, this is going to be fragile.

namerd/iface/control-http/src/test/scala/io/buoyant/namerd/iface/HttpControlServiceTest.scala
```diff
@@ -261,4 +261,44 @@ class HttpControlServiceTest extends FunSuite with Awaits {
...
+
+  test("delegate a path given a dtab") {
+    val service = newService()
+    val resp = await(service(Request("/api/1/delegate?dtab=/foo=>/bar&path=/foo")))
+    assert(resp.status == Status.Ok)
+    assert(resp.contentString == """
+      |{
+      |  "type":"delegate",
+      |  "path":"/foo",
+      |  "dentry":null,
```
olix0r: how come this is null if we've specified a dtab?

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
olix0r: at some point we had separate delegator/{core,http} projects?  seems like linkerd-admin and namerd-iface-http-controller should pull in delegator-http?

namerd/iface/control-http/src/test/scala/io/buoyant/namerd/iface/HttpControlServiceTest.scala
olix0r: nevermind, it's on the child...

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -143,12 +151,14 @@ class HttpControlService(storage: DtabStore, namers: Ns => NameInterpreter)
...
+    case BindUri(Some(ns), path) =>
       handleGetBind(ns, path)
-    case AddrUri(ns, path) =>
+    case AddrUri(Some(ns), path) =>
       handleGetAddr(ns, path)
-    case DelegateUri(ns, path) =>
+    case DelegateUri(Some(ns), path) =>
       handleGetDelegate(ns, path)
+    case DelegateUri(None, path) =>
+      delegateApiHander(req)
```
olix0r: fwiw, i expect that this is going to break in an upcoming change.  I expect it will work as follows:
- / is a valid (root) namespace.
- all delegation requests will accept a dtab (override) that is appended to the namespace's dtab.
- a special namespace, e.g. _/.empty_, will be supported to support requests that don't use an underlying dtab.

README.md
olix0r: perhaps warrants comments like
```
$ open http://localhost:9990  # open linkerd's admin interface in a browser
```

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
siggy: agree with you both, ticketed here:
https://github.com/BuoyantIO/linkerd/issues/326

i thought about going down this route, but decided to limit the scope of this PR to just not introducing any _more_ dependencies.

namerd/README.md
siggy: agreed: https://github.com/BuoyantIO/linkerd/issues/327

README.md
siggy: selfishly i find this quite helpful. i think for the uninitiated, it gets folks up and running very quickly, without having to know much about sbt. i'd like developers to go from clone to admin page in as few seconds as possible.

admin/src/main/resources/io/buoyant/admin/js/delegator.js
siggy: yup, agreed. fortunately all the code is in one place. ;)

https://github.com/BuoyantIO/linkerd/issues/328

README.md
esbie: It's helpful, would be even helpfuller with comments

olix0r: i think it's fine to be here, i just think it would be better with a little explanation if it's displayed so prominently in the docs.  (I worry that the README is getting to be a bit overwhelming)

README.md
```diff
@@ -27,6 +27,23 @@ centrally managing routing policy and fronting service discovery.
 * [linkerd](linkerd/README.md)
 * [namerd](namerd/README.md)
 
+## Quick start ##
```
adleong: These quickstart instructions are already in linkerd/README.md and namerd/README.md

README.md
```diff
@@ -42,57 +59,10 @@ $ ./sbt
 >
 ```
 
-The sbt project consists of many sub-projects:
+The sbt project consists of many sub-projects. To list all projects run:
```
adleong: fair enough

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -158,29 +159,29 @@ object DelegateApiHandler {
...
+    (dtab, path) match {
+      case (DtabStr(d), PathStr(p)) =>
+        Delegator(d, p, interpreter).values.toFuture().flatMap(Future.const).map { tree =>
```
adleong: Comment: io.buoyant.namerd.RichActivity adds a `toFuture` method to Activity.  Unfortunately, we don't want to depend on namerd here.  I wonder if we should create a utility project that we can move RichFuture into...

adleong:
:star: Looks good to me, acknowledging that much of this will change as we mess around with the dtab namespace format.

olix0r:
:star:

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
siggy: Fair point. I've noted this in #326. We may or may not bundle all that work together, but at least it's documented either way.

README.md
siggy: Fair, I've added links from here to those other two README's. I'd still like to keep these commands front and center, as it's the first thing folks see when they hit the repo.

PR #329, opened 2016-04-25
Make NsCache immutable

adleong:
Based on https://github.com/BuoyantIO/linkerd/pull/322

Problem:
NsCache, SvcCache, and Port each contain a Var which results in multiple nested levels of mutability.  This is complex and difficult to reason about.

Solution:
Make NsCache, SvcCache, and Port immutable.  Methods such as `update` now return a modified copy.

Result:
`watch` now returns an `Activity[NsCache]`.  Mutability is isolated to the `watch` method and separated from the data structure.

Also, we refactor `retryToActivity` to use `Var.async`.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -70,24 +70,31 @@ class EndpointsNamer(
      * Failed -> Ok
      */
     private[this] def retryToActivity[T](go: => Future[T]): Activity[T] = {
-      val state = Var[Activity.State[T]](Activity.Pending)
-      _retryToActivity(backoff, state)(go)
+      val state = Var.async[Activity.State[T]](Activity.Pending) { updates =>
+        val t = _retryToActivity(backoff, updates)(go)
+        Closable.make { _ =>
+          t.raise(new InterruptedException)
+          Future.Unit
+        }
+      }
       Activity(state)
     }
 
     private[this] def _retryToActivity[T](
       remainingBackoff: Stream[Duration],
-      state: Var[Activity.State[T]] with Updatable[Activity.State[T]] = Var[Activity.State[T]](Activity.Pending)
-    )(go: => Future[T]): Unit = {
-      val _ = go.respond {
+      updates: Updatable[Activity.State[T]]
+    )(go: => Future[T]): Future[T] = {
+      go.transform {
         case Return(t) =>
-          state() = Activity.Ok(t)
+          updates() = Activity.Ok(t)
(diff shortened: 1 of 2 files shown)


olix0r:
So I understands, this changes the k8s namer so that any update to a namespace triggers updates to _all_ observations on that namespace?  So if i'm observing _default/http/web_ and there's and the _default/http/users_ service is created, my original observation will receive an update?

adleong:
Yes, I think that's right.  Actually, I think this was already the case since we flatMap the NameTree with the Addr.  Any changes to the NameTree will therefore result in an update of your observation.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -28,25 +27,23 @@ class EndpointsNamer(
...
+                log.debug("k8s ns %s service %s port %s missing", nsName, serviceName, portName)
+                NameTree.Neg
 
-                case Some(port) =>
-                  log.debug("k8s ns %s service %s port %s found + %s", nsName, serviceName, portName, residual.show)
-                  NameTree.Leaf(Name.Bound(port.addr, idPrefix ++ id, residual))
-              }
+              case Some(port) =>
+                log.debug("k8s ns %s service %s port %s found + %s", nsName, serviceName, portName, residual.show)
+                NameTree.Leaf(Name.Bound(Var(port.addr), idPrefix ++ id, residual))
```
olix0r: i think this is going to be problematic -- we never update the var, we just return a new name tree every time the var updates.  this is going to be at-best costly and at-worst broken.

adleong: I think what this exercise has led me to is that I think I don't understand why finagle has both an Activity for the NameTree and a Var for the Addr.  Just watching the addr Var directly seems problematic because a change to the NameTree could invalidate that Var.  Therefore, it seems like you would always want to watch the NameTree flatMapped to the addr... in which case, why have both?

olix0r: 1. Name.Bound is not required to be a path (it may be a reference, for instance).
2. Multiple NameTrees may return the same Name.Bound -- "same" being determined by id equality.
3. clients are cached on this.   Load balancers observe updates to the Var[Addr] and update their list of endpoints accordingly.

It's probably helpful to review finagle's client stack (ignoring linkerd for the time being)--specifically, BindingFactory and how it's layered onto LoadBalancerFactory.

adleong:
I'm starting to see why the complexity in this implementation is necessary.  This was a helpful exercise.  Closing this PR.

PR #331, opened 2016-04-25
fix scaladoc reference

siggy:


namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/experimental/K8sDtabStore.scala
```diff
@@ -12,7 +12,7 @@ import java.util.concurrent.atomic.AtomicReference
 import scala.collection.breakOut
 
 /**
- * A [[DtabStore]] using the
+ * A [[io.buoyant.namerd DtabStore]] using the
```
olix0r: i would just remove the link -- linking to the package won't be helpful here.

gtcampbell:
⭐  I had the same comment as @olix0r: better to just remove the reference, which you did. I guess this is a (minor) downside to our current project structure.

olix0r:
:star:

PR #332, opened 2016-04-26
Breaking namerd thrift change: string-encode Dtabs

olix0r:
The thrift interface contains thrift-encoded dtab types. These types must be
mapped to/from finagle's types internally.

In the next finagle release, dtabs are changing to support wildcard prefixes.
Rather than reflect this change in the thrift structure, I propose that we
simply move to string-encoded dtabs.


adleong:
👍 🚢

klingerf:
⭐ 🐘  Works for me.

PR #333, opened 2016-04-26
Http addr endpoint should bind directly against namers.

### LINKED ISSUE #330: http addr interface should bind directly against namers
Namer'd http addr interface should bind directly against a namer (instead of a name interpreter).  This is how the thrift interface does it.

adleong:
Fixes #330 

The addr endpoint should only ever be given a bound path.  Therefore, it can bind against the namers directly instead of the name interpreter.  This is what the corresponding thrift endpoint does.


olix0r:
⭐

klingerf:
⭐ 🌊

PR #334, opened 2016-04-26
publish docker images upon successful ci master

### LINKED ISSUE #173: automate docker builds
should be able to automate `latest` and release images:
https://docs.docker.com/docker-hub/builds/

siggy:
fixes #173

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 machine:
   java:
     version: oraclejdk8
+  services:
+    - docker
 
 general:
   artifacts:
@@ -29,3 +31,9 @@ test:
   post:
   - ./sbt coverageAggregate
   - mkdir -p $CIRCLE_TEST_REPORTS/junit && find . -type f -regex ".*/target/test-reports/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
+
+deployment:
+  master:
+    branch: master
+    commands:
+    - ./project/ci-publish-docker.sh
--- a/project/ci-publish-docker.sh
+++ b/project/ci-publish-docker.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# Publish docker images from circleci
+
+# Usage:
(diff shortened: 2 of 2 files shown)


olix0r:
If we do this, we need to stop CI'ing jobs submitted from people outside of buoyant (otherwise, we expose our docker credentials to them...)

project/publish-docker.sh
```diff
@@ -0,0 +1,28 @@
...
+
+VERSION_TAG=$(docker images buoyantio/linkerd|sed -n 2p|awk '{print $2}')
+if [ -z "$VERSION_TAG" ]; then
+  echo "VERSION_TAG not found"
+  exit 1
+fi
+
+docker tag buoyantio/linkerd:$VERSION_TAG buoyantio/linkerd:latest
+docker tag buoyantio/namerd:$VERSION_TAG buoyantio/namerd:latest
+docker tag buoyantio/namerd:$VERSION_TAG-dcos buoyantio/namerd:latest-dcos
```
olix0r: I don't think we should update latest to snapshot tags...

project/publish-docker.sh
```diff
@@ -0,0 +1,28 @@
...
+
+docker push buoyantio/linkerd:$VERSION_TAG
+docker push buoyantio/namerd:$VERSION_TAG
+docker push buoyantio/namerd:$VERSION_TAG-dcos
```
olix0r: I think the sbt-docker plugin will do this for us.

olix0r:
I don't think we need a separate script to do docker pushing.  I think we can simply use something like:
`./sbt linkerd/dockerBuildAndPush` or `./sbt linkerd/dockerPush`

siggy:
Per offline discussion, disabling permissive forked builds. Forks can still ci, but won't have access to DOCKER_CREDENTIALS:
https://circleci.com/docs/fork-pr-builds/

olix0r:
⭐  looks good!

PR #339, opened 2016-04-27
Introduce response classifiers

### LINKED ISSUE #335: http response classifier module
linkerd should support pluggable response classifiers.

A default http classier should be provided as follows:

1. all 5xx responses are considered failures.
2. 5xx responses to _idempotent_ requests should be considered retryable.

From RFC2616:
> Methods can also have the property of "idempotence" in that (aside from error or expiration issues) the side-effects of N > 0 identical requests is the same as for a single request. The methods GET, HEAD, PUT and DELETE share this property. Also, the methods OPTIONS and TRACE SHOULD NOT have side effects, and so are inherently idempotent.

olix0r:
By default, finagle treats all non-exceptional responses as a "success".  This
means that, for instance, 5XX http responses do not compute in success rates.
We introduce response classifiers to mark 5XX responses as failures.

Furthermore, we can use response classifiers to mark some failures as retryable
(i.e. depending on the request type). Unfortunately, Finagle doesn't yet honor
response classification when retrying requests.

Initially, response classification can be used simply to inform success rate
calculation. Later (#361), this classification may be used to automatically
retry some class of requests.

This change introduces a protocol-specific base response classifier and, for
http, marks all 5XX responses as failures. A response classifier plugin type is
introduced to allow this logic to be configurable, but this is of marginal
utility until requeues are supported.

Fixes #335

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
@@ -156,9 +156,7 @@ trait RouterConfig {
 
   @JsonIgnore
   def responseClassifier: ResponseClassifier =
-    _responseClassifier.foldLeft(baseResponseClassifier) {
-      case (base, config) => config.mk orElse base
-    }
+    _responseClassifier.map(_.mk).getOrElse(PartialFunction.empty) orElse baseResponseClassifier
 
   @JsonIgnore
   def routerParams = Stack.Params.empty
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
@@ -6,7 +6,7 @@ import com.twitter.conversions.time._
 import com.twitter.finagle._
 import com.twitter.finagle.buoyant.DstBindingFactory
 import com.twitter.finagle.client.DefaultPool
-import com.twitter.finagle.service.{FailFastFactory, TimeoutFilter}
+import com.twitter.finagle.service.{FailFastFactory, ResponseClassifier, TimeoutFilter}
 import com.twitter.util.Closable
 import io.buoyant.namer.{InterpreterConfig, DefaultInterpreterConfig}
 import io.buoyant.router.RoutingFactory
@@ -38,11 +38,15 @@ trait Router {
 
   def withParams(ps: Stack.Params): Router = {
     val r = _withParams(ps)
-    // Copy stats and tracing params from router to servers
+
+    // Copy selected params from router to servers
     val param.Stats(stats) = r.params[param.Stats]
-    val srvStats = param.Stats(stats.scope(label, "srv"))
-    val tracer = r.params[param.Tracer]
-    r.withServers(servers.map(_.configured(srvStats).configured(tracer)))
+    def configure(s: Server) = s
+      .configured(param.Stats(stats.scope(label, "srv")))
+      .configured(r.params[param.ResponseClassifier])
+      .configured(r.params[param.Tracer])
+
+    r.withServers(servers.map(configure(_)))
   }
 
   def configured[P: Stack.Param](p: P): Router = withParams(params + p)
@@ -90,6 +94,7 @@ object Router {
     server.configured(param.Label(s"$ip/$port"))
       .configured(Server.RouterLabel(routerLabel))
       .configured(param.Stats(stats.scope(routerLabel, "srv")))
+      .configured(router.params[param.ResponseClassifier])
       .configured(router.params[param.Tracer])
   }
 }
@@ -113,6 +118,10 @@ trait RouterConfig {
   @JsonIgnore
   def label = _label.getOrElse(protocol.name)
 
+  /*
+   * interpreter controls how names are bound.
+   */
+
   @JsonProperty("interpreter")
   var _interpreter: Option[InterpreterConfig] = None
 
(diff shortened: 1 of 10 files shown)


klingerf:
⭐ 🐑  Looks good to me.

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -113,6 +118,10 @@ trait RouterConfig {
   @JsonIgnore
   def label = _label.getOrElse(protocol.name)
 
+  /*
+   * interpreter controls how names are bound.
+   */
+
   @JsonProperty("interpreter")
```
gtcampbell: is the extra space between the comment and the annotation intentional?

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -123,19 +132,42 @@ trait RouterConfig {
...
+  var _responseClassifier: Option[ResponseClassifierConfig] = None
+
+  @JsonIgnore
+  def baseResponseClassifier: ResponseClassifier =
+    ResponseClassifier.Default
+
+  @JsonIgnore
+  def responseClassifier: ResponseClassifier =
+    _responseClassifier.foldLeft(baseResponseClassifier) {
+      case (base, config) => config.mk orElse base
```
gtcampbell: `foldLeft` on `Option` is a little weird. maybe use `fold` instead, i.e.

```scala
_responseClassifier.fold(baseResponseClassifier)(_.mk)
```
?

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -141,10 +142,114 @@ routers:
...
+
+    } finally {
+      await(client.close())
+      await(downstream.server.close())
+      await(server.close())
+      await(router.close())
+    }
+  }
+
+  // TODO once repsonse classifiers are wired into requeues #361
```
gtcampbell: typo: "repsonse"

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -0,0 +1,124 @@
...
+     *   the methods OPTIONS and TRACE SHOULD NOT have side effects, and
+     *   so are inherently idempotent.
+     */
+    val Idempotent = ByMethod(Set(
+      Method.Get,
+      Method.Head,
+      Method.Put,
+      Method.Delete,
+      Method.Options,
+      Method.Trace
```
gtcampbell: tioli, tiny: I might extract the method sets into separate values and construct them as follows, to make it clear that `Idempotent` is a superset of `ReadOnly`:
```scala
private[this] val ReadOnlyMethods = Set(Method.Get, Method.Head, Method.Options, Method.Trace)
private[this] val IdempotentMethods = ReadOnlyMethods ++ Set(Method.Put, Method.Delete)

val ReadOnly = ByMethod(ReadOnlyMethods)
val Idempotent = ByMethod(IdempotentMethods)
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -0,0 +1,124 @@
...
+    ))
+  }
+
+  object Responses {
+
+    object Failure {
+      def unapply(rsp: Response): Boolean =
+        rsp.statusCode >= 500 && rsp.statusCode <= 599
```
gtcampbell: finagle-http provides very similar functionality in `Status.ServerError.unapply`, I think.

gtcampbell:
:star: looks good, my comments are all minor suggestions

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
adleong: maybe `_responseClassifier.map(_.mk).getOrElse(baseResponseClassifier` ?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    object Retryable {
+      // There are porbably some (linkerd-generated) failures that aren't
```
adleong: typo porbably

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -0,0 +1,124 @@
...
+    ResponseClassifier.named("RetryableReadFailures") {
+      case ReqRep(Requests.ReadOnly(), Return(Responses.Retryable()) | Throw(NonFatal(_))) =>
+        ResponseClass.RetryableFailure
+    }
+
+  /**
+   * Classifies 5XX responses and all exceptions as non-retryable
+   * failures.
+   */
+  val NonRetryableFailures: ResponseClassifier =
```
adleong: FYI this is the same as `c.t.f.http.service.HttpResponseClassifier.ServerErrorsAsFailures`

adleong:
Can you add response classifiers to linkerd/CONFIG.md too?

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
olix0r: neither of these are functionally equivalent ;) but I can change this since it seems to be confusing.

gtcampbell: ah, I misread it. I think what I would actually do is `flatMap`/`getOrElse` then.

olix0r:
I'd prefer to cover docs for this in the context of https://github.com/BuoyantIO/linkerd/pull/364.  Since we can't actually talk about retries yet, all of these classifiers are functionally equivalent (and so there's not really anything to configure...)

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
gtcampbell: I see that now! maybe `flatMap`/`getOrElse`?

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
olix0r: yeah, it's intended to basically be a section heading for the next few blocks. trying to provide visual cues about the sections

adleong:
:star: 🐰

PR #340, opened 2016-04-27
Deduplicate entries in streaming responses.

adleong:
Problem:
If the underlying Event updates multiple times with the same value, we will print out duplicate entries in streaming responses.

Solution:
Add a `.dedup` method which creates an Event with consecutive duplicates removed.

### DIFF
--- a/namerd/core/src/main/scala/io/buoyant/namerd/package.scala
+++ b/namerd/core/src/main/scala/io/buoyant/namerd/package.scala
@@ -1,6 +1,6 @@
 package io.buoyant
 
-import com.twitter.util.{Future, Activity}
+import com.twitter.util._
 
 package object namerd {
   type Ns = String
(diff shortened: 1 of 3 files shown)


siggy:
👍 ⭐  neat!
```
$ curl :4180/api/1/dtabs?watch=true
["pandoracorn"]
["pandoracorn","baz"]
```

namerd/core/src/main/scala/io/buoyant/namerd/package.scala
```diff
@@ -9,4 +9,19 @@ package object namerd {
...
+        var current: T = null.asInstanceOf[T] // scala fails to prove Null <: T for some reason
+        event.respond { t =>
+          if (t != current) {
+            current = t
+            s.notify(t)
+          }
+        }
+      }
+    }
+  }
```
olix0r: per conversation, Event already has a dedup :)

olix0r:
:Star:

PR #343, opened 2016-04-27
Turn on -Ywarn-value-discard

adleong:
Turn on the -Ywarn-value-discard compiler option and fix all value discard warnings (warnings are errors).

Most value discard warnings are fixed by assigning the value to a dummy val
```
val _ = discardThis
```
This explicitly signals that discarding the value is intentional.

We also introduce `assertThrows` for tests as a replacement for `intercept` that discards the resulting exception.


test-util/src/main/scala/io/buoyant/test/Exceptions.scala
```diff
@@ -0,0 +1,10 @@
+package io.buoyant.test
+
+import org.scalatest.Assertions
+
+trait Exceptions extends Assertions {
+
+  def assertThrows[T <: AnyRef: Manifest](f: => Any): Unit = {
+    val _ = intercept[T](f)
+  }
```
olix0r: Consider providing i.e. io.buoyant.test.FunSuite that mixes in scalatest's FunSuite with all of the helpers that we normally use (Awaits, for instance)...  Then we can simply use io.buoyant.test.FunSuite throughout our code.

namerd/iface/interpreter-thrift/src/test/scala/io/buoyant/namerd/iface/ThriftNamerInterfaceTest.scala
```diff
@@ -74,75 +74,79 @@ class ThriftNamerInterfaceTest extends FunSuite {
...
-    val isa = new InetSocketAddress("8.8.8.8", 4949)
-    addrs0() = Addr.Bound(Address(isa))
-    assert(rsp0.isDefined)
-
-    val boundAddr0 = {
-      val ip = ByteBuffer.wrap(isa.getAddress.getAddress)
-      val taddrs = Set(thrift.TransportAddress(ip, isa.getPort))
-      thrift.Addr(TStamp.mk(1), thrift.AddrVal.Bound(thrift.BoundAddr(taddrs)))
+  test("addr") {
+    val _ = new AddrCtx {
```
olix0r: tioli: may be nicer to do this sort of thing as:

```scala
val ctx = new AddrCtx
import ctx._
...
```

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -33,21 +33,21 @@ object Headers {
...
+    def set(headers: HeaderMap, id: TraceId): HeaderMap = {
       val bytes = TraceId.serialize(id)
       val b64 = Base64.getEncoder().encodeToString(bytes)
       headers.set(Key, b64)
     }
 
-    def clear(headers: HeaderMap): Unit = {
+    def clear(headers: HeaderMap): Option[String] = {
       headers.remove(Key)
     }
```
olix0r: I think it's preferable to return unit here, since these methods are actually side-effecty.  returning values makes them appear as if they are not side-effecty.

olix0r: looks like you changed the other set of set/clear functions, but these should still be `Unit`...

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -51,11 +51,11 @@ class CatalogNamer(
     def get(name: String): DcCache = synchronized {
       activity.sample() match {
         case Activity.Ok(snap) =>
-          snap.get(name).getOrElse {
+          snap.getOrElse(name, {
```
gtcampbell: FWIW, I tried to make this same change in #247 and @olix0r had [some objections](https://github.com/buoyantio/linkerd/pull/247#discussion_r58918627).

adleong: Thanks for pointing me to that discussion.  Option's `getOrElse` also takes a closure:
```def getOrElse[B >: A](default: => B): B```
Am I missing something?

File: linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala (hunk 3 above)
Comment (adleong) [inline linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala]: good catch

File: consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala (hunk 4 above)
Comment (olix0r) [inline consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala]: yeah, i tend to prefer explicit match statements, especially in synchronized blocks.

**Comment by olix0r** [comment]
:star: I do think we should use a base class for tests, but not a blocker

File: linkerd/admin/src/test/scala/io/buoyant/linkerd/admin/LinkerdAdminTest.scala [hunk 5]
```diff
@@ -17,45 +17,57 @@ class BuoyantMainTest extends TwitterServer {
...
-  test("serves buoyant admin at /")(new BuoyantMainTest {
-    override def main() {
-      val rsp = await(client(Request("/")))
-      assert(rsp.status == Status.Ok)
+  test("serves buoyant admin at /") {
+    val _ = new BuoyantMainTest {
+      override def main() {
+        val rsp = await(client(Request("/")))
+        assert(rsp.status == Status.Ok)
+      }
```
gtcampbell: tioli: it would be possible to slightly reduce this boilerplate by creating a mkMain method, but might not be worth the effort:

```scala
def testMain(body: Service[Request, Response] => Unit): Unit = {
  val _ = new BuoyantMainTest {
    override def main(): Unit = {
      body(client)
    }
  }
}
```

and then the tests would look like:

```scala
test("serves buoyant admin at /") {
  testMain { client =>
    val rsp = await(client(Request("/")))
    assert(rsp.status == Status.Ok)
  }
}
```
...not a huge deal either way, though.

namerd/iface/interpreter-thrift/src/test/scala/io/buoyant/namerd/iface/ThriftNamerInterfaceTest.scala
gtcampbell: agreed; could be a follow-up if that's easier

test-util/src/main/scala/io/buoyant/test/Exceptions.scala
```diff
@@ -0,0 +1,10 @@
+package io.buoyant.test
+
+import org.scalatest.Assertions
+
+trait Exceptions extends Assertions {
+
+  def assertThrows[T <: AnyRef: Manifest](f: => Any): Unit = {
```
gtcampbell: might be worth adding a comment here explaining that we provide this because `intercept` triggers value-discarded warnings?

gtcampbell:
⭐  looks good overall, my comments are non-blocking

PR #347, opened 2016-04-28
Make new dashboard default

rmars:
* Add navigation to the new dashboard
* Move the old dashboard to /legacy-dashboard, will remove code in a follow-up branch.

I'm not sure "router overview" is what we want that first nav to say, but leaving for now.

### DIFF
--- a/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
+++ b/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
@@ -29,7 +29,7 @@ object AdminHandler extends HtmlView {
             <li><a href="/metrics">metrics</a></li>
             <li><a href="/admin/logging">logging</a></li>
             <li><a href="https://linkerd.io/help/">help</a></li>
-            <!-- <li><a href="/dashboard">Beta</a></li> -->
+            <li><a href="/legacy-dashboard">legacy dashboard</a></li>
           </ul>
 
           <ul class="nav navbar-nav navbar-right">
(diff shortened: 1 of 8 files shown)


linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
```diff
@@ -29,7 +29,7 @@ object AdminHandler extends HtmlView {
             <li><a href="/metrics">metrics</a></li>
             <li><a href="/admin/logging">logging</a></li>
             <li><a href="https://linkerd.io/help/">help</a></li>
-            <!-- <li><a href="/dashboard">Beta</a></li> -->
+            <li><a href="/dashboard">Beta dashboard</a></li>
```
olix0r: Should this be removed now? Or make it a link to the "Old dashboard"

olix0r: ... or is this served from the old dashboard.  regardless, i think this link needs to change...

rmars: Yeah, this is served from the old dashboard - in theory people shouldn't arrive on this page unless they visit /legacy-dashboard

rmars: Oh, or the dtabs I guess.  Okay, I will make it point to the new one!

olix0r: aiui `/dashboard` is no longer a valid link.  Also probably shouldn't call it "beta" anymore ;)

adleong: This is served on all the legacy admin pages (dtabs, metrics, etc).  Maybe this should be a link to the legacy dashboard?  Users can already return to the new dashboard by clicking on the logo (which links to "/").

rmars: Yeah, that might be good

rmars:
Added prefixing clients with `/` here
screen shot 2016-04-28 at 11 04 05 am

olix0r:
⭐ 😍

adleong:
:shipit: 💐

PR #352, opened 2016-04-29
Hide dashboard clients section if there are no clients

rmars:
Also tweak sone font weights

screen shot 2016-04-29 at 10 28 08 am

screen shot 2016-04-29 at 10 41 06 am

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/router_clients.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/router_clients.js
@@ -10,15 +10,18 @@ var RouterClients = (function() {
 
   return function (metricsCollector, routers, $clientEl, routerName, clientTemplate, clientContainerTemplate, colors) {
     var clientToColor = assignColorsToClients(colors, routers.clients(routerName));
-
-    var routerClients = [];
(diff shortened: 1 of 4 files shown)


admin/src/main/resources/io/buoyant/admin/js/router_clients.js
```diff
@@ -16,9 +16,15 @@ var RouterClients = (function() {
 
     routers.onAddedClients(addClients);
 
-    _.map(routers.clients(routerName), initializeClient);
+    var clients = routers.clients(routerName);
+    if (clients.length > 0) {
+      _.map(clients, initializeClient);
```
gtcampbell: This should probably be a forEach since we're not doing anything with the result? except that we should probably actually just assign routerClients to the result of _.map rather than having initializeClient push onto routerClients below.

siggy:
⭐ 👍  lgtm pending other comments

admin/src/main/resources/io/buoyant/admin/js/router_clients.js
rmars: good point

klingerf:
⭐ 🍡

PR #353, opened 2016-04-29
[RFC] Introduce ObserverCache

adleong:
Introduce ObserverCache for caching stamping observers.  ObserverCache contains two levels of caching: active and inactive.

The active cache is for Observers that have current outstanding requests against them.
Therefore, any Observer in the active cache must not be closed as this would cause the
outstanding requests to hang.

The inactive cache is for Observers without outstanding requests against them.  These Observers
are kept open while in the inactive cache in case they receive a request and need to be moved
to the active cache.  The inactive cache has LRU eviction and Observers are closed upon
eviction.  Since Observers in the inactive cache have no outstanding requests against them,
this is safe to do.

When get is called, the Observer for that key is moved to the active cache if it exists or
created and placed in the active cache if it does not exist.  When the value of an Observer
in the active cache changes, the Observer is moved to the inactive cache.  If the inactive
cache is full at that time, an older inactive Observer will be evicted.


namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
```diff
@@ -0,0 +1,80 @@
...
+ * cache is full at that time, an older inactive Observer will be evicted.
+ *
+ * @param activeCapacity The maximum size of the active cache.  If get would cause the active cache
+ *                       to exceed this size, a MaximumObservationsReached exception is returned
+ *                       instead.
+ * @param inactiveCapacity The maximum size of the inactive cache.  LRU eviction is used to
+ *                         maintain this constraint.
+ * @param mkObserver The function to use to create new Observers if they are not in either cache.
+ */
+class ObserverCache[K <: AnyRef, T](activeCapacity: Int = 100, inactiveCapacity: Int = 10)(mkObserver: K => Observer[T]) {
```
adleong: I have no idea what reasonable values for the cache sizes are.

The inactiveCache can be pretty small, I think, because observations will typically only stay there for the short period between when the observation fires and when clients re-query it.

klingerf: Can you add stats to capture the size of each cache while they're running?  I think that will help to inform what the defaults should be.  Overall, I think the default for activeCapacity should be as high as is reasonably possible, since running out of capacity in that cache results in exception, not eviction.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def get(key: K): Try[Observer[T]] =
+    Option(activeCache.get(key)).map(Return(_)).getOrElse {
+      synchronized {
+        Option(activeCache.get(key))
+          .map(Return(_))
+          .getOrElse(makeActive(key))
+      }
+    }
```
klingerf: Should the second get call here happen against inactiveCache? As written you're fetching twice from activeCache.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
```diff
@@ -0,0 +1,80 @@
...
+  private[this] def makeActive(key: K): Try[Observer[T]] =
+    if (activeCache.size < activeCapacity) {
+      val obs = Option(inactiveCache.getIfPresent(key)).getOrElse {
+        mkObserver(key)
+      }
+      activeCache.put(key, obs)
+      inactiveCache.invalidate(key)
+      obs.nextValue().ensure {
+        makeInactive(key, obs)
+      }
```
klingerf: Would it be worth making the contents of this method explicitly synchronized?  As it is, it's only called inside a synchronized block above, but I'm concerned that we may not remember to wrap it with synchronized if we start calling it elsewhere.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
adleong: yup, good idea

namerd/iface/interpreter-thrift/src/test/scala/io/buoyant/namerd/iface/ObserverCacheTest.scala
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+    TestObserver(obs, event, closed)
+  }
+
+  test("observer cache caches") {
+
+    var count = 0
+
+    val cache = new ObserverCache[String, String](1, 0)({ _ =>
```
klingerf: Could you add a test explicitly tests that get fetches from inactive and promotes to active?

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
adleong: This is intentional.  Once we enter the synchronized block we check again that the key hasn't entered the active cache before making it active.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
adleong: Yeah, that's probably wise.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
klingerf: Ohhhhhh. Ok, that makes more sense. Maybe add a comment to explain that?

namerd/iface/interpreter-thrift/src/test/scala/io/buoyant/namerd/iface/ObserverCacheTest.scala
klingerf: Ah, nevermind, I now see the "observer cache reactivates observations" test below.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -116,6 +116,8 @@ object ThriftNamerInterface {
     final def apply(tstamp: TStamp): Future[(Stamp, T)] =
       apply(Stamp(tstamp))
 
+    final def nextValue(): Future[(Stamp, T)] = pending.flatMap(_.future)
```
olix0r: this this side-effecty? if not, i'd omit `()`

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
olix0r: I'd argue that the class shouldn't have defaults (because i don't think it can really know what a sane size is without considering how its being used.  Defaults should be applied higher up

olix0r: Also, i'd probably avoid the double-function constructor syntax here -- I don't think it makes anything easier to read elsewhere, especially given that you explicitly provide attribute names.  I'd just make a single constructor...

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -265,22 +257,13 @@ class ThriftNamerInterface(
...
-          obs
-      }
-    }
+  private[this] val bindingCache = new ObserverCache[(String, Dtab, Path), NameTree[Name.Bound]](
+    activeCapacity = 100,
+    inactiveCapacity = 10
+  )({
+    case (ns, dtab, path) =>
+      BindingObserver(interpreters(ns).bind(dtab, path), stamper)
+  })
```
olix0r: tioli I'd probably write this as:
```scala
{ k =>
  val (ns, dtab, path) = k
  BindingObserver(interpreters(ns).bind(dtab, path), stamper)
}
```

since this makes it clearer that this isn't a partial function...


Also, I'd be inclined to split this out into a separate function so it's clear:

```scala
private[this] case class BindingKey(ns: Ns, dtab: Dtab, name: path)
def mkBindingObserver(k: BindingKey) =
  BindingObserver(...)

private[this] val bindingCache = new ObserverCache[BindingKey, NameTree[Name.Bound]](
  activeCapacity = 100,
  inactiveCapacity = 10,
  mkBindingObserver
)
```

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -333,34 +317,25 @@ class ThriftNamerInterface(
...
+        case NameTree.Leaf(bound) => bound.addr
+        case NameTree.Empty => Var.value(Addr.Bound())
+        case NameTree.Fail => Var.value(Addr.Failed("name tree failed"))
+        case NameTree.Neg => Var.value(Addr.Neg)
+        case NameTree.Alt(_) | NameTree.Union(_) =>
+          Var.value(Addr.Failed(s"${id.show} is not a concrete bound id"))
       }
     }
+    AddrObserver(resolution, stamper)
+  })
```
olix0r: similarly, i think this would be a bit clearer as a separate function rather than inlined.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
```diff
@@ -0,0 +1,80 @@
...
+          .getOrElse(makeActive(key))
+      }
+    }
+
+  private[this] val activeCache = new ConcurrentHashMap[K, Observer[T]]
+  private[this] val inactiveCache = CacheBuilder.newBuilder()
+    .maximumSize(inactiveCapacity)
+    .removalListener(new RemovalListener[K, Observer[T]] {
+      override def onRemoval(notification: RemovalNotification[K, Observer[T]]): Unit =
+        if (notification.wasEvicted()) notification.getValue.close()
```
olix0r: nit: prefer `notification.wasEvicted` without parens

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  private[this] def makeInactive(key: K, obs: Observer[T]): Unit = {
+    synchronized {
+      activeCache.remove(key)
+      // insert obs into the inactive cache if it's not present
+      val _ = inactiveCache.get(key, new Callable[Observer[T]] {
+        def call = obs
+      })
+    }
+  }
```
olix0r: nit: needless braces

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
```diff
@@ -0,0 +1,80 @@
...
+  def get(key: K): Try[Observer[T]] =
+    Option(activeCache.get(key)).map(Return(_)).getOrElse {
+      synchronized {
+        Option(activeCache.get(key))
+          .map(Return(_))
+          .getOrElse(makeActive(key))
+      }
+    }
+
+  private[this] val activeCache = new ConcurrentHashMap[K, Observer[T]]
```
olix0r: i'd probably add a note about CHM is used to make reads lockless, but all updates are explicitly synchronized

olix0r:
:star: i really like this! all of my comments are about clarity and not about the general approach

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
```diff
@@ -0,0 +1,94 @@
...
+    .maximumSize(inactiveCapacity)
+    .removalListener(new RemovalListener[K, Observer[T]] {
+      override def onRemoval(notification: RemovalNotification[K, Observer[T]]): Unit =
+        if (notification.wasEvicted) notification.getValue.close()
+    })
+    .build[K, Observer[T]]()
+
+  private[this] val scope = "observer-cache"
+  private[this] val activeSize = stats.scope(scope).addGauge("active-size")(activeCache.size)
+  private[this] val inactiveSize = stats.scope(scope).addGauge("inactive-size")(inactiveCache.size)
```
klingerf: I see in the test that these stats are changing as the cache sizes change, but I can't actually figure out how that's happening based on initializing them here.  Can you help to shed some light?

klingerf:
⭐  Thanks for making those updates. My question aside, this looks good to me.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ObserverCache.scala
adleong: Yeah, of course!  The signature of `addGauge` is `def addGauge(name: String*)(f: => Float): Gauge` so the second parameter is a closure that can be repeatedly invoked.

klingerf: Ah ok, thanks.

adleong:
I'd appreciate if someone could take another look at this.  I plumbed a StatsReciever through namerd so that the ObserverCache could report cache sizes.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -335,11 +335,16 @@ class ThriftNamerInterface(
       }
     }
     AddrObserver(resolution, stamper)
-  })
+  }
+  private[this] val addrCache = new ObserverCache[Path, Option[Addr.Bound]](
+    activeCapacity = 100,
+    inactiveCapacity = 10,
+    stats = stats,
+    mkObserver = observeAddr
```
olix0r: these stats receives should be scoped to differentiate the caches, right?

adleong: There are two gauges (one for active size, one for inactive size) in the same scope.

olix0r: We have multiple observer caches, bindingCache and addrCache, right? Are these differentiated in stats scope?

adleong: good call

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -265,22 +259,14 @@ class ThriftNamerInterface(
...
-          bindingCache += (key -> obs)
-          obs
-      }
-    }
+  private[this] def observeBind(ns: Ns, dtab: Dtab, path: Path) =
+    BindingObserver(interpreters(ns).bind(dtab, path), stamper)
+  private[this] val bindingCache = new ObserverCache[(String, Dtab, Path), NameTree[Name.Bound]](
+    activeCapacity = 100,
+    inactiveCapacity = 10,
+    stats = stats.scope("binding-cache"),
```
olix0r: resultant stats will be e.g. "binding-cache/obsever-cache/active-size" -- can we tighten this up a bit?  Do we even need the observer-cache scope anymore?  Also, I would omit "-size" -- key verbosity comes with a cost at scale.

Can this just be bindingcache/active? (seems more idiomatic with rest of finagle stats)

olix0r:
:star: thanks!

PR #355, opened 2016-04-30
etcd Dtab store configuration.

gtcampbell:
Note: has not been verified to run against an actual etcd instance yet.


gtcampbell:
resolving in favor of #362.

PR #357, opened 2016-04-30
Implement put and update for EtcdDtabStore

olix0r:


namerd/storage/etcd/src/main/scala/io/buoyant/namerd/storage/etcd/EtcdDtabStore.scala
```diff
@@ -1,23 +1,33 @@
...
-  def update(ns: Ns, dtab: Dtab, version: Version): Future[Unit] = ???
-
-  def put(ns: Ns, dtab: Dtab): Future[Unit] = ???
+  def update(ns: Ns, dtab: Dtab, version: Version): Future[Unit] = {
+    val Buf.Utf8(vstr) = version
+    val index = vstr.toLong
+    val buf = Buf.Utf8(dtab.show)
+    val key = root.key(Path.Utf8(ns))
+    key.compareAndSwap(buf, prevIndex = Some(index)).unit.rescue {
+      case ApiError(ApiError.TestFailed, _, _, _) =>
```
adleong: What if the key does not exist?  We should return a DtabNamespaceDoesNotExistException in that case, I think.

PR #358, opened 2016-05-02
Fix etcd event observation

olix0r:
Before, event observation used the maximum modifiedIndex from the returned
set of nodes. There are no guarantees that this index is anywhere near current,
and so the watch could replay the history of updates or fail with an unknown
index.

By using etcd index (as returned in the header), we can ensure we're watching
from a recent index.

### DIFF
--- a/etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
+++ b/etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
@@ -129,11 +129,11 @@ class KeyTest extends FunSuite with ParallelTestExecution {
   }
 
   private[this] object Watch {
-    def unapply(params: Params): Option[(Boolean, Option[Int])] = {
+    def unapply(params: Params): Option[(Boolean, Option[Long])] = {
       val Recursive(recursive) = params
-      val idx = getParam(params, "wait").find(_ == "true") flatMap { _ =>
+      val idx = getParam(params, "wait").find(_ == "true").flatMap { _ =>
         getParam(params, "waitIndex") flatMap { waitIndex =>
-          Try(waitIndex.toInt).toOption
+          Try(waitIndex.toLong).toOption
         }
       }
       Some((recursive, idx))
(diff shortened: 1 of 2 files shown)


etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
```diff
@@ -409,12 +409,12 @@ class KeyTest extends FunSuite with ParallelTestExecution {
     val key = mkKey(base) {
       case (Method.Get, Watch(true, None)) if currentIndex < responses.length =>
         val idx = currentIndex.toInt
-        requested(idx).setDone()
-        responses(idx)
+        requested(idx.toInt).setDone()
+        responses(idx.toInt)
```
gtcampbell: Why do we need these `toInt`s here following the one on line 411?

gtcampbell:
:star: this looks fine other than my question about the repeated `toInt` calls

etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala
olix0r: I can drop it here, but otherwise:
```
[error] /Users/ver/b/l5d/etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala:416: type mismatch;
[error]  found   : Long
[error]  required: Int
[error]         requested(idx).setDone()
[error]                   ^
[error] /Users/ver/b/l5d/etcd/src/test/scala/io/buoyant/etcd/KeyTest.scala:417: type mismatch;
[error]  found   : Long
[error]  required: Int
[error]         responses(idx)
[error]                   ^
[error] two errors found
[error] (etcd/test:compileIncremental) Compilation failed
[error] Total time: 10 s, completed May 2, 2016 7:49:22 PM
```

gtcampbell: right, I get why it's needed on 416-7 - thanks.

gtcampbell:
⭐  again, thanks for the minor fixup

adleong:
:star: 🍉

PR #364, opened 2016-05-04
retry application failures in linkerd

### LINKED ISSUE #361: response classification should inform requeue policy
Currently, finagle does not use response classification to inform retry logic. this means, for instance, that 5xx responses cannot be retried.

Implement a RetryFilter that honors response classifiers.  If this works, feed it back to finagle.

olix0r:
When certain requests fail, we should be able to retry them automatically.  _ResponseClassifiers_ determine whether a request-response pair should be considered a retryable failure. Once we are able to [configure how responses are clarified](https://github.com/BuoyantIO/linkerd/pull/339), we need to [change requeing logic](https://github.com/twitter/finagle/pull/504) to take these failure classifications into account.

I have temporarily copied finagle's RequeueFilter, and its associated stack module, verbatim so that we can take advantage of this functionality before finagle's next release. Once finagle absorbs this functionality, we can remove the duplicated logic from linkerd.

Fixes #361

### DIFF
--- a/router/core/src/main/scala/com/twitter/finagle/buoyant/XXX_ClassifiedRequeueFilter.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/buoyant/XXX_ClassifiedRequeueFilter.scala
@@ -1,0 +1,183 @@
+package com.twitter.finagle.buoyant
+
+import com.twitter.finagle._
+import com.twitter.finagle.param.HighResTimer
+import com.twitter.finagle.service._
+import com.twitter.finagle.stats.{Counter, StatsReceiver}
+import com.twitter.util._
+
+/*
+ * Copied from com.twitter.finagle.service.RequeueFilter
+ * [https://github.com/twitter/finagle/pull/504].
+ * 
+ * XXX remove once the Pull Request is merged & released.
+ */
+object XXX_ClassifierRequeueFilter {
+  def module[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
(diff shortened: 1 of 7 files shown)


olix0r:
It occurs to me that, for the purposes of linkerd, we probably want to move the requeue filter up the stack: Right now requeueing occurs on the same client. Given a situation like /host/foo => 1 * /srv/foo & 9 * /srv/bar where /srv/foo fails some portion of requests, we should be able to retry onto /srv/bar. Furthermore, in an ideal world, we'd perform load balancing, failure accrual, etc in the same fashion...

olix0r:
reminder to add configuration docs for https://github.com/BuoyantIO/linkerd/pull/339

router/core/src/main/scala/com/twitter/finagle/buoyant/XXX_ClassifiedRequeueFilter.scala
```diff
@@ -0,0 +1,183 @@
...
+        )
+        svcFactory(retryBudget, filter, scoped, requeues, next)
+      }
+    }
+
+  // The upper bound on service acquisition attempts
+  private[this] val Effort = 25
+
+  // semi-arbitrary, but we don't want requeues to eat the entire budget
+  private[this] val MaxRequeuesPerReq = 0.2
```
adleong: Is hardcoding this going to be a problem?

olix0r: Maybe eventually.  This is what finagle does fwiw.

adleong:
:star: 🐘

klingerf:
⭐ 🐴

PR #365, opened 2016-05-04
Add support for router-wide retry configuration

### LINKED ISSUE #336: make retry budgets configurable
there is a good default retry budget, but we should support overriding it.  backoff strategies should be similarly configurable.

olix0r:
Allows users to configure backoffs and retry budgets.

Fixes #336

### DIFF
--- a/linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
+++ b/linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
@@ -1,7 +1,7 @@
 package io.buoyant.linkerd
 
-import com.twitter.finagle.buoyant.DstBindingFactory
 import com.twitter.finagle.{Dtab, Stack}
+import com.twitter.finagle.buoyant.DstBindingFactory
 import io.buoyant.config.Parser
 import io.buoyant.namer.{ConfiguredNamersInterpreter, InterpreterInitializer, TestInterpreterInitializer, TestInterpreter}
 import io.buoyant.router.RoutingFactory
@@ -95,5 +95,22 @@ servers:
     val router = parse(yaml, Stack.Params.empty)
     val DstBindingFactory.Namer(interpreter) = router.params[DstBindingFactory.Namer]
     assert(interpreter.isInstanceOf[TestInterpreter])
+  }
+
+  test("with retries") {
+    val yaml =
+      """|protocol: plain
+         |client:
+         |  retries:
+         |    backoff:
+         |      kind: jittered
+         |      minMs: 100
+         |      maxMs: 10000
+         |    budget:
+         |      ttlSecs: 30
+         |      minRetriesPerSec: 3
+         |      percentCanRetry: 0.33
+         |""".stripMargin
+    assert(parse(yaml, Stack.Params.empty) != null)
   }
 }
--- a/linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
+++ b/linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
@@ -11,14 +11,20 @@ import org.scalatest.FunSuite
 
 class RouterTest extends FunSuite with Exceptions {
 
+  def parseConfig(
+    yaml: String,
+    protos: Seq[ProtocolInitializer] = Seq(TestProtocol.Plain, TestProtocol.Fancy),
+    interpreters: Seq[InterpreterInitializer] = Seq(TestInterpreterInitializer)
+  ): RouterConfig =
+    Parser.objectMapper(yaml, Iterable(protos, interpreters)).readValue[RouterConfig](yaml)
+
   def parse(
     yaml: String,
     params: Stack.Params = Stack.Params.empty,
     protos: Seq[ProtocolInitializer] = Seq(TestProtocol.Plain, TestProtocol.Fancy),
     interpreters: Seq[InterpreterInitializer] = Seq(TestInterpreterInitializer)
   ): Router = {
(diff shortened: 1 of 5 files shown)


linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
```diff
@@ -96,4 +96,21 @@ servers:
...
+         |    backoff:
+         |      kind: jittered
+         |      minMs: 100
+         |      maxMs: 10000
+         |    budget:
+         |      ttlSecs: 30
+         |      minRetriesPerSec: 3
+         |      percentCanRetry: 0.33
+         |""".stripMargin
+    assert(parse(yaml, Stack.Params.empty) != null)
```
adleong: we should check the value of the `Retries.Budget` stack param in the parsed router.

adleong:
reminder to document this in config.md as part of #364

linkerd/docs/config.md
```diff
@@ -273,6 +273,46 @@ aperture supports the following options (see [here][aperture] for option semanti
...
+      allowed in order to accommodate clients that have just started
+      issuing requests as well as clients that do not issue many
+      requests per window. Must be non-negative and if `0`, then no
+      reserve is given. (Default: 10)
+    * *percentCanRetry* -- Optional. The percentage of calls that can
+      be retried. This is in addition to any retries allowed for via
+      `minRetriesPerSec`.  Must be >= 0 and <= 1000. As an example, if
+      `0.1` is used, then for every 10 non-retry calls , 1 retry will
+      be allowed. If `2.0` is used then every non-retry call, 2
+      retries. (Default: 0.2)
```
klingerf: I think this sentence is clearer as: "If 2.0 is used then every non-retry call will be allowed 2 retries."

linkerd/docs/config.md
```diff
@@ -273,6 +273,46 @@ aperture supports the following options (see [here][aperture] for option semanti
...
+    be used (see below).
+    * *kind* -- The name of a backoff algorithm. Either _constant_ or
+      _jittered_.
+
+The _constant_ backoff policy takes a single configuration parameter:
+* _ms_ -- The number of milliseconds to wait before each retry.
+
+The _jittered_ backoff policy uses a
+[decorrelated jitter](http://www.awsarchitectureblog.com/2015/03/backoff.html)
+backoff algorithm and requires takes two configuration parameters:
```
klingerf: Typo (no requires): "... and takes two configuration parameters:"

linkerd/docs/config.md
```diff
@@ -273,6 +273,46 @@ aperture supports the following options (see [here][aperture] for option semanti
...
+* _minMs_ -- Optional. The minimum number of milliseconds to wait
+  before each retry.
+* _maxMs_ -- The maximum number of milliseconds to wait before each
+  retry.
```
klingerf: As far as I can tell, both of these parameters are actually required, in which case I'd reword the sentence above, and remove "Optional".

olix0r: They're not. jackson will happily default them to zero.

linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
```diff
@@ -96,4 +102,23 @@ servers:
...
   }
+
+  test("with retries") {
+    val yaml =
+      """|protocol: plain
+         |client:
+         |  retries:
+         |    backoff:
+         |      kind: jittered
+         |      maxMs: 10000
```
klingerf: Related to my comment below, I'm surprised that this config is valid without minMs specified.  Given that the jittered backoff case class is defined as `JitteredBackoffConfig(minMs: Int, maxMs: Int)`, I thought that implies that both params are required -- what am I missing?

linkerd/docs/config.md
olix0r: i've added a test, though.

klingerf: Ah, in that case I wonder if we'd be better off providing our own defaults.  Something like:

```
case class JitteredBackoffConfig(minMs: Option[Int], maxMs: Option[Int]) extends BackoffConfig {
  def mk = Backoff.decorrelatedJittered(
    start = minMs.map(_.millis).getOrElse(Duration.Zero),
    maximum = maxMs.map(_.millis).getOrElse(100.millis)
  )
}
```

linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
olix0r: jackson doesn't enforce any sort of requirements, it will just leave null-values for missing fields. for Ints, this is 0 (which is good for minMs).  for options, it's None.

olix0r:
@klingerf good call.  writing better tests for this showed that both min and max have to be greater than 0, so they're both required.

klingerf:
⭐ 🍧 Great, thanks for making those updates.

linkerd/core/src/test/scala/io/buoyant/linkerd/RouterTest.scala
adleong: This test is a bit confusing.  It creates a RetriesConfig that will fail if `.mk` is called on it.  Since you test that error case elsewhere, I'd just give a fully valid config here.

olix0r: ah yeah, that's an oversight. good catch.

gtcampbell:
:ship: looks nice, and I agree that the tests are very helpful.

PR #368, opened 2016-05-05
Add announcer support to linkerd (and zk announcer)

adleong:
In the router section of the linkerd config, you can now specify a list of announcers.  Each server may then specify a list of names to announce.

### DIFF
--- a/linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
+++ b/linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
@@ -7,23 +7,13 @@ import io.buoyant.config.types.HostAndPort
 import io.buoyant.linkerd.Announcer
 import java.net.InetSocketAddress
 
-class ZkAnnouncer(zkAddrs: Seq[HostAndPort], pathPrefix: String) extends Announcer {
+class ZkAnnouncer(zkAddrs: Seq[HostAndPort], pathPrefix: Path) extends Announcer {
   override val scheme: String = "zk-serversets"
 
   val underlying = new FZkAnnouncer(DefaultZkClientFactory)
 
   private[this] val connect = zkAddrs.map(_.toString).mkString(",")
 
-  override def announce(addr: InetSocketAddress, name: String, version: Option[String] = None): Future[Announcement] =
-    version match {
-      case Some(v) => underlying.announce(addr, s"$connect!$pathPrefix/$name-$v!0")
-      case None => underlying.announce(addr, s"$connect!$pathPrefix/$name!0")
-    }
-
-  private[this] val prefix = Path.read(pathPrefix)
-  override def concreteName(name: String, version: Option[String] = None): Path =
-    version match {
-      case Some(v) => Path.Utf8("#", "io.l5d.serversets") ++ prefix ++ Path.Utf8(s"$name-$v")
-      case None => Path.Utf8("#", "io.l5d.serversets") ++ prefix ++ Path.Utf8(name)
-    }
+  override def announce(addr: InetSocketAddress, name: Path): Future[Announcement] =
+    underlying.announce(addr, s"$connect!${(pathPrefix ++ name).show}!0")
 }
--- a/linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
+++ b/linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
@@ -1,0 +1,24 @@
+package com.twitter.finagle.zookeeper.buoyant
+
+import com.twitter.finagle.{Path, Announcement}
+import com.twitter.finagle.zookeeper.{ZkAnnouncer => FZkAnnouncer, DefaultZkClientFactory}
+import com.twitter.util.Future
+import io.buoyant.config.types.HostAndPort
+import io.buoyant.linkerd.Announcer
+import java.net.InetSocketAddress
+
+class ZkAnnouncer(zkAddrs: Seq[HostAndPort], pathPrefix: String) extends Announcer {
+  override val scheme: String = "zk-serversets"
+
+  val underlying = new FZkAnnouncer(DefaultZkClientFactory)
+
+  private[this] val connect = zkAddrs.map(_.toString).mkString(",")
+
+  override def announce(addr: InetSocketAddress, name: String): Future[Announcement] =
+    underlying.announce(addr, s"$connect!$pathPrefix/$name!0")
+
+
+  private[this] val prefix = Path.read(pathPrefix)
+  override def concreteName(name: String): Path =
+    Path.Utf8("#", "io.l5d.serversets") ++ prefix ++ Path.Utf8(name)
+}
--- a/linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
+++ b/linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
@@ -14,11 +14,16 @@ class ZkAnnouncer(zkAddrs: Seq[HostAndPort], pathPrefix: String) extends Announc
 
   private[this] val connect = zkAddrs.map(_.toString).mkString(",")
 
-  override def announce(addr: InetSocketAddress, name: String): Future[Announcement] =
-    underlying.announce(addr, s"$connect!$pathPrefix/$name!0")
-
+  override def announce(addr: InetSocketAddress, name: String, version: Option[String] = None): Future[Announcement] =
+    version match {
+      case Some(v) => underlying.announce(addr, s"$connect!$pathPrefix/$name-$v!0")
+      case None => underlying.announce(addr, s"$connect!$pathPrefix/$name!0")
+    }
 
   private[this] val prefix = Path.read(pathPrefix)
-  override def concreteName(name: String): Path =
-    Path.Utf8("#", "io.l5d.serversets") ++ prefix ++ Path.Utf8(name)
+  override def concreteName(name: String, version: Option[String] = None): Path =
+    version match {
+      case Some(v) => Path.Utf8("#", "io.l5d.serversets") ++ prefix ++ Path.Utf8(s"$name-$v")
+      case None => Path.Utf8("#", "io.l5d.serversets") ++ prefix ++ Path.Utf8(name)
+    }
 }
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/AnnouncerInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/AnnouncerInitializer.scala
@@ -1,11 +1,27 @@
 package io.buoyant.linkerd
 
-import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.fasterxml.jackson.annotation.{JsonIgnore, JsonProperty, JsonTypeInfo}
+import com.twitter.finagle.Path
 import io.buoyant.config.ConfigInitializer
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
 trait AnnouncerConfig {
+
+  @JsonProperty("prefix")
+  var _prefix: Option[Path] = None
+
+  @JsonIgnore
+  def defaultPrefix: Path
+
+  @JsonIgnore
+  def prefix: Path = AnnouncerConfig.hash ++ _prefix.getOrElse(defaultPrefix)
+
+  @JsonIgnore
   def mk(): Announcer
+}
+
+object AnnouncerConfig {
+  val hash = Path.Utf8("#")
 }
 
 trait AnnouncerInitializer extends ConfigInitializer
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/AnnouncerInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/AnnouncerInitializer.scala
@@ -1,0 +1,11 @@
+package io.buoyant.linkerd
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import io.buoyant.config.ConfigInitializer
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
+trait AnnouncerConfig {
+  def mk(): Announcer
+}
+
+trait AnnouncerInitializer extends ConfigInitializer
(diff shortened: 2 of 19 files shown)


olix0r:
re: naming, i think we may want to call these namers (and announcers) io.l5d.zk.serverset -- serverset describes the scheme (esp the format registered in the nodes) -- there can be other schemes on zk.  (for instance, we will probably want to add _io.l5d.zk.leader_ discovery for communicating with marathon).

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -173,8 +179,10 @@ trait RouterConfig {
   def router(params: Stack.Params): Router = {
     val prms = params ++ routerParams
     val param.Label(label) = prms[param.Label]
+    val announcers = _announcers.toSeq.flatten.map(_.mk)
```
gtcampbell: tiny: prefer `flatMap` to `flatten.map`

linkerd/announcer/serversets/src/main/scala/io/l5d/announcer/Serversets.scala
```diff
@@ -0,0 +1,13 @@
...
+import com.twitter.finagle.Announcer
+import com.twitter.finagle.zookeeper.buoyant.ZkAnnouncer
+import io.buoyant.linkerd.{AnnouncerInitializer, AnnouncerConfig}
+
+class ServersetsAnnouncerInitializer extends AnnouncerInitializer {
+  override def configClass = classOf[Serversets]
+}
+
+case class Serversets(hosts: Seq[String], pathPrefix: Option[String]) extends AnnouncerConfig {
+  override def mk(): Announcer = new ZkAnnouncer(hosts.mkString(","), pathPrefix.getOrElse("/discovery"))
```
gtcampbell: tioli: I like defining defaults as constants in a companion class vs. having them hard-coded in the code, but obviously it's not a huge deal in this case where it's only used once.

gtcampbell:
:star: This looks good, although I don't have direct experience with the Finagle ZK announcer.

linkerd/docs/config.md
```diff
@@ -127,6 +127,20 @@ Each router must be configured as an object with the following params:
...
   handles paths of the form `/$/`.
   * protocol-specific module params, if any (the _default_ module has none)
+* *announcers* (optional) -- a list of service discovery announcers that servers can announce to.
+  Each announcer is an object with the following params:
+  * *kind* -- the name of the announcer module
+  * announcer specific params, if any
+  
+### Announcers
+
+#### zk
```
klingerf: I think this header should be the fully qualified name: io.l5d.announcer.Serversets.  Otherwise there's nowhere else in this documentation that provides that information.

klingerf:
⭐ 🐑  Looks good to me.

linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.twitter.util.Future
+import java.net.InetSocketAddress
+
+class ZkAnnouncer(hosts: String, pathPrefix: String) extends Announcer {
+  override val scheme: String = "zk-serversets"
+
+  val underlying = new FZkAnnouncer(DefaultZkClientFactory)
+
+  override def announce(addr: InetSocketAddress, name: String): Future[Announcement] =
+    underlying.announce(addr, s"$hosts!$pathPrefix/$name!0")
```
olix0r: is this really a name? it seems in many cases this will be a path like role/env/job -- we should probably be graceful (or at least error explicitly) about leading slashes

adleong: I think we have a choice about how we want to handle names that contain slashes

1. require that names start with / and treat them as paths
2. require that names don't start with / and treat them as paths
3. escape/encode slashes

linkerd/core/src/main/scala/io/buoyant/linkerd/AnnouncerInitializer.scala
```diff
@@ -0,0 +1,12 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.twitter.finagle.Announcer
+import io.buoyant.config.ConfigInitializer
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
+trait AnnouncerConfig {
+  def mk(): Announcer
+}
+
+trait AnnouncerInitializer extends ConfigInitializer
```
olix0r: It would be helpful to have some documentation somewhere on how announcers work - it's sort of spread around the implementation right now and, unless I'm missing it, I don't really see an overview of how this works.  From what I gather, each server must be announced separately, and there is no mechanism for i.e. registering multiple servers in a single node?

Concrete example:
```yaml
- protocol: http
  servers:
  - label: https
    ...
  - label: http
    ...
  baseDtab: |
    /host/foo => /$/inet/127.1/8080 ;
    /host/bar => /$/inet/127.1/8081 ;
    /http/1.1/* => /host ;
```

I assume if I want to register both 'foo' and 'bar' i might have an announcers block like:
```yaml
- kind: io.l5d.zk.serversets
    hosts:
    - ...
    path: /role/env/foo
- kind: io.l5d.zk.serversets
    hosts:
    - ...
    path: /role/env/bar
```

But i'd end up with paths /role/env/bar/http, /role/env/bar/https, /role/env/foo/http, , /role/env/foo/https?  This might be okay, but it diverges a bit from, for instance, the announcer scheme used by aurora.

linkerd/examples/acceptance-test.yaml
```diff
@@ -49,6 +49,11 @@ routers:
     httpUriInDst: true
   responseClassifier:
     kind: retryableIdempotent5XX
+  announcers:
+  - kind: io.l5d.announcer.Serversets
```
olix0r: i'll comment on your proposal, but I don't think we should have "announcer" in this -- we're in an announcers block, what would it mean to specify anything else?  Also, i'd prefer that we avoid capitalization.

olix0r: I'd prefer just io.l5d.zk.serversets -- the fact that it's an announcer is self-evident

linkerd/core/src/main/scala/io/buoyant/linkerd/AnnouncerInitializer.scala
adleong: In your example are you trying to announce the same server to two different names, foo and bar?

If registering multiple servers I'd expect the config to be
```
announcers:
- kind: io.l5d.whatever
  path: /role/env
servers:
- label: foo
  ...
- label: bar
  ...
```

linkerd/announcer/zk/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.twitter.util.Future
+import java.net.InetSocketAddress
+
+class ZkAnnouncer(hosts: String, pathPrefix: String) extends Announcer {
+  override val scheme: String = "zk"
+
+  val underlying = new FZkAnnouncer(DefaultZkClientFactory)
+
+  override def announce(addr: InetSocketAddress, name: String): Future[Announcement] =
+    underlying.announce(addr, s"$hosts!$pathPrefix/$name!0")
```
olix0r: what does `!0` mean?

linkerd/core/src/main/scala/io/buoyant/linkerd/AnnouncerInitializer.scala
olix0r: In this example I have two servers that may route over two local services...  my point is there's not necessarily a 1:1 relationship between servers and services

linkerd/core/src/main/scala/io/buoyant/linkerd/Server.scala
```diff
@@ -102,7 +111,8 @@ class ServerConfig { config =>
     label.getOrElse(routerLabel),
     port.map(_.port).getOrElse(pi.defaultServerPort),
     pi,
-    serverParams
+    serverParams,
+    announce.getOrElse(false)
```
olix0r: I think we should default to true.  If I have an announcers block, I want things to be announced.

adleong:
Going to hold off on this until kind naming changes go through

erdody:
Is this something that can be merged or is it still blocked on something else?

adleong:
hi @erdody.  we're still having some internal discussions as to what the right design for announcement in linkerd is.

erdody:
Thanks @adleong. I saw some movement recently, does that mean you've decided to go ahead with merging this or are you still discussing? 
Also, what do you think about adding and optional address to AnnouncerName, to have the ability to override the one coming from the Server? The use case is: we need the server to bind to all interfaces (0.0.0.0) to cover container IP and localhost but that, of course, won't work for announcement.

adleong:
Hi @erdody, good question regarding container address mapping.  Could you tell me more about the setup that you had in mind?  Would you have ips hardcoded into your linkerd config?  Are your containers being scheduled by any orchestration system?  Would the scheduler fill the ip into a linkerd config template when the container gets scheduled?

erdody:
@adleong, we are planning to run linkerd as an in-container side-car (initially), linkerd-to-linkerd, and the IP is passed as an environment variable. We have a container initialization script that runs confd and renders the linkerd yaml file with ip, service name, etc.
We use aurora and per-container IPs (IPs belong to the container). After I sent the message I realized that the outbound server is what needs to be bound to localhost. The inbound server can work with the container IP only (which is the one we want to announce), nobody needs to connect to it internally. Sorry about the confusion.

Are you planning to merge this soon?

adleong:
Thanks for the info, @erdody, that's really helpful.  If the inbound server is configured with the container IP, then does this just work as is?

We're planning to have this reviewed and merged early this week.

linkerd/core/src/main/scala/io/buoyant/linkerd/Server.scala
```diff
@@ -30,12 +31,16 @@ trait Server {
...
   def addr: InetSocketAddress = new InetSocketAddress(ip, port)
+
+  def announce: Seq[AnnouncerName]
 }
 
 object Server {
 
   case class RouterLabel(label: String)
 
+  case class AnnouncerName(service: String, version: Option[String])
```
olix0r: This feels a bit awkward to me. I think this should be a Path -- there's more information that needs to be encoded in some registry schemes (e.g. zone, environment, role), etc.  I don't think we should try to force this into a 2-tuple, and instead should provide a mechanism for these schemes to encode announcer-specific coordinates.

erdody:
@adleong Yes, it would work as-is, we would use the container IP for the inbound server which is what the announcer would advertise.
Thanks for the PR timeline, looking forward to it.

olix0r:
The relationship between announcers and announce names seems likely to be a problem. Announcers may encode or require attributes specific to a registration scheme.  I don't think linkerd should attempt to provide a single abstraction over all possible registration schemes, and so we need a way to provide tighter coupling between announced names and the backend.

I think, ultimately, this requires supporting typed announce names that are bound to types supported by configured announcers...

olix0r:
:star: I'm okay merging this as-is if we can communicate that this will almost certainly change soon...

adleong:
@olix0r what if we did away with configuring announcers and announce names separately?  Something like

```
routers:
- protocol: http
  servers:
  - ip: 0.0.0.0
    port: 4140
    announcers:
    - kind: ...
```

The `Announcer` interface would change to just accept an InetSocketAddress and any naming would be internal to the announcer plugin itself.  For example:

```
routers:
- protocol: http
  servers:
  - ip: 0.0.0.0
    port: 4140
    announcers:
    - kind: io.l5d.serversets
      zkAddrs:
      - host: zk.foo.co
        port: 2181
      name: /discovery/prod/foo-service
```

do you think that would be better?

olix0r:
That's more what I had in mind, yeah; though this requires that each announcer has its own client (or we try to maintain a cache of clients by... something (lik, their (type, dst)?)

But, yeah, that addresses my base concern around abstractions.

linkerd/announcer/serversets/src/main/scala/com/twitter/finagle/zookeeper/buoyant/ZkAnnouncer.scala
```diff
@@ -0,0 +1,19 @@
...
+
+class ZkAnnouncer(zkAddrs: Seq[HostAndPort], pathPrefix: Path) extends Announcer {
+  override val scheme: String = "zk-serversets"
+
+  val underlying = new FZkAnnouncer(DefaultZkClientFactory)
+
+  private[this] val connect = zkAddrs.map(_.toString).mkString(",")
+
+  override def announce(addr: InetSocketAddress, name: Path): Future[Announcement] =
+    underlying.announce(addr, s"$connect!${(pathPrefix ++ name).show}!0")
```
olix0r: Is it correct that this announces all instances with shard id=0?  Should we support a Path scheme that encodes shard ids? e.g. `/some/path/to/thing/0`

olix0r: At the very least, we should be careful to document this, since it may lead to surprising behavior in some environments

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -44,6 +44,16 @@ object Linkerd extends App {
...
+            for (name <- server.announce) {
+              val announcers = running.announcers.filter {
+                case (prefix, announcer) => name.startsWith(prefix)
+              }
+              for ((prefix, announcer) <- announcers) {
+                log.info("announcing %s as %s to %s", server.addr, name.show, announcer.scheme)
+                announcer.announce(server.addr, name.drop(prefix.size)).onSuccess(closeOnExit)
+              }
+              if (announcers.isEmpty) log.warning("no announcer found for %s", name.show)
+            }
```
olix0r: Thinking aloud here -- perhaps it makes sense to pass announcers to routers/servers rather than have servers expose their paths to main.

My thought here is that it's an implementation detail that the list of endpoints is statically provided.  We could envision a model where routers can register additional services at runtime.  If routers have access to the announcers -- or an "AnnouncerDispatcher" that deals with prefix mapping/removal -- then the static-ness of announcement is an implementation detail rather than an fundamental assumption.

Perhaps YAGNI.  I'm fine with this as-is, just thinking about some possible future use cases.

adleong: The main reason to plumb this to main is so that we can `closeOnExit` the announcements.  But yes, I could imagine refactoring this so that routers expose a generic list of Closables.  I'm open to refactoring this in the future as the need arises.

olix0r:
:star: this turned out great

PR #370, opened 2016-05-05
Move requeueing up into the path stack

olix0r:
By default, the requeueing filter is installed on the client stack. However,
this means that retries must be performed on the same client, which isn't
always desirable. By moving retries into the (logical) path stack, retry
requests may issued on an client (i.e. if the name tree includes unions).

Furthermore, this adds some basic stats at the logical level so that we have
top-level successrate, etc.

Depends on #364

### DIFF
--- a/router/core/src/main/scala/io/buoyant/router/Router.scala
+++ b/router/core/src/main/scala/io/buoyant/router/Router.scala
@@ -4,7 +4,7 @@ import com.twitter.finagle._
 import com.twitter.finagle.buoyant._
 import com.twitter.finagle.client._
 import com.twitter.finagle.server.StackServer
-import com.twitter.finagle.service.{FailFastFactory, Retries}
+import com.twitter.finagle.service.{FailFastFactory, Retries, StatsFilter}
 import com.twitter.finagle.stack.Endpoint
 import com.twitter.finagle.stats.DefaultStatsReceiver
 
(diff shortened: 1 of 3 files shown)


router/core/src/main/scala/io/buoyant/router/Router.scala
```diff
@@ -253,12 +253,17 @@ object StackRouter {
...
+    def mkStack[Req, Rsp](orig: Stack[ServiceFactory[Req, Rsp]]): Stack[ServiceFactory[Req, Rsp]] = {
+      val stk = new StackBuilder[ServiceFactory[Req, Rsp]](orig)
+      stk.push(TlsClientPrep.nop[Req, Rsp])
+      stk.result.remove(Retries.Role)
+    }
   }
 
   def newPathStack[Req, Rsp]: Stack[ServiceFactory[Req, Rsp]] = {
     val stk = new StackBuilder[ServiceFactory[Req, Rsp]](stack.nilStack)
+    stk.push(StatsFilter.module)
```
adleong: how is this stats filter scoped?

olix0r: It uses the stats receiver for the path stack which is scoped by `rt/<rt>/dst/path/<name>`.  The HttpEndToEndTest illustrates this.

adleong:
:star: this is cool 🐯 

To make sure I understand, suppose we had a union: `/foo => /srv/a & /srv/b`
With this change, a (retryable) failure from service a gets retried in the path stack and the retry could get sent to either a or b.  How does finagle load balance over the union?  I assume that a and b each have their own load balancer which is not shared, which means that 50% of requests will also go to each client, even if one is slower than the other?

adleong:
is it possible to write a test that exercises the case where a request is retried on a different client?

olix0r:
Your understanding matches mine.  Currently load balancing does **not** factor into unions: name binding is distributed randomly/weighted (see c.t.f.factory.NameTreeFactory).

We can't really write a deterministic test but we can probably write a test with unions... We probably have two options:

1. write a test that issues up to N-large requests until a retry hits the alternate
2. write a test where we do the following:
  1. Use a dtab like `/req => /srv/a | /srv/b`
  2. issue request, which gets routed to _a_. don't respond
  3. cause `/srv/a` to go neg
  4. respond from _a_ with a retryable failure
  5. ensure that retry goes to _b_

I think (1) is easier, though it's a lilttle "sloppier".

klingerf:
⭐ 🌱  Afaict, this change is actually pretty small once #364 has shipped.

olix0r:
replaced by https://github.com/BuoyantIO/linkerd/pull/403

PR #371, opened 2016-05-05
Package 'basic' http identifier with the http protocol

olix0r:
There's no real benefit to packaging the 'basic' http identifier seperately
from the http protocol.


klingerf:
⭐ 🐚  Works for me.

adleong:
:star: 📦

PR #372, opened 2016-05-05
Fix etcd KeyTest

adleong:
Key creates use the PUT verb, not POST.

Also turned off parallel test execution which was causing test failures not to be reported.


klingerf:
⭐  Whoa, nice catch.

olix0r:
:star:

PR #374, opened 2016-05-05
Volume mounting change in README Docker Example.

tonyd3:
I was having issues getting a relative path to work for me, I think it might be clearer if the example just used an absolute path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ _path/to/myapp/linkerd.yml_, you could start linkerd in docker with
 the following command:
 
 ```
-$ docker run -p 4140:4140 -v /absolute/path/to/myapp:/myapp -w /myapp buoyantio/linkerd:0.0.10-SNAPSHOT linkerd.yml
+$ docker run -p 4140:4140 -v /absolute/path/to/myapp:/myapp buoyantio/linkerd:0.0.10-SNAPSHOT /myapp/linkerd.yml
 ```
 
 The list of image names may be changed with a command like:
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ _path/to/myapp/linkerd.yml_, you could start linkerd in docker with
 the following command:
 
 ```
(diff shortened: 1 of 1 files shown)


olix0r:
thanks, tony! would you mind signing the [CLA](https://buoyant.io/cla/) (if you haven't)?

tonyd3:
signed!

olix0r:
:star:

README.md
```diff
@@ -243,7 +243,7 @@ _path/to/myapp/linkerd.yml_, you could start linkerd in docker with
 the following command:
 
 ```
-$ docker run -p 4140:4140 -v path/to/myapp:/myapp -w /myapp buoyantio/linkerd:0.0.10-SNAPSHOT linkerd.yml
+$ docker run -p 4140:4140 -v /absolute/path/to/myapp:/myapp -w /myapp buoyantio/linkerd:0.0.10-SNAPSHOT linkerd.yml
```
klingerf: (apologies in advance for bike shedding on a 10-character change)

I actually think this example would be clearer if you volume mounted the file instead of the whole directory, and got rid of the -w flag.  So...

    $ docker run -p 4140:4140 -v /absolute/path/to/myapp/linkerd.yml:/etc/linkerd.yml buoyantio/linkerd:0.0.10-SNAPSHOT /etc/linkerd.yml

tonyd3: This is a big weakness in Docker(and I think the same in most container) volume mounting is that you can only mount a directory, not specific files. 
I do like getting rid of the -w and using absolute path, and it's what I did when I tried to run the example. I'll make that change.

klingerf: Ah, hmm -- I think we're probably using different versions of docker.  Mine supports mounting files [[link](https://docs.docker.com/engine/userguide/containers/dockervolumes/#mount-a-host-file-as-a-data-volume)].  But no matter, if it's not universally supported, I agree that mounting a directory and using an absolute path to get rid of the -w flag is the right approach.  Thanks!

klingerf:
⭐ 👍  Thanks for updating.

PR #380, opened 2016-05-06
add /admin/metrics/prometheus stats endpoint

siggy:
this adds a new `/admin/metrics/prometheus` endpoint to export stats in a format prometheus understands. the data mirrors the `/admin/metrics.json` endpoint.

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
@@ -1,0 +1,58 @@
+package io.buoyant.admin
+
+import com.twitter.common.metrics.Metrics
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{MediaType, Request, Response}
+import com.twitter.io.Charsets
+import com.twitter.util.Future
+import org.jboss.netty.buffer.ChannelBuffers
+import scala.collection.JavaConverters.mapAsScalaMapConverter
+import scala.util.matching.Regex
+
+/**
+ * PrometheusStatsHandler
+ *
+ * Export Stats in a Prometheus-compatible format. Prometheus (http://prometheus.io/)
+ * exports stats in a somewhat different format than the Finagle's MetricsExporter. Specifically,
+ * instead of delimiting scopes by forward slashes, a colon is used. For stats, instead of exporting
+ * percentiles (p50, p99, etc.) and other quantities (max, min, etc.) as individual stats, they
+ * should be exported as labels in the Prometheus format.
+ */
+private[admin] object PrometheusStatsHandler {
+  private[this] case class Escape(regex: Regex, replace: String) {
+    private[admin] def replaceAllIn(str: String) = regex.replaceAllIn(str, replace)
+  }
+  private[this] val escape1 = Escape("[!,]".r, "-")
+  private[this] val escape2 = Escape("[/]".r, ":")
+  private[this] val statPattern = """(.*)\.(count|sum|avg|min|max|stddev|p50|p90|p95|p99|p9990)$""".r
+
+  private[this] val escape3 = Escape("[.\\-$]".r, "_")
+
+  private[admin] def formatKey(key: String) = {
+    val escapedKey = escape2.replaceAllIn(escape1.replaceAllIn(key))
+
+    val parsedKey = escapedKey match {
+      case statPattern(label, stat) => s"""$label{stat="$stat"}"""
+      case _ => escapedKey
+    }
+
+    escape3.replaceAllIn(parsedKey)
+  }
+}
+
+private[admin] class PrometheusStatsHandler(registry: Metrics) extends Service[Request, Response] {
+  import PrometheusStatsHandler._
+
+  override def apply(request: Request): Future[Response] = {
+    val rsp = Response()
+    rsp.contentType = MediaType.Txt
+
+    val output = new StringBuilder
+    registry.sample().asScala.foreach {
+      case (key, value) => output ++= s"""${formatKey(key)} $value\n"""
+    }
+    output ++= "\n"
+    rsp.contentString = output.toString
+    Future.value(rsp)
+  }
+}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
   are marked as failures.
 * Add a `retries` client config section supporting configurable retry
   budgets and backoffs.
+* Add an `/admin/metrics/prometheus` stats endpoint.
 * Automatically retry certain types of failures, as determined by
   response classifiers.
 * Remove TLS support from the k8s namer in favor of using `kubectl proxy` for
--- a/admin/src/main/scala/io/buoyant/admin/Admin.scala
+++ b/admin/src/main/scala/io/buoyant/admin/Admin.scala
@@ -3,6 +3,7 @@ package io.buoyant.admin
 import com.twitter.app.{App => TApp}
 import com.twitter.finagle._
 import com.twitter.finagle.http.{HttpMuxer, Request, Response}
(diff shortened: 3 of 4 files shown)


admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
```diff
@@ -0,0 +1,52 @@
...
+ * PrometheusStatsHandler
+ *
+ * Export Stats in a Prometheus-compatible format. Prometheus (http://prometheus.io/)
+ * exports stats in a somewhat different format than the Finagle's MetricsExporter. Specifically,
+ * instead of delimiting scopes by forward slashes, a colon is used. For stats, instead of exporting
+ * percentiles (p50, p99, etc.) and other quantities (max, min, etc.) as individual stats, they
+ * should be exported as labels in the Prometheus format.
+ */
+class PrometheusStatsHandler extends Service[Request, Response] {
+  private[this] lazy val registry = MetricsStatsReceiver.defaultRegistry
```
olix0r: i'd take a registry in the constructor so this can be tested.  Something like:
```scala
class PrometheusStatsHandler(registry: MetricsRegistry) {
  ...
}
```

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
```diff
@@ -0,0 +1,52 @@
...
+      .replaceAll(",", "-") // chaining multiple hosts with a comma is not parsed
+      .replaceAll("/", ":")
+
+    val parsedKey = escapedKey match {
+      case StatPattern(label, stat) if ValidStatNames.contains(stat) =>
+        """%s{stat="%s"}""".format(label, stat)
+      case _ => escapedKey
+    }
+
+    parsedKey.replaceAll("\\.", "_").replaceAll("-", "_").replaceAll("\\$", "_")
```
stevej: replaceAll's first parameter is a regex that is compiled each time this is called. IIRC, you can create a Pattern in a static variable and use that instead to reduce allocations and cpu time.

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  import scala.collection.JavaConverters.mapAsScalaMapConverter
+
+  override def apply(request: Request): Future[Response] = {
+    val rsp = Response()
+    rsp.contentType = MediaType.Txt
+
+    val output = registry.sample().asScala.map {
+      case (key, value) => "%s %s".format(formatKey(key), value)
+    }.toList.sorted
```
olix0r: does this really need to be sorted? that's potentially a bunch of work...

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    rsp.contentString = output.mkString("\n") + '\n'
```
olix0r: @adleong any thoughts on mkString versus StringBuilder, etc?

olix0r: Generally, It would be preferable to not have to iterate over this thing many times.

can we just build the response string in the formatting loop?

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
stevej: Since these are all single characters, there's probably a more efficient way than using a regex. I'll dig around.

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  private[this] def formatKey(key: String) = {
+    val escapedKey = key.replaceAll("!", "-") // '!inet' address strings (connecting thrift clients) are not parsed by prometheus.
+      .replaceAll(",", "-") // chaining multiple hosts with a comma is not parsed
+      .replaceAll("/", ":")
+
+    val parsedKey = escapedKey match {
+      case StatPattern(label, stat) if ValidStatNames.contains(stat) =>
+        """%s{stat="%s"}""".format(label, stat)
```
olix0r: use string interpolation over .format. `s"""$label{stat="$stat"}"""`

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
olix0r: I think we could just do this in a single regex.  But, unfortunately this is a lossy transformation and will make it difficult to correlate with other systems.

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
```diff
@@ -0,0 +1,58 @@
...
+ * percentiles (p50, p99, etc.) and other quantities (max, min, etc.) as individual stats, they
+ * should be exported as labels in the Prometheus format.
+ */
+private[admin] object PrometheusStatsHandler {
+  private[this] case class Escape(regex: Regex, replace: String) {
+    private[admin] def replaceAllIn(str: String) = regex.replaceAllIn(str, replace)
+  }
+  private[this] val escape1 = Escape("[!,]".r, "-")
+  private[this] val escape2 = Escape("[/]".r, ":")
+  private[this] val statPattern = """(.*)\.(.*?)$""".r
```
adleong: Could fold the stat names in here maybe?
```"""(.*)\.(count|sum|avg|min|max|stddev|p50|p90|p95|p99|p9990)"""`

File: admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala [hunk 7]
```diff
@@ -0,0 +1,58 @@
...
+
+  override def apply(request: Request): Future[Response] = {
+    val rsp = Response()
+    rsp.contentType = MediaType.Txt
+
+    val output = new StringBuilder
+    registry.sample().asScala.foreach {
+      case (key, value) => output ++= s"""${formatKey(key)} $value\n"""
+    }
+    rsp.contentString = s"""$output\n"""
```
adleong: Two newlines at the end is intentional?

If so we can use the same StringBuilder
``` output ++= "\n" ```

File: admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala [hunk 8]
```diff
@@ -0,0 +1,58 @@
...
+
+private[admin] class PrometheusStatsHandler(registry: Metrics) extends Service[Request, Response] {
+  import PrometheusStatsHandler._
+
+  override def apply(request: Request): Future[Response] = {
+    val rsp = Response()
+    rsp.contentType = MediaType.Txt
+
+    val output = new StringBuilder
+    registry.sample().asScala.foreach {
```
adleong: Do we need to worry about latching logic here?  Do we want deltas for counters?  

Is there any value in building a `SampledValues` and feeding that through a Prometheus `StatsFormatter`?

olix0r: I think it's preferable to expose raw sums of counters. this way the exporter doesn't have to know anything about its polling interval. deltas should be handled by the stats system (i.e. at read time)

adleong:
:star: 🏡

olix0r:
:star: you may want to add an `Also-by: Vincent Chu <...@.....>` note to the bottom of your commit message so he gets credit ;)

PR #381, opened 2016-05-06
Fix issue where metrics were disappearing and reappearing

### LINKED ISSUE #376: client request & connection counts disappearing and reappearing in dashboard
I see some weird behavior on the dashboard which I may or may not be able to reproduce, where the "requests" and "connections" number for one particular client are disappearing then reappearing at exactly 1 Hz.

(The labels are there, just not the numbers.)

There are multiple clients, and this is just happening to the last one.

rmars:
* Fixes #376 
* Refactor some repeated metrics code into partials

The bug:
* The templates re-render every 1000ms when we fetch new data.
* When both metrics are null, there was no display value, which causes the displayed text to adjust upwards.
* Adding a default value of 0 if the metric has no value fixes this.

One disadvantage to the template approach is that when data takes 1 second to load, all of the metrics display their default values, such as "metric 0" as opposed to something like "requests 0"

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/template/metric.partial.template
+++ b/admin/src/main/resources/io/buoyant/admin/template/metric.partial.template
@@ -1,0 +1,16 @@
+<div class="{{containerClass}}" style={{#if borderColor}}"border-left: 1px solid {{borderColor}}"{{/if}}>
+  <div class="metric-header">
+    {{#if description}}
+      {{description}}
+    {{else}}
+      metric
+    {{/if}}
+  </div>
+  <div class="{{metricClass}}">
+    {{#if value}}
+      {{value}}
+    {{else}}
+      0
+    {{/if}}
+  </div>
+</div>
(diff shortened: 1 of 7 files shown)


admin/src/main/resources/io/buoyant/admin/template/metric.partial.template
```diff
@@ -0,0 +1,16 @@
+<div class="{{containerClass}}" style={{#if borderColor}}"border-left: 1px solid {{borderColor}}"{{/if}}>
+  <div class="metric-header">
+    {{#if description}}
+      {{description}}
+    {{else}}
+      metric
+    {{/if}}
```
esbie: since "metric" doesn't say a lot, drop the if/else

esbie:
⭐  🌟

admin/src/main/resources/io/buoyant/admin/template/metric.partial.template
rmars: I added an `update()` before we call `setInterval` so now, when we load the page, we shouldn't see blank spots in the client section.  We still see it when we add a new client, as we need to wait for the next call to `update()`... so I'll leave "metric" in as a placeholder for this case.

klingerf:
⭐ 🔧

PR #383, opened 2016-05-09
Trace response body contents if the http status code is an error.

esbie:
I'd like to surface more information in tracing about why a request fails,
starting with the service's failure response


rmars:
:star: 🔑

klingerf:
⭐ ✒️

olix0r:
Unfortunately, I think this is a dangerous feature.  We have no control over what sort of data is in response bodies -- it could potentially contain passwords, credit card numbers, health information, etc.  Instead, I think we should use headers to communicate errors.  linkerd, for instance, sets an `l5d-err` header with linkerd-generated errors.  I think we should use something like this rather than arbitrary response bodies.

PR #384, opened 2016-05-09
Add etcd announcer

adleong:
Based off of #368 

Add an etcd announcer that uses TTLs.  The announcer refreshes a key in etcd with a TTL at regular intervals.


etcd/src/main/scala/io/buoyant/etcd/Key.scala
```diff
@@ -133,6 +133,38 @@ class Key(key: Path, client: Service[Request, Response]) {
   }
 
   /**
+   * Create a key in the current directory with an in-order name.
+   *
+   * If value is None, the key is treated as a directory.  In order to
+   * create an empty data node, use `Some(Buf.Empty)`.
```
gtcampbell: Rather than requiring this explanatory comment, could we consider using a more structured data type for this and other similar APIs in this class? something like:

```scala
sealed trait KeyType

object Directory extends KeyType
object Empty extends KeyType
case class Named(value: Buf) extends KeyType
```

adleong: I like this idea as a refactor, but I think it's out of scope for this change.

gtcampbell: sounds reasonable

linkerd/docs/config.md
```diff
@@ -142,6 +142,20 @@ Announce to ZooKeeper using the serverset format.
...
+
+* *host* -- (optional) the etcd API host.  (default: localhost)
+* *port* -- (optional) the etcd API port.  (default: 2379)
+* *pathPrefix* -- (optional) the etcd path under which services should be registered. (default:
+  /discovery)
+* *ttlSecs* -- (optional) the TTL in seconds for announcements to live in etcd.  Announcements will
+  disappear from etcd after this amount of time if not refreshed.  (default: 2 minutes)
+* *refreshSecs* -- (optional) the frequency in seconds to refresh the TTL.  This value should be set
+  smaller than the TTL to ensure that the announcement does not disappear from etcd. (default: 1
+  minute)
```
gtcampbell: should we add a warning or error in the event that this is larger than the TTL? (ignore this if there already is one, but I couldn't spot it)

gtcampbell:
:star: looks good

Ashald:
I'd be happy to contribute Consul announcer (actually, port this one into linkerd https://github.com/ncbi/finagle-consul/blob/develop/src/main/scala/gov/nih/nlm/ncbi/finagle/consul/catalog/ConsulAgentCatalogAnnouncer.scala) once this PR is merged

adleong:
Closing this until we have more specific requirements around an etcd service registry format.

PR #385, opened 2016-05-10
Remove TLS support from k8s namer.

adleong:
Starting a local `kubectl proxy` is the recommended way to securely talk to the k8s cluster API.  See https://github.com/BuoyantIO/linkerd-examples/pull/12

### DIFF
--- a/linkerd/docs/config.md
+++ b/linkerd/docs/config.md
@@ -662,12 +662,8 @@ experimental.
 
 The Kubernetes namer is configured with kind `io.l5d.experimental.k8s`, and these parameters:
 
-* *host* -- the Kubernetes master host. (default: kubernetes.default.svc.cluster.local)
-* *port* -- the Kubernetes master port. (default: 443)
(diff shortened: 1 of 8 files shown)


linkerd/docs/config.md
```diff
@@ -662,23 +662,20 @@ experimental.
...
-  (default: /var/run/secrets/kubernetes.io/serviceaccount/token)
+* *host* -- the Kubernetes master host. (default: localhost)
+* *port* -- the Kubernetes master port. (default: 8001)
 
 For example:
 ```yaml
 namers:
 - kind: io.l5d.experimental.k8s
   host: kubernetes.default.svc.cluster.local
   port: 443
```
klingerf: I think it would be worthwhile to change this to `host: localhost` `port: 8001`, since 443 is typically SSL and we stopped supporting that.

klingerf:
⭐ 🐼  I like it -- just had that one documentation nit.

olix0r:
:star: we'll to update the examples repo as well (we can do this before the next release)

PR #388, opened 2016-05-11
Add configurable http path identifier

klingerf:
I'm adding a new http identifier that constructs destinations based on a variable number of "/" separated segments at the beginning of the request's http path.  As part of this change I'm refactoring the default http identifier slightly, moving it into the io.buoyant.linkerd.protocol.http package.

From the updated `config.md` file:

HTTP requests are routed based on a configurable number of "/" separated segments from the start of their HTTP path.

For instance, here's a router configured with an http path identifier:

```yaml
routers:
- protocol: http
  dstPrefix: /custom/prefix
  identifier:
    kind: path
    segments: 2
  servers:
    port: 5000
```

A request to `:5000/true/love/waits.php` will be identified as `/custom/prefix/true/love`.


esbie:
⭐ 🐫

router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -0,0 +1,22 @@
...
+  segments: Int = 1,
+  baseDtab: () => Dtab = () => Dtab.base
+) extends RoutingFactory.Identifier[http.Request] {
+
+  def apply(req: http.Request): Future[Dst] =
+    Path.read(req.path) match {
+      case path if path.size >= segments =>
+        Future.value(Dst.Path(prefix ++ path.take(segments), baseDtab(), Dtab.local))
+      case _ =>
+        Future.exception(new IllegalArgumentException(s"not enough segments in path ${req.path}"))
```
adleong: You could make an argument for just taking as many segments as there are, instead of erroring out.  Not sure which is better, I could see it either way.

adleong:
:star: 🐐 just a TIOLI on how to handle the case when there are fewer segments than requested

router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
klingerf: Yeah, I could go either way, but am inclined to explicitly fail the request since having fewer segments than expected is probably a sign that something is misconfigured.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/PathIdentifierConfig.scala
```diff
@@ -0,0 +1,28 @@
...
+
+class PathIdentifierInitializer extends IdentifierInitializer {
+  val configClass = classOf[PathIdentifierConfig]
+  override val configId = PathIdentifierConfig.kind
+}
+
+object PathIdentifierInitializer extends PathIdentifierInitializer
+
+object PathIdentifierConfig {
+  val kind = "path"
```
adleong: Let's use io.l5d.path

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -33,7 +33,7 @@ object TracingFilter extends SimpleFilter[Request, Response] {
     Trace.recordBinary("http.rsp.version", rsp.version.toString)
     rsp.contentLength.foreach(Trace.recordBinary("http.rsp.content-length", _))
     if (rsp.status.code >= 500 && rsp.status.code < 600) {
-      Trace.recordBinary("http.rsp.body", rsp.content.slice(0, 64 * 1000 /*64kb*/))
+      Trace.recordBinary("http.rsp.body", rsp.content.slice(0, 64 * 1000 /*64kb*/ ))
```
olix0r: unrelated

klingerf: It's auto formatting from scalariform.

olix0r:
lgtm, modulo module name change

klingerf:
Ok, have fixed merge conflicts with master and switched to io.l5d.path.

adleong:
:star: ⛲

PR #390, opened 2016-05-11
more verbose logging for invalid request errors

### LINKED ISSUE #389: gracefully handle invalid requests
We've observed `/info?txtAirPlay&txtRAOP` requests causing a stack trace with:
```
io.buoyant.router.RoutingFactory$UnknownDst: Unknown destination: invalid http request
```

It's preferable if we gracefully log more helpful info.

siggy:
This change adds more exception logging to both `RoutingFactory` and `DefaultIdentifier`. I erred on the side of redundant logging of the request, as I thought it prudent to capture requests in cases where we're throwing an `UnknownDst` exception, as well as when an `HTTP/1.1` request is missing a host header.

Example exception:
```
io.buoyant.router.RoutingFactory$UnknownDst: Unknown destination: Request("GET /info?txtAirPlay&txtRAOP", from /127.0.0.1:55704)
HTTP/1.1 request missing hostname: Request("GET /info?txtAirPlay&txtRAOP", from /127.0.0.1:55704)
```

Multiline exception format inspired by:
https://github.com/BuoyantIO/linkerd/blob/master/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala#L85

fixes #389

### DIFF
--- a/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
+++ b/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
@@ -10,7 +10,7 @@ object RoutingFactory {
   val description = "Performs per-request name binding"
 
   case class UnknownDst[Req](request: Req, cause: Throwable)
-    extends Exception(s"Unknown destination: ${cause.getMessage}", cause)
+    extends Exception(s"Unknown destination: $request / ${cause.getMessage}", cause)
     with NoStacktrace
 
   /**
--- a/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
@@ -33,7 +33,7 @@ object TracingFilter extends SimpleFilter[Request, Response] {
     Trace.recordBinary("http.rsp.version", rsp.version.toString)
     rsp.contentLength.foreach(Trace.recordBinary("http.rsp.content-length", _))
     if (rsp.status.code >= 500 && rsp.status.code < 600) {
-      Trace.recordBinary("http.rsp.body", rsp.content.slice(0, 64 * 1000 /*64kb*/))
+      Trace.recordBinary("http.rsp.body", rsp.content.slice(0, 64 * 1000 /*64kb*/ ))
     }
     rsp.contentType.foreach(Trace.recordBinary("http.rsp.content-type", _))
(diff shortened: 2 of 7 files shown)


adleong:
⭐ 💣

gtcampbell:
:star: looks good, although I wonder if any of this would be amenable to unit testing

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -33,7 +33,7 @@ object TracingFilter extends SimpleFilter[Request, Response] {
     Trace.recordBinary("http.rsp.version", rsp.version.toString)
     rsp.contentLength.foreach(Trace.recordBinary("http.rsp.content-length", _))
     if (rsp.status.code >= 500 && rsp.status.code < 600) {
-      Trace.recordBinary("http.rsp.body", rsp.content.slice(0, 64 * 1000 /*64kb*/))
+      Trace.recordBinary("http.rsp.body", rsp.content.slice(0, 64 * 1000 /*64kb*/ ))
```
olix0r: hm, not related to your change, but i'm very dubious about this feature in general. what if response is chunked? What if response contains sensitive data? I'd be more in favor of checking specific headers...

olix0r: also 64 * 1000 isn't 64kb... anyway, i don't think this line should have been touched by this change ;)

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -10,7 +10,7 @@ object RoutingFactory {
   val description = "Performs per-request name binding"
 
   case class UnknownDst[Req](request: Req, cause: Throwable)
-    extends Exception(s"Unknown destination: ${cause.getMessage}", cause)
+    extends Exception(s"Unknown destination: $request\n${cause.getMessage}", cause)
```
olix0r: not a strong objection, but i'm slightly dubious about putting newlines in exception messages... i've worked on things in the past that try to parse exceptions from logs and this sort of things make it super painful.

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
klingerf: The change in this branch is a result of scalariform reformatting.

PR #394, opened 2016-05-11
ZooKeeper Leader namer

### LINKED ISSUE #360: Support marathon master discovery via ZK
In HA mode marathon stores the master(s) address in zk using a format very similar to server sets. While this isn't an officially supported format and could possibly change, it would be a nice feature for environments where a constant http address for marathon does not necessarily exist.

adleong:
Fixes #360 

A namer backed by ZooKeeper leader election.  The path processed by this namer is treated as the ZooKeeper path of a leader group.  The namer resolves to the address stored in the data of the leader.

This can be used, for example, to discover the marathon master.

### Try it out ###

Start zookeeper and marathon locally:
```
bin/zkServer.sh start
./bin/start --master local --zk zk://localhost:2181/marathon
```
Run linkerd
```./sbt linkerd-examples/marathon:run```
Visit the dtab playground at `localhost:9990/delegator`.  View the delegation for `/io.l5d.zkLeader/marathon/leader`

### CODE REVIEW

**Comment by siggy** [comment]
⭐ 👍  works!

this is intended to be used in conjunction with a marathon namer, correct? if so, i'm having trouble crafting a config yaml that would leverage this setup, mind adding one to the repo?

**Comment by adleong** [comment]
@siggy done: check out marathon.yaml

**Comment by siggy** [comment]
This is really helpful, thanks!

I'd prefer to preserve the original marathon.yaml, as a domain name like `marathon.mesos` is more user-friendly than `/$/io.buoyant.namer.zkLeader/localhost:2181/marathon/leader`. Mind setting this up as a separate example yaml?

**Comment by siggy** [comment]
⭐ 👍 great work!

File: namer/marathon/src/test/scala/io/l5d/MarathonTest.scala [hunk 1]
```diff
@@ -11,7 +11,7 @@ class MarathonTest extends FunSuite {
 
   test("sanity") {
     // ensure it doesn't totally blowup
-    val _ = marathon(None, None, None, None).newNamer(Stack.Params.empty)
+    val _ = marathon(None, None, None, None, None).newNamer(Stack.Params.empty)
```
gtcampbell: tioli: I usually prefer named parameters when we have this many parameters

namer/zk-leader/src/main/scala/io/buoyant/namer/zkLeader/ZkLeaderNamer.scala
```diff
@@ -0,0 +1,85 @@
...
+          bind(path.take(n - 1), residual ++ path.drop(n - 1))
+        } else {
+          Activity.value(NameTree.Neg)
+        }
+    }
+  }
+
+  private[this] def leaderAddr(candidate: Candidate): Option[Addr] = {
+    try {
+      Option(candidate.getLeaderData.orNull).map { data =>
```
gtcampbell: curious: why the need for the orNull + Option wrapping here?

gtcampbell:
:star: looks good.

namer/zk-leader/src/main/scala/io/buoyant/namer/zkLeader/ZkLeaderNamer.scala
adleong: getLeaderData returns a `com.google.common.base.Optional`.  This is this simplest way I could think of to convert it to a `scala.Option`

PR #395, opened 2016-05-12
Namer paths now start with /#

### LINKED ISSUE #393: Prefix bound names with # path segment
In order to easily differentiate bound paths (ie paths that are to be processed by a namer rather than by delegation) we should apply the convention where bound paths start with a # path segment.

e.g. /#/io.l5d.fs/foobar

That is, /# should be prepended to the prefix of all namers.

This makes delegations easier to understand because it will be more immediately obvious which paths are "terminal".

adleong:
Fixes #393 

The path prefix `/#` now indicates that the path should be processed by a namer.  A namer matches a path starting with `/#/<prefix>`.  This makes it easier to quickly distinguish between "terminal" paths and paths that will be further processed by delegation.

Code change is in NamerInitializer.scala.
All other changes are fixing tests and docs.


esbie:
Cool! I'd really like to see a partial re-implementation of Namer.global with this prefix. As the review currently stands, our service discovery namers are `/#/namerId` but inet is still `/$/inet/host/port` and our http helpers are `/$/io.buoyant.http.domainToPathPfx/marathonId`

adleong:
The idea here was to differentiate between global namers (which start with /$/) and configured namers (which start with /#/).  This isn't set in stone; it would certainly be possible to use /$/ for both.  To me, this is less confusing and makes it a bit more explicit why you need a `io.l5d.fs` namer section, for example, but there's no `inet` namer section.

esbie:
Hm ok. I still think a single symbol representing terminality is more useful and clearer than two symbols representing terminality and/or configurability.

esbie:
⭐ I don't want to stand in the way of progress, but I'm interested in what others think here

wmorgan:
I kinda like the # vs $ distinction, but IANADE.

olix0r:
lgtm. i agree that it makes some sense to differentiate class-namers from configured-namers.

olix0r:
let's hold off on merging backwards-incompatible changes for the moment. there appear to be some issues in 0.5.0 that we should diagnose first.

adleong:
Thanks all for the input.  I'll go forward with this for now but am happy to revisit if we find that having 2 different prefix symbols becomes confusing.

PR #398, opened 2016-05-12
Fix disallowed chars in prometheus and add test

lsjostro:
Fix for disallowed characters in prometheus key names and added test.

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
@@ -22,21 +22,17 @@ private[admin] object PrometheusStatsHandler {
   private[this] case class Escape(regex: Regex, replace: String) {
     private[admin] def replaceAllIn(str: String) = regex.replaceAllIn(str, replace)
   }
-  private[this] val escape1 = Escape("[!,]".r, "-")
-  private[this] val escape2 = Escape("[/]".r, ":")
+  private[this] val delimiter = Escape("[/]".r, ":")
   private[this] val statPattern = """(.*)\.(count|sum|avg|min|max|stddev|p50|p90|p95|p99|p9990)$""".r
+  private[this] val disallowedChars = Escape("[^a-zA-Z0-9:]".r, "_")
 
-  private[this] val escape3 = Escape("[.\\-$]".r, "_")
+  private[this] def escapeKey(key: String) = {
+    disallowedChars.replaceAllIn(delimiter.replaceAllIn(key))
+  }
 
-  private[admin] def formatKey(key: String) = {
-    val escapedKey = escape2.replaceAllIn(escape1.replaceAllIn(key))
-
-    val parsedKey = escapedKey match {
-      case statPattern(label, stat) => s"""$label{stat="$stat"}"""
-      case _ => escapedKey
-    }
-
-    escape3.replaceAllIn(parsedKey)
+  private[admin] def formatKey(key: String) = key match {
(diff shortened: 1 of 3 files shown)


olix0r:
@lsjostro thanks for the fix! Would you mind signing our [Contributor License Agreement](https://buoyant.io/cla)?

olix0r:
:star: ignore the CI failure -- we need to change to coverage publishing script to exit gracefully for pull requests...

lsjostro:
@olix0r np, signed! 👍

admin/src/main/scala/io/buoyant/admin/PrometheusStatsHandler.scala
```diff
@@ -22,21 +22,17 @@ private[admin] object PrometheusStatsHandler {
...
-
+    val escapedKey = disallowedChars.replaceAllIn(delimiter.replaceAllIn(key))
     val parsedKey = escapedKey match {
       case statPattern(label, stat) => s"""$label{stat="$stat"}"""
       case _ => escapedKey
     }
-
-    escape3.replaceAllIn(parsedKey)
+    parsedKey
   }
```
klingerf: Nice, this is much easier to follow. Tiny nit -- the parsedKey val doesn't need to be assigned, and instead you could just return directly.  Something like:

```
private[admin] def formatKey(key: String) = {
  val escapedKey = disallowedChars.replaceAllIn(delimiter.replaceAllIn(key))
  escapedKey match {
    case statPattern(label, stat) => s"""$label{stat="$stat"}"""
    case _ => escapedKey
  }
}
```

Or perhaps even more concisely:

```
private[admin] def formatKey(key: String) =
  disallowedChars.replaceAllIn(delimiter.replaceAllIn(key)) match {
    case statPattern(label, stat) => s"""$label{stat="$stat"}"""
    case escapedKey => escapedKey
  }
```

lsjostro: Thanks! much cleaner! I will fix!

lsjostro:
@klingerf more cleaner and scala:ish now. ;-)

siggy:
@lsjostro this is awesome, thanks for cleaning this up, particularly for adding tests!

after reading through your tests, i realized how subtle this transformation is. i found an edge case on our side which fails with this change (not your fault at all, as this test never existed until now ;) ):
```scala
  test("Preserve non-histogram stats") {
    val key = PrometheusStatsHandler.formatKey("fd_count")
    assert(key == "fd_count")
  }
```

i took a crack at a couple more changes to try to address this, let me know what you think:
```scala
  private[this] val statPattern = """(.*)\.(count|sum|avg|min|max|stddev|p50|p90|p95|p99|p9990)$""".r
  private[this] val disallowedChars = Escape("[^a-zA-Z0-9:]".r, "_")

  private[this] def escapeKey(key: String) = {
    disallowedChars.replaceAllIn(delimiter.replaceAllIn(key))
  }

  private[admin] def formatKey(key: String) = key match {
    case statPattern(label, stat) => s"""${escapeKey(label)}{stat="$stat"}"""
    case _ => escapeKey(key)
  }
```

lsjostro:
nice catch @siggy! I like your solution. I'll go ahead and add that.

lsjostro:
@siggy should be fixed now.

klingerf:
⭐  Looks great -- thanks for making those updates.

siggy:
⭐ 👍 thank you for doing this @lsjostro !

siggy:
(un)related: ticket to get ci to pass: #399

PR #400, opened 2016-05-12
Split config.md into multiple files

adleong:
I found `config.md` was getting large and hard to navigate, so I split it into multiple files with relative links between them.


esbie:
I like this and think it's a good call. I think the more tedious work will be reworking the pipeline to get it into https://linkerd.io/
Any thoughts there?

adleong:
I haven't tried it yet, but it might just be a matter of updating `pull-linkerd-docs` to pull the whole docs directory instead of just `config.md`

esbie:
⭐  Will likely also require linkerd site navigation update. I'd test that the inter-file links work in linkerd-site before going through with the merge

adleong:
It looks as if we'll have to do a bit of munging to get links to work correctly.  I'll get that working on a branch in linkerd-site, then merge this, then merge the linkerd-site chagnes.

adleong:
:star: 
self-star
I don't mind updating or reverting this if we later decide that the split out files aren't working well.

PR #401, opened 2016-05-13
Update Identifier kind name

adleong:
* Move the default http identifier out of the io.l5d package.
* Change the kind from `default` to `io.l5d.default`.


olix0r:
while where here... should we choose another name besides default?  I.e. we might change the default at some point...  Also there may be identifiers for other plugins (is it confusing to have io.l5d.default be different across protocols)?  Perhaps io.l5d.http?

olix0r:
let's hold off on merging backwards-incompatible changes for the moment. there appear to be some issues in 0.5.0 that we should diagnose first.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpIdentifierConfig.scala
```diff
@@ -1,9 +1,10 @@
-package io.l5d
+package io.buoyant.linkerd.protocol
```
klingerf: In #388 I'm moving this file to:

io/buoyant/linkerd/protocol/http/DefaultIdentifierConfigTest.scala

Instead of:

io/buoyant/linkerd/protocol/DefaultHttpIdentifierConfigTest.scala

I think I have a slight preference for the former, given that's how we handle other http-specific configuration, but could switch to the latter if you prefer.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/MethodAndHostIdentifierTest.scala
```diff
@@ -0,0 +1,16 @@
...
+import org.scalatest.FunSuite
+
+class MethodAndHostIdentifierTest extends FunSuite {
+  test("sanity") {
+    val _ = new MethodAndHostIdentifierConfig().newIdentifier(Path.empty)
+  }
+
+  test("service registration") {
+    assert(LoadService[IdentifierInitializer].exists(_.isInstanceOf[MethodAndHostIdentifierInitializer]))
+  }
```
klingerf: These two tests are also present in linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/MethodAndHostIdentifierConfigTest.scala -- maybe just delete this file?

olix0r:
hmmm... so sorry to :bike::house: naming, but technically this is versionAndMethodAndMaybeHostAndMaybeURI...  other ideas: httpMeta, httpEnvolope...?

adleong:
I feel like MethodAndHost captures the essence of the identifier even if it's not a totally complete description of everything the identifier uses.  Envelope reminds me of SOAP and meta reminds me of the html meta tag.

olix0r:
I guess my main quibble is that host isn't included for http/1.0 requests

olix0r:
:star:

klingerf:
⭐ 🔨  Thanks for updating.

PR #402, opened 2016-05-13
Make HTTP tracing more efficient

olix0r:
The HTTP tracing filter is inefficient:
- It is always installed, even if tracing is disabled.
- It always annotates requests, even if the request is not sampled.
- It needlessly allocates anonymous functions.

In order to fix these issues:
- TracingFilter.module now only adds the filter is a non-nil tracer is present
- Annotations are only added if the request is actively being traced (i.e. it is sampled).
- Anonymous functions have been converted to non-allocating alternatives.

### DIFF
--- a/router/core/src/main/scala/com/twitter/finagle/buoyant/DstTracing.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/buoyant/DstTracing.scala
@@ -21,9 +21,11 @@ object DstTracing {
       private[this] val localDtabShow = dst.localDtab.show
       private[this] val pathShow = dst.path.show
       override def apply(conn: ClientConnection) = {
-        Trace.recordBinary("namer.dtab.base", baseDtabShow)
-        Trace.recordBinary("namer.dtab.local", localDtabShow)
-        Trace.recordBinary("namer.path", pathShow)
+        if (Trace.isActivelyTracing) {
+          Trace.recordBinary("namer.dtab.base", baseDtabShow)
+          Trace.recordBinary("namer.dtab.local", localDtabShow)
+          Trace.recordBinary("namer.path", pathShow)
+        }
         self(conn)
       }
     }
@@ -44,8 +46,10 @@ object DstTracing {
       private[this] val idStr = dst.idStr
       private[this] val path = dst.path.show
       override def apply(conn: ClientConnection) = {
-        Trace.recordBinary("dst.id", idStr)
-        Trace.recordBinary("dst.path", path)
+        if (Trace.isActivelyTracing) {
+          Trace.recordBinary("dst.id", idStr)
+          Trace.recordBinary("dst.path", path)
+        }
         self(conn)
       }
     }
--- a/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
@@ -1,43 +1,83 @@
 package io.buoyant.router.http
 
-import com.twitter.finagle.{Service, SimpleFilter}
-import com.twitter.finagle.http.{Request, Response}
+import com.twitter.conversions.storage._
+import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack, Stackable, param}
+import com.twitter.finagle.client.StackClient
+import com.twitter.finagle.http.{Request, Response, Status}
 import com.twitter.finagle.tracing.Trace
+
+object TracingFilter {
+  val role = StackClient.Role.protoTracing
+  val module: Stackable[ServiceFactory[Request, Response]] =
+    new Stack.Module1[param.Tracer, ServiceFactory[Request, Response]] {
+      val role = TracingFilter.role
+      val description = "Traces HTTP-specific request metadata"
+      def make(_tracer: param.Tracer, next: ServiceFactory[Request, Response]) = {
+        val param.Tracer(tracer) = _tracer
+        if (tracer.isNull) next
+        else (new TracingFilter).andThen(next)
+      }
+    }
+
+  val TransferEncoding = "transfer-encoding"
+}
 
 /**
  * Annotates HTTP method, uri, and status code, content-type, and content-length.
  */
-object TracingFilter extends SimpleFilter[Request, Response] {
+class TracingFilter extends SimpleFilter[Request, Response] {
+  import TracingFilter.TransferEncoding
 
   def apply(req: Request, service: Service[Request, Response]) = {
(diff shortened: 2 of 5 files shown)


olix0r:
cc @stevej

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -1,43 +1,83 @@
...
+class TracingFilter extends SimpleFilter[Request, Response] {
+  import TracingFilter.TransferEncoding
 
   def apply(req: Request, service: Service[Request, Response]) = {
     recordRequest(req)
     service(req).onSuccess(recordResponse)
   }
 
-  private[this] def recordRequest(req: Request): Unit = {
+  private[this] def recordRequest(req: Request): Unit = if (Trace.isActivelyTracing) {
```
klingerf: Just an aesthetic comment, but it feels like the type signatures of `recordRequest` and `recordResponse` should be similar.  With your change we have:

    private[this] def recordRequest(req: Request): Unit
    private[this] val recordResponse: Response => Unit

Maybe convert `recordRequest` to a functional val too?

Also, TIOLI, but rather than checking `Trace.isActivelyTracing` in both functions, you could just do it once inside the apply method.

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -1,43 +1,83 @@
...
     Trace.recordRpc(req.method.toString)
     // http.uri is used here for consistency with finagle-http's tracing filter
     Trace.recordBinary("http.uri", req.uri)
     Trace.recordBinary("http.req.method", req.method.toString)
-    req.host.foreach(Trace.recordBinary("http.req.host", _))
+    req.host match {
+      case Some(h) => Trace.recordBinary("http.req.host", h)
+      case None =>
+    }
```
klingerf: I'm going to be sad if we can no longer use the collections API on options in our codebase.  Maybe we could test this change in isolation to see if it's necessary?  I feel like the big efficiency gains from this branch are going to come from the Trace.isActivelyTracing change and the change to only install the filter if tracing is configured.

olix0r: We don't have to eliminate them entirely, but as in finagle, we should _avoid_ using anonymous functions in the request path. We know it creates more objects, and we should strive to minimize allocation.  Don't fear imperative code ;)  They're fine to use in other places (i.e. configuration code, admin interfaces, etc), but we need to be disciplined about allocation.

router/core/src/main/scala/com/twitter/finagle/buoyant/DstTracing.scala
```diff
@@ -21,9 +21,11 @@ object DstTracing {
       private[this] val localDtabShow = dst.localDtab.show
       private[this] val pathShow = dst.path.show
       override def apply(conn: ClientConnection) = {
-        Trace.recordBinary("namer.dtab.base", baseDtabShow)
-        Trace.recordBinary("namer.dtab.local", localDtabShow)
-        Trace.recordBinary("namer.path", pathShow)
+        if (Trace.isActivelyTracing) {
```
adleong: `record` checks `isActivelyTracing` so we already shouldn't be annotating the request.  This does save us from constructing the annotation though. 

With this change. `debugTrace()` will no longer print annotations for non-sampled requests.

olix0r: Yeah, the goal here to save on needless object construction.

For debug trace, I think we should make the debug tracer a full fledged tracer like:

```yaml
tracers:
- kind: io.l5d.debug
```

This solves this problem nicely so that isActivelyTracing will be true.

Note that isActivelyTracing returns true when:

1. at least one tracer is configured
2. and sampling is _not false_.

klingerf: Big 👍  to `io.l5d.debug` tracer.

adleong:
:star: ✏️

klingerf:
⭐ 🐑

olix0r:
Would like to stagger this behind https://github.com/BuoyantIO/linkerd/pull/413, which moves some of the router tracing around.

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -1,43 +1,86 @@
...
-    // http.uri is used here for consistency with finagle-http's tracing filter
-    Trace.recordBinary("http.uri", req.uri)
-    Trace.recordBinary("http.req.method", req.method.toString)
-    req.host.foreach(Trace.recordBinary("http.req.host", _))
-    Trace.recordBinary("http.req.version", req.version.toString)
-    req.contentLength.foreach(Trace.recordBinary("http.req.content-length", _))
-    req.contentType.foreach(Trace.recordBinary("http.req.content-type", _))
-    req.headerMap.get("transfer-encoding").foreach { te =>
-      Trace.recordBinary("http.req.transfer-encoding", te)
+  private[this] def recordRequest: Request => Unit =
```
adleong: I think this should be a val

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
```diff
@@ -1,43 +1,86 @@
...
     }
-  }
 
-  private[this] def recordResponse(rsp: Response): Unit = {
-    Trace.recordBinary("http.rsp.status", rsp.status.code)
-    Trace.recordBinary("http.rsp.version", rsp.version.toString)
-    rsp.contentLength.foreach(Trace.recordBinary("http.rsp.content-length", _))
-    if (rsp.status.code >= 500 && rsp.status.code < 600) {
-      Trace.recordBinary("http.rsp.body", rsp.content.slice(0, 64 * 1000 /*64kb*/ ))
+  private[this] val recordResponse: Response => Unit =
```
adleong: TIOLI:
 ```
def recordResponse(rsp: Response): Unit = { ... }
val recordResponseFn: Response => Unit = recordResponse _
```

router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
olix0r: you're right good catch. it should be a val. and it can be a method but kl asked me to change it.  I think it makes sense as a method, so I'll change it.

adleong:
:star: ✒️

PR #403, opened 2016-05-15
Move classification-based retries into the router's RetryFilter

olix0r:
Introduced a ClassifiedRetries module that installs a ResponseClassifier-based RetryFilter.

Separate backoff logic: the requeue filter now uses _no_ backoff strategy,
while the configured backoff strategy controls the router's application-aware
retries.


router/core/src/main/scala/io/buoyant/router/Router.scala
```diff
@@ -255,11 +283,12 @@ object StackRouter {
      */
     def mkStack[Req, Rsp](orig: Stack[ServiceFactory[Req, Rsp]]): Stack[ServiceFactory[Req, Rsp]] =
       (orig ++ (TlsClientPrep.nop[Req, Rsp] +: stack.nilStack))
-        .replace(Retries.Role, XXX_ClassifierRequeueFilter.module[Req, Rsp])
   }
 
   def newPathStack[Req, Rsp]: Stack[ServiceFactory[Req, Rsp]] = {
     val stk = new StackBuilder[ServiceFactory[Req, Rsp]](stack.nilStack)
+    stk.push(StatsFilter.module)
+    stk.push(ClassifiedRetries.module)
```
olix0r: It's unclear to me if this is the proper ordering. it's consistent with finagle's client metrics, but I'm concerned that it will be difficult to compute a proper success rate with stats measured below retries.  I think for now we should move the StatsFilter above retries.  If we need stats below, we can add another module later.

adleong:
I've soft of lost track of where we've diverge from finagle's default stack in terms of requeues and retries.  A refresher would be helpful.  Are you sure that it's safe to not have any backoff for requeues?  What about if an entire cluster is overloaded and is trying to recover?  Is "requeue-storm" a problem?

olix0r:
After [input from kevino](https://github.com/twitter/finagle/pull/504#issuecomment-218615637), I think requeueing without backoffs is sane default behavior.  We may want to to introduce requeue backoffs later, but I don't think it's necessary to do this immediately (and it certainly complicates configuration).  There are other mechanisms---failure accrual, failfast, budgets themselves---that help guard against requeue storms.  In the majority of cases, requeues occur when a downstream service has not received or process the request at all, unlike application-level retries.  In the case of Nacks, the finagle server has rejected the request before it has been admitted into application code.  Requeues are generally safe to retry immediately.

At this point, we'll be closer to finagle's default stack in terms of requeues; and we introduce an application-level retry filter within the path stack.  It shares a retry budget because retry budgets should be coordinated (otherwise when the requeue filter exhausts its budget, the classified retry filter would kick in continue to retry --- we can think of retry budgets as limiting _client_ work as much as limiting server work, and so we we want to limit the max amount of extra work this client does.  This explains why retry budgets are intended to be shared across clients (even in finagle).  We care about being able do backoff application requests because these requests generally touch application code, and so we want a mechanism to give the application time to recover. This does not apply the same way for requeues.

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -202,10 +213,16 @@ case class RetriesConfig(
...
+  def mkBackoff: Option[ClassifiedRetries.Backoffs] =
+    backoff.map(_.mk).map(ClassifiedRetries.Backoffs(_))
+
+  // We use an empty backoff for Retries.Budget, since this informs
+  // _requeue_ delay. Requeues are explicitly for Nacks and
+  // non-application-level failures, and so we want to reenqueue
+  // these as quickly as possible.
+  @JsonIgnore
+  def mkBudget: Option[Retries.Budget] =
+    budget.map { b => Retries.Budget(b.mk, Backoff.const(Duration.Zero)) }
```
adleong: TIOLI could use `Retries.Budget.emptyBackoffSchedule` instead of `Back.const(Duration.Zero)`

adleong:
:star: 🎱

router/core/src/main/scala/io/buoyant/router/ClassifiedRetries.scala
```diff
@@ -0,0 +1,56 @@
...
+object ClassifiedRetries {
+  val role = Stack.Role("ClassifiedRetries")
+
+  case class Backoffs(backoff: Stream[Duration])
+  implicit object Backoffs extends Stack.Param[Backoffs] {
+    val default = Backoffs(Backoff.const(Duration.Zero))
+  }
+
+  /**
+   * A retry policy that uses a response classifier to compute
```
olix0r: incomplete sentence

olix0r:
:shipit: myself

PR #404, opened 2016-05-15
Remove the 'Connection' header from downstream requests.

olix0r:
Per RFC2616:

> The Connection general-header field allows the sender to specify options that
> are desired for that particular connection and MUST NOT be communicated by
> proxies over further connections.


klingerf:
⭐ 🐚  Looks good to me.

adleong:
:star: 🔗

PR #407, opened 2016-05-16
Constrain retryable exceptions

olix0r:
Not all exception types are safe or wise to retry. Instead, limit retries to
known connection-level events including timeouts and disconnects.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
@@ -1,11 +1,11 @@
 package io.buoyant.linkerd.protocol.http
 
-import com.twitter.finagle.{ChannelClosedException, Failure, TimeoutException, WriteException}
+import com.twitter.finagle.service.ResponseClassifier
 import com.twitter.finagle.service.RetryPolicy.{TimeoutAndWriteExceptionsOnly, ChannelClosedExceptionsOnly}
 import com.twitter.finagle.http.{Method, Request, Response, Status}
 import com.twitter.finagle.http.service.HttpResponseClassifier
 import com.twitter.finagle.service.{ResponseClass, ReqRep, ResponseClassifier}
-import com.twitter.util.{NonFatal, Return, Throw, Try, TimeoutException => UtilTimeoutException}
+import com.twitter.util.{NonFatal, Return, Throw, Try}
 import io.buoyant.config.ConfigInitializer
 import io.buoyant.linkerd.{ResponseClassifierConfig, ResponseClassifierInitializer}
 
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
@@ -1,9 +1,11 @@
 package io.buoyant.linkerd.protocol.http
 
+import com.twitter.finagle.{ChannelClosedException, Failure, TimeoutException, WriteException}
+import com.twitter.finagle.service.RetryPolicy.RetryableWriteException
 import com.twitter.finagle.http.{Method, Request, Response, Status}
 import com.twitter.finagle.http.service.HttpResponseClassifier
 import com.twitter.finagle.service.{ResponseClass, ReqRep, ResponseClassifier}
-import com.twitter.util.{NonFatal, Return, Throw}
+import com.twitter.util.{NonFatal, Return, Throw, Try, TimeoutException => UtilTimeoutException}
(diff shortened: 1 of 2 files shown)


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -58,14 +60,32 @@ object ResponseClassifiers {
...
+  object RetryableException {
+    def unapply(e: Throwable): Boolean = e match {
+      case RetryableWriteException(_) => true
+      case Failure(Some(_: TimeoutException)) => true
+      case Failure(Some(_: UtilTimeoutException)) => true
+      case _: TimeoutException => true
+      case _: UtilTimeoutException => true
+      case _: ChannelClosedException => true
+      case _ => false
+    }
```
klingerf: This is almost identical to Finagle's `TimeoutAndWriteExceptionsOnly` partial function, defined here:

https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/service/RetryPolicy.scala#L162

That function does not catch ChannelClosedException.  But maybe you could do something like:

    TimeoutAndWriteExceptionsOnly.orElse(ChannelClosedExceptionsOnly)

I think tying ourselves to their definition of retryable exceptions is worthwhile, since it may change over time.

olix0r: this signatures of finagle's things are a bit funky.. i'll see if i can reuse it, though

olix0r: retry policies take a `Try[Nothing]` and we are handling a `Try[Any]` (from ReqRep), so I'm not sure they can be reused here:

```
[error] /Users/ver/b/l5d/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala:77: type mismatch;
[error]  found   : PartialFunction[com.twitter.util.Try[Nothing],Boolean]
[error]  required: PartialFunction[com.twitter.util.Try[Any],Boolean]
[error]       TimeoutAndWriteExceptionsOnly.orElse(ChannelClosedExceptionsOnly)
[error]   
```

klingerf: Hmm, how about?

```
  object RetryableException {
    def unapply(e: Throwable): Boolean =
      RetryPolicy.TimeoutAndWriteExceptionsOnly
        .orElse(RetryPolicy.ChannelClosedExceptionsOnly)
        .isDefinedAt(Throw(e))
  }
```

You could constantize the `RetryPolicy.TimeoutAndWriteExceptionsOnly.orElse(RetryPolicy.ChannelClosedExceptionsOnly)`.

adleong:
```
def convert(nothing: PartialFunction[Try[Nothing], Boolean]): PartialFunction[Try[Any], Boolean] = { case th@Throw(_) if nothing.isDefinedAt(th.cast[Nothing]) => nothing(th.cast[Nothing]) }
```
I think this gives you a pf that's defined at the same places as the original.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -1,9 +1,11 @@
 package io.buoyant.linkerd.protocol.http
 
+import com.twitter.finagle.{ChannelClosedException, Failure, TimeoutException, WriteException}
```
klingerf: Thanks for updating. These imports are no longer used.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -1,9 +1,11 @@
...
+import com.twitter.finagle.service.RetryPolicy.{TimeoutAndWriteExceptionsOnly, ChannelClosedExceptionsOnly}
 import com.twitter.finagle.http.{Method, Request, Response, Status}
 import com.twitter.finagle.http.service.HttpResponseClassifier
 import com.twitter.finagle.service.{ResponseClass, ReqRep, ResponseClassifier}
-import com.twitter.util.{NonFatal, Return, Throw}
+import com.twitter.util.{NonFatal, Return, Throw, Try, TimeoutException => UtilTimeoutException}
```
klingerf: Nor is UtilTimeoutException.

adleong:
:star: ⚾️

klingerf:
⭐ 🐑  Thanks for updating.

PR #408, opened 2016-05-17
Fix zkLeader package and class name conflict.

adleong:
io.buoyant.namer.zkLeader referred to both a package and a class.  This seems to cause non-deterministic compiler errors.

### DIFF
--- a/linkerd/examples/marathon-leader.yaml
+++ b/linkerd/examples/marathon-leader.yaml
@@ -2,7 +2,7 @@ namers:
 - kind:      io.l5d.experimental.marathon
   prefix:    /io.l5d.marathon
   # Discover the marathon master by leader election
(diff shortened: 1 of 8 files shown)


olix0r:
wait where's the package?  it would be nice to keep the name lowercased to match the other finagle namers.  perhaps ...zk.leader?

olix0r:
Should the name in the config initializer change as well to keep it consistent? Can we write a simple sanity test to ensure the config loads?

adleong:
I don't think we need to update the config initializer name.  That convention that we've seemed to have fallen into is that configured namers can have more verbose names since these don't show up in configs ever (only kind names do), while classpath namers should have more concise (lowercase) names.

olix0r:
:star:

gtcampbell:
:ship:

klingerf:
⭐ 🐚

PR #418, opened 2016-05-18
Rename namerd iface kinds

adleong:
Update to match current conventions:

* kind names should be namespaced (io.l5d for 1st party plugins)


klingerf:
⭐

olix0r:
:star:

PR #421, opened 2016-05-18
Don't mark commons dependencies as 'twitter libs'

olix0r:
The withTwitterLib helper is used to build against snapshots of the twitter
tree for: util, finagle, scrooge, and twitter-server. However, commons
dependencies are not pushed with the same regularity, so we can't build against
snapshot versions of commons dependencies.


olix0r:
:shipit: trivial

PR #423, opened 2016-05-18
Use a static RoutingService in RoutingFactory

olix0r:
RoutingFactory currently builds a new RoutingService for each connection. The
server-side connection is passed into the DstBindingFactory.  This is wasteful,
considering that none of the server-side connection state matters to the
router. By simplifying this to use a static RoutingService with a nil
connection, we see substantial performance improvements in the face of
serverside connection churn.


stevej:
l5d is much much colder when connections are churning with this change. I'll keep the test running overnight so we can see if things unexpectedly melt down.

adleong:
:star: :100:

stevej:
⭐ I don't think pullapprove cares what I think but here we go.

olix0r:
@stevej au contraire, mon frere

PR #426, opened 2016-05-18
Move kind names out of doc headers.

adleong:
With certain stylings, headers are displayed in all caps which makes it impossible to see the casing of kind names.  We move kind names out of the headers to avoid this.


olix0r:
:star:

klingerf:
⭐ 🌈  Thanks for putting this together!

PR #430, opened 2016-05-20
Add Host header to k8s requests

### LINKED ISSUE #427: Requests to k8s API do not include a Host header
Requests that linkerd makes to the k8s API do not include a `Host` header, and requests like these are – for better or worse – [rejected by all Go HTTP servers](https://github.com/golang/go/blob/master/src/net/http/server.go#L782-L784) 😞. This is also in contravention of the HTTP 1.1 spec which [says](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23):
>A client MUST include a `Host` header field in all HTTP/1.1 request messages

obeattie:
Fixes #427. I have tested this in my environment, but apologies I haven't included any actual (automated) tests 🐼.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
@@ -23,7 +23,12 @@ trait ClientConfig {
 
   protected def mkClient(
     params: Stack.Params = Stack.Params.empty
-  ) = Http.client.withParams(Http.client.params ++ params).withStreaming(true)
+  ) = {
+    val setHost = new SetHostFilter(getHost, getPort)
+    Http.client.withParams(Http.client.params ++ params)
+      .withStreaming(true)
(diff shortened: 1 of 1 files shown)


olix0r:
@obeattie Thanks for the patch! Would you please fill out the [Contributor License Agreement](https://buoyant.io/cla) so we can accept this ;)

obeattie:
Done 🖋 😄

olix0r:
:star:

adleong:
:star: thanks, @obeattie

olix0r:
@obeattie thanks! fyi, we've published an updated kubectl to [buoyantio/kubectl:1.2.3](https://hub.docker.com/r/buoyantio/kubectl/tags/) and it sounds like the kubernetes folks are going to  start publishing kubectl regularly (https://github.com/kubernetes/kubernetes/issues/25900).

PR #435, opened 2016-05-22
Use "nightly" docker tags in CI.

olix0r:
Only publish docker tags from CI when the `NIGHTLY` build parameter is set to 1.

When the `TWITTER_DEVELOP` build parameter is also set to 1, publish the "unstable" tag.


klingerf:
⭐ 🌃

adleong:
:star:

PR #437, opened 2016-05-23
Fix Tls integration test by turning off retries

adleong:
When client tls is used with an incorrect common name, the server responds with a nack, which is retryable.  We set the retry budget to 0 in tests to avoid this retry.


olix0r:
:star:

PR #440, opened 2016-05-23
Bump EndToEndTest timeouts to 5 seconds

olix0r:
We still see way too many CI failures due to timeouts, presumably because the
CI infrastructure is slooooow. So let's set a ridiculously high timeout to work
around it.


adleong:
:star: ⌚

PR #441, opened 2016-05-24
Change client latency graph to be a success rate graph

### LINKED ISSUE #411: Display success rate rather than latency in main linkerd dashboard
* Replace the per-client latency line graph (which updates once per minute, making it somewhat confusing) with a success rate line graph. The graph should be pinned at the top to 100%, i.e. the top line should always be 100%, but the bottom can autoscale.

* Replace the "successes" counter with a success rate. Make this the top left element of the per-client table, i.e. the top row is "Success rate" and "Connections", and the bottom row is "Requests" and "Failures".

* Have the line colors next to each cell in the per-client table just be the color of the client line.

* Have the success rate text color change flip to red once it drops below 90%. (Hacky, but it's a start.)

rmars:
Fixes #411

I had to do some smoothie chart hacks to get the gray line to not be drawn over the client's coloured line - advice on that appreciated.

Some screenshots, with random data:
screen shot 2016-05-23 at 5 54 09 pm

screen shot 2016-05-23 at 5 02 53 pm

screen shot 2016-05-23 at 5 01 47 pm


admin/src/main/resources/io/buoyant/admin/css/dashboard-shared.css
```diff
@@ -36,3 +36,22 @@
 .failure-metric {
   color: #D84A4A;
 }
+
+/*
```
rmars: These are from `summary.css`, which I believe we can delete once we delete the old dashboard.

adleong:
Nice.  Some questions about the screenshots:

* In the 100% screenshot, I don't actually see a line at 100% in the line graph.
* In the other screenshots, the y-axis labels range from 0.0 to 1.0 instead of 0% to 100%
* The line graphs look very different client to client but the numerical success rate is the same for each client.  Is this just an artifact of the fake data?

olix0r:
question: typically, successrate will hover hover somewhere between 95% and 100%, or even between 99% and %100...  How do the graphs behave when success rate is at ~99.9%?  It's probably not a blocker, but ideally we should be able to show not-quite 100% success clearly.

rmars:
>  In the 100% screenshot, I don't actually see a line at 100% in the line graph.

In the 100% case, the top and bottom of the smoothie charts show 100% so the bottom line in that case is the 100% line.  I am looking into how to make this less weird, I don't think smoothie deals well with this.

> In the other screenshots, the y-axis labels range from 0.0 to 1.0 instead of 0% to 100%

My fault!  I updated the axes late, after I'd taken some screenshots.

> The line graphs look very different client to client but the numerical success rate is the same for each client. Is this just an artifact of the fake data?

Correct!  I mocked the numerical data to show 3 cases - good (green) warn (yellow) and bad (red) and I mocked the graph data with a Math.random() - I should have been more clear about this!

rmars:
@olix0r here are some more screenshots demonstrating what it would look like in various situations close to 100%!

Success rate between 99.0% and 99.9%:
alternating 99 9 and 99 0

Success rate between 99.90% and 99.95%:
alternating 99 9 and 99 5

At 100%:
100 success

At 99.9%:
99 9 success

olix0r:
lgtm! thanks, @rmars

adleong:
:star: It would be cool if there was a way to force the y-axis range to have a minimum of 0.001%.  That way, a constant 100% would have a y-axis range of [99.999, 100] and the line would show on the top instead of the bottom.

admin/src/main/resources/io/buoyant/admin/js/client_success_graph.js
```diff
@@ -0,0 +1,69 @@
...
+
+  function createChartLegend(successLineColor) {
+    return {
+      topLine: neutralLineColor,
+      successRate: successLineColor
+    }
+  }
+
+  function initializeChart($chartEl, dataKeys, timeseriesParamsFn) {
+    var $canvas = $("<canvas id='canvas' height='141'></canvas>");
```
esbie: lol we should not have so many charts all with id of "canvas" (id is supposed to guarantee uniqueness)

esbie:
⭐ 🍤

admin/src/main/resources/io/buoyant/admin/js/client_success_graph.js
rmars: omg, yes let me update that

rmars:
I added a chart hack that will create the reference line at 99.99% if the successRate is 100% - let me know what you think:

screen shot 2016-05-24 at 12 14 09 pm

screen shot 2016-05-24 at 12 10 19 pm

adleong:
:star: awesome! 🏯

PR #444, opened 2016-05-25
Remove problematic scaladoc links

adleong:
These external scaladoc links are causing scaladoc generation to fail.

### DIFF
--- a/router/core/src/main/scala/io/buoyant/router/Router.scala
+++ b/router/core/src/main/scala/io/buoyant/router/Router.scala
@@ -256,8 +256,8 @@ trait StdStackRouter[Req, Rsp, This <: StdStackRouter[Req, Rsp, This]]
 object StackRouter {
 
   /**
-   * A single budget needs to be shared across a [[RequeueFilter]] and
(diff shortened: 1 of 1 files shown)


rmars:
:star:

gtcampbell:
:star: note for a potential follow-up that there is an `apiMappings` feature in `sbt` that might allow things like this to work in the future, but I haven't verified it.

router/core/src/main/scala/io/buoyant/router/Router.scala
```diff
@@ -256,8 +256,8 @@ trait StdStackRouter[Req, Rsp, This <: StdStackRouter[Req, Rsp, This]]
 object StackRouter {
 
   /**
-   * A single budget needs to be shared across a [[RequeueFilter]] and
-   * a [[RetryFilter]] for debiting purposes, but we only want one of
+   * A single budget needs to be shared across a RequeueFilter and
```
stevej: can you add backticks to RequeueFilter and RetryFilter for code styling?

olix0r:
👍

PR #448, opened 2016-05-26
[RFC] Expose a list of possible bound names

adleong:
Looking for feedback on the general approach.

We create a sub-trait of `Namer` called `EnumeratingNamer` which defines `getAllNames`, an interface for getting all the bound names from service discovery.  Implementing this interface is optional; namers may implement either `Namer` or `EnumeratingNamer`.  We add an admin endpoint called `bound-names.json` which exposes the bound names from all the `EnumeratingNamer`s.  

`WatchingNamer` has been updated to implement `EnumeratingNamer` as a proof-of-concept.  Most other namers could be updated similarly.


namer/core/src/main/scala/io/buoyant/namer/EnumeratingNamer.scala
```diff
@@ -0,0 +1,8 @@
+package io.buoyant.namer
+
+import com.twitter.finagle.{Name, NameTree, Path, Namer}
+import com.twitter.util.{Activity, Future}
+
+trait EnumeratingNamer extends Namer {
+  def getAllNames: Future[Set[Path]]
```
olix0r: should this expose an Activity as well? Polling for a set is a special case of watching, and I imagine it makes sense to watch this for many applications

adleong: good idea

admin/src/main/scala/io/buoyant/admin/names/BoundNamesHandler.scala
```diff
@@ -0,0 +1,18 @@
...
+
+class BoundNamesHandler(namers: Seq[EnumeratingNamer]) extends Service[Request, Response] {
+  override def apply(req: Request): Future[Response] = {
+    Activity.collect(namers.map(_.getAllNames)).map { names =>
+      val json = names.toSet.flatten.map(_.show).mkString("[\"", "\",\"", "\"]")
+      val resp = req.response
+      resp.contentString = json
+      resp.contentType = MediaType.Json
+      resp
+    }.values.toFuture.flatMap(Future.const)
```
olix0r: tioli: i'd be inclined to write this as

`Activity.collect(namers.map(_.getAllNames)).values.toFuture.flatMap(Future.const).map { names =>`

(You don't really want an Activity of responses ever -- you want to get a list of names and then build a response). Not a huge deal.

Also nit: we use `rsp` throughout linkerd.  Prefer `rsp` over `resp` (since this isn't go ;p)

namerd/main/src/main/scala/io/buoyant/namerd/NamerdAdmin.scala
```diff
@@ -20,6 +27,7 @@ class NamerdAdmin(app: App, config: NamerdConfig, namerd: Namerd) extends Admin(
     )),
     "/" -> new DtabListHandler(namerd.dtabStore),
     "/delegator.json" -> new DelegateApiHandler(namerd.namers.toSeq),
-    "/dtab/" -> new DtabHandler(namerd.dtabStore)
+    "/dtab/" -> new DtabHandler(namerd.dtabStore),
+    "/bound-names.json" -> new BoundNamesHandler(enumeratingNamers.toSeq)
```
olix0r: should this be on the admin or as a part of the HTTP controller API?  I'd expect it to be on the controller api.

adleong: it's now on the linkerd admin and on the namerd http control iface.

olix0r:
⭐ lgtm

namer/fs/src/main/scala/io/buoyant/namer/fs/FsInitializer.scala
```diff
@@ -1,9 +1,10 @@
 package io.buoyant.namer.fs
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.Stack.Params
 import com.twitter.finagle.{Stack, Path}
 import io.buoyant.config.types.Directory
-import io.buoyant.namer.{NamerConfig, NamerInitializer}
+import io.buoyant.namer.{EnumeratingNamer, NamerConfig, NamerInitializer}
```
klingerf: Unused input EnumeratingNamer?

namer/fs/src/main/scala/io/buoyant/namer/fs/FsInitializer.scala
```diff
@@ -1,9 +1,10 @@
 package io.buoyant.namer.fs
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.Stack.Params
```
klingerf: Unused import?

klingerf:
⭐ 🐑  Looks good to me.

Issue #447, opened 2016-05-25
Better namerd admin for 0 namespaces


esbie:
Right now if there are no namespaces defined for namerd, the admin page displays nothing but the word "namespaces:" which looks very broken. Adding a link to docs on namespaces or using namerctl would be more useful

wmorgan:
Should the linkerd dtab UI just call out to namerd for resolution when it's configured, and we ditch the namerd ui entirely?

PR #450, opened 2016-05-27
Add a link to the delegator UI in the no hosts error message.

### LINKED ISSUE #445: have linkerd HTTP error messages link to the dtab playground
As per discussion on Slack:

New users often have trouble configuring the dtab, and the error behavior currently is a somewhat obscure plain-text message:

```No hosts are available for /http/1.1/POST/api-close, Dtab.base=[/http/1.1/*=>/#/io.l5d.k8s/linkerd/http], Dtab.local=[]. Remote Info: Not Available```

Ideally, this error message would have a link to a pre-populated dtab playground (including the path and Dtab.base) so that users could quickly debug the dtab trace. This wouldn't capture non-dtab issues, but it would speed up the debugging process when trying to configure things the first time.

adleong:
Fixes #445 

Changed the no hosts exception message from 

```No hosts are available for /http/1.1/POST/api-close, Dtab.base=[/http/1.1/*=>/#/io.l5d.k8s/linkerd/http], Dtab.local=[]. Remote Info: Not Available```

to

```Error resolving /http/1.1/GET/foobar.  See 0.0.0.0:9991/delegator?dtab.local=%2Fhost%2Ffoobar%3D%3E%2Fnope&router=http#%2Fhttp%2F1.1%2FGET%2Ffoobar for details. Remote Info: Not Available```

which is a link to a pre-populated delegator UI.

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/names/DelegateHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/names/DelegateHandler.scala
@@ -12,8 +12,15 @@ class DelegateHandler(
   namers: Seq[(Path, Namer)]
 ) extends Service[Request, Response] {
 
-  def apply(req: Request): Future[Response] =
-    dtabs().map(render).flatMap(view.mkResponse(_))
+  def apply(req: Request): Future[Response] = {
+    val dtabLocal = req.params.get("dtab.local") match {
+      case Some(dtabStr) => Dtab.read(dtabStr)
+      case None => Dtab.empty
+    }
+    dtabs().map { dtabs =>
+      render(dtabs.mapValues(_ ++ dtabLocal))
+    }.flatMap(view.mkResponse(_))
+  }
 
   def render(dtab: Map[String, Dtab]) =
     view.html(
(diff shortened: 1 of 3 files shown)


olix0r:
"See 0.0.0.0:9991" -- you can't actually visit that ;) it's going to be tricky to provide a routable address in a link? Furthermore, this link isn't particularly helpful when namerd is in play...  I'm not against the change but I'm skeptical how helpful it's actually going to be in practice.

wmorgan:
How about putting something like `<linkerd address>:9991`? That would give a hint about how to debug this.

Re: not being useful when namerd is in play, per my comment in #447, I think the linkerd dtab playground should be able to use namerd for resolution.

olix0r:
@wmorgan re: driving delegator UI from namerd: I agree in principal, but this is quite a bit trickier than it sounds.

olix0r:
also, we should consider that linkerd's admin UI may not be externally routable/discoverable (in typical k8s & mesos deploys).  We rely on namerd being served at a well-known place, but outside of development, I think it's atypical for users to have easy access to linkerd's admin UI.

adleong:
closing in favor of #453

PR #452, opened 2016-05-28
Add context links to docs

### LINKED ISSUE #451: add more context to split config file docs
Now that we've split config into sections with 0.6.0, the individual "split" pages are missing some context.

E.g. https://linkerd.io/doc/0.6.0/linkerd/retries/ doesn't mention that the block itself is called `retries`, or that it belongs in the `client` section. You need to navigate to another page to get to that.

At a minimum, we should add the above two bits of information to each of the split pages.

adleong:
Fixes #451

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/ProtocolInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/ProtocolInitializer.scala
@@ -92,44 +92,4 @@ trait ProtocolInitializer extends ConfigInitializer {
   def defaultServerPort: Int
 }
 
-object ProtocolInitializer {
 
-  /**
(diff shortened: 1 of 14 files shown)


esbie:
⭐ 
tioli: a wordier intro, eg "This documentation section describes the configuration options for the loadbalancer key"

olix0r:
lgtm.

I think it would be even clearer to include a full snippet of how each of these is used in the context of a configuration file.  E.g. for tls:

```yaml
routers:
- ...
  client:
    tls:
      kind: io.l5d.static
      commonName: foo.example.com
```

gtcampbell:
:star: LGTM

olix0r:
:star:

PR #454, opened 2016-06-01
Better defaults

stevej:
Removing JVM_HEAP and instead using a JVM_MIN_HEAP and JVM_MAX_HEAP allows linkerd to grow the heap as traffic increases. slow cooking it showed that memory use was pretty consistent at 1000 qps and didn't grow without bounds.

I removed `-XX:ReservedCodeCacheSize=32m` because that's already the default.
I removed `-XX:CICompilerCount=2` to let the JVM tune itself.

At 10 qps, RSS is 105mb, cpu is 4%
At 100 qps, RSS is 110mb, cpu is 10%
At 1000 qps, RSS is 140mb, cpu is 40%


olix0r:
👍 thanks @stevej

adleong:
:star: 💻

PR #455, opened 2016-06-02
Hold cached vars open

adleong:
# Problem

In ThriftNamerClient, we cache the bind Activity and addr Var.  However, each request passing through linkerd will observe an Activity/Var and then close the observation.  This means that the Var.async loop gets restarted on every request.  The result is that 2 calls to namerd are made for each linkerd request which causes extreme memory blowup.

# Solution

Register a dummy observer for each cached Var to hold it open.

Note: when we add eviction to ThiftNamerClient, we will also need to add a mechanism to close observations on the evicted Vars.


olix0r:
> In ThriftNamerClient, we cache the bind Activity and addr Var. However, each request passing through linkerd will observe an Activity/Var and then close the observation.

This is surprising to me.  My understanding is that `DynBoundFactory` (as created from `DstBindingFactory`) should be responsible for holding the observation open.

adleong:
Closing in favor of https://github.com/BuoyantIO/linkerd/pull/457

PR #457, opened 2016-06-02
Increase bindcache capacity

adleong:
If the number of distinct paths linkerd handles is larger than the path cache, evictions will cause a lot of churn.

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
@@ -143,6 +143,12 @@ trait RouterConfig {
   def bindingTimeout = _bindingTimeoutMs.map(_.millis).getOrElse(10.seconds)
 
   /*
+   * binding cache size
+   */
+
+  var bindingCache: Option[BindingCacheConfig] = None
+
+  /*
    * responseClassifier categorizes responses to determine whether
    * they are failures and if they are retryable.
    */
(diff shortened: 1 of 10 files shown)


gtcampbell:
How did we settle on these numbers (besides their being 10x the old ones)?

adleong:
Good question.  We want the cache capacity to be big enough to accommodate the number of distinct paths and bounds that linkerd will be receiving.  In practice, I don't really know what that is.  I pulled the 500 paths number out of thin air, but it feels like the right ballpark.  I welcome any input, though.

olix0r:
:star: though, we should follow up with easier to understand documentation about these parameters for users

adleong:
I've bumped the linkerd binding cache size down to 100 and the namerd binding cache size up to 1000.  This is a configuration I've load tested and am confidant works.  It should also be sufficient for small/medium sized organizations.  A total of 1000 logical names is probably plenty for many use cases.

I've also made all of these parameters configurable for those who these defaults aren't sufficient.

adleong:
filed https://github.com/BuoyantIO/linkerd/issues/463 as well

olix0r:
This looks good.  I still suspect the DstBindingFactory capacities are slightly on the low side -- certainly conceivable that modest-sized deployments could hit these limits.  Also, I'm curious how the namerd sizes relate to linkerd capacities -- right now the linkerd capacity is 1/10th the size of the namerd capacity. What's the motivation for limiting namerd caches at 100/10 -- was there any notable difference from 500/50?

adleong:
linkerd was certainly able to handle a DstBindingFactory path cache of size 500 no problem.  My motivation for sizing this down was that a) in a side-car model, I'd imagine that any particular linkerd wouldn't target very many logical paths.  i.e. the number of logical downstreams of a service is probably < 100 (or at least that's my unqualified intuition) and b) this cache doesn't have any time-based eviction so I wanted some mechanism to limit the number of "one-off" watches that are kept active.  e.g. if a developer sends some test traffic through with a different logical path, we probably don't want to keep a watch open on that path forever.  Shrinking the cache limits the number of these one-off watches.

olix0r:
@adleong thanks, that makes sense.

PR #460, opened 2016-06-03
Use scala.util.control.NoStackTrace instead of ctf.NoStackTrace

adleong:
ctf.NoStackTrace no longer exists in the latest version of finagle


gtcampbell:
:star: seems like a good idea

olix0r:
:star:

PR #462, opened 2016-06-06
Use :: as the zk host delimiter in the zk leader namer

### LINKED ISSUE #461: Impossible to specify multiple zk hosts to the zk leader namer
Because finagle paths do not admit commas, it is not possible to to specify multiple zk hosts to the zk leader namer.

adleong:
Fixes #461 

Use `::` as the zk host delimiter instead of `,` because commas are not permitted in finagle Paths.


olix0r:
LGTM. Though, I worry about how this will work with IPV6 addresses (where :: is perfectly legal). For better or worse, our current method doesn't work well with IPV6 anyway, so it's probably fine to proceed with this.  Would it be a better fix to simply admit commas in paths?

adleong:
Admitting commas in paths feels like a more invasive sweeping change.  I could imagine there existing comma-separated lists of paths out there somewhere...

PR #464, opened 2016-06-07
Forcibly disable coverage when building util in CI

olix0r:
Until https://github.com/twitter/util/pull/166 is fixed.


olix0r:
:shipit: self-ship trivial

PR #466, opened 2016-06-09
Dashboard style tweaks related to responsiveness

### LINKED ISSUE #234: Make dashboard responsive
Or at least deal with how it looks on widths less than full width.

Currently it looks like this:
screen shot 2016-04-04 at 3 34 06 pm

Part of #183

### LINKED ISSUE #456: Admin dashboard graphs lag over time
Am seeing an issue where the scrolling dashboard graphs begin to lag behind actual clock time when left open for a few hours.  The net effect of this is that the most recently displayed graph value does not match the summary value that's displayed elsewhere in the value.

rmars:
Despite the branch name and ticket name, note that there was not actually a lag bug :)

* removed the 1181 widths - since the dashboard expands on larger screens, having this as the width didn't have much meaning
* charts now determine their widths in a more sensible manner

Should make the dashboard display a little better on small screens, and fix the bug where part of the graph is cut off.

We definitely do some weird things when we co-opt the bootstrap grid (e.g. row margins to 0, container fluid padding to 0) - the -27px row margin was causing the weird white column on the right that you saw (that made it look like the graph was lagging).

Fixes  #456 
I'll say this also ~ fixes #234

screen shot 2016-06-09 at 1 28 52 pm

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/css/dashboard.css
+++ b/admin/src/main/resources/io/buoyant/admin/css/dashboard.css
@@ -24,7 +24,7 @@ body {
 }
 
 .row {
-  margin: 0 -27px 0 0;
+  margin: 0;
 }
 
 .container-fluid {
@@ -43,7 +43,6 @@ nav {
 
 .navbar-container {
   padding: 20px 27px 9px 27px;
-  width: 1215px; /*1161 + 27 + 27 (padding)*/
 }
 
 .nav {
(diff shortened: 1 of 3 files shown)


wmorgan:
Yay!

esbie:
⭐ 🌔

klingerf:
⭐ 🚴  Looks great! And no more horizontal scrollbar at smaller widths.  One thing I noticed is that the scrollbar reappears in Chrome when the window is wider than ~1380px, but that's trivial and doesn't need to be fixed as part of this branch, or at all really.

olix0r:
👍

PR #467, opened 2016-06-09
Add server success rate graph to dashboard

### LINKED ISSUE #436: unify dashboard server UI with client UI
Once we have client success rates displayed per #411, we should display server success rates as well. (Note that these may be higher than client success rates due to retries.)

I also notice there is a bug in the server stats, which is that success rate is always 100% even in the presence of errors. While fixing this, I think we should sanitize this like how we've done clients above--have it be success rate, requests, and failure (just those three), then a line graph that aligns with the per-client success rate graphs.

rmars:
Also fix the bug where server success rate was displaying 100% even in the presence of errors.

I'll rearrange the columns to make it more in line with the client section in a follow up branch - there's code reuse that can probably happen there.

Fixes #436

screen shot 2016-06-09 at 4 40 04 pm

screen shot 2016-06-09 at 4 30 23 pm


klingerf:
Awesome -- I like this change a lot.  For the new success rate graphs that you're adding and all of the existing ones, would it be possible to add a gray label to say that the chart is displaying success rate?  Similar to this label on the RPS graph:

screen shot 2016-06-09 at 6 18 27 pm

As it stands now there isn't a clear indicator in the UI of what is being displayed.  Also totally fine to do this in a separate branch or take a different approach altogether.

rmars:
Ooooh yes good point - let me add that.

wmorgan:
W00t! I have a minor tweak I'd like to make--I think the y axis should extend to 101% or something so that the 100% line doesn't look so much like a UI element--but that doesn't have to be part of this branch.

rmars:
@wmorgan here you go:
screen shot 2016-06-10 at 11 29 42 am

wmorgan:
IT'S BEAUTIFUL

olix0r:
👍

admin/src/main/resources/io/buoyant/admin/js/success_rate_graph.js
```diff
@@ -1,65 +1,76 @@
...
       successRate: successLineColor
     }
   }
 
-  function initializeChart($chartEl, dataKeys, timeseriesParamsFn) {
+  function getChartWidthFn(isFullWidthChart, $chartEl) {
+    if (isFullWidthChart) {
+      return function() { return $chartEl.width(); }
+    }
+    else {
```
esbie: nit:
} else {

admin/src/main/resources/io/buoyant/admin/js/success_rate_graph.js
```diff
@@ -1,65 +1,76 @@
...
+    return { min: min, max: max };
+  }
+
+  function initializeChart($chartEl, timeseriesParamsFn, chartWidthFn) {
     var $canvas = $("<canvas id='client-success-canvas' height='141'></canvas>");
     $chartEl.append($canvas);
 
     var chart = new UpdateableChart(
       {
-        maxValue: 100.000001, // the 0.000001 is a SmoothieCharts hack
```
esbie: Is there a reason this maxValue, 100.000001 is different from the yRange max of 100.001? Might help to pull these into constants

admin/src/main/resources/io/buoyant/admin/js/router_server.js
```diff
@@ -49,31 +52,50 @@ var RouterServer = (function() {
...
+      if(_.isFunction(defn.getRate)) {
+        defn.rate = defn.getRate(lookup);
+      }
+    });
+
+    return populatedMetrics;
   }
+
+  function getSuccessRate(data) {
+    var suc = _.find(data, function(d) { return d.metricSuffix === "success" });
```
esbie: tioli: more succinct as `_.find(data, "metricSuffix", "success")`

esbie:
👍  non-blocking comments above

admin/src/main/resources/io/buoyant/admin/js/success_rate_graph.js
rmars: Yeah, this maxValue doesn't make sense any more now that we're fully customizing the y range.  I got rid of this maxValue - the .001 is so that the graph line markers extend past the 100% success rate line (to make the graph look less confusing in the case of a pure 100% success rate). Before I was using a custom yRangeFunction, I was using the maxValue of 100.000001 to pin the graph's upper limit so that the 100% line would appear at the top and look reasonable, but the custom yRangeFunction does this better. I can pull these out into constants!

PR #469, opened 2016-06-10
linkerd delegator UI uses namerd for delegation

adleong:
Refator delegator to be a subtrait of NameInterpreter called DelegatingNameInterpreter.

Fix linkerd delegator UI when using namerd for name interpretation.  Add thrift endpoints to the ThriftNamerInterface for getting delegations and dtabs.  Have ThriftClientInterface extend DelegatingNameInterpreter and use these new thrift methods.  Refactor delegator UI to differentiate baseDtab (editable) and the dtab applied by the namer interpreter (un-editable).


olix0r:
at first blush, this looks like really nice.  it's a big change, so I'd like to spend a little more time understanding it (and check it out etc).

i noticed there's are typos in the commit history (e.g. "Refator") that you should probably rebase out of.

Also, I wonder if we can write some better tests for the tree-map bijections--I know I didn't do a great job of that with the original code, but it seems complex enough to warrant some dedicated unit tests.

olix0r:
Also, (bikeshedding), I'm curious if we should expose a long-polling-style API from delegator.json so that the page can receive updates as namerd change.  Fine to do in another/later branch, just thinking aloud.

admin/src/main/resources/io/buoyant/admin/js/delegator.js
```diff
@@ -19,19 +19,23 @@ var Delegator = (function() {
     return templates.node(obj);
   }
 
-  return function($root, dtab, t) {
+  return function($root, namespace, dtab, dtabBase, t) {
     templates = t
     $root.html(templates.delegator({}));
 
-    var dtabViewer = new DtabViewer(dtab, templates.dentry);
+    $("#dtab").html(dtab.map(function(e,_i) {
```
esbie: nit, I wouldn't bother listing _i here

esbie:
👍  js readability looks good

namer/core/src/main/scala/io/buoyant/namer/DelegatingNameInterpreter.scala
```diff
@@ -0,0 +1,21 @@
+package io.buoyant.namer
+
+import com.twitter.finagle.naming.NameInterpreter
+import com.twitter.finagle._
+import com.twitter.util.Activity
+
+trait DelegatingNameInterpreter extends NameInterpreter {
```
olix0r: Does this need to extend NameInterpreter? I think this trait could simply express delegation features, i.e. `Delegator`.  The NameInterpreter-ness of this type isn't really related to Delegator-ness.  `ConfiguredNamersInterpreter` would extend `NameInterpreter with Delegator`, and `DelegateApiHandler` can take a `String => Option[Delegator]`... tioli.

linkerd/admin/src/test/scala/io/buoyant/linkerd/admin/names/DelegateApiHandlerTest.scala
```diff
@@ -58,15 +59,15 @@ class DelegateApiHandlerTest extends FunSuite with Awaits {
   }
 
   test("invalid path results in 400") {
```
olix0r: Is it easy to test what happens when we request for a non-delegating interpreter?

admin/src/main/scala/io/buoyant/admin/names/DelegateHandler.scala
```diff
@@ -1,21 +1,37 @@
...
+                .map { interpreterDtab =>
+                  ns -> interpreterDtab
+                }
+            case _ => Future.value(ns -> Dtab.empty)
+          }
+      }.toSeq
+    }.map(_.toMap).map { interpreterDtabs =>
+      render(interpreterDtabs, baseDtabs)
+    }.flatMap(view.mkResponse(_))
+  }
```
olix0r: I generally prefer to break up complex blocks into a few clear statements:

e.g.

```scala
def apply(req: Request): Future[Response] = {
  val dtabs = baseDtabs.map {
    case (ns, dtab) => getDtab(ns).map(ns -> _)
  }
  Future.collect(dtabs.toSeq)
    .map { ds => render(ds.toMap, baseDtabs) }
    .flatMap(view.mkResponse(_))
}
```

I sometimes find it hard to understand chains of collect/map/flatmap, and I find it helpful to see them split out.

namer/core/src/main/scala/io/buoyant/namer/DelegatingNameInterpreter.scala
adleong: Yeah, probably.  I'll try it.

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -169,29 +175,40 @@ object DelegateApiHandler {
...
+            rsp.contentType = MediaType.Json
+            rsp
+          }
       case _ => err(Status.BadRequest)
     }
 }
 
 class DelegateApiHandler(
-  namers: Seq[(Path, Namer)]
+  interpreters: String => NameInterpreter
```
olix0r: I think DelegateApiHandler can now be expressed in terms of Delegators.  `delegatorByNs: String => Option[Delegator]`...

olix0r:
lgtm!

klingerf:
⭐  Tested this out and it works great.

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
adleong: Sure, it just moves where you have to handle the case where the interpreter is not a delegator.  Having that handled here means we can handle it in a consistent way.

PR #470, opened 2016-06-10
Overhaul linkerd header handling

olix0r:
As we introduce additional contextual information to be passed between linkerd
instances, we have an opportunity to establish better conventions that allow
for extension.

Specifically:

- All linkerd-generated headers are to be encoded as `l5d-ctx-*`.  Apps should be
  able to safely forward this prefix and only this prefix for linkerd to
  operate normally. This won't be the case until the next finagle release.
- The `l5d-ctx` header has been renamed to `l5d-ctx-trace`.
- The `l5d-ctx-deadline` header encodes deadlines.  Deadlines are not yet
  enforced.
- We now accept `l5d-dtab` in place of `dtab-local`.  Currently, `l5d-dtab` is
  appended to the `dtab-local` value. In a future release, we will probably
  stop honoring `dtab-local` in linkerd.
- We are now more gracious in accepting `l5d-sample` -- negative numbers are
  treated as 0 and values greater than 1 are treated as 1.
- We stop encoding `l5d-dst-*` headers on responses. These response headers are
  particularly meaningless in linkerd-to-linkerd mode, and generally serve as
  a potential information leak to clients.  These values are reported on
  downstream requests, but do not need to be forwarded by applications.

TODO
- [x] update changes
- [x] update documentation

### DIFF
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -96,12 +96,15 @@ _Context headers_ (`l5d-ctx-*`) are generated and read by linkerd
 instances. Applications should forward all context headers in order
 for all linkerd features to work. These headers include:
 
-- `dtab-local`: Currently (until the next release of finagle), the
+- `dtab-local`: currently (until the next release of finagle), the
   `dtab-local` header is used to propagate dtab context. In an
   upcoming release this header will no longer be honored, in favor of
   `l5d-ctx-dtab` and `l5d-dtab`.
-- `l5d-ctx-deadline`:
-- `l5d-ctx-trace`:
+- `l5d-ctx-deadline`: describes time bounds within which a request is
+  expected to be satisfied. Currently deadlines are only advisory and
+  do not factor into request cancellation.
+- `l5d-ctx-trace`: encodes zipkin-style trace IDs and flags so that
+  trace annotations emitted by linkerd may be correlated.
 
 Edge services should take care to ensure these headers are not set
 from untrusted sources.
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -85,3 +85,58 @@ routers:
 
 A request to `:5000/true/love/waits.php` will be identified as
 `/custom/prefix/true/love`.
+
+## HTTP Headers
+
+linkerd reads and sets several headers prefixed by `l5d-`.
+
+### Context Headers
+
+_Context headers_ (`l5d-ctx-*`) are generated and read by linkerd
+instances. Applications should forward all context headers in order
+for all linkerd features to work. These headers include:
+
+- `dtab-local`: Currently (until the next release of finagle), the
+  `dtab-local` header is used to propagate dtab context. In an
+  upcoming release this header will no longer be honored, in favor of
+  `l5d-ctx-dtab` and `l5d-dtab`.
+- `l5d-ctx-deadline`:
+- `l5d-ctx-trace`:
+
+Edge services should take care to ensure these headers are not set
+from untrusted sources.
+
+### User Headers
+
+_User headers_ are useful to allow user-overrides
+
+- `l5d-dtab`: a client-specified delegation override
+- `l5d-sample`: a client-specified trace sample rate override
+
+Note that if linkerd process incoming requests for applications
+(i.e. in linker-to-linker configurations), applications do not need to
+provide special treatment for these headers since linkerd does _not_
+forward these headers (and instead translates them into context
+headers).
+
+Edge services should take care to ensure these headers are not set
+from untrusted sources.
+
+### Informational Request Headers
+
+In addition to the context headers, linkerd may emit the following
+headers on outgoing requests:
+
+- `l5d-dst-path`: the logical name of the request as identified by linkerd
+- `l5d-dst-bound`: the concrete client name after delegation
+- `l5d-dst-residual`: an optional residual path remaining after delegation
+- `l5d-reqid`: a token that may be used to correlate requests in a
+               callgraph across services and linkerd instances
+
+### Informational Response Headers
+
+And in addition to the context headers, lay may emit the following
+headers on outgoing responses:
+
+- `l5d-err`: indicates a linkerd-generated error. Error responses
+             that do not have this header are application errors.
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -137,6 +137,10 @@ headers on outgoing requests:
 Applications are not required to forward these headers on downstream
 requests.
 
+The value of the _dst_ headers may include service discovery
+information including host names.  Operators may opt to remove these
+headers from requests sent to the outside world.
+
 ### Informational Response Headers
 
 linkerd may emit the following _informational_ headers on outgoing
@@ -147,3 +151,7 @@ responses:
 
 Applications are not required to forward these headers on upstream
 responses.
+
+The value of this header may include service discovery information
+including host names. Operators may opt to remove this header from
+responses sent to the outside world.
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -113,11 +113,12 @@ _User headers_ are useful to allow user-overrides
 - `l5d-dtab`: a client-specified delegation override
 - `l5d-sample`: a client-specified trace sample rate override
 
-Note that if linkerd process incoming requests for applications
+Note that if linkerd processes incoming requests for applications
 (i.e. in linker-to-linker configurations), applications do not need to
 provide special treatment for these headers since linkerd does _not_
 forward these headers (and instead translates them into context
-headers).
+headers). If applications receive traffic directly, they _should_
+forward these headers.
 
 Edge services should take care to ensure these headers are not set
 from untrusted sources.
@@ -127,16 +128,22 @@ from untrusted sources.
 In addition to the context headers, linkerd may emit the following
 headers on outgoing requests:
 
-- `l5d-dst-path`: the logical name of the request as identified by linkerd
-- `l5d-dst-bound`: the concrete client name after delegation
+- `l5d-dst-logical`: the logical name of the request as identified by linkerd
+- `l5d-dst-concrete`: the concrete client name after delegation
 - `l5d-dst-residual`: an optional residual path remaining after delegation
 - `l5d-reqid`: a token that may be used to correlate requests in a
                callgraph across services and linkerd instances
 
+Applications are not required to forward these headers on downstream
+requests.
+
 ### Informational Response Headers
 
-And in addition to the context headers, lay may emit the following
-headers on outgoing responses:
+linkerd may emit the following _informational_ headers on outgoing
+responses:
 
 - `l5d-err`: indicates a linkerd-generated error. Error responses
              that do not have this header are application errors.
+
+Applications are not required to forward these headers on upstream
+responses.
--- a/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
+++ b/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
@@ -30,8 +30,8 @@ import java.util.Base64
  * In addition to the context headers, linkerd may emit the following
  * headers on outgoing requests:
  *
- *   - `l5d-dst-path`: the logical name of the request as identified by linkerd
- *   - `l5d-dst-bound`: the concrete client name after delegation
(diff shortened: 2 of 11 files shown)


linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,46 +1,232 @@
...
-        traceId <- get(header).toOption
-      } yield traceId
-
-    def set(headers: HeaderMap, id: TraceId): Unit = {
-      val bytes = TraceId.serialize(id)
-      val b64 = Base64.getEncoder().encodeToString(bytes)
-      val _ = headers.set(Key, b64)
+    val serverModule: Stackable[ServiceFactory[Request, Response]] =
+      new Stack.Module0[ServiceFactory[Request, Response]] {
+        val role = Stack.Role("DeadlineExtract")
```
adleong: role and description should indicate that other context is extracted as well (eg dtab)

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,46 +1,232 @@
...
+      def get(headers: HeaderMap): Option[FDeadline] =
+        headers.getAll(Key).foldLeft[Option[FDeadline]](None) { (d0, v) =>
+          (d0, read(v).toOption) match {
+            case (Some(d0), Some(d1)) => Some(FDeadline.combined(d0, d1))
+            case (d0, d1) => d0.orElse(d1)
+          }
+        }
+
+      def write(d: FDeadline): String =
+        s"${d.timestamp.inNanoseconds} ${d.deadline.inNanoseconds}""
```
adleong: I think there's a quoting typo here

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,46 +1,232 @@
...
+       * Extract the deadline from the request and, if it exists, use
+       * either that deadline or the one already on the local context,
+       * whichever is stricter.
+       */
+      class ServerFilter extends SimpleFilter[Request, Response] {
+        def apply(req: Request, service: Service[Request, Response]) =
+          get(req.headerMap) match {
+            case None => service(req)
+            case Some(reqDeadline) =>
+              clear(req.headerMap)
```
adleong: just curious: why do we clear the header after reading it?

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,46 +1,232 @@
...
+      def set(dtab: FDtab, msg: Message): Unit =
+        if (dtab.nonEmpty) {
+          val _ = msg.headerMap.set(CtxKey, dtab.show)
+        }
+
+      /**
+       * Extract a Dtab from the L5d-Ctx-Dtab and L5d-Dtab headers (in
+       * that order) and append them to the local context.
+       *
+       * The L5d-Ctx-Dtab header is intended to be set by a linkerd
```
adleong: what's the purpose of differentiating these?  the contents of l5d-dtab will get moved into l5d-ctx-dtab after one hop, right?

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
olix0r: yep

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
olix0r: not sure we _need_ to do this, cleared so that it's not passed-through automatically and the client-side is responsible for encoding any downstream deadlines.

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
olix0r: i want all of the context headers to be l5d-ctx-dtab, and also i'd like that these be assumed to be linkerd-generated (so that, for instance, all all l5d-ctx can be stripped from external traffic.  l5d-dtab is intended to be user-facing.  Not 100% sure the distinction is necessary, but i wanted the user-facing header to be terse.

adleong: Personally, I'd favor simplicity over brevity.

olix0r: Yeah i understand.  I think this is simple because we can clarify the context of this header more clearly:

l5d-ctx headers are ONLY to be generated by linkerd. They should always be stripped at the edge.  l5d-dtab and l5d-sample are ONLY to be set by users, and only admitted from trusted sources.

In the same way that l5d-sample influences l5d-ctx-trace, l5d-dtab influences l5d-ctx-dtab.  l5d-ctx headers should be totally opaque to users and services.

linkerd/docs/protocol-http.md
```diff
@@ -85,3 +85,58 @@ routers:
...
+### Informational Request Headers
+
+In addition to the context headers, linkerd may emit the following
+headers on outgoing requests:
+
+- `l5d-dst-path`: the logical name of the request as identified by linkerd
+- `l5d-dst-bound`: the concrete client name after delegation
+- `l5d-dst-residual`: an optional residual path remaining after delegation
+- `l5d-reqid`: a token that may be used to correlate requests in a
+               callgraph across services and linkerd instances
```
gtcampbell: TIOLI: should we provide guidance here on whether these headers should be stripped at the edge? (i.e. is there any risk they leak sensitive information?)

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,121 +1,427 @@
...
+     */
+    object Deadline {
+      val Key = Prefix + "deadline"
+
+      def read(v: String): FDeadline = {
+        val values = v.split(' ')
+        val timestamp = Time.fromNanoseconds(values(0).toLong)
+        val deadline = Time.fromNanoseconds(values(1).toLong)
+        FDeadline(timestamp, deadline)
+      }
```
gtcampbell: tioli: should this return an Option[FDeadline] to match the signature of `unmarshalDeadline` from `finagle-http`?

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,121 +1,427 @@
...
     }
 
-    def clear(headers: HeaderMap): Unit = {
-      val _ = headers.remove(Key)
+    /**
+     * There are two headers used to control local Dtabs in linkerd:
+     *
+     *   1. `l5d-ctx-dtab` is read and _written_ by linkerd. It is
+     *      intended to managed entirely by linked, and applications
```
gtcampbell: typo: "linked" -> "linkerd"

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,121 +1,427 @@
...
+     */
+    object Dtab {
+      val CtxKey = Ctx.Prefix + "dtab"
+      val UserKey = Headers.Prefix + "dtab"
+
+      private val EmptyReturn = Return(FDtab.empty)
+
+      def get(headers: HeaderMap, key: String): Try[FDtab] =
+        if (!headers.contains(key)) EmptyReturn
+        else Try { FDtab(headers.getAll(key).view.flatMap(FDtab.read(_)).toIndexedSeq) }
```
gtcampbell: I assume the `view` call here is to save on allocating the intermediate collection? If so, IIRC using `breakOut` would be even cheaper, but I haven't done the microbenchmark recently:
```scala
else Try { FDtab(headers.getAll(key).flatMap(FDtab.read)(breakOut)) }
```

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/HttpTraceInitializer.scala
```diff
@@ -6,88 +6,87 @@ import com.twitter.finagle.http._
 import com.twitter.finagle.tracing._
 
 /**
- * Typically, finagle clients initialize trace ids to capture a
- * request-response flow.  This doesn't really fit with what we want
- * to capture in the router.  Specifically, we want to capture the
- * 'request' and 'response' portions of a trace individually--ingress
- * to egress in each direction.
+ * Uses [[Headers.Ctx.Trace]] and [[Headers.Sample]] headers
+ * propagate/control tracing.
```
gtcampbell: typo: I assume this should read "to propagate/control"

gtcampbell:
:star: my comments are largely cosmetic. I appreciate the attention to detail in the documentation!

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
olix0r: I generally prefer to use wrapper types only where they are necessary.  I.e. calling read should fail with an exception if it can't be read.  In get, we call Try(read(v)).toOption to wrap it appropriately.  Much in the way that Path.read and Dtab.read throw exceptions on illegal input.

gtcampbell:
:star: thanks for updates and answers!

linkerd/docs/protocol-http.md
```diff
@@ -85,3 +85,73 @@ routers:
...
+- `l5d-dst-concrete`: the concrete client name after delegation
+- `l5d-dst-residual`: an optional residual path remaining after delegation
+- `l5d-reqid`: a token that may be used to correlate requests in a
+               callgraph across services and linkerd instances
+
+Applications are not required to forward these headers on downstream
+requests.
+
+The value of the _dst_ headers may include service discovery
+information including host names.  Operators may opt to remove these
```
adleong: This would require something in between linkerd and the outside world (like nginx).  If we want to position linkerd as being able to be used on the edge, maybe we need to configurably not emit these

adleong:
:star: 💇

linkerd/docs/protocol-http.md
```diff
@@ -85,3 +85,73 @@ routers:
...
+_Context headers_ (`l5d-ctx-*`) are generated and read by linkerd
+instances. Applications should forward all context headers in order
+for all linkerd features to work. These headers include:
+
+- `dtab-local`: Currently (until the next release of finagle), the
+  `dtab-local` header is used to propagate dtab context. In an
+  upcoming release this header will no longer be honored, in favor of
+  `l5d-ctx-dtab` and `l5d-dtab`.
+- `l5d-ctx-deadline`:
+- `l5d-ctx-trace`:
```
klingerf: Do you want to include an explanation after the colons here to explain what the deadline and trace headers represent?

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/Headers.scala
```diff
@@ -1,121 +1,428 @@
...
+     * A serverside stack module that extracts contextual information
+     * from requests and configures the local
+     * `com.twitter.finagle.Context` appropriately. Currently this includes:
+     *   - Deadline
+     *   - Dtab
+     *
+     * Note that currently, the dtabs read by this module are
+     * appeneded to that specified by the `l5d-dtab` header.  The
+     * `dtab-local` header should be considered deprecated in favor of
+     * `l5d-dtab, and will not be supported in the future.
```
klingerf: Nit: missing backtick here after l5d-dtab.

klingerf:
⭐  Looks great.

linkerd/docs/protocol-http.md
olix0r: Yeah, agreed.

klingerf:
⭐  Thanks for updating.

PR #471, opened 2016-06-10
Dashboard: Discard the first few data points from metrics

### LINKED ISSUE #379: dashboard occasionally shows a huge first datapoint
While traffic is going through linkerd, the first time you load the dashboard, the first data point on the request volume graph is sometimes really high. I think this a function of how the aggregation code works for metrics.json. One short-term workaround might be to ignore the first value.

rmars:
... as the request count we get back from the metrics endpoint seems to be very high for the first few data points, causing a spike in the graph

Fixes #379 (hackily)

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
@@ -42,10 +42,19 @@ var CombinedClientGraph = (function() {
     var desiredMetrics = _.map(Query.filter(query, metricsCollector.getCurrentMetrics()), clientToMetric);
     chart.setMetrics(desiredMetrics, timeseriesParams, true);
 
-    metricsCollector.registerListener(function(data) {
-      var filteredData = Query.filter(query, data.specific);
-      chart.updateMetrics(filteredData);
-    }, function(metrics) { return Query.filter(query, metrics); });
+    var count = 0;
+    var metricsListener = function(data) {
+      if (count < 5) {
+        // Hacky bug fix: discard the first few data points to fix the issue
+        // where the first values from /metrics are very large [linkerd#485]
+        count++;
+      } else {
+        var filteredData = Query.filter(query, data.specific);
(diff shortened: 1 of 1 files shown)


esbie:
⭐  looks good code-wise. I'd be interested to understand why this happens

olix0r:
My understanding is that we should only have to drop _the first_ data point (since it is absolute and not a delta to the prior). Is this not the case?

rmars:
When I tested it out, it looked like more than the first data point had an unusually large value - so I left out the first couple to be sure. Let me check again. It does mean a more seconds of waiting for the graph after loading the dashboard page.

rmars:
Yep, checked again... the first 4 ish values are way too high.

olix0r:
:star: very odd... i'd like to understand why that happens but this seems like a good workaround

admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
```diff
@@ -45,10 +45,19 @@ var CombinedClientGraph = (function() {
...
     chart.setMetrics(desiredMetrics, timeseriesParams, true);
 
-    metricsCollector.registerListener(function(data) {
-      var filteredData = Query.filter(query, data.specific);
-      chart.updateMetrics(filteredData);
-    }, function(metrics) { return Query.filter(query, metrics); });
+    var count = 0;
+    var metricsListener = function(data) {
+      if (count < 5) {
+        // Hacky bug fix: discard the first few data points to fix the issue
```
adleong: Might be worth opening an issue for investigating why this happens.  Even if we never do it, at least there's a record documenting why we drop data points.

adleong:
:star: ▶️

admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
rmars: Good point! Created #485, I'll add it here.

PR #472, opened 2016-06-11
Add the ability to go to a specific routers dashboard

rmars:
... via visiting /routers/<routername> in the url.
Removes use of getSelectedRouter() from the dashboard code which didn't work anyway.

It definitely still needs UI improvements. The blue summary banner at the top still displays metrics for all the routers, so maybe we should remove it or clarify that that is across all routers.

(Not even sure if we even still want single router pages, but here's a stab at it.)

screen shot 2016-06-10 at 5 27 06 pm

### DIFF
--- a/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
+++ b/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
@@ -6,20 +6,25 @@ import com.twitter.util.Future
 import io.buoyant.linkerd.Build
 
 private[admin] class DashboardHandler extends Service[Request, Response] {
-  lazy val html = dashboardHtml
+  import DashboardHandler._
 
   override def apply(req: Request): Future[Response] = req.path match {
-    case "/" =>
-      AdminHandler.mkResponse(html)
+    case "/" | "/routers" =>
+      AdminHandler.mkResponse(dashboardHtml())
+    case rexp(router) =>
+      AdminHandler.mkResponse(dashboardHtml(router))
     case _ =>
       Future.value(Response(Status.NotFound))
   }
 
-  def dashboardHtml = {
+  def dashboardHtml(router: String = "") = {
     AdminHandler.html(
       content = s"""
         <div class="request-totals"></div>
-        <div class="server-data" data-linkerd-version="${Build.load().version}" style="visibility:hidden"></div>
+        <div class="server-data"
+          data-linkerd-version="${Build.load().version}"
+          data-router-name="${router}"
+          style="visibility:hidden"></div>
         <div class="dashboard-container"></div>
         <div class="row proc-info">
         </div>
@@ -44,4 +49,8 @@ private[admin] class DashboardHandler extends Service[Request, Response] {
       navbar = AdminHandler.version2NavBar
     )
   }
+}
+
+object DashboardHandler {
+  val rexp = "^/routers/(.*)/?".r
 }
--- a/admin/src/main/resources/io/buoyant/admin/js/dashboard.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/dashboard.js
@@ -1,5 +1,5 @@
 "use strict";
-/* globals getSelectedRouter, MetricsCollector, ProcInfo, RequestTotals, Routers, RouterController */
+/* globals MetricsCollector, ProcInfo, RequestTotals, Routers, RouterController */
 
 /**
  * Number of millis to wait between data updates.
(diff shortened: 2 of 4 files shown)


linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
```diff
@@ -6,19 +6,22 @@ import com.twitter.util.Future
...
     case _ =>
       Future.value(Response(Status.NotFound))
   }
 
-  def dashboardHtml = {
+  def dashboardHtml(router: String = "") = {
     AdminHandler.html(
       content = s"""
         <div class="request-totals"></div>
+        <div class="router-name-data" data-router-name="${router}" style="visibility:hidden"></div>
```
adleong: just curious: what's the difference between doing this vs `<script id="data-router-name" type="application/json">$router</script>`?

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
```diff
@@ -45,3 +48,7 @@ private[admin] class DashboardHandler extends Service[Request, Response] {
     )
   }
 }
+
+object DashboardHandler {
+  val rexp = "/routers/(.*)/?".r
```
adleong: do we need ^ and $ anchors so that this doesn't match longer urls that include this?

rmars: Ah yes! Thanks :)

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
rmars: Hm, good question! I am not really sure what the most idiomatic way to pass these variables along is, between building up an object, and passing that, or using html5 data attributes. Since we're only passing 2 variables, it probably doesn't make too much of a difference. If someone has a strong opinion they can chime in :)

We could make global json object for our constants, something like `<script>var DASHBOARD_CONSTANTS = { router: "", linkerdVersion: ""}</script>`, and since our application's js is loaded after our `$content` it'd work... but if we ever moved those scripts, then the variables would be defined before they're used. If we did `<script id="data-router-name" type="application/json">$router</script>` we'd need to find the div to get the router name anyway, similar to what we're doing now.

Passing the data as part of the div does mean we're responsible for naming/obtaining the variables in the dashboard js code instead of here, which I kind of like, though it does mean we use an extra jquery selector to find the div.

We could make it its own js file, which we'd have to generate in the scala and then pass in in our list of `javaScripts` - but since we're only passing 2 variables it seems much. 

This did make me think, though, I don't need a separate div for the new data, so I'll just put all the data under one div, so I can reuse the jquery selector on the frontend.

rmars:
I've added a little indicator to the request totals when we're in a specific router's page - hopefully it will make it clearer that the request totals are still for all routers.  It looks weird but I think maybe less confusing. lmk what you think:

screen shot 2016-06-14 at 11 44 17 am

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
adleong: Cool, thanks for the explanation.  I don't really have an opinion but I agree with you that it doesn't seem like a big deal either way, it's just always nice to be consistent.

adleong:
:star: 🌌

esbie:
my preference would be to hide the blue summary banner when on an individual router page

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/DashboardHandler.scala
```diff
@@ -6,20 +6,27 @@ import com.twitter.util.Future
...
 private[admin] class DashboardHandler extends Service[Request, Response] {
-  lazy val html = dashboardHtml
+  import DashboardHandler._
 
   override def apply(req: Request): Future[Response] = req.path match {
     case "/" =>
-      AdminHandler.mkResponse(html)
+      AdminHandler.mkResponse(dashboardHtml())
+    case "/routers" =>
+      AdminHandler.mkResponse(dashboardHtml())
```
esbie: these cases can be combined: `case "/" | "/routers"`

rmars:
Updated summary banner: (hide on router page)
screen shot 2016-06-14 at 2 24 42 pm

esbie:
👍 looks good, thanks!

PR #477, opened 2016-06-14
Include plugins when running linkerd from sbt

adleong:
Include any plugins in the plugins directory when running linkerd or namerd from sbt.


gtcampbell:
:ship:

olix0r:
⭐

PR #478, opened 2016-06-14
Install ClassifiedTracing filter after TraceInitializerFilter

klingerf:
## Problem

The ClassifiedTracing filter was being applied before the TraceInitializerFilter in Router's client stack, such that response classification annotations were being set on the parent server request span, rather than on the client span that was being classified.

## Solution

Move the ClassifiedTracing filter to immediately after the TraceInitializerFilter.  This creates a 1-to-1 mapping of client request to classification annotation.


olix0r:
⭐  lgtm. thanks for adding tests!

esbie:
⭐ 🎯

PR #483, opened 2016-06-15
[RFC] Add metrics exporter interface

adleong:
Proof of concept.

We define a new plugin type `MetricsExporter` and add a config section for a list of metrics exporters.  At a configurable interval, a dump of the current metrics is given to each exporter.


linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -74,7 +77,8 @@ object Linker {
     namers: Option[Seq[NamerConfig]],
     routers: Seq[RouterConfig],
     tracers: Option[Seq[TracerConfig]],
-    admin: Option[AdminConfig]
+    admin: Option[AdminConfig],
+    metrics: Option[Seq[MetricsExporterConfig]]
```
olix0r: bikeshed:  It seems like a _metrics_ section might have some other things (like, configuration that controls how metrics are _recorded_ in addition to exported.  Does it make sense to structure this as:

```yaml
metrics:
  exporters:
  - ... 
```

linkerd/core/src/main/scala/io/buoyant/linkerd/MetricsExporterInitializer.scala
```diff
@@ -0,0 +1,22 @@
...
+trait MetricsExporterConfig {
+
+  @JsonProperty("periodSecs")
+  var _periodSecs: Option[Int] = None
+
+  @JsonIgnore
+  def period = _periodSecs.map(_.seconds).getOrElse(1.minute)
+
+  @JsonIgnore
+  def export(metrics: mutable.Map[String, Number]): Future[Unit]
```
olix0r: I'd be inclined to separate configuration types from the type that actually does the work. I'd prefer to see something like `def mk(timer: Timer): MetricsExporter` here.

linkerd/core/src/main/scala/io/buoyant/linkerd/MetricsExport.scala
```diff
@@ -0,0 +1,24 @@
...
+  exporters: Seq[MetricsExporterConfig],
+  metrics: MetricProvider = MetricsStatsReceiver.defaultRegistry,
+  timer: Timer = DefaultTimer.twitter
+) extends Closable {
+
+  private[this] val tasks = exporters.map { exporter =>
+    timer.schedule(exporter.period) {
+      val _ = exporter.export(metrics.sample().asScala)
+    }
+  }
```
olix0r: I'd be inclined to push this logic into each exporter.  There's nothing that says stats export has to be on a fixed interval that's known up front.  Let the exporter make that decision.  MetricsExporter could simply expose Closable.  Then, all exporters can be joined with `Closable.all(exporters:_*)`

olix0r: That said, it definitely seems like a good idea to provide PeriodicMetricsExporter and PeriodicMetricsExporterConfig types to make it easy to write these.

adleong:
Closing in favor of https://github.com/BuoyantIO/linkerd/pull/580

PR #484, opened 2016-06-15
linkerd-protocol-http: Handle client exceptions in the path stack

olix0r:
Problem

When an exception is raised in the linkerd client, it is thrown into the
server.  This means that client connection errors are measuerd as server
connection errors in stats, etc.

Solution

Handle exceptions at the top of the router's path stack so that the server
isn't exposed to client-generated exceptions.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -20,6 +20,7 @@ class HttpInitializer extends ProtocolInitializer.Simple {
     val pathStack = Http.router.pathStack
       .prepend(Headers.Dst.PathFilter.module)
       .replace(StackClient.Role.prepFactory, DelayedRelease.module)
+      .prepend(http.ErrorResponder.module)
     val boundStack = Http.router.boundStack
       .prepend(Headers.Dst.BoundFilter.module)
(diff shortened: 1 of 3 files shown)


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -29,8 +33,9 @@ class HttpInitializer extends ProtocolInitializer.Simple {
   }
 
   protected val defaultServer = {
-    val stk = http.ErrorResponder +: Http.server.stack
+    val stk = Http.server.stack
       .replace(HttpTraceInitializer.role, HttpTraceInitializer.server)
+      .prepend(http.ErrorResponder)
```
klingerf: Is this still required on the default server stack, now that it's on the default router stack?  Probably doesn't hurt to include it, but just curious.

klingerf:
⭐  Makes sense to me.

adleong:
:star: 🍳 TIOLI: is possible to add a test where an exception is thrown in the client and verify the right stats get incremented?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
olix0r: Yep. For instance, when an identifier cannot name a request, errors are thrown on the server stack.

PR #487, opened 2016-06-17
Use guava cache for activity caching in HttpControlService

adleong:
Use guava instead of a homemade cache for activities.  Note that the request handlers will take care of observing the activities and closing the observations when they're done, so the cache doesn't need to worry about this.  Therefore, there's very little cost to keeping activities around in cache (if they're not observed, they won't receive updates).

(see the streaming section of https://github.com/BuoyantIO/linkerd/issues/310)


gtcampbell:
:star: I like this approach!

olix0r:
lgtm! we should look at moving our guava cache stuff to [caffeine](https://github.com/ben-manes/caffeine), since this is where the rest of the finagle codebase is going. not a blocker

PR #489, opened 2016-06-17
Delete the legacy dashboard

### LINKED ISSUE #198: Remove remnants of old dashboard
* Delete old dashboard code
* Remove things that say // TODO: remove when we kill the old dashboard 
* Move rendering of process info to process_info.template (esbie's comment in #192)

Basically get rid of todos in `git grep TODO -- '*.js'`
Remove `summary.css`
To be done after we verify the new dashboard.
Part of #183

### LINKED ISSUE #377: dashboard "metrics" link should just point to metrics.json
Right now clicking on "metrics" points to the old metrics browser. I'd prefer we point it to /admin/metrics.json directly. The browser is not really in a usable state (e.g. almost every metric name is too long), isn't themed to match the dashboard, and I'd rather show off the programmatic endpoint than push people into that experience.

(Longer term we can consider whether we want to update the browser or discard it.)

rmars:
Also remove the "metrics" link in the nav.
I have not removed the metrics browser code, I know we were also considering removing that.

Fixes #198
Fixes #377

screen shot 2016-06-17 at 4 29 17 pm


adleong:
:star: 🔥 yeaaaahhhhhh!!

esbie:
⭐ 🚒  sweet

PR #490, opened 2016-06-18
Turn off tracing for namer calls

adleong:


siggy:
👍 ⭐ thanks!

klingerf:
Rather that calling `Trace.letClear` everywhere, do you think it would be possible to turn off tracing in the client stack when the namer is initialized?  I haven't tested it, but if we don't want tracing in any of our namers, you might be able to just change this line to use a NullTracer instead:

https://github.com/BuoyantIO/linkerd/blob/master/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala#L96

olix0r:
:star: This is okay for the time being if it fixes a problem, but I think we really do want to include namer tracing when it blocks the request path.

[Approved with PullApprove](https://pullapprove.com/BuoyantIO/linkerd/pull-request/490/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=BuoyantIO/linkerd)

klingerf:
⭐  ♨️  Thanks for updating.

siggy:
👍 ⭐ confirmed `NullTracer` + `Trace.letClear` works!

PR #493, opened 2016-06-20
Make config initializers abstract classes for better java compat

adleong:
To implement a plugin, you need to extend the appropriate initialize class.  Extending traits in java is difficult so we make these abstract classes instead to make plugin development in java easier.


olix0r:
Won't there be a whole bunch of other problems writing plugins in Java?

adleong:
nope: https://github.com/BuoyantIO/linkerd-examples/compare/alex/plugin?expand=1

gtcampbell:
LGTM

olix0r:
lgtm

PR #494, opened 2016-06-21
Configure server-side timeouts

olix0r:
Router timeouts are only applied in the client stack, which is below retries, etc. This means that retries could persist forever, since they would never exceed a global timeout.

To fix this, the server is now configured with the router's timeout value, to impose a global timeout requirement.

In the future, it may make sense to admit client and server-specific timeouts.


adleong:
:star: ⏰

gtcampbell:
🚢

PR #496, opened 2016-06-21
Push ClassifiedTracing filter further down the client stack

klingerf:
In #478 I moved the ClassifiedTracing filter below the TraceInitializerFilter in the client stack, so that classification annotations would happen after we set the span id for client request spans.  This put the ClassifiedTracing filter between the TraceInitializerFilter and the ClientTracingFilter, but we actually need to install it below the ClientTracingFilter, as advised by this comment in the Finagle codebase:

https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/client/StackClient.scala#L334

The ClientTracingFilter sets the ClientRecv annotation, at which point finagle-zipkin considers the trace to be complete and flushes the span, so no other annotating filters should be above it in the stack.  In this branch I'm moving ClassifiedTracing to immediately before the ProtoTracing filter, which seems like the right place to put it.

For my own reference, the client stack now looks something like: ... => TraceInitializerFilter => ClientTracingFilter => ProcessFailure => ClassifiedTracing => ProtoTracing => ...

### DIFF
--- a/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
+++ b/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
@@ -64,6 +64,18 @@ class HttpEndToEndTest extends FunSuite with Awaits {
         |  - port: 0
         |""".stripMargin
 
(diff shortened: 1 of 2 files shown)


olix0r:
lgtm! thanks, @klingerf

PR #497, opened 2016-06-21
Start to unify the different pages of the linkerd admin site

rmars:
* consolidate versions of the navbar that we use
* add router name toggle to linkerd dashboard page
* handle router name as params in the dashboard page, like we do in the delegator

screen shot 2016-06-21 at 3 12 50 pm
screen shot 2016-06-21 at 3 26 30 pm


linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
```diff
@@ -20,7 +20,7 @@ object AdminHandler extends HtmlView {
           </button>
 
           <a class="navbar-brand-img" href="/">
-            <img alt="Logo" src="/files/images/logo.svg">
```
siggy: does this mean we can delete logo.svg?

rmars: Ooh, yes we can!

siggy:
⭐ 👍 great cleanup!

i noticed the navbars are still styled slightly differently between the main dashboard and the dtab page. is that expected? followup branch?

rmars:
Yes indeed, I intend to keep unifying!
Fonts, backgrounds still to come. Wanted to keep it incremental.

adleong:
:star: looks great!  Are you planning on doing the "help" page as well in a followup?  

There's also the issue of the logging page which is straight up, twitter server.  Not sure if we want to bother restyling that or what it would take.

rmars:
Thanks!! Good questions.
* The Help section is a link to linkerd.io - so I think if we wanted to really unify it we would make a version of the docs available offline. IIRC we did want to do this at some point?
* You're correct about the logging page! So I looked into restyling it and got this far:
screen shot 2016-06-22 at 2 53 59 pm

I do this by writing a filter before LoggingHandler that injects our own css and a bit of js into the logging page html, such that I can now control the theming. It is maybe a sketchy solution, what do you think?

adleong:
Interesting.  What do you think about making an offline copy of just the help page and styling it to match?    The links to the slack and github repo wouldn't work if you're offline, but I think that's probably okay...

adleong:
I Am Not A Designer, but I think that logging page is an improvement!

rmars:
I think that would be fine. If we want the complete site's docs I think we'll need to figure out overall navigation, updating content, etc which is probably not trivial, but I will look into it.

esbie:
⭐ looking good

PR #498, opened 2016-06-22
Support client- and server-timeout configuration

olix0r:
Add `timeoutMs` configurations to the client and server sections of router
configs.

The server timeout configures an explicit global timeout for all requests
received on the given server. If not specified, the base router timeout may
be used.

The client timeout configures individual request timeouts. This timeout is
applied between retries, whereas the server timeout applies across retries. If
the base router timeout is specified, the lesser timeout value is applied.

This change introduces three new stack parameters: Router.RouterTimeout,
Router.ServerTimeout, nad Router.ClientTimeout. These timeouts are converted to
a TimeoutFilter.Param by stack modules, Router.ServerTimeout.module and
Router.ClientTimeout.module.

Furthermore, this changes the manner in which servers are configured by
linkerd.  Instead of updating server params as router configurations are
updated, we simply compute server configuration during router initialization.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,10 @@
 * linkerd routers' `timeoutMs` configuration now applies on the
   server-side, so that the timeout acts as a global timeout rather
   than an individual request timeout.
+  * Added server `timeoutMs` configuration, which controls global
+    timeout.
+  * Added client `timeoutMs` configuration, which controls individual
+    request timeout.
 
 ## 0.6.0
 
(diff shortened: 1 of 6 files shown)


adleong:
Are client timeouts particularly useful?  For any of the built in request classifiers, timeouts are not retryable, so having a timeout underneath retries doesn't really accomplish anything.  Of course, it's a different story if we can classify requests as idempotent, but none of the built in classifiers do that.  And if we DID do that then I think BackupRequestFilter is probably a better way to cut down on tail latency for idempotent requests than retries are.

I also think that server or client timeouts should override router timeouts instead of taking the minimum, in case you need to lengthen the timeout on a particular server for some reason.

PR #499, opened 2016-06-22
Make the delegator page look more like the dashboard page

rmars:
`parent: rmars/style-unification`

* delete the css files under /styleguide, as we no longer care to keep the admin dashboard in sync with linkerd.io
* the linkerd delegator page now is consistent with the dashboard (and now looks like the namerd one)

screen shot 2016-06-22 at 1 06 40 pm

Follow up to #497


adleong:
:star: 👍

admin/src/main/scala/io/buoyant/admin/names/DelegateHandler.scala
```diff
@@ -47,7 +47,7 @@ class DelegateHandler(
         <script id="dtab-base-data" type="application/json">${DelegateApiHandler.Codec.writeStr(dtabBase)}</script>
       """,
       javaScripts = Seq("dtab_viewer.js", "delegator.js", "delegate.js"),
-      csses = Seq("styleguide/styleguide.css", "styleguide/local.css", "admin.css", "delegator.css", "dashboard-shared.css")
+      csses = Seq("admin.css", "delegator.css", "dashboard-shared.css", "dashboard.css")
```
esbie: if dashboard-shared is going to be shared across all pages, should we add it to AdminHandler instead?

rmars:
(Addressing esbie's comment in #505)

PR #501, opened 2016-06-22
Make the logging page look more like the dashboard page

rmars:
`parent: rmars/style-followup`

Follow up to #499 and #497

I've added our own js/css by creating a filter that adds things to the response body. Let me know your opinions on this.

Also colors/styles can be tweaked, opinions welcome.

screen shot 2016-06-22 at 4 33 31 pm

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/css/logger-overrides.css
+++ b/admin/src/main/resources/io/buoyant/admin/css/logger-overrides.css
@@ -1,0 +1,53 @@
+/*
+* Overrides the twitter server logging page styles
+*/
+
+#navbar {
+  float: left;
+  padding-left: 0px;
+  padding-right: 0px;
+}
+
+#navbar ul.nav>li {
+  padding-left: 0px;
+  padding-right: 0px;
+}
+
+#navbar ul.nav>li>a {
+  font-size: 18px;
+  font-weight: 300;
+  line-height: 1.1;
+}
+
+#navbar ul.nav {
+  margin-top: -12px;
+}
+
+#contents {
+  margin-left: 0px;
+  padding-top: 0px;
+}
+
+.table-striped>tbody>tr:nth-child(odd)>td, .table-striped>tbody>tr:nth-child(odd)>th {
+  background-color: #161963;
+}
+
+.table-striped>tbody>tr>td {
+  border-top: 2px solid gray;
+}
+
+.table-striped>thead>tr {
+  border-bottom: 3px solid gray;
+}
+
+.table>thead>tr>th {
+  border-bottom: none;
+}
+
+.table h5 {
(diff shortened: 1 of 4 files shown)


admin/src/main/scala/io/buoyant/admin/StyleOverrideFilter.scala
```diff
@@ -0,0 +1,24 @@
...
+  val stylesAndJavascripts =
+    s"""
+    <link type="text/css" href="/files/css/logger-overrides.css" rel="stylesheet"/>
+    <link type="text/css" href="/files/css/dashboard.css" rel="stylesheet"/>
+    <link type="text/css" href="/files/css/dashboard-shared.css" rel="stylesheet"/>
+    <script src="/files/js/logger-overrides.js"></script>
+      """
+
+  def apply(req: Request, svc: Service[Request, Response]) = {
+    svc(req).map(response => {
```
adleong: style nit:

```
svc(req).map { response =>
  ...
}

**Comment by adleong** [comment]
:star: 🎄

File: admin/src/main/resources/io/buoyant/admin/css/logger-overrides.css [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+.table-striped>tbody>tr:nth-child(odd)>td, .table-striped>tbody>tr:nth-child(odd)>th {
+  background-color: #161963;
+}
+
+.table-striped>tbody>tr>td {
+  border-top: 2px solid gray;
+}
+
+.table-striped>thead>tr {
+  border-bottom: 3px solid gray;
```
esbie: my preference for these borders is `1px solid dimgray` or `1px solid slategray`

admin/src/main/resources/io/buoyant/admin/js/logger-overrides.js
```diff
@@ -0,0 +1,26 @@
...
+  '        </div>' +
+  '        <div id="navbar" class="collapse navbar-collapse">' +
+  '          <ul class="nav navbar-nav">' +
+  '            <li><a href="/delegator">dtab</a></li>' +
+  '            <li><a href="/admin/logging">logging</a></li>' +
+  '            <li><a href="https://linkerd.io/help/">help</a></li>' +
+  '          </ul>' +
+  '        </div>' +
+  '      </div>' +
+  '    </nav>"';
```
esbie: Can we send this html down as a string inside a script tag? Like:
```
<script id="linkerd-navbar" type="application/json">{"navbarToInject": "<nav class=..."}</script>
```

This way ideally we can reuse the nav bar from other pages without duplication.

rmars:
(addressing esbie's comments in #505)

PR #504, opened 2016-06-24
Track status codes on each linkerd http server

### LINKED ISSUE #503: Track response code to client in /srv/
Currently we only track response codes from downstream in /dst/

This is in response to a client issue where they are seeing high 504s but we don't find evidence of that in the metrics.json they've provided us.

olix0r:
We track status codes on clients, but we do not track server side statuses.

Fixes #503.


adleong:
:star: 🍃

PR #507, opened 2016-06-27
Standardize the admin help page

rmars:
* Add a static help page, rather than link to linkerd.io
* Unify dashboard.css and dashboard-shared.css, since we were using styles from both anyway
* Address some comments from previous dashboard branch reviews

I just copied the linkerd help page, but the content of this is easy to change. Suggestions welcome.

Going to move this to a template now or very soon.

screen shot 2016-06-27 at 4 35 02 pm

Follow up to #501, #499, #497,  etc.


admin/src/main/resources/io/buoyant/admin/js/help-page.js
```diff
@@ -0,0 +1,15 @@
+$(function() {
+  var helpText = '<h1 class="title">Need some help?</h1>' +
```
adleong: The help page on linkerd.io was updated recently, I think this is missing those changes.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/HelpPageHandler.scala
```diff
@@ -0,0 +1,19 @@
...
+import com.twitter.util.Future
+
+private[admin] class HelpPageHandler extends Service[Request, Response] {
+
+  override def apply(req: Request): Future[Response] = AdminHandler.mkResponse(html)
+
+  def html = {
+    AdminHandler.html(
+      content = s"""
+        <div class="container main help-page"></div>
```
adleong: Just curious why we populate this with javascript instead of hardcoding it here

rmars: If we use js / load this from a file, then we can make changes to the html and not have to rebuild to pick up changes!

admin/src/main/resources/io/buoyant/admin/js/help-page.js
rmars: Good spot, updated!

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -36,6 +36,7 @@ class LinkerdAdmin(app: App, linker: Linker, config: LinkerConfig) extends Admin
     "/delegator" -> new DelegateHandler(AdminHandler, dtabs, interpreterForRouter),
     "/delegator.json" -> new DelegateApiHandler(interpreterForRouter),
     "/metrics" -> MetricsHandler,
+    "/help" -> new HelpPageHandler,
```
adleong: What about 
```
"/help.html" -> (StaticFilter andThen ResourceHandler.fromDirectoryOrJar(...)
```

rmars: hmmmm that's a good idea, though I wanted to use the AdminHandler to avoid copying the navbar html, as well as the stylesheets to also be inserted into the html for this page...

adleong: good point

adleong:
:star: 🆘

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/HelpPageHandler.scala
esbie: I had the same question as @adleong. Given that this page is 100% static, no template variable needed, I think it should be hardcoded here. (I know it's a pain to develop that way, but the help page won't need to change too often)

rmars: It is true - it won't change too often besides this first time. I can hardcode it!

esbie:
⭐  Thank you!

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/HelpPageHandler.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{Request, Response, Status}
+import com.twitter.util.Future
+
+private[admin] class HelpPageHandler extends Service[Request, Response] {
+
+  override def apply(req: Request): Future[Response] = AdminHandler.mkResponse(html)
+
+  def html = {
```
adleong: tiny non-blocking nit:

could use val instead of def
could omit braces for single expression

adleong:
:star: 📞

PR #508, opened 2016-06-28
Remove leading spaces from code tag code (minor)

rmars:
I forgot when transferring the html that the code tag leaves the leading spaces in. Small fix.

screen shot 2016-06-28 at 11 55 21 am
screen shot 2016-06-28 at 11 55 35 am


adleong:
:star: ⬅️

PR #515, opened 2016-06-30
k8s namer should implement EnumeratingNamer

adleong:


k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -55,6 +56,21 @@ class EndpointsNamer(
       Activity.value(NameTree.Neg)
   }
 
+  override def getAllNames: Activity[Set[Path]] = {
```
olix0r: I think this can be a (lazy?) val.  Since vars/activities are populated lazily, it's safe to build statically.  Furthermore, this ensures that multiple consumers share observations, potentially reducing redundant work.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -410,7 +410,7 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
   private[this] val enumeratingNamers = namers.values.collect {
     case namer: EnumeratingNamer => namer
   }.toSeq
-  private[this] val boundNames =
+  private[this] def boundNames =
     Activity.collect(enumeratingNamers.map(_.getAllNames))
       .map(_.toSet.flatten)
```
olix0r: Similarly, is this necessary to be a def?  I'd expect it be to be fine/good for this to be static (potentially a lazy val?).

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
olix0r: oh, i see. thanks

olix0r: Wait -- doesn't this mean that that this activity can't actually behave properly if new namespaces are added?

adleong:
You're right.  I've made the cache a `Var` instead.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -336,4 +354,13 @@ private object EndpointsNamer {
         }
       }
   }
+
+  implicit class ActSet[A](val actSet: Activity[Set[A]]) extends AnyVal {
```
olix0r: can this be private?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -97,17 +113,19 @@ class EndpointsNamer(
...
     def get(name: String): NsCache = synchronized {
-      caches.get(name) match {
+      caches.sample.get(name) match {
         case Some(ns) => ns
         case None =>
           val ns = new NsCache(name)
           val closable = retryToActivity { watch(name, ns) }
           _watches += (name -> closable)
-          caches += (name -> ns)
+          caches() = caches.sample + (name -> ns)
```
olix0r: We may want to note (in a comment) that `caches` must be updated with `synchronized` -- otherwise this could be potentially unsafe.

olix0r:
lgtm

PR #516, opened 2016-06-30
Turn off decompression

### LINKED ISSUE #514: by default, linkerd should not de/compress bodies.
linkerd should be able to operate blissfully unaware of a message's payload.  while linkerd must transfer-encoding (i.e. to support streaming), it does not need decompress or compress bodies.

Finagle exposes two params to control this:
- `Http.param.CompressionLevel(0)`
- `Http.param.Decompression(false)`

However cursory testing has shown that setting these parameters on both the client and server do not work as hoped -- we still see zip/deflate in cpu profiles...

Figure out how to disable all body de-/re-compression.

adleong:
Fixes #514 

By turning off decompression on the client side, linkerd will no longer decompress compressed payload responses, instead passing the uncompressed payload along.  Profiling with Java Flight Recorder without this change, serving gzipped responses showed `JdkZlibEncoder.encode` and `InfCodes.inflate_fast` as Hot Methods.  With this change, those methods were absent.

I chose not to disable compression on the server side.  This means that if linkerd gets a request for compressed data but forwards to a service that serves an uncompressed response, linkerd will compress it.  This seems like a useful feature.

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/Http.scala
+++ b/router/http/src/main/scala/io/buoyant/router/Http.scala
@@ -73,6 +73,7 @@ object Http extends Router[Request, Response] with FinagleServer[Request, Respon
     val defaultParams: Stack.Params =
       StackServer.defaultParams +
         FinagleHttp.param.Streaming(true) +
+        FinagleHttp.param.Decompression(false) +
         ProtocolLibrary("http")
   }
 
--- a/router/http/src/main/scala/io/buoyant/router/Http.scala
+++ b/router/http/src/main/scala/io/buoyant/router/Http.scala
@@ -1,6 +1,5 @@
 package io.buoyant.router
 
-import com.twitter.finagle.Http.param.Decompression
 import com.twitter.finagle.client.StackClient
 import com.twitter.finagle.http.{Request, Response, TlsFilter}
 import com.twitter.finagle.param.ProtocolLibrary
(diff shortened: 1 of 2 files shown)


router/http/src/main/scala/io/buoyant/router/Http.scala
```diff
@@ -37,7 +38,8 @@ object Http extends Router[Request, Response] with FinagleServer[Request, Respon
     val defaultParams: Stack.Params =
       StackRouter.defaultParams +
         FinagleHttp.param.Streaming(true) +
-        ProtocolLibrary("http")
+        ProtocolLibrary("http") +
+        Decompression(false)
```
olix0r: we should just use `FinagleHttp.param.Decompression` like we do for `Streaming`... makes it slightly clearer

olix0r:
Should we disable decompression on the serverside as well so that bodies sent to linkerd are not decompressed ndedlessly?

Have we validated that this doesn't end up double-encoding bodies?  Perhaps an end-to-end test?

adleong:
I've manually validated that there's no double-encoding.

I tried sending a bunch of compressed request bodies but couldn't get any compression or inflation to show up on the profile.  Let me know if you see otherwise.

olix0r:
awesome! as i said, i believe that we want to disable decompression on the serverside as well, as--as I understand it-- should disable decompresing posted bodies.

olix0r:
:star: thanks!

Would you mind updating CHANGES while you're here?

klingerf:
⭐ 🚲  Awesome.

PR #526, opened 2016-07-07
Fix sbt command for `namerd-main`

cab:
The command should use a slash to indicate a task within a project.


adleong:
Good catch!  I think we can remove this command altogether since it doesn't bundle up the namerd plugins and therefore isn't very useful.

olix0r:
:star:

PR #528, opened 2016-07-07
Support unions in delegator UI

### LINKED ISSUE #524: dtab playground doesn't display resolution for conjunction rules
When I have a rule with an "&" operator, the resulting dtab resolution graph appears to just skip over that rule.

screen shot 2016-07-07 at 6 31 23 am

Note the `/a/potato` evaluation, which has no rule displayed, but is listed as a bound path, and no subtree.

I think that we should show a full resolution of both branches for these rules (even with weights). 

(This is with 0.7.0 and namerd providing the top it of the dtab.)

adleong:
Fixes #524 

screen shot 2016-07-07 at 12 44 13 pm

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/delegator.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/delegator.js
@@ -7,14 +7,18 @@ var Delegator = (function() {
     $('.result').html(renderNode(resp));
   }
 
-  function renderNode(obj){
+  function renderNode(obj, weight = 1.0){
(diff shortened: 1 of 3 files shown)


rmars:
:star:
lgtm! I think it's failing because we need to specify in our eslintrc that we'd like es6 syntax: http://eslint.org/docs/user-guide/configuring

rmars:
oh! though we don't use es6 here yet... might just wanna take out that default syntax

olix0r:
lgtm

Issue #520, opened 2016-07-05
Add overrides support for namerd HTTP Control Interface


Ashald:
When using `$NAMERD_URL/api/1/delegate/foo` endpoint there is now way to specify a 'Dtab-override'. Or in Finagle terms, Dtab-Local is not honored.

Talking to @adleong and @olix0r on Slack we figured out that it might be a nice improvement - either through headers or through an extra query parameter.

Ashald:
Addressed in #529

PR #529, opened 2016-07-07
Optional dtab query parameter for selected Namerd HTTP Control API endpoints

Ashald:
For next Namerd HTTP Control API endpoints:
* `/bind/:ns`
* `/delegate/:ns`

added an optional dtab query parameter for that can be used to supply an extra delegation table that will be appended to the delegation table configured for the namespace.

Addresses #520, implemented according to decisions made in Slack during discussion with @adleong and @olix0r

### DIFF
--- a/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
+++ b/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
@@ -160,7 +160,7 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
     case AddrUri(Some(ns), path) =>
       handleGetAddr(ns, path, req)
     case DelegateUri(Some(ns), path) =>
-      handleGetDelegate(ns, path)
+      handleGetDelegate(ns, path, req.getParam("dtab", ""))
     case DelegateUri(None, path) =>
       delegateApiHander(req)
     case _ if req.path == s"$apiPrefix/bound-names" && req.method == Method.Get =>
@@ -392,12 +392,12 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
     }
   }
 
-  private[this] def handleGetDelegate(ns: String, path: Path): Future[Response] = {
+  private[this] def handleGetDelegate(ns: String, path: Path, extraDtab: String): Future[Response] = {
     getDtab(ns).flatMap {
       case Some(dtab) =>
         delegate(ns) match {
           case delegator: Delegator =>
-            DelegateApiHandler.getDelegateRsp(dtab.dtab.show, path.show, delegator)
+            DelegateApiHandler.getDelegateRsp((dtab.dtab ++ Dtab.read(extraDtab)).show, path.show, delegator)
           case _ =>
             val rsp = Response(Status.NotImplemented)
             rsp.contentString = s"Name Interpreter for $ns cannot show delegations"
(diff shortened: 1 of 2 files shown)


namerd/iface/control-http/src/test/scala/io/buoyant/namerd/iface/HttpControlServiceTest.scala
```diff
@@ -285,6 +285,27 @@ class HttpControlServiceTest extends FunSuite with Awaits {
     assert(resp.contentString == bound + "\n")
   }
 
+  test("bind with an extra dtab") {
+    val ni = new NameInterpreter {
+      override def bind(dtab: Dtab, path: Path): Activity[NameTree[Bound]] = {
+        assert(dtab.show == "/foo=>/bar")
+        val (act, witness) = Activity[NameTree[Bound]]()
+        witness.notify(Return(NameTree.Leaf(Name.Bound(Var(null), s"/io.l5d.namer${dtab.lookup(path).show}"))))
+        act
```
adleong: Could use `Activity.value(...)` here instead of creating an activity with a witness.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -316,31 +316,33 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
     }
 
   private[this] val bindingCacheMu = new {}
-  private[this] var bindingCache: Map[(String, Path), Activity[NameTree[Name.Bound]]] = Map.empty
-  private[this] def getBind(ns: String, path: Path): Activity[NameTree[Name.Bound]] =
+  private[this] var bindingCache: Map[(String, Path, String), Activity[NameTree[Name.Bound]]] = Map.empty
```
adleong: I don't think we want to key off of the extra dtab in the cache.  What do you think about making extraDtab a `Option[String]` in `getBind`?  If the extraDtab is present, then skip the cache and just return `delegate(ns).bind(...)` directly.  If it's absent, then use the cache like before.

Does that seem reasonable?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -316,31 +316,33 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
...
-          val act = delegate(ns).bind(Dtab.empty, path)
+          val act = delegate(ns).bind(Dtab.read(extraDtab), path)
           bindingCache += (key -> act)
           act
       }
     }
 
-  private[this] def handleGetBind(ns: String, path: Path, req: Request): Future[Response] =
+  private[this] def handleGetBind(ns: String, path: Path, req: Request): Future[Response] = {
+    val extraDtab = req.getParam("dtab", "")
```
adleong: `req.params.get("dtab")` will return an `Option[String]` if that's easier to work with

adleong:
This looks great!  Just a couple suggestions, mostly around caching.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
Ashald: Given that `Dtab.read` works well with an empty string I feel like `req.getParam("dtab", "")` is good enough. OTOH if you prefer that `Option` is used I can change the code.

namerd/iface/control-http/src/test/scala/io/buoyant/namerd/iface/HttpControlServiceTest.scala
Ashald: Thanks!

adleong:
:star: This is great!  Thanks for putting this together

olix0r:
thanks for your contribution, @Ashald!  Could you please sign our [Contributor License Agreement](https://buoyant.io/cla) so that we're able to accept this?

PR #532, opened 2016-07-08
Upgrade to finagle 6.35

adleong:
The sampling logic for tracing has changed slightly.  (see: https://github.com/twitter/finagle/commit/1b29f4e334f5f24f1c1ff2766369a75cad63561)

`isActivelyTracing` now checks to see if any of the tracers are actively tracing instead checking the `sampled` bit on the trace itself.  Therefore, BufferedTracer will always sample a trace.  On the other hand, SamplingTracer respects the `sampled` bit so I think this only messes up our tests.

### DIFF
--- a/namer/core/src/test/scala/io/buoyant/namer/DelegateTreeTest.scala
+++ b/namer/core/src/test/scala/io/buoyant/namer/DelegateTreeTest.scala
@@ -13,18 +13,13 @@ class DelegateTreeTest extends FunSuite {
         Delegate(Path.read("/b"), Dentry.read("/a => /b"),
           Alt(Path.read("/b"), Dentry.read("/b => /c|/d"),
             Leaf(Path.read("/c"), Dentry.read("/b => /c|/d"), Path.read("/c")),
-            Leaf(Path.read("/d"), Dentry.read("/b => /c|/d"), Path.read("/d"))
-          )
-        )
-      )
+            Leaf(Path.read("/d"), Dentry.read("/b => /c|/d"), Path.read("/d")))))
 
     val simplified =
       Delegate(Path.read("/a"), Dentry.nop,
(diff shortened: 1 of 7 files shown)


namer/core/src/test/scala/io/buoyant/namer/DelegateTreeTest.scala
```diff
@@ -13,18 +13,13 @@ class DelegateTreeTest extends FunSuite {
         Delegate(Path.read("/b"), Dentry.read("/a => /b"),
           Alt(Path.read("/b"), Dentry.read("/b => /c|/d"),
             Leaf(Path.read("/c"), Dentry.read("/b => /c|/d"), Path.read("/c")),
-            Leaf(Path.read("/d"), Dentry.read("/b => /c|/d"), Path.read("/d"))
-          )
-        )
-      )
+            Leaf(Path.read("/d"), Dentry.read("/b => /c|/d"), Path.read("/d")))))
```
adleong: whatever you say, scalariform

router/core/src/test/scala/io/buoyant/router/ClassifiedTracingTest.scala
```diff
@@ -104,13 +104,4 @@ class ClassifiedTracingTest extends FunSuite with Awaits {
     assert(tracer.iterator.map(_.annotation).toSeq ==
       Seq(Annotation.BinaryAnnotation("l5d.success", 0.84)))
   }
-
-  test("does not trace when is false") {
```
adleong: I think these tests are not actually correct.

Respect for the `sampled` bit of the trace is a property of the Tracer itself so I don't think it makes sense to test this here.  Indeed, BufferedTracer no longer respects it.

olix0r:
⭐  nice!  There are some additional changes i'd like to get in before we do another release, but this is a great first step. thanks!

PR #533, opened 2016-07-11
Fix netty merge strategy

adleong:
Each netty dependency includes a `META-INF/io.netty.versions.properties` file that includes information such as build time.  Since these jars were not all built at the same time, the contents of these files differ and therefore conflict.


klingerf:
⭐  Nice catch.

olix0r:
⭐

PR #534, opened 2016-07-11
Ensure default server configuration is preserved

olix0r:
When configuring servers, the protocol-specific defaults were elided.  Ensure
that they are maintained.


olix0r:
It would be nice to test this stuff better but it seems a little complicated.

adleong:
:star: good sleuthing 🔎

klingerf:
⭐

PR #536, opened 2016-07-12
Fix handling of Addr.Pending in ServersetNamer

adleong:
# Problem

If `resolveServerset` returns `Addr.Pending`, we incorrectly assume that it has successfully bound this path.  This means that this namer may appear to bind paths that it cannot, or might bind entire paths when it should only bind a prefix and leave a residual.

# Solution

Convert the `Var[Name.Bound]` into an Activity to properly reflect the pending state.


olix0r:
⭐  thanks!

as always, i'm curious what we can do to test these things better (but not a blocker)

tonyd3:
🚢

PR #537, opened 2016-07-12
Update identifier interface to return a modified request

adleong:
The Identifier interface now returns a 2-tuple of `(Dst, Request)`.  This allows Identifiers to pass along a modified request in place of the original.  This can be useful for Identifiers that consume segments of the URI, for example.


olix0r:
👍

PR #539, opened 2016-07-13
Add a consume option to the io.l5d.path identifier

### LINKED ISSUE #458: support HTTP routing based on first path component, and strip that component from the request
We've seen a couple use cases where we want to map a call to `localhost:4140/a/b/c` to a call to `/b/c` on service `a`. This is useful when setting the Host header is difficult, but when explicitly setting the destination service to `localhost:4140/a` is easy.

We can either have a general path rewriting mechanism or we can explicitly support the "route on first path component" case, which may capture the majority of usage.

adleong:
Fixes #458

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
@@ -8,12 +8,16 @@ import io.buoyant.router.RoutingFactory
 case class PathIdentifier(
   prefix: Path,
   segments: Int = 1,
+  consume: Boolean = false,
   baseDtab: () => Dtab = () => Dtab.base
 ) extends RoutingFactory.Identifier[http.Request] {
 
   def apply(req: http.Request): Future[(Dst, http.Request)] =
     Path.read(req.path) match {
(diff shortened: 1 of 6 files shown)


router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -8,12 +8,16 @@ import io.buoyant.router.RoutingFactory
...
   segments: Int = 1,
+  consume: Boolean = false,
   baseDtab: () => Dtab = () => Dtab.base
 ) extends RoutingFactory.Identifier[http.Request] {
 
   def apply(req: http.Request): Future[(Dst, http.Request)] =
     Path.read(req.path) match {
       case path if path.size >= segments =>
+        if (consume) {
+          req.uri = req.uri.split("/").drop(segments + 1).mkString("/", "/", "")
```
olix0r: TIOLI: there are probably ways of doing this with fewer copies (e.g. `req.uri.drop(captured.show.length)`) -- probably not worth optimizing yet

olix0r:
lgtm

PR #541, opened 2016-07-13
Experimental Implementation of Consul Storage for Namerd

Ashald:
Although there are no integration tests I verified that it works with local Consul installation.


Ashald:
I'm a super-newby in Scala and I believe this code make use of a lot of refactoring and optimizations. I'd highly appreciate code review and change requests so I can bring this code to the level of quality you will agree to merge into the upstream.

Thanks!

samuel:
This is awesome.

The one thing it seems though is that this PR is using the KV store rather than Consul's service registration. Using the services rather than kv would probably be better as it's the standard way to handle this with Consul as it then allows using the health checks and DNS resolution.

https://www.consul.io/docs/agent/http/catalog.html

Ashald:
@samuel The Consul-enabled `io.l5d.consul` namer in already implemented in Linkerd https://linkerd.io/doc/0.7.0/linkerd/namer/ and it uses Catalog API.

This PR targets the Dtab storage interface rather than namer. And it cannot be implemented with the Catalog API.

samuel:
Ahh, cool, my bad. Thanks for the link.

Ashald:
@samuel you're welcome ;)

adleong:
From giving this a read-through, it looks really good.  I have a bunch of style comments and suggestions but overall it looks great.

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -44,7 +49,7 @@ package object v1 {
     Try(mapper.readValue[T](bytes, begin, end - begin))
   }
 
-  object Api {
+  object CatalogApi {
```
adleong: Could we refactor this into an `Api` base trait and have `CatalogApi` and `KvApi` inherit from it?  Could also split these into 3 files (Api, CatalogApi, KvApi).

Ashald:
@adleong once you have some time, can you please share your comments and suggestions so I can improve the code?

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -121,6 +126,128 @@ package object v1 {
...
+    client: Client,
+    uriPrefix: String,
+    backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+    stats: StatsReceiver = DefaultStatsReceiver
+  ) extends Closable {
+    def close(deadline: Time) = client.close(deadline)
+
+    val kvPrefix = s"$uriPrefix/kv"
+
+    private[this] val infiniteRetryFilter = new RetryFilter[http.Request, http.Response](
```
adleong: infiniteRetryFilter and retryClient could be factored into the `Api` common trait

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -121,6 +126,128 @@ package object v1 {
...
+      val req = mkreq(
+        http.Method.Get,
+        s"$kvPrefix$path",
+        "keys" -> Some(true.toString),
+        "separator" -> separator,
+        "index" -> blockingIndex,
+        "dc" -> datacenter
+      )
+      Trace.letClear(retryClient(retry)(req)).flatMap {
+        case rsp if rsp.status == http.Status.NotFound => Future.exception(NotFound(rsp))
```
adleong: Can all the status code checking be moved into `Indexed.mk`?

Ideally this could just be 
``` 
Trace.letClear(...).flatMap { rsp =>
  Future.const(Indexed.mk[Seq[String]](rsp)
}
```

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -121,6 +126,128 @@ package object v1 {
...
+    )
+
+    def retryClient(retry: Boolean) = {
+      val retryFilter = if (retry) infiniteRetryFilter else Filter.identity[http.Request, http.Response]
+      retryFilter andThen client
+    }
+
+    // https://www.consul.io/docs/agent/http/kv.html#single
+    def list(
+      path: String,
```
adleong: might be worth noting in a comment/scaladoc that path must begin with a slash (I assume?)

consul/src/main/scala/io/buoyant/consul/v1.scala
Ashald: You mean classes, right? Or objects as well? I may miss something but with regards to objects I see only 1 'generalization option' - add a type parameter to the base one and implement `apply` there...

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -121,6 +126,128 @@ package object v1 {
...
+    ): Future[Indexed[String]] = {
+      val req = mkreq(
+        http.Method.Get,
+        s"$kvPrefix$path",
+        "raw" -> Some(true.toString),
+        "index" -> blockingIndex,
+        "dc" -> datacenter
+      )
+
+      Trace.letClear(retryClient(retry)(req)) flatMap {
```
adleong: style nit (here and elsewhere): avoid infix notation for methods that take a block.

e.g.
do
`foo.flatMap {`

not
`foo flatMap {`

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -121,6 +126,128 @@ package object v1 {
...
+        case rsp if rsp.status == http.Status.Ok =>
+          Future.value(Indexed[String](rsp.getContentString(), rsp.headerMap.get(Headers.Index)))
```
adleong: why not use `Indexed.mk`?

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -121,6 +126,128 @@ package object v1 {
...
+      val req = mkreq(
+        http.Method.Put,
+        s"$kvPrefix$path",
+        "cas" -> cas,
+        "dc" -> datacenter
+      )
+      req.setContentString(value)
+      Trace.letClear(retryClient(retry)(req)).flatMap {
+        case rsp if rsp.status == http.Status.Ok =>
+          Future.const(readJson[Boolean](rsp.content))
```
adleong: Could use `Indexed.mk` here too.  Of course, the return type of this method would change to `Future[Indexed[Boolean]]` and the index would always be `None`, but I think that's okay.

consul/src/main/scala/io/buoyant/consul/v1.scala
```diff
@@ -132,6 +259,7 @@ package object v1 {
         readJson[T](rsp.content).map { t =>
           Indexed[T](t, rsp.headerMap.get(Headers.Index))
         }
+      case http.Status.Forbidden => throw Forbidden(rsp)
```
adleong: could move all status code handling to here

consul/src/test/scala/io/buoyant/consul/KvApiTest.scala
```diff
@@ -0,0 +1,225 @@
...
+    rsp.content = buf
+    rsp.headerMap.set("X-Consul-Index", "4")
+    lastUri = req.uri
+    Future.value(rsp)
+  }
+
+  test("list returns an indexed seq of key names") {
+    val service = stubService(listBuf)
+
+    val result = await(KvApi(service).list("foo/"))
```
adleong: it doesn't really impact this test, but I think path needs to start with a slash or else you'll get some weird uris.

consul/src/test/scala/io/buoyant/consul/KvApiTest.scala
```diff
@@ -0,0 +1,225 @@
...
+
+  test("list reports wrong paths with an error") {
+    val failureService = Service.mk[Request, Response] { req =>
+      val rsp = Response()
+      rsp.headerMap.set("X-Consul-Index", "42")
+      rsp.setStatusCode(404)
+      lastUri = req.uri
+      Future.value(rsp)
+    }
+    try {
```
adleong: this could be

`assertThrows[NotFound](await(...))`

(you have to import `io.buoyant.test.Exceptions`)

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,102 @@
+package io.buoyant.namerd
+package storage.consul
+
+import com.twitter.finagle.{Dtab, Path}
+import com.twitter.io.Buf
+import com.twitter.util._
+import io.buoyant.consul.NotFound
+import io.buoyant.consul.v1._
+
```
adleong: don't need this blank line

consul/src/main/scala/io/buoyant/consul/v1.scala
Ashald: `Indexed.mk` tries to parse data with JSON - this endpoint with given parameters returns value as is. If it's an arbitrary string JSON will fail since not everything can be a top-level object in JSON. But I like the idea...

consul/src/main/scala/io/buoyant/consul/v1.scala
adleong: I don't think we would factor up `apply`, but definitely `infiniteRetryFilter` and `retryClient` could be factored up.

```
trait Api {
  ...
}

object CatalogApi extends Api {
  ...
}

object KvApi extends Api {
  ...
}
```

consul/src/main/scala/io/buoyant/consul/v1.scala
adleong: what about creating a `Indexed.mkString` that takes a `http.Response` and returns a `Try[Indexed[String]]`.  It would share the same status code and index handling with `Indexed.mk` but would use the raw string instead of json parsing it.  what do you think?

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreInitializer.scala
```diff
@@ -0,0 +1,48 @@
+package io.buoyant.namerd.storage.consul
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.tracing.NullTracer
+import com.twitter.finagle.{Http, Path}
+import io.buoyant.config.types.Port
+import io.buoyant.consul.SetHostFilter
+import io.buoyant.consul.v1.KvApi
+
```
adleong: don't need this blank line

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,102 @@
...
+      @volatile var running = true
+
+      def cycle(index: Option[String]): Future[Unit] =
+        if (running)
+          api.list(s"${root.show}/", blockingIndex = index) transform {
+            case Return(result) =>
+              val namespaces = result.value.map(namespace).toSet
+              updates() = Activity.Ok(namespaces)
+              cycle(result.index)
+            case Throw(e) =>
```
adleong: I think we need to handle NotFounds here and turn them into an empty namespace list

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,102 @@
...
+    api.get(key) transform {
+      case Return(_) => api.delete(key).unit
+      case Throw(e: NotFound) => Future.exception(new DtabNamespaceDoesNotExistException(ns))
+      case Throw(e) => Future.exception(e)
+    }
+  }
+
+  def update(ns: Ns, dtab: Dtab, version: Version): Future[Unit] = {
+    val Buf.Utf8(vstr) = version
+    api.put(s"${root.show}/$ns", dtab.show, cas = Some(vstr)) flatMap { result =>
```
adleong: consul throws a 500 if the cas value isn't a number.  we should check here that vstr can be parsed as a number, and immediately return a version mismatch exception if not

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,102 @@
...
+        if (running)
+          api.get(key, blockingIndex = index) transform {
+            case Return(result) =>
+              val version = Buf.Utf8(result.index.get)
+              val dtab = Dtab.read(result.value)
+              updates() = Activity.Ok(Some(VersionedDtab(dtab, version)))
+              cycle(result.index)
+            case Throw(e: NotFound) =>
+              updates() = Activity.Ok(None)
+              cycle(None)
```
adleong: in the case of a NotFound, we still want to extract the `X-Consul-Index` and set the index, otherwise we'll just send keep sending requests here.

Ashald:
@adleong I hope I didn't miss anything and addressed all your comments. Can you please take another look? Although now there are so much changes I believe it makes more sense to look just at the code under `consul` and `namerd/storage/consul` rather than at the diff.

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -0,0 +1,76 @@
...
+    val req = http.Request(path, params: _*)
+    req.method = method
+    req
+  }
+
+  private[this] val mapper = new ObjectMapper with ScalaObjectMapper
+  mapper.registerModule(DefaultScalaModule)
+  mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+
+  def parseJson[T: Manifest](buf: Buf): T = {
```
adleong: I think this should still be wrapped in a `Try` like before, to help protect against malformed responses

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -0,0 +1,76 @@
...
+    val Buf.ByteArray.Owned(bytes, begin, end) = Buf.ByteArray.coerce(buf)
+    mapper.readValue[T](bytes, begin, end - begin)
+  }
+
+  private[v1] def execute[T: Manifest](req: http.Request, retry: Boolean): Future[Indexed[T]] = {
+    Trace.letClear(getClient(retry)(req)).map { rsp =>
```
adleong: I think the idiomatic way to mesh this with `parseJson` returning `Try` would be
```
for {
  rsp <- Trace.letClear(getClient(retry)(req))
  value <- Future.const(parseJson[T](rsp.content))
} yield Indexed[T](value, rsp.headerMap.get(Headers.Index))
```

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -0,0 +1,114 @@
...
+    blockingIndex: Option[String] = None,
+    retry: Boolean = false
+  ): Future[Indexed[String]] = {
+    val req = mkreq(
+      http.Method.Get,
+      s"$kvPrefix$path",
+      "index" -> blockingIndex,
+      "dc" -> datacenter
+    )
+    execute[Seq[ConsulKey]](req, retry).map { idx =>
```
adleong: by using the `executeRaw` method I suggest above, we could go back to using raw mode here, which seems nicer

consul/src/main/scala/io/buoyant/consul/v1/package.scala
```diff
@@ -0,0 +1,31 @@
...
+  type Client = Service[http.Request, http.Response]
+
+  trait ConsulApiError extends Throwable {
+    def rsp: http.Response
+  }
+  case class UnexpectedResponse(rsp: http.Response) extends ConsulApiError
+  case class NotFound(rsp: http.Response) extends ConsulApiError
+  case class Forbidden(rsp: http.Response) extends ConsulApiError
+
+  private[v1] val apiErrorFilter = new SimpleFilter[http.Request, http.Response] {
```
adleong: good call making this a filter, I like it

consul/src/main/scala/io/buoyant/consul/v1/package.scala
```diff
@@ -0,0 +1,31 @@
...
+
+    def apply(request: http.Request, service: Client): Future[http.Response] = {
+      service(request).map { response: http.Response =>
+        response.status match {
+          case http.Status.Ok => response
+          case http.Status.Forbidden => throw Forbidden(response)
```
adleong: prefer to use `Future.exception` instead of `throw`:

```
service(request).flatMap { ...
  ...
    case http.Status.Ok => Future.value(response)
    case http.Status.Forbidden => Future.exception(Forbidden(response))
    ...
```

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: You mean an alternative version of `execute` that doesn't require de-serializng? Ok.

consul/src/test/scala/io/buoyant/consul/v1/CatalogApiTest.scala
```diff
@@ -99,11 +95,8 @@ class ApiTest extends FunSuite with Awaits {
       lastUri = req.uri
       Future.value(rsp)
     }
-    try {
-      await(Api(failureService).serviceMap(datacenter = Some("non-existant dc")))
-      assert(false)
-    } catch {
-      case e: UnexpectedResponse => assert(true)
-    }
+    assertThrows[UnexpectedResponse](
```
adleong: nice, thanks

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+
+import com.twitter.finagle.{Dtab, Path}
+import com.twitter.io.Buf
+import com.twitter.util._
+import io.buoyant.consul.v1._
+
+import io.buoyant.namerd.DtabStore.{DtabNamespaceAlreadyExistsException, DtabNamespaceDoesNotExistException}
+
+class ConsulDtabStore(api: KvApi, root: Path) extends DtabStore {
+  import DtabStore.{DtabVersionMismatchException, Version}
```
adleong: use absolute imports and put them at the top of the file

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  override def list(): Activity[Set[Ns]] = {
+    def namespace(key: String): Ns = key.stripPrefix("/").stripSuffix("/").substring(root.show.length)
+
+    val run = Var.async[Activity.State[Set[Ns]]](Activity.Pending) { updates =>
+      @volatile var running = true
+
+      def cycle(index: Option[String]): Future[Unit] =
+        if (running)
+          api.list(s"${root.show}/", blockingIndex = index) transform {
```
adleong: `).transform {`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+              cycle(e.rsp.headerMap.get(Headers.Index))
+            case Throw(e) =>
+              updates() = Activity.Failed(e)
+              cycle(None)
+          }
+        else
+          Future.Unit
+      val pending = cycle(None)
+
+      Closable make { _ =>
```
adleong: `Closable.make`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+        pending.raise(new FutureCancelledException)
+        Future.Unit
+      }
+    }
+
+    Activity(run)
+  }
+
+  def create(ns: Ns, dtab: Dtab): Future[Unit] =
+    api.put(s"${root.show}/$ns", dtab.show, cas = Some("0")) flatMap { result =>
```
adleong: `).flatMap {`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+      if (result) Future.Done else Future.exception(new DtabNamespaceAlreadyExistsException(ns))
+    }
+
+  def delete(ns: Ns): Future[Unit] = {
+    val key = s"${root.show}/$ns"
+    api.get(key) transform {
```
adleong: `).transform {`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+      case Return(_) => api.delete(key).unit
+      case Throw(e: NotFound) => Future.exception(new DtabNamespaceDoesNotExistException(ns))
```
adleong: as far as I can tell, consul seems to always return 200 and "true", even if the key doesn't exist.  *shrug*

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: Ooops... :)

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: Another ooops.. :)

consul/src/main/scala/io/buoyant/consul/v1/package.scala
Ashald: Thanks! :)

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+      case Throw(e) => Future.exception(e)
+    }
+  }
+
+  def update(ns: Ns, dtab: Dtab, version: Version): Future[Unit] = {
+    val Buf.Utf8(vstr) = version
+    if (!vstr.forall(_.isDigit)) Future.exception(new DtabVersionMismatchException)
```
adleong: what about `if (Try(vstr.toLong)).isThrow)`?

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+
+    api.put(s"${root.show}/$ns", dtab.show, cas = Some(vstr)) flatMap { result =>
```
adleong: `).flatMap {`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+    api.put(s"${root.show}/$ns", dtab.show).unit
+
+  def observe(ns: Ns): Activity[Option[VersionedDtab]] = {
+    val key = s"${root.show}/$ns"
+    val run = Var.async[Activity.State[Option[VersionedDtab]]](Activity.Pending) { updates =>
+      @volatile var running = true
+
+      def cycle(index: Option[String]): Future[Unit] =
+        if (running)
+          api.get(key, blockingIndex = index) transform {
```
adleong: `).transform {`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -0,0 +1,106 @@
...
+              cycle(e.rsp.headerMap.get(Headers.Index))
+            case Throw(e) =>
+              updates() = Activity.Failed(e)
+              cycle(None)
+          }
+        else
+          Future.Unit
+      val pending = cycle(None)
+
+      Closable make { _ =>
```
adleong: `Closable.make {`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreInitializer.scala
```diff
@@ -0,0 +1,47 @@
...
+  pathPrefix: Option[Path]
+) extends DtabStoreConfig {
+  import ConsulConfig._
+
+  @JsonIgnore
+  override val experimentalRequired = true
+
+  @JsonIgnore
+  override def mkDtabStore: DtabStore = {
+    val serviceHost = host getOrElse DefaultHost
```
adleong: `host.getOrElse(DefaultHost)`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreInitializer.scala
```diff
@@ -0,0 +1,47 @@
...
+    val servicePort = (port getOrElse DefaultPort).port
```
adleong: `port.getOrElse(DefaultPort)`

namerd/storage/consul/src/test/scala/io/buoyant/namerd/storage/consul/EtcdConfigTest.scala
```diff
@@ -0,0 +1,29 @@
+package io.buoyant.namerd.storage.consul
```
adleong: I think this file has the wrong name

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: Yeah, that's for `delete`. that's why I added a `get` first so we can raise a `DtabNamespaceDoesNotExistException ` as needed.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: I like your version better - I'm just not sure about consul index boundaries. But it should fit into Long, right? :)

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
adleong: oh, yeah, that makes sense.  good call

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
adleong: I think you're missing either a `return Future.exception(...)` or an else block.  Currently the `Future.exception` isn't returned, it's just dropped.

Ashald: Thanks!

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: Argh, I'm angry at myself

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: I'm sorry. :(

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: I'm so sorry. :(

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: I'm super sorry. :(

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreInitializer.scala
Ashald: I just copied it from `etcd`. :D of course will fix. :)

namerd/storage/consul/src/test/scala/io/buoyant/namerd/storage/consul/EtcdConfigTest.scala
Ashald: Ooops, thanks

Ashald:
Another `Try`. :D

adleong:
:star: This is great!  Thanks for all your hard work on this, @Ashald!

[Approved with PullApprove](https://pullapprove.com/BuoyantIO/linkerd/pull-request/541/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=BuoyantIO/linkerd)

Ashald:
Thanks for the help!

PR #542, opened 2016-07-13
Support a Netty4 engine for HTTP routers

olix0r:
Finagle-6.36 introduces a usable netty4 backend for HTTP. linkerd users should
be able to opt-in to this new backend so that it may be introduced in a
controlled manner (until it's proven to be a suitable default).

Client and Server sections of the HTTP router configuration admit an `engine`
block that controls which HTTP implementation is used.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
   recompress bodies.
 * Add a `consume` option to the `io.l5d.path` identifier to strip off the path
   segments that it reads from the URI.
+* Introduce a configurable Netty4 http implementation.
 
 ## 0.7.0
 
@@ -28,7 +29,6 @@
   * Added server success rate graphs to the dashboard, improved responsiveness
   * Added the ability to navigate to a specific router's dashboard
   * Standardized the look and feel of the admin pages
-* Introduce a configurable Netty4 http implementation.
 
 ## 0.6.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,7 @@
   * Added server success rate graphs to the dashboard, improved responsiveness
   * Added the ability to navigate to a specific router's dashboard
   * Standardized the look and feel of the admin pages
+* Introduce a configurable Netty4 http implementation.
 
 ## 0.6.0
 
--- a/linkerd/examples/http.yaml
+++ b/linkerd/examples/http.yaml
@@ -8,14 +8,31 @@ namers:
 # A simple HTTP router that looks up host header values in io.l5d.fs.
 routers:
 - protocol: http
-  identifier:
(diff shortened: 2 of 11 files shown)


CHANGES.md
```diff
@@ -26,6 +26,7 @@
   * Added server success rate graphs to the dashboard, improved responsiveness
   * Added the ability to navigate to a specific router's dashboard
   * Standardized the look and feel of the admin pages
+* Introduce a configurable Netty4 http implementation.
```
adleong: this should go in the x.x.x section above

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -52,13 +52,27 @@ class HttpInitializer extends ProtocolInitializer.Simple {
 
 object HttpInitializer extends HttpInitializer
 
+case class HttpClientConfig(
+  engine: Option[HttpEngine]
+) extends ClientConfig {
+  override def clientParams =
+    engine.foldLeft(super.clientParams) { (params, engine) => engine.mk(params) }
```
adleong: TIOLI: it's probably just me, but for some reason I get confused when folding over options.  
```
engine match {
  case Some(eng) => eng.mk(super.clientParams)
  case None => super.clientParams
}```

is maybe more obvious? FFTI

File: project/Base.scala [hunk 3]
```diff
@@ -36,6 +36,7 @@ class Base extends Build {
     version := Git.version,
     homepage := Some(url("https://linkerd.io")),
     scalaVersion in GlobalScope := "2.11.7",
+    ivyScala := ivyScala.value.map(_.copy(overrideScalaVersion = true)),
```
adleong: what does this do/is this for?

olix0r: Oh, we get a ton of warnings from sbt always because, for instance jackson, pulls in alternate scala versions.  I put this in at some point to pin scala versions to ours

adleong:
:star: 🚒  pending moving the changelog entry

linkerd/examples/http.yaml
```diff
@@ -8,14 +8,31 @@ namers:
...
+    /host => /srv;
     /http/1.1/* => /host;
   servers:
   - port: 4140
     ip: 0.0.0.0
+    engine:
+      kind: netty3
+  client:
+    engine:
+      kind: netty3
```
klingerf: Does it make sense to also update `linkerd/examples/acceptance-test.yaml` to include engine configs?

klingerf:
⭐  Thanks for updating, looks good.

PR #546, opened 2016-07-15
Add tag support to consul namer

### LINKED ISSUE #230: Support Consul Tags and make DC optional
Setup I'm using heavily depends on Consul for service discovery.
I've looked at linkerd as a tool that would greatly help with that setup. 
But right now it seems to me that Linkerd consul integration is lacking some important features to be 100% usable.

dc - should be optional - as by default consul agent resolves queries to local 
you should be able to query services by tags, 

I also have mixed feeling about linkerd using v1/catalog consul api for service discovery instead of '/health/service'

As I'm new to both scala and finagle (including dtabs) I might have missed something so excuse my ignorance.

adleong:
Fixes #230 

The io.l5d.consul namer now has a `includeTag` property.  If set to true, the namer admits paths in the form `/<dc>/<tag>/<serviceName>`.

Filtering based on multiple tags is not supported (this is similar to consul dns).


olix0r:
:star: thanks, alex!

PR #547, opened 2016-07-15
Memoize status stats

olix0r:
Finagle's status stats module does not reuse counters and stats. Furthermore,
it provides a per-status latency histogram, which is pretty wasteful in the
typical case.

So, we memoize stats components; and we only track status-class latency
histograms.

Reported by @stevej.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/StatusCodeStatsFilter.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/StatusCodeStatsFilter.scala
@@ -1,8 +1,9 @@
 package io.buoyant.linkerd.protocol.http
 
-import com.twitter.finagle.http.filter.StatsFilter
-import com.twitter.finagle.http.{Request, Response}
-import com.twitter.finagle.{ServiceFactory, Stack, Stackable, param}
+import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack, Stackable, param}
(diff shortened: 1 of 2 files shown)


olix0r:
needs tests

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/StatusCodeStatsFilter.scala
```diff
@@ -15,7 +16,55 @@ object StatusCodeStatsFilter {
...
         val param.Stats(statsReceiver) = _stats
         if (statsReceiver.isNull) next
-        else new StatsFilter[Request](statsReceiver) andThen next
+        else new StatusCodeStatsFilter(statsReceiver).andThen(next)
       }
     }
 }
+
+/**
+ * Modified from com.twitter.finagle.http.filter.StatsFilter.
```
adleong: I notice we don't stat response size here.  Is that intentional?

Should we just contribute memoizaiton to c.t.f.http.filter.StatsFilter upstream?

olix0r: That is intentional, since response size computation is totally broken for chunked responses, which is really misleading. We should track _request_ and response body sizes and durations, but I'm happy to drop them for the time being.  Should ticket the fix, though.

adleong:
:star: 📊

adleong:
This will significantly clean up debug level log output as well 😁

PR #549, opened 2016-07-16
Improve java compat for Namers

adleong:
Since `Namer` is a trait, it is very difficult to extend in java which makes it difficult to write namer plugins in java.  We provide `JNamer` which is easier to extend from java as well as switching NamerConfig to an abstract class.

With these changes I was able to write a namer plugin in java. It was horrible, but possible.


olix0r:
:star:

PR #552, opened 2016-07-21
Add a Client Stack Module to rewrite HTTP Host Header/Consul integration for it

Ashald:
This PR introduces following features:
1) Consul Catalog Namer now has an extra option `domain` - when set it makes it to embed `l5d-http-host` meta field with value `$name.service.$dc.$domain` into `Addr.Bound`
2) Namerd Thrift Interface/Client now can pass around `Addr.Metadata` for `Addr.Bound` and `Addr.Inet` but only those with values of `String` type
3) `clientStack` within `io.buoyant.linkerd.protocol.HttpInitializer#defaultRouter` now includes `RewriteHostHeader` module that rewrites outgoing request's `Host` header but only if `l5d-http-host` meta field is present within `Addr.Bound` metadata

All changes are covered with tests.

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
@@ -1,0 +1,40 @@
+package io.buoyant.consul.v1
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.http
+import com.twitter.finagle.service.Backoff
+import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
+import com.twitter.util.{Closable, Duration, Future}
+
+object AgentApi {
+  def apply(c: Client): AgentApi = new AgentApi(c, s"/$versionString")
+}
+
+class AgentApi(
+  val client: Client,
+  val uriPrefix: String,
+  val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+  val stats: StatsReceiver = DefaultStatsReceiver
+) extends BaseApi with Closable {
+
+  val agentPrefix = s"$uriPrefix/agent"
+
+  // https://www.consul.io/docs/agent/http/agent.html#agent_self
+  def localAgent(retry: Boolean = false): Future[LocalAgent] = {
+    val req = mkreq(http.Method.Get, s"$agentPrefix/self")
+    executeJson[LocalAgent](req, retry).map(_.value)
+  }
+}
+
+// Classes below are used for JSON parsing hence field names case have to match the on in JSON
+
+// Represents configuration and member information of the local agent
+// For now only Config is parsed as other fields are unused
+case class LocalAgent(
+  Config: Option[Config] // configuration of the local agent stored under the Config key
+)
+
+// Represents configuration of the local agent
+case class Config(
+  Domain: Option[String] // TLD for DNS zone handled by consul, e.g. consul.acme.co.
+)
--- a/consul/src/test/scala/io/buoyant/consul/v1/AgentApiTest.scala
+++ b/consul/src/test/scala/io/buoyant/consul/v1/AgentApiTest.scala
@@ -1,0 +1,32 @@
+package io.buoyant.consul.v1
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{Request, Response}
+import com.twitter.io.Buf
+import com.twitter.util.Future
+import io.buoyant.test.{Awaits, Exceptions}
+import org.scalatest.FunSuite
+
+class AgentApiTest extends FunSuite with Awaits with Exceptions {
+  val localAgentBuf = Buf.Utf8("""{ "Config": { "Domain": "acme.co." } }""")
+  var lastUri = ""
+
+  override val defaultWait = 2.seconds
+
+  def stubService(buf: Buf) = Service.mk[Request, Response] { req =>
+    val rsp = Response()
+    rsp.setContentTypeJson()
+    rsp.content = buf
+    rsp.headerMap.set("X-Consul-Index", "4")
+    lastUri = req.uri
+    Future.value(rsp)
+  }
+
+  test("localAgent returns LocalAgent instance") {
+    val service = stubService(localAgentBuf)
+
+    val result = await(AgentApi(service).localAgent())
+    assert(result.Config.get.Domain.get == "acme.co.")
+  }
+}
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/CatalogNamer.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/CatalogNamer.scala
@@ -1,16 +1,24 @@
-package io.buoyant.consul
+package io.buoyant.namer.consul
 
 import com.twitter.finagle._
+import com.twitter.logging.Logger
 import com.twitter.util._
-import io.buoyant.consul.v1.{UnexpectedResponse, ServiceNode}
+import io.buoyant.consul._
+import io.buoyant.consul.v1.{ServiceNode, UnexpectedResponse}
+import io.buoyant.namer.Metadata
 
 class CatalogNamer(
   idPrefix: Path,
-  api: v1.CatalogApi,
-  includeTag: Boolean = false
+  catalogApi: v1.CatalogApi,
+  agentApi: v1.AgentApi,
+  includeTag: Boolean = false,
+  setHost: Boolean = false
 ) extends Namer {
 
   import CatalogNamer._
+
+  private[this] def domainFuture(): Future[Option[String]] =
+    agentApi.localAgent().map { la => la.Config.flatMap(_.Domain).map(_.stripPrefix(".").stripSuffix(".")) }
 
   /**
    * Accepts names in the form:
@@ -38,7 +46,12 @@ class CatalogNamer(
         Activity.value(NameTree.Neg)
     }
 
-  def lookup(dcName: String, svcKey: SvcKey, id: Path, residual: Path): Activity[NameTree[Name]] = {
+  def lookup(
+    dcName: String,
+    svcKey: SvcKey,
+    id: Path,
+    residual: Path
+  ): Activity[NameTree[Name]] = {
     log.debug("consul lookup: %s %s", id.show)
     Activity(Dc.get(dcName).services).map { services =>
       log.debug("consul dc %s initial state: %s", dcName, services.keys.mkString(", "))
@@ -83,7 +96,8 @@ class CatalogNamer(
   }
 
   /**
-   * Contains all cached serviceNodes responses for a particular serviceName in a particular datacenter
+   * Contains all cached serviceNodes responses for a particular serviceName
+   * in a particular datacenter
    */
   case class SvcCache(datacenter: String, key: SvcKey, updateableAddr: VarUp[Addr]) {
 
@@ -97,7 +111,7 @@ class CatalogNamer(
     }
 
     def mkRequest(): Future[Seq[ServiceNode]] =
-      api
+      catalogApi
         .serviceNodes(key.name, datacenter = Some(datacenter), tag = key.tag, blockingIndex = Some(index), retry = true)
         .map { indexedNodes =>
           indexedNodes.index.foreach(setIndex)
@@ -112,26 +126,37 @@ class CatalogNamer(
 
     def serviceNodeToAddr(node: ServiceNode): Option[Address] = {
       (node.Address, node.ServiceAddress, node.ServicePort) match {
-        case (_, Some(addr), Some(port)) if !addr.isEmpty =>
-          Some(Address(addr, port))
-        case (Some(addr), _, Some(port)) if !addr.isEmpty =>
-          Some(Address(addr, port))
+        case (_, Some(serviceIp), Some(port)) if !serviceIp.isEmpty =>
+          Some(Address(serviceIp, port))
+        case (Some(nodeIp), _, Some(port)) if !nodeIp.isEmpty =>
+          Some(Address(nodeIp, port))
         case _ => None
       }
     }
 
     def update(nodes: Seq[ServiceNode]): Future[Unit] = {
-      synchronized {
-        try {
-          val socketAddrs = nodes.flatMap(serviceNodeToAddr).toSet
-          updateableAddr() = if (socketAddrs.isEmpty) Addr.Neg else Addr.Bound(socketAddrs)
-        } catch {
-          // in the event that we are trying to parse an invalid addr
-          case e: IllegalArgumentException =>
-            updateableAddr() = Addr.Failed(e)
+      val metaFuture =
+        if (setHost)
+          domainFuture().map { domainOption =>
+            val domain = domainOption.getOrElse("consul")
+            Addr.Metadata(Metadata.authority -> s"${key.name}.service.$datacenter.$domain")
+          }
+        else
+          Future.value(Addr.Metadata.empty)
+
+      metaFuture.flatMap { meta =>
+        synchronized {
+          try {
+            val socketAddrs = nodes.flatMap(serviceNodeToAddr).toSet
+            updateableAddr() = if (socketAddrs.isEmpty) Addr.Neg else Addr.Bound(socketAddrs, meta)
+          } catch {
+            // in the event that we are trying to parse an invalid addr
+            case e: IllegalArgumentException =>
+              updateableAddr() = Addr.Failed(e)
+          }
         }
+        mkRequest().flatMap(update).handle(handleUnexpected)
       }
-      mkRequest().flatMap(update).handle(handleUnexpected)
     }
 
     val handleUnexpected: PartialFunction[Throwable, Unit] = {
@@ -165,7 +190,7 @@ class CatalogNamer(
     }
 
     def mkRequest(): Future[Seq[SvcKey]] =
-      api
+      catalogApi
         .serviceMap(datacenter = Some(name), blockingIndex = Some(index), retry = true)
         .map { indexedSvcs =>
           indexedSvcs.index.foreach(setIndex)
@@ -257,5 +282,5 @@ class CatalogNamer(
 object CatalogNamer {
   type VarUp[T] = Var[T] with Updatable[T]
   type ActUp[T] = VarUp[Activity.State[T]]
+  private val log = Logger.get(getClass.getName)
 }
-
--- a/namerd/iface/interpreter-thrift-idl/src/main/thrift/namer.thrift
+++ b/namerd/iface/interpreter-thrift-idl/src/main/thrift/namer.thrift
@@ -81,6 +81,7 @@ struct AddrReq {
 }
 
 struct AddrMeta {
+  1: optional string authority // HTTP/1.1 Host or HTTP/2.0 :authority
 }
 
 struct TransportAddress {
--- a/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
+++ b/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
@@ -9,7 +9,7 @@ import com.twitter.io.Buf
 import com.twitter.logging.Logger
 import com.twitter.util._
 import com.twitter.util.TimeConversions._
-import io.buoyant.namer.{DelegateTree, Delegator}
+import io.buoyant.namer.{Metadata, DelegateTree, Delegator}
 import io.buoyant.namerd.iface.{thriftscala => thrift}
 import java.net.{InetAddress, InetSocketAddress}
 
@@ -146,6 +146,17 @@ class ThriftNamerClient(
     mk(ttree.root)
   }
 
+  /**
+   * Converts Thrift AddrMeta into Addr.Metadata
+   */
+  private[this] def convertMeta(thriftMeta: Option[thrift.AddrMeta]): Addr.Metadata = {
+    // TODO handle non-String metadata
+    thriftMeta match {
+      case Some(thrift.AddrMeta(Some(authority))) => Addr.Metadata(Metadata.authority -> authority)
+      case _ => Addr.Metadata.empty
+    }
+  }
+
   private[this] def watchAddr(id: TPath): Var[Addr] = {
     val idPath = mkPath(id).show
 
@@ -162,17 +173,16 @@ class ThriftNamerClient(
             Trace.record("namerd.client/addr.neg")
             loop(stamp1)
 
-          case Return(thrift.Addr(stamp1, thrift.AddrVal.Bound(thrift.BoundAddr(taddrs, _)))) =>
+          case Return(thrift.Addr(stamp1, thrift.AddrVal.Bound(thrift.BoundAddr(taddrs, boundMeta)))) =>
             val addrs = taddrs.map { taddr =>
-              val thrift.TransportAddress(ipbb, port, _) = taddr
+              val thrift.TransportAddress(ipbb, port, addressMeta) = taddr
               val ipBytes = Buf.ByteArray.Owned.extract(Buf.ByteBuffer.Owned(ipbb))
               val ip = InetAddress.getByAddress(ipBytes)
-              // TODO convert metadata
-              Address(new InetSocketAddress(ip, port))
+              Address.Inet(new InetSocketAddress(ip, port), convertMeta(addressMeta))
             }
             // TODO convert metadata
             Trace.recordBinary("namerd.client/addr.bound", addrs)
-            addr() = Addr.Bound(addrs.toSet)
+            addr() = Addr.Bound(addrs.toSet[Address], convertMeta(boundMeta))
             loop(stamp1)
 
           case Throw(e@thrift.AddrFailure(msg, retry, _)) =>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 ## x.x.x
 
-* Add `resolve` endpoint to Namerd HTTP API 
+* Add `resolve` endpoint to Namerd HTTP API
+* Add `authority` metadata field to re-write HTTP host/:authority on demand
+* Add `setHost` parameter for Consul CatalogNamer to set `authority` metadata 
 * Add file-system based name interpreter.
 * Add auth `token` parameter to Consul Namer & Dtab Store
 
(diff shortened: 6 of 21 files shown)


Ashald:
@olix0r please take a look at this PR whenever you'll have time. I'm interested in your 'conceptual' comments along with code review as I feel that my code if far away from being pretty. :/

olix0r:
@Ashald thanks for your patience waiting for us to get to this.

My initial understanding of this feature was that the Host value you'd like to set was actually stored in Consul.  I now see that this is provided to the namer configuration statically in the namerd/linkerd configuration.   So, effectively, all names resolved through this namer must receive the same domain value.  If that's the case, why does need any integration at all with the consul namer?  Couldn't we just add a module to linkerd that sets the host value?  It seems odd to me that this is set globally on the namer rather than on a per-service basis.

Ashald:
Hi @olix0r, thanks for taking your time to look a this one.

The short answer that configured value in namer is not a static thing that will be set for all requests but rather a required suffix to generate Consul-ish host value. I tired to explain that briefly within `linkerd/docs/namer.md`.

When Consul is used as a DNS server it resolves names that end with pre-configured TLD. By default it's `.consul` but can be changed to anything, for instance, `acme.co`. Given that TLD Consul will resolve following name 'patterns':
1) `$name.service.$dc.$tld` - some service (e.g., `foobar.service.us-east.acme.co`)
2) `$node.node.$dc.$tld` - some machine (e.g., `web-worker1.node.us-east.acme.co`)

As I said above, you can configure TLD in Consul config but unfortunately it doesn't return it when you query its HTTP API for service info. It only uses it to distinguish between names it's supposed to handle and others that it cannot handle.
We're heavily relying on HTTP `Host` header in order to distinguish between different services/versions of service hosted on single `ip:port`. I believe other Consul users might have similar use-cases.

The idea behind this PR is to allow to set the aforementioned TLD in Consul namer and use it to set `Host` header in the form `$name.service.$dc.$tld` whenever `linkerd` forwards request based on Consul concrete name.

Having this feature and services `foo-1` and `foo-2` we will be able to switch a version by just setting a dtab like `/foo=>/foo-1` and we won't need to generate additional web server configs to handle a 'canonical' name like `foo.service.us-east.acme.co`. Moreover it enables us to do an a/b or canary testing with different versions sitting on the same `ip:port`. I think it won't be useful only to us but is the crucial feature for many Consul users who rely on `Host` value.

Ashald:
@olix0r upon further digging Consul API I found an endpoint that reports the domain value we need. I think it will make a lot of sense to integrate with it so we can fetch the domain from Consul directly. Given that the option in `Namer` will become a flag. The question is how to name it: `override_host`? Or something else?

olix0r:
@Ashald thanks for the background (and sorry for missing your other explanation in the docs ;)

I agree that it would be preferable to fetch this value from consul directly, rather than have to synchronize configuration across consul and linkerd.

So:
- the consul namer propagates a domain suffix
- linkerd has some sort of stripDomain option that causes this value to be used to modify downstream requests

Ashald:
@olix0r thanks for the review. So right now I have it implemented following way:
1) if given `l5d-http-host` meta field on bound addr `linekrd` will always set that as a host
2) there is an option on `ConsulCatalogNamer` whether to set that meta field or not

I see the benefit of such implementation in (1) that other components may use this feature as well (2) and you can enabled/disable that on per-namer basis.

Although what you mentioned above is supposed to work in an 'opposite direction' I don't mind changing the PR to make it work as you suggested. Though I'll need your advice on how to add that parameter to `linkerd`.

BTW, are you OK with meta field called `l5d-http-host`?

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -2,16 +2,20 @@ package io.buoyant.consul
...
-  includeTag: Boolean = false
+  catalogApi: v1.CatalogApi,
+  agentApi: v1.AgentApi,
+  includeTag: Boolean = false,
+  setHost: Boolean = false
 ) extends Namer {
 
   import CatalogNamer._
 
+  lazy val domainFuture: Future[String] = agentApi.self().map(_.Config.get.Domain.get.stripPrefix(".").stripSuffix("."))
```
olix0r: please try to wrap lines around 100 chars.  In this case I'd just move `agentApi...` onto the following line.

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -89,49 +93,42 @@ class CatalogNamer(
...
-
-    def mkRequest(): Future[Seq[ServiceNode]] =
-      api
-        .serviceNodes(key.name, datacenter = Some(datacenter), tag = key.tag, blockingIndex = Some(index), retry = true)
-        .map { indexedNodes =>
-          indexedNodes.index.foreach(setIndex)
-          indexedNodes.value
-        }
+    def mkRequest(index: Option[String]) =
+      catalogApi.serviceNodes(key.name, datacenter = Some(datacenter), tag = key.tag, blockingIndex = index, retry = true)
```
olix0r: similarly, I'd break this up like:

```scala
catalogApi.serviceNodes(
  key.name,
  datacenter = ...,
  tag = ...,
  blockingIndex = index,
  retry = true
)
```

namerd/iface/interpreter-thrift-idl/src/main/thrift/namer.thrift
```diff
@@ -81,6 +81,7 @@ struct AddrReq {
 }
 
 struct AddrMeta {
+  1: map<string, string> fields
```
olix0r: I'd prefer to enumerate the set of supported metadata, until we really need an arbitrary map.

Let's have an optional `authority` field (to match HTTP/2 conventions) that captures what we're doing here.

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
```diff
@@ -0,0 +1,31 @@
+package io.buoyant.consul.v1
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.http
+import com.twitter.finagle.service.Backoff
+import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
+import com.twitter.util.{Closable, Duration, Future}
+
+object AgentApi {
+  def apply(c: Client): AgentApi = new AgentApi(c, s"/$versionString")
```
olix0r: Where does _versionString_ come from

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
```diff
@@ -0,0 +1,31 @@
...
+}
+
+class AgentApi(
+  override val client: Client,
+  override val uriPrefix: String,
+  override val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+  override val stats: StatsReceiver = DefaultStatsReceiver
```
olix0r: Omit `override`?  I don't think these actually override anything...

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
```diff
@@ -0,0 +1,31 @@
...
+  // https://www.consul.io/docs/agent/http/agent.html#agent_self
+  def self(retry: Boolean = false): Future[Self] = {
+    val req = mkreq(http.Method.Get, s"$agentPrefix/self")
+    executeJson[Self](req, retry).map(_.value)
+  }
+}
+
+case class Config(Domain: Option[String])
+
+case class Self(Config: Option[Config])
```
olix0r: please lowercase field names (`domain`, `config`).

please include a ~1-sentence summary comment of each of these types so that a maintainer can get their bearings without having to read consul docs

olix0r:
Thanks again for all the work on this.

Right now I think the main blocker is that I'd like to change is to move away from an arbitrary map of attributes to something typed.  Since the read side of this is a part of the core linkerd http functionality, I think it makes sense to introduce the metadata as a first-class primitive.

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
Ashald: Regarding case of field names - here I followed example from https://github.com/BuoyantIO/linkerd/blob/master/consul/src/main/scala/io/buoyant/consul/v1/CatalogApi.scala#L62 - aren't field names should match JSON field names for parser to work properly? Or it can figure out the thing? Because in Consul JSON output they're starting from upper case.

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
Ashald: The same as for other APIs (I inherited it from the original implementation) - from `consul/src/main/scala/io/buoyant/consul/v1/package.scala` - it defines `package object v1` with common stuff.

Ashald:
@olix0r I pushed some changes and updated the PR as per your comments. Can you please take another look?

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -83,55 +93,58 @@ class CatalogNamer(
...
+      )
 
     def clear(): Unit = synchronized {
       updateableAddr() = Addr.Pending
     }
 
-    def init(): Future[Unit] = mkRequest().flatMap(update).handle(handleUnexpected)
+    def init(): Future[Unit] = domainFuture.flatMap { domain =>
+      val meta = if (setHost)
+        Addr.Metadata(("authority", s"${key.name}.service.$datacenter.$domain"))
```
olix0r: I'd prefer if we had one val'd version of this string so that typos were compilation errors.

Also minor style preference for `Addr.Metadata("authority" -> s"...")`.  Neither are blockers, just sharing...

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -83,55 +93,58 @@ class CatalogNamer(
...
+      else
+        Addr.Metadata.empty
+      mkRequest(None).flatMap(nodes => update(nodes, meta)).handle(handleUnexpected)
```
olix0r: tioli: it would be fine to write this as `mkRequest(None).flatMap(update(_, meta)).handle(...)`

Ashald: That's neat - thanks!

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -2,16 +2,21 @@ package io.buoyant.consul
...
+  catalogApi: v1.CatalogApi,
+  agentApi: v1.AgentApi,
+  includeTag: Boolean = false,
+  setHost: Boolean = false
 ) extends Namer {
 
   import CatalogNamer._
 
+  lazy val domainFuture: Future[String] =
+    agentApi.self().map(_.Config.get.Domain.get.stripPrefix(".").stripSuffix("."))
```
adleong: avoid `Option.get`.  instead do something like:
```
agentApi.self().flatMap { self =>
  val domain = self.Config
    .flatMap(_.Domain)
    .map(_.stripPrefix(".").stripSuffix("."))
  domain match {
    case Some(d) => Future.value(d)
    case None => Future.exception(???)
  }
}
```

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
Ashald: You mean, within this file? Or across the whole `linkerd` codebase?

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -83,55 +93,58 @@ class CatalogNamer(
...
-    }
-
-    def mkRequest(): Future[Seq[ServiceNode]] =
-      api
-        .serviceNodes(key.name, datacenter = Some(datacenter), tag = key.tag, blockingIndex = Some(index), retry = true)
-        .map { indexedNodes =>
-          indexedNodes.index.foreach(setIndex)
-          indexedNodes.value
-        }
+    def mkRequest(index: Option[String]) =
```
adleong: is this `index` refactor unrelated to the rest of the change?  If so, I'd prefer it split into a separate PR.

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
olix0r: Ideally, it would be great to have the known metadata keys somewhere (as we do for headers, for instance) -- we can be incremental to getting it somewhere global, though.

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
```diff
@@ -0,0 +1,40 @@
...
+    val req = mkreq(http.Method.Get, s"$agentPrefix/self")
+    executeJson[Self](req, retry).map(_.value)
+  }
+}
+
+// Classes below are used for JSON parsing hence field names case have to match the on in JSON
+
+// Represents configuration and member information of the local agent
+// For now only Config is parsed as other fields are unused
+case class Self(
```
adleong: What do you think about calling this `LocalAgent` or something like that?  `Self` (and `self` above) is a bit confusing because of scala's self-types.

olix0r:
modulo nits and whatever alex has, this is looking about ready to me

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -119,6 +120,10 @@ object HttpControlService {
     val prefix = s"$apiPrefix/addr"
   }
 
+  object ResolveUri extends NsPathUri {
```
adleong: I think this Resolve stuff snuck in from a different PR

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
```diff
@@ -162,17 +173,16 @@ class ThriftNamerClient(
...
-          case Return(thrift.Addr(stamp1, thrift.AddrVal.Bound(thrift.BoundAddr(taddrs, _)))) =>
+          case Return(thrift.Addr(stamp1, thrift.AddrVal.Bound(thrift.BoundAddr(taddrs, boundMeta)))) =>
             val addrs = taddrs.map { taddr =>
-              val thrift.TransportAddress(ipbb, port, _) = taddr
+              val thrift.TransportAddress(ipbb, port, addressMeta) = taddr
               val ipBytes = Buf.ByteArray.Owned.extract(Buf.ByteBuffer.Owned(ipbb))
               val ip = InetAddress.getByAddress(ipBytes)
-              // TODO convert metadata
-              Address(new InetSocketAddress(ip, port))
+              Address.Inet(new InetSocketAddress(ip, port), convertMeta(addressMeta)).asInstanceOf[Address]
```
adleong: I don't think you need this asInstanceOf cast

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
Ashald: Woul it be OK if I'll put it into an object in consul namespace for starters?

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
Ashald: Sure

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
Ashald: Thanks!

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
Ashald: Yeah, a bit of it. I just saw a way to get rid of variable and changed the code a bit. But can put this into a separate PR.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
Ashald: Ooops, you right - I'll drop changes these from this PR

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
Ashald: It didn't compile without it. :( Is there is a better way around?

adleong: Address.Inet is a subclass of Address so something weird must be going on here...

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
Ashald: According to Consul API spec the `Domain` field cannot be empty and thus `Future.exception(???)` should never be executed. Would be OK to leave it as is or I should replace `???` with some concrete exception (and which one would be appropriate in such case)?

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
Ashald:
```
[error] .../linkerd/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala:185: overloaded method value apply with alternatives:
[error]   (addrs: scala.collection.immutable.Set[com.twitter.finagle.Address],metadata: com.twitter.finagle.Addr.Metadata)com.twitter.finagle.Addr.Bound <and>
[error]   (addrs: com.twitter.finagle.Address*)com.twitter.finagle.Addr
[error]  cannot be applied to (scala.collection.immutable.Set[com.twitter.finagle.Address.Inet], com.twitter.finagle.Addr.Metadata)
[error]             addr() = Addr.Bound(addrs.toSet, convertMeta(boundMeta))
[error]                           ^
[error] one error found
[error] (namerd-iface-interpreter-thrift/compile:compileIncremental) Compilation failed
```

adleong: ah, I think scala just needs some help with the type unification.  Try:

```
addr() = Addr.Bound(addrs.toSet[Address], convertMeta(boundMeta))
```

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -2,16 +2,30 @@ package io.buoyant.consul
...
   import CatalogNamer._
 
+  private[this] lazy val domainFuture: Future[String] =
+    agentApi.self().flatMap { self =>
+      val domain = self.Config
+        .flatMap(_.Domain)
+        .map(_.stripPrefix(".").stripSuffix("."))
+      domain match {
+        case Some(d) => Future.value(d)
+        case None => Future.exception(???)
```
adleong: Sorry, I meant that you would use some real exception here to indicate that the consul response was malformed, or didn't contain the right info, or whatever.

Ashald: I asked about that in a comment above. :)

I see 2 options:
1) Some concrete exception - in that case can you recommend one? :)
2) Rather than exception fallback to `consul` - the default value according to Consul docs.

What would you prefer?

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
```diff
@@ -2,16 +2,30 @@ package io.buoyant.consul
...
-  includeTag: Boolean = false
+  catalogApi: v1.CatalogApi,
+  agentApi: v1.AgentApi,
+  includeTag: Boolean = false,
+  setHost: Boolean = false
 ) extends Namer {
 
   import CatalogNamer._
 
+  private[this] lazy val domainFuture: Future[String] =
```
adleong: Something to note: the domain request will be fired once when the catalog namer does `init`, and then never again.  This means if that request fails for any reason, the domain will never be populated.  Worse, anything flatMapped from the domain future (like the rest of the body of `init` will never be executed).  

Similarly, this also means that we will never pick up changes if the domain value in consul changes.

Does the consul self API support long polling?  Can we have the domain be populated by a Var.async instead?

@olix0r any thoughts on this?

Ashald: Unfortunately it doesn't support long polling. Would you like a `Var.async` with some interval? If yes, what the interval would be ok? Something like `10.seconds`?

consul/src/main/scala/io/buoyant/consul/v1/AgentApi.scala
```diff
@@ -0,0 +1,40 @@
...
+  val client: Client,
+  val uriPrefix: String,
+  val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+  val stats: StatsReceiver = DefaultStatsReceiver
+) extends BaseApi with Closable {
+
+  val agentPrefix = s"$uriPrefix/agent"
+
+  // https://www.consul.io/docs/agent/http/agent.html#agent_self
+  def self(retry: Boolean = false): Future[LocalAgent] = {
```
adleong: I'd rename this method to `localAgent` too

project/LinkerdBuild.scala
```diff
@@ -12,7 +12,8 @@ object LinkerdBuild extends Base {
   val Bundle = config("bundle") extend Minimal
   val Dcos = config("dcos") extend Bundle
 
-  val consul = projectDir("consul")
+  lazy val consul = projectDir("consul")
```
adleong: We try to avoid lazy vals in our build file to keep dependancies explicit.  Adding a dependency on namer/core actually tangles things up a bit.  I actually think the best way to resolve this is to move the CatalogNamer out of `consul` and into `namer/consul`.

Ashald: OK, will do.

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
adleong: Option 3) have domainFuture be a `Future[Option[String]]`.  When you use this value elsewhere, you can substitute a default value if it's None.

Ashald: Thanks!

consul/src/main/scala/io/buoyant/consul/CatalogNamer.scala
adleong: 10s seems like a reasonable polling interval to me.  Just a warning that this is going to bit tricky to get right because `Var.async`s will stop updating if they don't have any listeners which will be true most of the time.  We probably don't want the `Var.async` to restart and fire a new request for every consul change.  (or maybe we do, in which case we can ditch `Var.async` and just make a Future call each time.)

Ashald: Hm, sounds more complicated than I though first. In this case, would you mind making it 'just an another Future call each time' for now. Once we will get some usage experience with this we will be able to see the performance impact and if it will be significant then we will contribute another PR with an improvement.

What do you think?

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
Ashald: Thanks!

Ashald:
@adleong I addressed you latest comments - can you take another look please?

namer/consul/src/main/scala/io/buoyant/namer/consul/CatalogNamer.scala
```diff
@@ -108,30 +122,43 @@ class CatalogNamer(
       updateableAddr() = Addr.Pending
     }
 
-    def init(): Future[Unit] = mkRequest().flatMap(update).handle(handleUnexpected)
+    private[this] def metaFuture() =
```
adleong: I don't think this refactor is necessary.  You should be able to just add a call to domainFuture in update (if setHost is set, of course).

namer/consul/src/main/scala/io/buoyant/namer/consul/CatalogNamer.scala
```diff
@@ -108,30 +122,43 @@ class CatalogNamer(
...
+        }
+      else
+        Future.value(Addr.Metadata.empty)
+
+    def cycle(): Future[Unit] =
+      for {
+        services <- mkRequest()
+        meta <- metaFuture()
+        result <- update(services, meta)
+      } yield cycle().handle(handleUnexpected)
```
adleong: This isn't quite right; `cycle` returns a Future so it shouldn't be in the yield position.

Ashald: Can I do `yield None` and then a call to `cycle` on the next line? Or that's idiomatically incorrect?

namer/consul/src/main/scala/io/buoyant/namer/consul/CatalogNamer.scala
Ashald: So should I roll-back or this way it also OK?

adleong: I suggest rolling back the refactor and just adding the domain call in update.

consul/src/test/scala/io/buoyant/consul/v1/AgentApiTest.scala
```diff
@@ -0,0 +1,32 @@
...
+  def stubService(buf: Buf) = Service.mk[Request, Response] { req =>
+    val rsp = Response()
+    rsp.setContentTypeJson()
+    rsp.content = buf
+    rsp.headerMap.set("X-Consul-Index", "4")
+    lastUri = req.uri
+    Future.value(rsp)
+  }
+
+  test("self returns Self instance") {
```
adleong: update test name to not reference `Self` object

Ashald: Done

adleong:
:star: Thanks for being patient with all my requests!  This looks good to me!

olix0r:
👍  looks great!

Ashald:
Thanks for your review guys! And sorry that this PR took so much of your time. :/ I'll try to go with smaller PRs in future.

PR #553, opened 2016-07-22
HTTP Router module to discard proxy side-effects

Ashald:
* Strip host from URI
* Drop `Proxy-*` requests

Request considered being proxied if its URI is absolute

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/Http.scala
+++ b/router/http/src/main/scala/io/buoyant/router/Http.scala
@@ -6,7 +6,7 @@ import com.twitter.finagle.param.ProtocolLibrary
 import com.twitter.finagle.server.StackServer
 import com.twitter.finagle.{Http => FinagleHttp, Server => FinagleServer, http => fhttp, _}
 import io.buoyant.router.Http.param.HttpIdentifier
-import io.buoyant.router.http.{ForwardedFilter, MethodAndHostIdentifier, StripConnectionHeader}
+import io.buoyant.router.http.{DiscardProxySideEffects, ForwardedFilter, MethodAndHostIdentifier, StripConnectionHeader}
 import java.net.SocketAddress
 
 object Http extends Router[Request, Response] with FinagleServer[Request, Response] {
@@ -69,6 +69,7 @@ object Http extends Router[Request, Response] with FinagleServer[Request, Respon
   object Server {
     val stack: Stack[ServiceFactory[Request, Response]] =
       (ForwardedFilter.module +: FinagleHttp.Server.stack)
+        .insertBefore(http.TracingFilter.role, DiscardProxySideEffects.module)
 
     val defaultParams: Stack.Params =
       StackServer.defaultParams +
--- a/router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala
@@ -1,0 +1,39 @@
+package io.buoyant.router.http
+
+import com.twitter.finagle.http.{Request, Response}
+import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack}
+import java.net.URI
+
+object DiscardProxySideEffects {
+
+  /**
+   * Strips host from request URI and drops Proxy-* headers
+   * Those side-effects are added by proxy-aware clients
+   */
+  class DiscardProxySideEffects extends SimpleFilter[Request, Response] {
+    def apply(req: Request, svc: Service[Request, Response]) = {
+      val uri = new URI(req.uri)
+
+      if (uri.isAbsolute) { // Proxied request
+        // TODO: HTTP/2.0 support once finagle supports it
+        val relativeUri = new URI(null, null, uri.getRawPath, uri.getRawQuery, uri.getRawFragment)
+        req.host = uri.getAuthority
+        req.uri = relativeUri.toString
+      }
+
+      req.headerMap.retain { (k, v) => !k.startsWith("Proxy-") }
+      svc(req)
+    }
+  }
+
+  object DiscardProxySideEffects {
+    def apply() = new DiscardProxySideEffects()
+  }
+
+  object module extends Stack.Module0[ServiceFactory[Request, Response]] {
+    val role = Stack.Role("DiscardProxySideEffects")
+    val description = "Strips host from request URI and drops Proxy-* headers"
+    def make(next: ServiceFactory[Request, Response]) =
+      DiscardProxySideEffects() andThen next
+  }
+}
--- a/router/http/src/test/scala/io/buoyant/router/http/DiscardProxySideEffectsTest.scala
+++ b/router/http/src/test/scala/io/buoyant/router/http/DiscardProxySideEffectsTest.scala
@@ -1,0 +1,64 @@
+package io.buoyant.router.http
+
+import com.twitter.finagle._
+import com.twitter.finagle.http.{Status, _}
+import com.twitter.util.Future
+import io.buoyant.test.Awaits
+import org.scalatest.FunSuite
+
+class DiscardProxySideEffectsTest extends FunSuite with Awaits {
+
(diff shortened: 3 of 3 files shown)


Ashald:
@olix0r, another PR to your attention - as we agreed yesterday :)

router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * Strips host from request URI and drops Proxy-* headers
+   * Those side-effects are added by proxy-aware clients
+   */
+  object filter extends SimpleFilter[Request, Response] {
+    def apply(req: Request, svc: Service[Request, Response]) = {
+      if (!req.uri.startsWith("/")) { // Proxied request
+        // split URI like http://acme.co/foo/bar?x=1 into host 'acme.co' and uri 'foo/bar?x=1'
+        val Array(_, _, host, uri) = req.uri.split("/", 4)
```
olix0r: this will throw an exception if there aren't 4 elements returned.  we should be more tolerant of unexpected input:

```
req.uri.split("/", 4) match {
  case Array(...) =>
  case _ => // doesn't match ...
}

Comment (olix0r) [inline router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala]: If at all possible i'd prefer to use a real url parser.  URLs come in some unexpected shapes and sizes.  Does java.net have something that does this job?

File: router/http/src/test/scala/io/buoyant/router/http/DiscardProxySideEffectsTest.scala [hunk 2]
```diff
@@ -0,0 +1,70 @@
...
+    req.headerMap.set("Proxy-Authorization", "Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==")
+    req.headerMap.set("Proxy-Authenticate", "Basic")
+    req.headerMap.set("Proxy-Foo", "Bar")
+    req.headerMap.set("Pragma", "42")
+
+    val result = await(service(req))
+    assert(result.getContentString() == "/foo;matrix=uri;/bar?x=1&y=2")
+    assert(
+      result.headerMap.toSet == Set(
+        ("Proxy-Connection", "Keep-Alive"),
```
olix0r: We should probably drop Proxy-Connection regardless of whether it's detected to be a proxy request or not.  Connection headers should _never_ be forwarded by http intermediaries.

router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack}
+import com.twitter.finagle.http.{Request, Response}
+
+object DiscardProxySideEffects {
+
+  /**
+   * Strips host from request URI and drops Proxy-* headers
+   * Those side-effects are added by proxy-aware clients
+   */
+  object filter extends SimpleFilter[Request, Response] {
```
olix0r: i know we've done this elsewhere, but now i have a slight preference for making this a class, e.g.

```scala
class DiscardProxySideEffects extends SimpleFilter... {
}

object DiscardProxySideEffects {
  ...
}
```

So that the filter can easily be extended.

olix0r:
All in all this looks good! A few small comments...

router/http/src/main/scala/io/buoyant/router/Http.scala
```diff
@@ -32,6 +32,7 @@ object Http extends Router[Request, Response] with FinagleServer[Request, Respon
     val client: StackClient[Request, Response] = FinagleHttp.client
       .transformed(StackRouter.Client.mkStack(_))
       .transformed(_.replace(http.TracingFilter.role, http.TracingFilter.module))
+      .transformed(_.insertBefore(http.TracingFilter.role, DiscardProxySideEffects.module))
```
olix0r: Finally, I think that this needs to be installed on the serverside and not the client-side.

We want this rewriting to happen before we apply an Identifier to the request (which happens above the client).

router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala
Ashald: Yeah, that makes sense. At first I thought about keeping it simple but I was still in doubts. I'll dig API, thanks for an advice.

router/http/src/test/scala/io/buoyant/router/http/DiscardProxySideEffectsTest.scala
Ashald: Also though about that - will do. :)

So I see to options: 
1) always drop all `Proxy-*` headers
2) always drop `Proxy-Connection` header and deal with other `Proxy-*` headers if we're in 'proxy mode'

what would you prefer?

router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala
Ashald: OK, will do.

router/http/src/main/scala/io/buoyant/router/Http.scala
Ashald: Looks like I failed and put it in wrong place although you initially told me to put it onto the server side. %) Will fix, thanks.

Ashald:
@olix0r pushed an updated - please take another look

router/http/src/main/scala/io/buoyant/router/Http.scala
```diff
@@ -1,13 +1,14 @@
...
+
 import com.twitter.finagle.client.StackClient
 import com.twitter.finagle.http.{Request, Response, TlsFilter}
 import com.twitter.finagle.param.ProtocolLibrary
 import com.twitter.finagle.server.StackServer
 import com.twitter.finagle.{Http => FinagleHttp, Server => FinagleServer, http => fhttp, _}
 import io.buoyant.router.Http.param.HttpIdentifier
-import io.buoyant.router.http.{ForwardedFilter, MethodAndHostIdentifier, StripConnectionHeader}
-import java.net.SocketAddress
+import io.buoyant.router.http.{DiscardProxySideEffects, ForwardedFilter, MethodAndHostIdentifier, StripConnectionHeader}
```
olix0r: super minor: can yiou please sort imports without breaking out sections.

From [Effective Scala](https://twitter.github.io/effectivescala/)
> _Sort import lines alphabetically_
> This makes it easy to examine visually, and is simple to automate.

IntelliJ, etc have settings to configure this.

router/http/src/main/scala/io/buoyant/router/http/DiscardProxySideEffects.scala
```diff
@@ -0,0 +1,38 @@
+package io.buoyant.router.http
+
+import java.net.URI
+
+import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack}
+import com.twitter.finagle.http.{Request, Response}
```
olix0r: sort without breaking...

router/http/src/test/scala/io/buoyant/router/http/DiscardProxySideEffectsTest.scala
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+    await(
+      stk.make(Stack.Params.empty)()
+    )
+  }
+
+  test("filter rewrites uri and drops Proxy-* headers for proxied requests") {
+    val req = Request(Method.Get, "http://acme.co:8080/foo;matrix=uri;/bar?x=1&y=2")
+    req.host = "acme.co:8080"
```
olix0r: This test should have a different host value so that the test validates that the URI's host value is used.

olix0r:
Sorry, just a few more nits.  Once we fix that test, I think this is good to go.

olix0r:
lgtm!

PR #554, opened 2016-07-22
linkerd docker image now on 32-bit JVM

stevej:
* Moves linkerd to the debian 32-bit docker image
* adds a static namer example for even easier local testing
* Updates README to point people towards the simple static_namer config for convenience

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,12 @@ the following command:
 $ docker run -p 4140:4140 -p 9990:9990 -v /absolute/path/to/myapp:/myapp buoyantio/linkerd:0.0.10-SNAPSHOT /myapp/linkerd.yml
 ```
 
+For local testing convenience, we supply a config that routes to a single backend on _localhost:8080_.
+
+```
+$ docker run -p 4140:4140 -p 9990:9990 -v /path/to/linkerd/linkerd/examples:/config buoyantio/linkerd:0.7.1-SNAPSHOT /config/static_namer.yaml
+```
+
 The list of image names may be changed with a command like:
 
 ```
--- a/linkerd/examples/static_namer.yaml
+++ b/linkerd/examples/static_namer.yaml
@@ -1,0 +1,9 @@
+# Routes 'default' to a single backend on localhost:8080
+routers:
+- protocol: http
+  baseDtab: |
+    /host/default => /$/inet/127.0.0.1/8080;
+    /http => /host/default;
+  servers:
+  - port: 4140
+    ip: 0.0.0.0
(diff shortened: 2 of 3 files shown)


wmorgan:
Sweeet

README.md
```diff
@@ -247,6 +247,12 @@ the following command:
 $ docker run -p 4140:4140 -p 9990:9990 -v /absolute/path/to/myapp:/myapp buoyantio/linkerd:0.0.10-SNAPSHOT /myapp/linkerd.yml
 ```
 
+We supply a config that routes to a single backend on localhost:8080 for your convenience.
+
+```
+$ docker run -p 4140:4140 -p 9990:9990 -v /path/to/linkerd/linkerd/examples:/config buoyantio/linkerd:0.7.1-SNAPSHOT /config/static_namer.yaml
```
klingerf: I find this a little bit misleading to appear in the Docker section, since the static_namer.yaml isn't actually "supplied" as part of the image -- it still needs to be volume mounted.  That makes it no different from any of the other example config files provided in `linkerd/examples/`.  In fact, without docker, there's a much more straightforward way to run this config, as follows:

```
$ ./sbt linkerd-examples/static_namer:run
```

wmorgan: I like the idea of having a one-liner "just run the docker image" in here though.

klingerf: Sure -- maybe we should just modify the command on line 247 to do that though.

klingerf:
⭐  I tested out the new 32-bit images and everything appears to be working as expected. Nice work!

README.md
stevej: Ok, I have changed the wording to make it clear this is for local testing convenience. I added this config because I wanted to test running the container with a simple static config and I felt new users would appreciate that.

Happy to make other changes.

klingerf:
⭐  Thanks for updating, changes look good to me.

pypeng:
Is there a Dockerfile for buoyantio/debian-32-bit ?

We have a problem running java using this image:

```
docker run -it --rm --entrypoint=java buoyantio/debian-32-bit
Error occurred during initialization of VM
java/lang/NoClassDefFoundError: java/lang/Object
```

klingerf:
Hey @pypeng -- the dockerfile is available here: https://github.com/BuoyantIO/debian-32-bit

When I run that command the container works as expected.

```
kl ~ $ docker run -it --rm --entrypoint=java buoyantio/debian-32-bit | head -n4
Usage: java [-options] class [args...]
           (to execute a class)
   or  java [-options] -jar jarfile [args...]
           (to execute a jar file)
kl ~ $ echo $?
0
```

Is it possible that you're modifying the container's environment in such a way (like setting `JAVA_HOME`) that would make it not load properly?

pypeng:
thx a lot.

we are still working on it. it seems to be a xfs/overlayfs/kernel(3.18) related problem.

PR #235, opened 2016-04-05
[RFC] Composed Dtabs

adleong:
We introduce the notion that slash delimited namespaces are hierarchical paths and that dtabs have a "composed" view where they are composed with each of their ancestors.  

For example, if the namespace `alex` contains the dtab `/foo => /bar` and the namespace `alex/leong` contains the dtab `/foo => /bas` then the composed view of `alex/leong` would be `/foo => /bar; /foo => /baz`.  

This allows a sort of dtab inheritance where base dentries are set up in a parent and inherited by a child.  Child dtabs must come after ancestor dtabs in the composed view so that they may "override" dentries from the ancestor.

PR #555, opened 2016-07-22
Optional recursive fetch strategy in ConsulDtabStore

Ashald:
This change gives us "composite" dtabs support with Consul.
Given the namespace `ConsulDtabStore` will fetch all the keys
below the namespace recursively and combine then in alphabetical
order of keys while dropping empty ones.
It watches for changes the whole tree of keys.

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
@@ -61,17 +61,11 @@ trait BaseApi extends Closable {
     Try(mapper.readValue[T](bytes, begin, end - begin))
   }
 
-  private[v1] def executeJson[T: Manifest](req: http.Request, retry: Boolean): Future[Indexed[T]] = {
+  private[v1] def execute[T: Manifest](req: http.Request, retry: Boolean): Future[Indexed[T]] = {
     for {
       rsp <- Trace.letClear(getClient(retry)(req))
       value <- Future.const(parseJson[T](rsp.content))
     } yield Indexed[T](value, rsp.headerMap.get(Headers.Index))
-  }
-
-  private[v1] def executeRaw(req: http.Request, retry: Boolean): Future[Indexed[String]] = {
-    Trace.letClear(getClient(retry)(req)).map { rsp =>
-      Indexed[String](rsp.contentString, rsp.headerMap.get(Headers.Index))
-    }
   }
 }
 
--- a/consul/src/main/scala/io/buoyant/consul/v1/CatalogApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/CatalogApi.scala
@@ -22,7 +22,7 @@ class CatalogApi(
   // https://www.consul.io/docs/agent/http/catalog.html#catalog_datacenters
   def datacenters(retry: Boolean = false): Future[Seq[String]] = {
     val req = mkreq(http.Method.Get, s"$catalogPrefix/datacenters")
-    executeJson[Seq[String]](req, retry).map(_.value)
+    execute[Seq[String]](req, retry).map(_.value)
   }
 
   // https://www.consul.io/docs/agent/http/catalog.html#catalog_services
@@ -37,7 +37,7 @@ class CatalogApi(
       "index" -> blockingIndex,
       "dc" -> datacenter
     )
-    executeJson[Map[String, Seq[String]]](req, retry)
+    execute[Map[String, Seq[String]]](req, retry)
   }
 
   // https://www.consul.io/docs/agent/http/catalog.html#catalog_service
@@ -55,7 +55,7 @@ class CatalogApi(
       "dc" -> datacenter,
       "tag" -> tag
     )
-    executeJson[Seq[ServiceNode]](req, retry)
+    execute[Seq[ServiceNode]](req, retry)
   }
 }
 
--- a/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
@@ -1,21 +1,58 @@
 package io.buoyant.consul.v1
 
+import java.util.Base64
+
 import com.twitter.conversions.time._
 import com.twitter.finagle.http
 import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
 import com.twitter.util._
 
-object KvApi {
-  def apply(c: Client): KvApi = new KvApi(c, s"/$versionString")
+trait KvApi {
+  def list(
+    path: String,
+    datacenter: Option[String] = None,
+    blockingIndex: Option[String] = None,
+    separator: Option[String] = None,
+    retry: Boolean = false
+  ): Future[Indexed[Seq[String]]]
+
+  def get(
+    path: String,
+    datacenter: Option[String] = None,
+    blockingIndex: Option[String] = None,
+    recurse: Boolean = false,
+    retry: Boolean = false
+  ): Future[Indexed[Seq[Key]]]
+
+  def put(
+    path: String,
+    value: String,
+    datacenter: Option[String] = None,
+    cas: Option[String] = None,
+    retry: Boolean = false
+  ): Future[Boolean]
+
+  def delete(
+    path: String,
+    datacenter: Option[String] = None,
+    cas: Option[String] = None,
+    recurse: Boolean = false,
+    retry: Boolean = false
+  ): Future[Boolean]
+
 }
 
-class KvApi(
+object KvApiV1 {
+  def apply(c: Client): KvApiV1 = new KvApiV1(c, s"/$versionString")
+}
+
+class KvApiV1(
   override val client: Client,
   override val uriPrefix: String,
   override val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
   override val stats: StatsReceiver = DefaultStatsReceiver
-) extends BaseApi with Closable {
+) extends KvApi with BaseApi with Closable {
   val kvPrefix = s"$uriPrefix/kv"
 
   // https://www.consul.io/docs/agent/http/kv.html#single
@@ -23,7 +60,7 @@ class KvApi(
     path: String,
     datacenter: Option[String] = None,
     blockingIndex: Option[String] = None,
-    separator: Option[String] = Some("/"),
+    separator: Option[String] = None,
     retry: Boolean = false
   ): Future[Indexed[Seq[String]]] = {
     val req = mkreq(
@@ -34,7 +71,7 @@ class KvApi(
       "index" -> blockingIndex,
       "dc" -> datacenter
     )
-    executeJson[Seq[String]](req, retry)
+    execute[Seq[String]](req, retry)
   }
 
   /**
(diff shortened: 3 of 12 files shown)


Ashald:
For an expanded explanation of the feature please see the change in `namerd/docs/storage.md` or for example - `io.buoyant.namerd.storage.consul.ConsulDtabStoreTest`

Ashald:
Within our infrastructure we have dozens of services that we deploy in automated way.
At any given moment of time we usually have dozens of versions of each service deployed.
All of them go through various stages of testing and eventually can get activated - promoted to production.

Given all of that our dtabs are usually structured following way (an abstract example):
1) general rules (environments, datacenters, fallbacks etc)
2) dtab for an application A
3) dtab for an application B
4) dtab for an application 3
and so on.

For each application the dtab can be just a simple mapping like `/service/A=>/service/A-version-1` or it can be more complicated if we need a canary testing/cross-datacenter fallbacks/etc.
We need to update configuration for each app during our deployment (and activation) which is fully automated and doesn't involve humans.

Having the whole Dtab as a single peace of text (that's how ConsulDtabStore and others work right now) is not very convenient for automation. This means that we need somehow parse the Dtab contents, find the right place to to edit, drop the old peace of dtab and put a new one.

With proposed optional feature for ConsulDtabStore it becomes possible to split Dtab into peaces that are assembled together in a stable and predictable way. In our case we would do something like this (another abstract example):

```
namerd/
       dtabs/
             production = "/current=>/datacenter/us-east; /fallback=>/datacenter/us-west;"  # some super-generic stuff
             production/
                        A-services = "/service=>/current"  # some other generic stuff
                        ...
                        X-app-foo = "/app/foo=>/service/foo-version-1"
                        X-app-bar = "/app/bar=>/service/bar-version-23"
                        X-app-search = "/app/search=> 0.5 * /app/search/a & 0.5 * /app/search/b "
                        X-app-search/
                                     a = "/app/search/a=>/service/bio-search-version-3"
                                     b = "/app/search/b=>/service/new-bio-search-version-1"
```

The example is totally made up but I think it gives an idea: we separate 'generic' dtabs parts into separate keys and each application is 'configured' by its own key. So during deployment/activation we just need to override a specific key and that's all.
Of course those 'app configs' can also be 'dirs' so we can configure different aspects of app version resolution by overwriting only those peaces that are required.


Given comments that I received last week I made the feature optional and also adjusted behavior so it's preserves semantics of other backends and can be freely used with `namerd` to CRUD any parts.

Ashald:
I'm working on additional tests but would love to hear your opinion on this and would appreciate code-review.

Ashald:
For convenience, here is the quote from updated readme for Consul storage:

## Consul

`io.l5d.consul`

*experimental*

Stores the dtab in Consul KV storage.  Supports the following options

* *host* -- Optional. The location of the etcd API.  (default: localhost)
* *port* -- Optional. The port used to connect to the etcd API.  (default: 8500)
* *pathPrefix* -- Optional.  The key path under which dtabs should be stored.  (default: "/namerd/dtabs")
* *recursive* -- Optional.  Whether to use recursive strategy to fetch namespaces.  (default: "false")

In `recursive` mode the storage will look not only at keys that are immediate children of the `pathPrefix`
but it will traverse the whole tree beneath recursively. The resulting dtab is composed from all non-empty
valid values beneath the given namespace (sorted in alphabetical order of keys).

This may be very useful when you need to modify your dtab in runtime in automated manner and you don't want
to bother with dealing with dtabs syntax - you can just put mutable parts into separate keys and update them as needed.  

Although `recursive` strategy may sound as something complicated at first it's actually very simple. Let's just
look at an example and see how it will work in `recusrive` and non-`recursive` modes. 

Let's assume that we configured storage `pathPrefix=/namerd/dtabs` and the KV in Consul is as below:

```
namerd/
       dtabs/
             sample = "/foo=>/bar"
             nested/
                    example = "/fiz=>/baz"
                    broken = "invalid dtab"
             mydtab = "/1st=>line"
             mydtab/
                    a = "/2nd=>/line"
                    a/
                      a = "/3rd=>/line"
                      b = "/4th=>/line"
                    b = "/5th=>/line"
                    c/
                      1 = "/6th=>/line"
                    empty = ""
```

If we will list namespaces in non-recursive mode we will see just 2 of them (with their Dtabs given in brackets):
* `sample` (`"/foo=>/bar"`)
* `mydtab` (`"/1st=>line"`)

On the other hand in recursive mode we will see slightly bigger list of namespaces:
* `sample` (`"/foo=>/bar"`)
* `nested/` (`"/fiz=>/baz"`)
* `nested/example` (`"/fiz=>/baz"`)
* `nested/broken` (`""`)
* `mydtab` (`"/1st=>line"`)
* `mydtab/` (`/1st=>/line;/2nd=>/line;/3rd=>/line;/4th=>/line;/5th=>/line;/6th=>/line;`)
* `mydtab/a` (`"/2nd=>/line"`)
* `mydtab/a/` (`"/3rd=>/line;/4th=>/line"`)
* `mydtab/a/a` (`"/3rd=>/line"`)
* `mydtab/a/b` (`"/4th=>/line"`)
* `mydtab/b` (`"/5th=>/line"`)
* `mydtab/c/` (`"/6th=>/line"`)
* `mydtab/c/1` (`"/6th=>/line"`)
* `mydtab/empty` (`""`)

So the rule is simple: everything that doesn't end with `/` is a key that can has some value.
Everything that ends with `/` is a 'virtual namespace' that aggregates other keys.
 
With regards to `namerd` API there is only 1 restriction - you cannot create/update 'virtual namespaces'.
But you can read and delete them (removal of a 'virtual namespace' will recursively remove everything beneath it).

olix0r:
Thanks for submitting this, @Ashald.

This is a neat approach, but I'd like to find a more general purpose solution to these problems that is exposed explicitly via the API.  The proposed approach strikes me as a bit complex.  I don't think that subtleties like trailing slashes or lexical sorting should have substantial semantic meaning in the API.

Let's take a step back and enumerate the problems we are trying to solve:

#### Dtab Syntax
It sounds like the root problem is that you don't want to have to write dtabs as plaintext and you'd like some object notation to describe dtabs and name trees (so that, for instance, it is easier to mutate them programmatically). This is a great idea. To address this, I think it would be preferable to add additional encodings to namerd's HTTP API to accept & emit JSON nametrees.  (Similarly to what is done by the delegator API -- in fact, I think the API already supports JSON encodings but doesn't sufficiently encode nametrees?). This would impose no constraints on the storage layer; and makes it easier to build generic tools on namerd.

#### Dtab Composition
I think there needs to be a way to have an application use multiple dtabs so that it's easy to change one part of routing policy without creating unintended consequences; but I feel strongly that this behavior is made explicit by the API. @adleong has done some design exploration on this front which seems really promising.  We're being careful about introducing this, since we're hoping that we can begin to stabilize the namerd API over the next few months, and so we have been reticent to introduce a public API only to have to break it shortly thereafter.  We'll keep you in the loop on this as it progresses.

Ashald:
Hey @olix0r , thanks for your feedback.

With regards to *Dtab Syntax* - it looks like we have a little miscommunication there. Although what you described sounds like a great feature it's not on the list of our requirements.

Our real requirement is being able to decompose a Dtab into set of independent peaces so that each of them can be edited separately. 

> I think there needs to be a way to have an application use multiple dtabs

Can you explain this in more details please?
Is it about using different namespaces for different services in runtime without prior configuration? 

Because with the proposed feature whenever we need to add a service we don't have to pre-configure anything - just put an extra key (or in terms of `namerd` API - create a new 'nested' namespace) with few Dentries in it and we're done. What also important is that we neither introduce any new syntax (e.g., how to define 'composed' dtabs) nor we break the `namerd` API semantics - assuming that namespaces are opaque strings for `namerd` with recursive mode enabled you can do exactly the same set of CRUD operations.

> @adleong has done some design exploration on this front which seems really promising

Are you talking about #235 or something else? The #235 really looks interesting but I think it's clear that it solves a totally different problems compared to our use case/this PR.

> We're being careful about introducing this, since we're hoping that we can begin to stabilize the namerd API over the next few months, and so we have been reticent to introduce a public API only to have to break it shortly thereafter. 

That's really sounds great. And we do care about APIs stability a lot as well - that's exactly the reason why we tried our best to implement the proposed feature in least intrusive way (making it optional and keeping the existing API semantics) hoping that you can consider it at least as a temporary solution for a use-case such as ours unless `namerd` will get some even better way to deal with it (which might take more time after API stabilizes).

I hope this clarifies our use cases a bit more.

In our situation, we need to have a solution for the problem we described. We would gladly prefer some generic solution on the overall `namerd` API level but since it doesn't provide such an option we though about _an optional non-intrusive feature that's 100% compatible with the existing API_. Our assumption is that it can be dropped in the future when the problem would be solved on `namerd` level and an optional feature in _experimental_ backend shouldn't hurt.

I understand that you hesitant about this feature since you have much broader perspective than me but I also would like to understand what are the real stoppers here.

kmt:
First, thanks @olix0r for looking into this, highly appreciated!

My thoughts are that while "Dtab Syntax" perhaps partially helps our use case (well, at least outsources the parsing to a json parser, after which you'd just navigate a structure, modify a node and serialize again), the "Dtab composition", at least as described in @adleong 's RFC sounds like a different feature from what we need.

To be clear about our feature request: given that we'd be dealing with a large number of frequently deployed services, we'd like to be able to update only the part of the Dtab relevant to the service being currently deployed, rather than read/deserialize/modify/serialize/write the whole Dtab.  Is "Dtab composition", the way you envision it, going to address that?  If yes, we'd be happy, regardless of how exactly it's done technically.

Ashald:
The more I think about it the more I realize that I failed to describe our use-case properly.

Within our environment we have _a lot_ of services. Right now we run about 100 services registered in Consul be we also have ~7K legacy services that pending migration to Consul.
Given that we want to use `linkerd` / `namerd`for service routing and, obviously, configure it using `Dtab`.

Another thing that I need to mention that we're doing a lot of deployments (which are fully automated): with 100 services being currently running we have about 50 deployments per day.

With this approach the resulting `Dtab` can be pretty huge and so we have few concerns:
1) getting all of the data each time we need to change the little peace of it and then putting it back has a lot of overhead
2) with concurrent updates we need be careful version checks so we won't discard other concurrent changes (which would be a huge problem with frequent deployments with 7K services)
3) we will be veeery close to Consul KV limit for max key's value: it's 512k - having ~7K services we get about 75 bytes per service config in `Dtab`

All of that being said, I believe that our scale is by far is not the biggest in the world and other people using Consul at such or greater scale will have same concerns.

Analyzing the situation we figured out that it'd be great if we could split `Dtab` into tiny peaces that can be edited separately (reducing concurrency), reduce overhead traffic and discard any concerns about size restrictions in Consul KV.

Looking at `namerd` and its API we made a few assumptions:
1) from `namerd` perspective namespace names are opaque strings - given the diverse nature of backends that seems to be the case as otherwise it can limit integration options with some backends
2) using `namerd` API one should be able to do full CRUD with any namespace exposed by a backend
3) any `namerd` backend shouldn't rely/require anything else except what's defined in API/trait
4) the proposed feature might be an 'advanced mode' that must be enabled separately (an analogy to `experimental` flag - confirm that you know what you're doing)
5) this is an optional feature that may be obsoleted in future and as well removed given the fact `namerd` will get a 'generic' way to deal with such a use-case

Given all of that we implemented a proposed feature and presented it within the current PR. I think you see how our assumptions are reflected into the implementation but let me know if any of that requires additional clarification.
Hope this use-case description will make more sense the previous one. :)

Issue #518, opened 2016-07-05
namerd - "session expired"


JonathanBennett:
Error message gets thrown over and over in the logs and gets raised as a HTTP 502 to the client.

`E 0704 16:09:06.818 THREAD22: service failure
Failure(session expired, flags=0x100000000)
    with NoSources
    at com.twitter.finagle.NoStacktrace(Unknown Source)`

From @olix0r : 
fwiw we it would ​_seem_​ like a zk exception:
 ```:; git grep -lF session\ expired
finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/ZkSession.scala```
but we probably should try to prevent that from bubbling up into the request flow if we can

adleong:
Hey @JonathanBennett, I'm having trouble reproducing this.  As I understand it, namerd's zk client sends a heartbeat to keep the zk session alive.  Do you have any idea what could cause the session to expire?  By default, the client requests a 10 second session timeout.

Is the error message you pasted from linkerd or namerd?  If it's from linkerd, is there anything interesting in the namerd logs?

JonathanBennett:
Hey,
I don't have a reason why zk would have failed to respond to a session. Rather annoyingly due to our log rotation in dev I don't have the logs for this as it looks to have been failing for quite a while so I'll be on the lookout to try and reproduce it; only thing I can think of is maybe ZK was underload and therefore failed to respond but I would imagine that as unlikely.

In such a case how should namerd recover? Should I be restarting namerd or should it be more resilient than this in failing from a zk timeout? From my logs I do have (about 4 days of failures over a weekend) I just see all lookups/sets failing with an error similar to the following:

``` 
E 0701 01:47:17.364 THREAD25: binding name /http/1.1/GET/zeppelin.xxxxx.co
java.lang.Exception: session expired
	at com.twitter.finagle.serverset2.buoyant.ZkSession$$anonfun$watchedOperation$1$$anonfun$com$twitter$finagle$serverset2$buoyant$ZkSession$$anonfun$$loop$2$1$$anonfun$2.apply(ZkSession.scala:162)
	at com.twitter.finagle.serverset2.buoyant.ZkSession$$anonfun$watchedOperation$1$$anonfun$com$twitter$finagle$serverset2$buoyant$ZkSession$$anonfun$$loop$2$1$$anonfun$2.apply(ZkSession.scala:142)
	at com.twitter.util.Witness$$anon$17.notify(Event.scala:434)
[...]

PR #556, opened 2016-07-25
linkerd should use last known good data if it get errors from namerd

adleong:
# Problem

If namerd encounters an error during binding, linkerd will go into an error state and fail to bind requests.

# Solution

If linkerd gets error updates from namerd, it should not update its state and keep the last known good data.

This is a partial solution for #518.  It doesn't fix the zk session expired issue, but it does mean that linkerd will continue to function with last known good data in the case of such errors.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## x.x.x
 
 * Add support for tags in the `io.l5d.consul` namer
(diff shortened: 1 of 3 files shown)


olix0r:
lgtm

olix0r:
please update CHANGES

PR #557, opened 2016-07-26
Add maxResponseKB to http client configuration

olix0r:
By default, Finagle and Netty set a 10MB limit for an unchunked response. This
is done to protect against having to buffer arbitrarily large payloads, which
can lead to performance degradation.

This change introduces a new configuration option for HTTP clients, allowing
this limit to be increased on a per-router basis.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 * Add support for tags in the `io.l5d.consul` namer
 * linkerd should use last known good data if it get errors from namerd
 * Fix exceptions when k8s namer encounters unexpected end of stream #551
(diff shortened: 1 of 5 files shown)


adleong:
could you add this to CHANGES.md and the config docs?  Maybe with a warning in the docs that increasing this limit can leads to performance issues.

seekely:
Thanks for adding the option!

wmorgan:
@seekely Note that using this flag to handle large responses is not "great" for performance reasons (they take up a big blob of memory). Using chunked encoding is better in the long run.

adleong:
:star: 🏥

PR #561, opened 2016-07-30
Allow protocol-specific parameters to be inherited on servers

olix0r:
Currently, only a few non-protocol-specific parameters are shared from a router
to its servers.  This makes it impossible to set router-levelcodec parameters
(e.g. MaxChunkSize).

Add a `configureServer` method to ProtocolInitializer so that each protocol may
provide configuration-sharing logic as needed.  By default, we share HTTP codec
parameters from routers to servers.

<hr />
Manually tested locally: uploading a 38M file fails with a 413 unless `maxRequestKB` is set appropriately.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -38,10 +38,9 @@ class HttpInitializer extends ProtocolInitializer.Simple {
   }
 
   /**
-   * Apply the router's codec coonfiguration applies to the server.
+   * Apply the router's codec configuration parameters to a server.
    */
-  override protected def configureServer(router: Router, server: Server): Server = {
-    println(s"configuring server with ${router.params}")
+  override protected def configureServer(router: Router, server: Server): Server =
     super.configureServer(router, server)
       .configured(router.params[hparam.MaxChunkSize])
       .configured(router.params[hparam.MaxHeaderSize])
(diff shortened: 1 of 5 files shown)


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -37,6 +37,21 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .configured(RoutingFactory.DstPrefix(Path.Utf8(name)))
   }
 
+  /**
+   * Apply the router's codec coonfiguration applies to the server.
+   */
+  override protected def configureServer(router: Router, server: Server): Server = {
+    println(s"configuring server with ${router.params}")
```
gtcampbell: println

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -37,6 +37,21 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .configured(RoutingFactory.DstPrefix(Path.Utf8(name)))
   }
 
+  /**
+   * Apply the router's codec coonfiguration applies to the server.
```
gtcampbell: typo: "coonfiguration" (and I think the "applies" may be extraneous, unless I misunderstand its intent?)

gtcampbell:
:star: is it possible to write an automated test for this?

olix0r:
whoops, thanks for those catches (that's what i get for pushing a fix from the bus ;p)

as for tests -- it's certainly possible to test but about 3x as much work as the fix.  I'll ry to get tests together...

olix0r:
⭐  self-review

PR #564, opened 2016-08-04
Resolve endpoint for namerd HTTP control interface

Ashald:
Usage example:

```
$ curl -s "$NAMERD_URL/api/1/resolve/default?path=/http/1.1/GET/users"
{"addrs":[{"addr":"192.168.10.23:8080","metadata":{}}],"metadata":{"authority": "users.service.dc1.consul.acme.co"}}%
```

Under the hood, combines `bind` and `addr` and uses `DelegateApiHandler.Codec` to serialize result into JSON.

### DIFF
--- a/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
+++ b/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
@@ -3,10 +3,11 @@ package io.buoyant.namerd.iface
 import com.fasterxml.jackson.databind._
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
-import com.google.common.cache.{CacheLoader, RemovalNotification, RemovalListener, CacheBuilder}
+import com.google.common.cache.{CacheBuilder, CacheLoader}
 import com.twitter.finagle.Name.Bound
 import com.twitter.finagle.http._
 import com.twitter.finagle.naming.NameInterpreter
+import com.twitter.finagle.util.{Rng, Drv}
 import com.twitter.finagle.{Status => _, _}
 import com.twitter.io.Buf
 import com.twitter.util._
@@ -120,6 +121,10 @@ object HttpControlService {
     val prefix = s"$apiPrefix/addr"
   }
 
+  object ResolveUri extends NsPathUri {
+    val prefix = s"$apiPrefix/resolve"
+  }
+
   object DelegateUri extends NsPathUri {
     val prefix = s"$apiPrefix/delegate"
   }
@@ -167,6 +172,8 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
       handleGetBind(ns, path, req)
     case AddrUri(Some(ns), path) =>
       handleGetAddr(ns, path, req)
+    case ResolveUri(Some(ns), path) =>
+      handleGetResolve(ns, path, req)
     case DelegateUri(Some(ns), path) =>
       handleGetDelegate(ns, path, req.params.get("dtab"))
     case DelegateUri(None, path) =>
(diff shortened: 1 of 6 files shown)


olix0r:
Thanks, @Ashald. 

Isn't the delegate API sufficient for this? It exposes a parseable object that includes all of the information needed to determine what the bound addr(s) apply (we do this is in the dtab playground UI, for instance)

How do you expect to use this API endpoint?  It doesn't seem like the output is all that parseable...

Ashald:
Thanks for looking at the PR.

The issue with `delegate` API is that it returns the whole tree and consumers of the API will have to implement the logic to interpret it.
Although that might be suitable for the needs you mentioned it will make consuming the API harder.
We though about using `bind` + `addr` but that doesn't work for few reasons:
1) if `bind` returns and `alt` (like `0.50*/#/io.l5d.consul/dc1/myservice & 0.50*/#/io.l5d.consul/dc2/myservice`) that its output cannot be directly fed to `addr` endpoint
2) that would require 2 calls and will make workign with the API harder

As I mentioned this before, in #linkerd during discussion with Alex, and during our chat - the use-case for this is to allow apps that cannot benefit from linkerd (due to various reasons - legacy code, protocols not supported by linkerd and so on) to use namerd as a generic resolver. IIRC, you and William agreed that it sounds like a valid usecase.

As for output format - yeah, I agree that it's by far not the best one. I just used the same idea as in `addr` endpoint. Now looking again, as per your comment, at `delegate` API output I see that we can re-use the same serialization logic from `DelegateApiHandler.getDelegateRsp` for this proposed endpoint and we will get a pretty easy parseable JSON.

olix0r:
Thanks for providing more context---this really helps to have all of this captured in the PR so that it's easy to reference.  I have a better understanding now.

I wasn't aware that the addr API worked like that! That's unfortunate.  We should fix it... I think we'll need to take a pass at cleaning up this API in general, but let's leave that for another issue.

I agree that the API endpoint makes sense.  Since we're returning a non-trivial data structure, though, I'd prefer that we use a proper serialization format.  Hopefully some of the DelegateApi structures can be reused/copied to satisfy this.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -417,6 +424,48 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
...
+        }
+    }
+  }
+
+  private[this] def handleGetResolve(ns: String, path: Path, req: Request): Future[Response] = {
+    val extraDtab = req.params.get("dtab")
+
+    val activity = getBind(ns, path, extraDtab).flatMap { tree =>
+      flattenTree(tree) match {
+        case p: Some[Path] => addrCache.get((ns, p.get))
```
adleong: I think this line should be `case Some(p) => addrCache.get((ns, p))`

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -417,6 +424,48 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
...
+        case None => Activity.value(Addr.Neg)
+      }
+    }
+
+    if (isStreaming(req)) {
+      streamingResp(activity.values) { addr => Buf.Utf8(addr.toString + "\n") }
```
adleong: I think you should use the `renderAddr` method here.  As @olix0r mentioned, our serialization format isn't great right now, but if you use `renderAddr` then we can add proper serialization in one place.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -417,6 +424,48 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
...
+    } else {
+      activity.toFuture.map { addr =>
+        // maybe we should return 404 on Addr.Neg and 302 with Retry-After header on Addr.Pending?..
+        val rsp = Response()
+        rsp.content = Buf.Utf8(addr.toString + "\n")
```
adleong: I think `renderAddr` can be used here too.

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
Ashald: Thanks!

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
Ashald: For now I changed it to call to `DelegateApiHandler.Codec` but I see #575 - once it's merged (assuming it would be merged before this one) I'd be happy to change it to updated `renderAddr` that makes the same thing.

Ashald:
@olix0r @adleong can you please take another look at this PR? In first place I'm interested in the interface part - are you OK with calling the endpoint `resolve`? Are you OK with such output format? And are you OK with parameters it takes and their names?

Once you OK with all those things I'd like to ask you whether I need to improve code so it could be merged. Thanks!

adleong:
This approach makes sense to me, and I'm okay with the name `resolve`.  The output format should be the same as addr, which is getting a facelift in #575.  The parameters (`dtab` and `path`) match the format of the other endpoints, so I think that's good.

Ashald:
@adleong @olix0r I rebased the PR on master after #575 was merged - please let me know what you think about it now

adleong:
:star: I think this turned out quite nicely

Ashald:
As per agreement with @adleong I pushed an extra commit to fix `JsonAddr.mk` behavior and include `Address.Inet` meta into the JSON representation.

PR #569, opened 2016-08-09
modify the delegate api to provide bound addresses

siggy:
problem: the http /delegate endpoint would not provide bound addresses
unless a namespace parameter was provided.

solution: modify the behavior of /delegate to use available namers when
a namespace is not provided.

note: this does not change the behavior of the /delegate/[namespace]/
api, though it may be prudent in the future to harmonize or converge
these code paths.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## x.x.x
 
+* Modify namerd's `/delegate` http endpoint to return bound names #569.
+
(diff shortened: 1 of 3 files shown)


adleong:
:star: 🍰 great!

adleong:
oh, almost forgot: would you mind adding something to CHANGES.md for this?

PR #573, opened 2016-08-12
Add filesystem based name interpreter

adleong:
The file-system interpreter (io.l5d.fs) resolves names via the configured namers, just like the default interpreter, but also uses a dtab read from a file on the local file-system.  The specified file is watched for changes so that the dtab may be edited live.


olix0r:
LGTM!  Would you mind adding something to the acceptance configuration for this?

PR #574, opened 2016-08-12
Dashboard: add toggling to the router clients 

### LINKED ISSUE #572: Dashboard becomes slow with many clients
In my setup, I have 130+ backend services, and each of these is represented by a separate client in linkerd. At this number of clients, the dashboard slows to the point of being unusable. I'm not a front-end engineer and I haven't investigated this hypothesis, but I suspect this is due to the number of charts being updated in real-time.

rmars:
Problem:
With large numbers of clients, there are many listeners registered to the metrics collector (each client registers with the metrics collector when initialized). After each stats fetch, the metrics collector goes through its list of handlers and calls `listener.handler`. That's a lot of work.

Solution:
* Add toggling to the router clients. When a section is toggled, register/deregister the listener with the metrics collector. This lets us have some control over how much work is being done.
* All clients under a router now collapse if there are more than 6 clients.
* Collapsed clients remove themselves from the list of metricsListeners on metricsCollector,
which should result in much less work being done for a large number of clients.

Fixes #572

screen shot 2016-08-12 at 3 48 23 pm

screen shot 2016-08-12 at 3 29 50 pm

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/css/dashboard.css
+++ b/admin/src/main/resources/io/buoyant/admin/css/dashboard.css
@@ -236,6 +236,11 @@ nav ul.navbar-nav {
   padding: 8px 0;
 }
 
+.client-toggle {
+  font-size: 14px;
+  cursor: pointer;
+}
+
 .router-latencies-container {
   padding-top: 18px;
 }
--- a/admin/src/main/resources/io/buoyant/admin/js/router_client.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/router_client.js
@@ -83,17 +83,42 @@ var RouterClient = (function() {
     return summary;
   }
 
-  return function (metricsCollector, routers, client, $metricsEl, routerName, clientTemplate, metricPartial, $chartEl, colors) {
+  return function (metricsCollector, routers, client, $metricsEl, routerName, clientTemplate, metricPartial, $chartEl, colors, $toggleLinks, shouldExpandInitially) {
     template = clientTemplate;
     Handlebars.registerPartial('metricPartial', metricPartial);
     var clientColor = colors.color;
     var latencyLegend = createLatencyLegend(colors.colorFamily);
     var metricDefinitions = getMetricDefinitions(routerName, client.label);
 
-    renderMetrics($metricsEl, client, [], [], clientColor);
-    var chart = SuccessRateGraph($chartEl, colors.color);
+    var $expandLink = $toggleLinks.find(".client-expand");
+    var $collapseLink = $toggleLinks.find(".client-collapse");
 
-    var metricsHandler = function(data) {
+    renderMetrics($metricsEl, client, [], [], clientColor);
+    var chart = SuccessRateGraph($chartEl.find($(".client-success-rate")), colors.color);
+
+    // collapse client section by default (deal with large # of clients)
+    if(shouldExpandInitially) {
+      toggleClientDisplay(true);
+    } else {
+      toggleClientDisplay(false);
+    }
+
+    $expandLink.click(function() { toggleClientDisplay(true); });
+    $collapseLink.click(function() { toggleClientDisplay(false); });
+
+    function toggleClientDisplay(expand) {
+      if (expand) {
+        metricsCollector.registerListener(metricsHandler, getDesiredMetrics);
+      } else {
+        metricsCollector.deregisterListener(metricsHandler);
+      }
+      $metricsEl.toggle(expand);
+      $chartEl.toggle(expand);
+      $collapseLink.toggle(expand);
+      $expandLink.toggle(!expand);
+    }
+
+    function metricsHandler(data) {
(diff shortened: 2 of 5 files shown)


admin/src/main/resources/io/buoyant/admin/css/dashboard.css
```diff
@@ -236,6 +236,10 @@ nav ul.navbar-nav {
   padding: 8px 0;
 }
 
+.client-toggle {
+  font-size: 14px;
```
siggy: would you mind adding `cursor: pointer;` ?

rmars: ooh yes, good call

admin/src/main/resources/io/buoyant/admin/js/router_client.js
```diff
@@ -83,17 +83,47 @@ var RouterClient = (function() {
...
+      metricsCollector.deregisterListener(metricsHandler);
+      toggleClientDisplay(false);
+    }
+
+    function toggleClientDisplay(expand) {
+      $metricsEl.toggle(expand);
+      $chartEl.toggle(expand);
+      $collapseLink.toggle(expand);
+      $expandLink.toggle(!expand);
+    }
```
siggy: tioli: consider moving the register/deregister calls into `toggleClientDisplay` and removing show/hideClient methods.

if this could cause handlers to be registered more than once, we could modify `registerListener` to check for existence before adding.

if that causes `registerListener` to be expensive, maybe we could change its underlying implementation to a hash?

realizing this may be a lot of work as i'm typing this, just some things to think about.

siggy:
👍 ⭐ looks great would expand again !

admin/src/main/resources/io/buoyant/admin/js/router_client.js
rmars: oo, interesting. would that consolidation be less confusing?

originally I separated it into showClient/hideClient to make the steps clearer:
* handle the listener
* update the UI

additionally, in the initialization section in line 100, I didn't want to have to call deregisterListener on a listener that we know doesn't exist if on initialization we don't want to register that listener.

what do you think?

siggy: hmm, yeah, that all makes sense. my thinking was that if the state of client display is tightly coupled to register/deregister, then be maximally explicit about that. if i were altering this code, i wouldn't be sure whether to call toggleClientDisplay directly or show/hide client. by removing that extra level of indirection, i wouldn't have a choice, which i think is a good thing.

all that said, this is a minor thing, and i understand the intent. i think it's totally cool to leave as is, but thank you for the explanation!

admin/src/main/resources/io/buoyant/admin/js/router_client.js
```diff
@@ -83,17 +83,47 @@ var RouterClient = (function() {
...
+
     renderMetrics($metricsEl, client, [], [], clientColor);
-    var chart = SuccessRateGraph($chartEl, colors.color);
+    var chart = SuccessRateGraph($chartEl.find($(".client-success-rate")), colors.color);
+
+    // collapse client section by default (deal with large # of clients)
+    if(shouldExpandInitially) {
+      showClient();
+    } else {
+      toggleClientDisplay(false);
```
esbie: I actually agree with siggy here. This would be clearer as `if(shouldExpandInitially) showClient() else hideClient()`

Deregistering something that's not registered is a no-op. What's the worry here?

admin/src/main/resources/io/buoyant/admin/js/router_client.js
rmars: ah thanks for explaining, that's helpful. 
the extra indirection probably isn't worth it. will move!

admin/src/main/resources/io/buoyant/admin/js/router_client.js
rmars: Yep indeed it's a no-op, I just didn't want to go through the list of listeners when I already know we won't find the item.

Agreed that it would read more clearly in the if/else if it is just show vs hide.

esbie: If you're worried about the list of listeners' performance, I bet an implementation based off of  `addEventListener`/`dispatchEvent` would be more performant

esbie:
⭐  I really like this solution, can definitely be shipped as is

admin/src/main/resources/io/buoyant/admin/js/router_client.js
rmars: Yeah! I do think we should get to it at some point. I think I'll make a ticket to keep track of this for later.

esbie:
lgtm

siggy:
👍 ⭐ thanks!

olix0r:
lgtm

PR #576, opened 2016-08-15
Consul API refactoring

Ashald:
This is the followup PR after #555 was declined.

As discussed previously with @olix0r this PR provides few improvements and optimizations for Consul API implementation making it more efficient and extensible.

The main changes are:
* `CatalogNamer` / `ConsulDtabStore` now can be configured with an extra `token` parameter that would set an auth token HTTP header for each request (done with a filter)
* `ConsulDtabStore` now can be configured with an extra `datacenter` parameter that would instruct Consul to forward requests to given datacenter rather than use the datacenter of local agent
* `io.buoyant.consul.v1.KvApi` - got rid of `list` method. This method is actually doesn't exist and just and alternative mode of operation for `get` method. The same result could be achieved by using `get` in recursive mode. Bonus: less code. :)
* `ConsulDtabStore` - remove code duplication (now it's less than 90 LOC), properly handle long-polling cancellation (get rid of tracebacks in logs when requests are canceled),  don't fail on malformed dtabs (report them as empty rather than crash), more tests
* `HttpControlService` / `DtabListHandler` - sort namespaces before showing them to user - no significant impact but just removes 'items jumping' in UI upon refresh which is really annoying when working with Named Admin UI

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
@@ -1,41 +1,52 @@
 package io.buoyant.consul.v1
 
+import java.util.Base64
+
 import com.twitter.conversions.time._
 import com.twitter.finagle.http
 import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
 import com.twitter.util._
 
-object KvApi {
-  def apply(c: Client): KvApi = new KvApi(c, s"/$versionString")
-}
+trait KvApi {
 
-class KvApi(
-  override val client: Client,
-  override val uriPrefix: String,
-  override val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
-  override val stats: StatsReceiver = DefaultStatsReceiver
-) extends BaseApi with Closable {
-  val kvPrefix = s"$uriPrefix/kv"
-
-  // https://www.consul.io/docs/agent/http/kv.html#single
-  def list(
+  def get(
     path: String,
     datacenter: Option[String] = None,
     blockingIndex: Option[String] = None,
-    separator: Option[String] = Some("/"),
+    recurse: Boolean = false,
     retry: Boolean = false
-  ): Future[Indexed[Seq[String]]] = {
-    val req = mkreq(
-      http.Method.Get,
-      s"$kvPrefix$path",
-      "keys" -> Some(true.toString),
-      "separator" -> separator,
-      "index" -> blockingIndex,
-      "dc" -> datacenter
-    )
-    executeJson[Seq[String]](req, retry)
-  }
+  ): Future[Indexed[Seq[Key]]]
+
+  def put(
+    path: String,
+    value: String,
+    datacenter: Option[String] = None,
+    cas: Option[String] = None,
+    retry: Boolean = false
+  ): Future[Boolean]
+
+  def delete(
+    path: String,
+    datacenter: Option[String] = None,
+    cas: Option[String] = None,
+    recurse: Boolean = false,
+    retry: Boolean = false
+  ): Future[Boolean]
+
+}
+
+object KvApiV1 {
+  def apply(c: Client): KvApiV1 = new KvApiV1(c, s"/$versionString")
+}
+
+class KvApiV1(
+  val client: Client,
+  val uriPrefix: String,
+  val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+  val stats: StatsReceiver = DefaultStatsReceiver
+) extends KvApi with BaseApi with Closable {
+  val kvPrefix = s"$uriPrefix/kv"
 
   /**
    * Get the key value
@@ -48,16 +59,17 @@ class KvApi(
     path: String,
     datacenter: Option[String] = None,
     blockingIndex: Option[String] = None,
+    recurse: Boolean = false,
     retry: Boolean = false
-  ): Future[Indexed[String]] = {
+  ): Future[Indexed[Seq[Key]]] = {
     val req = mkreq(
       http.Method.Get,
       s"$kvPrefix$path",
-      "raw" -> Some(true.toString),
       "index" -> blockingIndex,
-      "dc" -> datacenter
+      "dc" -> datacenter,
+      "recurse" -> (if (recurse) Some(recurse.toString) else None)
     )
-    executeRaw(req, retry)
+    execute[Seq[Key]](req, retry)
   }
 
   /**
@@ -81,7 +93,7 @@ class KvApi(
       "dc" -> datacenter
     )
     req.setContentString(value)
-    executeJson[Boolean](req, retry).map(_.value)
+    execute[Boolean](req, retry).map(_.value)
   }
 
   /**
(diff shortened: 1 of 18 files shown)


consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -95,15 +107,30 @@ class KvApi(
...
+      "recurse" -> (if (recurse) Some(recurse.toString) else None),
       "dc" -> datacenter
     )
-    executeJson[Boolean](req, retry).map(_.value)
+    execute[Boolean](req, retry).map(_.value)
   }
+}
+
+object Key {
+  def apply(key: String, value: String): Key = Key(Some(key), Some(Base64.getEncoder.encodeToString(value.getBytes)))
```
adleong: It's a little confusing that `Key(k, v)` expects a unencoded value but `Key(Some(k), Some(v))` expects a base64 encoded value.  Consider naming this something other than `apply` (perhaps `mk`?).  This reinforces the idea that a `Key` stores base64 encoded values, but it has a helper method to construct it from an unencoded value.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -95,15 +107,30 @@ class KvApi(
...
+}
 
+case class Key(
+  Key: Option[String],
```
adleong: lowercase `key` and `value` please.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -95,15 +107,30 @@ class KvApi(
...
+  Value: Option[String]
+) {
+  lazy val decoded: Option[String] = Value match {
```
adleong: You can use `value.map` here

namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
```diff
@@ -50,11 +52,16 @@ case class ConsulConfig(
    */
   @JsonIgnore
   def newNamer(params: Stack.Params): CatalogNamer = {
+    val filters = if (token.isDefined)
```
adleong: it's idiomatic to avoid `Option.get`:

```
val filters = token match {
  case Some(t) => new SetHostFilter(...) andThen new SetAuthTokenFilter(t)
  case None => ...
}
```
or
```
val authFilter = token match {
  case Some(t) => new SetAuthTokenFilter(t)
  case None => Filter.identity
}
val filters = new SetHostFilter(...) andThen authFilter
```

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -1,106 +1,86 @@
...
 import com.twitter.util._
 import io.buoyant.consul.v1._
 import io.buoyant.namerd.DtabStore.{DtabNamespaceAlreadyExistsException, DtabNamespaceDoesNotExistException, DtabVersionMismatchException, Version}
 
-class ConsulDtabStore(api: KvApi, root: Path) extends DtabStore {
+class ConsulDtabStore(api: KvApi, root: Path, datacenter: Option[String] = None) extends DtabStore {
 
-  override def list(): Activity[Set[Ns]] = {
-    def namespace(key: String): Ns = key.stripPrefix("/").stripSuffix("/").substring(root.show.length)
+  def path(ns: Ns) = s"${root.show}/$ns"
```
adleong: can this be made `private[this]`?

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -1,106 +1,86 @@
...
 
-    val run = Var.async[Activity.State[Set[Ns]]](Activity.Pending) { updates =>
-      @volatile var running = true
+  def getNs(ns: Ns, index: Option[String] = None) =
```
adleong: can this be made private[this]?

adleong:
I don't think your replacement of `list` with `get` is correct.  By specifying `?keys` and `?separator`, we get a list of keys only (not values) from a single directory (not recursive).  Take a look at https://www.consul.io/docs/agent/http/kv.html#single

adleong:
Would you mind splitting this up into a PR for each of your bullet points in the description?  Some of these are obvious wins (ability to pass a dc into consul store, ability to add auth token, etc) whereas the larger refactors are more controversial.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: Got the idea - thanks, will refactor it into `mk`.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: Those fields _must_ match the case of fields in JSON produced by Consul for mapper to be able to decode it. I'll add the comment explaining that. Other Consul APIs in the same package using the same approach - that's not 'my invention' but was done like that before my first PR. I'd be happy to change that to something else if you can advice me on another way of decoding JSON objects.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: Ah, great! Thanks!

namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
Ashald: Your version looks definitely better - thanks! :)

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: Yeah, will do.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
adleong: Ah, yes, you're right.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
Ashald: Are you ok with making it `protected`? The reason why I'm asking is one of the reasons for this refactoring - when we had chat with @olix0r and @wmorgan I asked about submitting a PR to make this refactoring and make it easier to extend ConsulDtabStore so we can get our own plugin that would do #555 but without the need to re-write the whole store.

Ashald:
> I don't think your replacement of list with get is correct. By specifying ?keys and ?separator, we get a list of keys only (not values) from a single directory (not recursive). Take a look at https://www.consul.io/docs/agent/http/kv.html#single

Yeah, that's how it was originally. As I mentioned above, the idea behind this refactoring to switch to from `GET .../kv/some-key?raw` to `GET .../kv/some-key` so it would be easier to implement extensions for `ConsulDtabStore` (make API more generic).

But once I had that in place I realized that (1) `GET .../kv/namerd/dtabs/?recursive` would give us exactly the same result as `?keys`. Yeah, you're right about `separator` but:
1) it actually doesn't filter out dirs - we still need to do manual filtering
2) since namespaces expected to be simple keys the current implementation if sufficient enough
Although I can put check like `.filter(!_.Key.contains("/"))` back.

Ashald:
> Would you mind splitting this up into a PR for each of your bullet points in the description? Some of these are obvious wins (ability to pass a dc into consul store, ability to add auth token, etc) whereas the larger refactors are more controversial.

Yeah, will do.

adleong:
This has been split into smaller PRs.

PR #578, opened 2016-08-15
An option to set auth token for Consul backends

Ashald:
Both for ConsulNamer and ConsulDtabStore an optional token parameter can
be provided to set `X-Consul-Token` auth header for every request

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## x.x.x
 
 * Add file-system based name interpreter.
+* Add auth `token` parameter to Consul Namer & Dtab Store
 
 ## 0.7.3
 
--- a/consul/src/main/scala/io/buoyant/consul/SetAuthTokenFilter.scala
+++ b/consul/src/main/scala/io/buoyant/consul/SetAuthTokenFilter.scala
@@ -1,0 +1,12 @@
+package io.buoyant.consul
+
+import com.twitter.finagle.http.{Request, Response}
+import com.twitter.finagle.{Service, SimpleFilter}
+
+class SetAuthTokenFilter(token: String) extends SimpleFilter[Request, Response] {
+
+  def apply(req: Request, svc: Service[Request, Response]) = {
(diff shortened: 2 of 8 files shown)


Ashald:
@adleong as per your request in #576 I started splitting it into smaller chunks. This one is about auth token and essentially is pretty small one - could you please take another look? I changed that way how filters are composed as per your advice.

adleong:
:star: This looks great, thanks!  Would you mind adding an entry to CHANGES.md?

Ashald:
Sure! Sorry that I'm constantly forgetting about it. :(

Ashald:
I updated the commit to include `CHANGES.md` entry. Waiting on this to be merged so I can follow-up with other pieces of #576 

Thanks for the review!

olix0r:
This looks good to me, thanks!

One practical question before merging:  If this is a secret token, would it make more sense for this to be read from another file rather than encoded in linkerd configuration? I'm envisioning linkerd's config as non-sensitive (so you can share them easily, for instance), and secrets should be provided separately.  In this case, we'd change `token` to i.e. `tokenFile`.

I don't feel strongly--whatever seems most practical.

Ashald:
That's interesting point. There are lot of things that we might consider on the 'security' side of the things - including the fact that the token would be passed in plain text along with request unless HTTPS is used.

While working on this PR I made few assumptions that are common for environment I have to deal with - such us dedicated cluster of instances `namerd` (therefore you don't need to pass the token around `linkerd` instance configs) as well as automated deployment with config files filled with token stored in secure store and a service account to run the process together with restricted access on `namerd` config file.

At the end of the day that's up to you: would you prefer `tokenFile` - let me know and I will update the PR. Or we can proceed with a simple `token` field and think about `tokenFile` option if there will be demand for it.

olix0r:
Yeah, I think it's fine to start with `token` and introduce `tokenFile` later. :star::shipit:

PR #579, opened 2016-08-15
Sort namespaces before showing them to user

Ashald:
Wouldn't hurt but removes items jumping in UI - that's super annoying when working a lot with Namerd Admin UI.


Ashald:
@adleong another piece of #576 - a micro change to make Namerd UI a bit more convenient. Let me know what you think please.

adleong:
:star: Makes sense!

Ashald:
What would be the next step for this PR? Merge or we're waiting for an additional review?

Thanks!

PR #581, opened 2016-08-16
Add dyn binding timeout for naming

adleong:
# Problem

If a namer stays in the Pending state forever, no timeout applies and requests to linkerd hang.

# Solution

Apply the binding timeout to DynBindingFactory.

# Result

If a namer stays Pending forever, requests to linkerd will timeout after the binding timeout.


olix0r:
lgtm!

klingerf:
⭐  Makes sense to me.

PR #583, opened 2016-08-16
Path identifier should only parse as many segments as requested

### LINKED ISSUE #563: label char expected but '!' found at <url-to-handle>
While using io.l5d.path indentifier, linkerd is evaluating the complete path and throwing an error when it encounters a problematic character.

In my precise case - a rich frontend app is using "!" as a delimiter to construct its internal bread-crumb location corresponding to the shown tab/window. When a user hits page-refresh, linkerd error shows up:

`label char expected but '!' found at /production/dev-team/service/my-app/projects/789/[!]/list/something`

Is it possible that linkerd ignores the "validity" of the rest of the url (after stripping and processing the X segments it needs to do the forwarding?)

Thanks!

My linkerd configuration is below:

```
admin:  
  port: 9990
  
namers:
- kind:         io.l5d.marathon
  experimental: true           
  prefix:       /io.l5d.marathon  
  host:         marathon.mesos 
  port:         8080           
  ttlMs:        300            
  
routers:
- protocol: http               
  baseDtab: |                  
    /http => /#/io.l5d.marathon;    
  dstPrefix: /http             
  identifier:                  
    kind: io.l5d.path          
    segments: 4
    consume: true              
  httpAccessLog: logs/access.log  
  label: int                   
  servers:
  - port: 4140                 
    ip: 0.0.0.0
```

adleong:
Fixes #563

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
@@ -13,16 +13,15 @@ case class PathIdentifier(
 ) extends RoutingFactory.Identifier[http.Request] {
 
   def apply(req: http.Request): Future[(Dst, http.Request)] =
-    Path.read(req.path) match {
+    req.path.split("/").drop(1) match {
       case path if path.size >= segments =>
         if (consume) {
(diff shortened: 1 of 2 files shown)


gtcampbell:
👍

olix0r:
lgtm!

PR #585, opened 2016-08-17
Allow 'empty tags' in ConsulNamer with includeTag=true

Ashald:
When `includeTag` mode is enabled there is no way to request service
without filtering by tags. This change makes behavior of
`/dc1/_/my-service` with `includeTag=true` equal to `/dc1/my-service` with `includeTag=false`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## x.x.x
 
 * Add file-system based name interpreter.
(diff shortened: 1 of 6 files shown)


Ashald:
@adleong can you take a look at this PR as well please?

Ashald:
It turned out that it's doable by adding the same namer with another prefix and config.

PR #586, opened 2016-08-17
Add multiGet to Consul KvApi

Ashald:
The main difference is that `mulltiGet` always returns a list (even when single element is requested) and can be used to fetch keys recursively.

This PR also adds optional `recurse` parameter to `delete`.

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
@@ -1,5 +1,7 @@
 package io.buoyant.consul.v1
 
+import java.util.Base64
+
 import com.twitter.conversions.time._
 import com.twitter.finagle.http
 import com.twitter.finagle.service.Backoff
@@ -61,6 +63,30 @@ class KvApi(
   }
 
   /**
+   * Get key(s)
+   *
+   * https://www.consul.io/docs/agent/http/kv.html#single
+   *
+   * @param path path to the key, must start with /
+   */
+  def multiGet(
+    path: String,
+    datacenter: Option[String] = None,
+    blockingIndex: Option[String] = None,
+    recurse: Option[Boolean] = None,
+    retry: Boolean = false
+  ): Future[Indexed[Seq[Key]]] = {
+    val req = mkreq(
+      http.Method.Get,
+      s"$kvPrefix$path",
+      "index" -> blockingIndex,
+      "dc" -> datacenter,
+      "recurse" -> recurse.map(_.toString)
+    )
+    executeJson[Seq[Key]](req, retry)
+  }
+
+  /**
    * Store the key value
    *
    * https://www.consul.io/docs/agent/http/kv.html#single
@@ -95,15 +121,27 @@ class KvApi(
     path: String,
     datacenter: Option[String] = None,
     cas: Option[String] = None,
+    recurse: Option[Boolean] = None,
     retry: Boolean = false
   ): Future[Boolean] = {
     val req = mkreq(
       http.Method.Delete,
       s"$kvPrefix$path",
       "cas" -> cas,
+      "recurse" -> recurse.map(_.toString),
       "dc" -> datacenter
     )
     executeJson[Boolean](req, retry).map(_.value)
   }
+}
 
+object Key {
+  def mk(key: String, value: String): Key = Key(Some(key), Some(Base64.getEncoder.encodeToString(value.getBytes)))
+}
+
+case class Key(
+  Key: Option[String],
+  Value: Option[String]
+) {
+  lazy val decoded: Option[String] = Value.map { raw => new String(Base64.getDecoder.decode(raw)) }
 }
--- a/consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
+++ b/consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
@@ -11,8 +11,11 @@ import org.scalatest.FunSuite
 class KvApiTest extends FunSuite with Awaits with Exceptions {
   val listBuf = Buf.Utf8("""["foo/bar/", "foo/baz/"]""")
   val getBuf = Buf.Utf8("""foobar""")
+  val multiGetBuf = Buf.Utf8("""[{"LockIndex":0,"Key":"sample","Flags":0,"Value":"Zm9vYmFy","CreateIndex":10,"ModifyIndex":12}]""")
   val putOkBuf = Buf.Utf8("""true""")
   val putFailBuf = Buf.Utf8("""false""")
+  val deleteOkBuf = Buf.Utf8("""true""")
+  val deleteFailBuf = Buf.Utf8("""false""")
   var lastUri = ""
 
   override val defaultWait = 2.seconds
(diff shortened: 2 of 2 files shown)


consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -95,15 +98,27 @@ class KvApi(
...
+    execute[Boolean](req, retry).map(_.value)
   }
+}
+
+object Key {
+  def mk(key: String, value: String): Key = Key(Some(key), Some(Base64.getEncoder.encodeToString(value.getBytes)))
+}
 
+case class Key(
+  Key: Option[String],
```
Ashald: I wonder whether I can drop `Option` from those fields as they 'must be set at all times'?

adleong:
I like the convenience of having a single-value get that returns a String.  Rather than modifying `get` to support multi-gets, would you consider adding a new method called `multiGet` or something like that?

Ashald:
How about making it `multiGet` but making `get` to work work as a call to `multiGet` and taking 1st item from the list?

adleong:
Is the first item guaranteed to be the one that matches the key exactly?  It seems simpler to keep them separate.

Ashald:
The short answer is: yes.

The long answer:
1) in non-recursive mode (default-one) it's always either 1-item list or `404`
2) in recursive mode it's a list of lexicographically ordered items filtered by prefix match therefore the requested item is always 1st in the list

The `CatalogNamer` doesn't care about recursive mode pretty much.

On a side note, IMO it's better to have a single `execute` that works with the json rather than `execute` and `executeJson` as it's now.

adleong:
Unless the requested key doesn't exist 😜 
In that case, taking the first item from a recursive get is definitely not what you want.

Is lexicographical ordering a guarantee of the api?  I couldn't find that in the docs.

Ashald:
> Unless the requested key doesn't exist 😜 
> In that case, taking the first item from a recursive get is definitely not what you want.
But that applies only to recursive mode - in non-recursive mode that would result in 404. ;)
My proposal is to 'use default mode by default' but add a flag to switch to recursive mode.
With it `CatalogNamer`'s `observe` (or the simple `get` as proposed above) will use the `non-recursive` mode.

> Is lexicographical ordering a guarantee of the api? I couldn't find that in the docs.
I don't recall seeing that in docs as well but it appears that that's the case. That how items displayed in Consul UI and that's how they are returned in API calls. I will ask their support and if yes maybe they can add that to docs. But, again, that applies only to recursive mode which is not used by `CatalogNamer` but will be used by the extension we planing to build with it (we applying an extra sort just in case in it).

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -61,6 +63,30 @@ class KvApi(
...
+   *
+   * https://www.consul.io/docs/agent/http/kv.html#single
+   *
+   * @param path path to the key, must start with /
+   */
+  def multiGet(
+    path: String,
+    datacenter: Option[String] = None,
+    blockingIndex: Option[String] = None,
+    recurse: Boolean = false,
```
adleong: for multiGet, recurse should always be true, right?

adleong: should allow the separator to be specified as well

adleong: TIOLI but I think it's more ergonomic to have this be
```
recurse: Option[Boolean] = None
```

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -95,15 +121,27 @@ class KvApi(
     path: String,
     datacenter: Option[String] = None,
     cas: Option[String] = None,
+    recurse: Boolean = false,
     retry: Boolean = false
```
adleong: probably want to be able to specify separator here too

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: > should allow the separator to be specified as well
It works only with `list` query parameter.

adleong:
This is good.  I think you need to update the title/description, however.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: > for multiGet, recurse should always be true, right?

I'd say it _can_ but not _must_. Do you mind leaving it optional?
I expect usage as:
```
def getNs(path: String, index: Option[String] = None) =
    api.get(path, blockingIndex = index, recurse = path.endsWith("/"))
```

so it'd be much more convenient to switch the `recurse` flag rather than use another method.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: The same situation here as well - separator is accepted only for `get` method and only when `list` query parameter is used.

Ashald:
@adleong thanks - totally forgot about description. :( I updated both title and description.

Please let me know whether I need to fix anything else so it can be merged.

Thanks!

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -61,6 +63,30 @@ class KvApi(
...
+    blockingIndex: Option[String] = None,
+    recurse: Option[Boolean] = None,
+    retry: Boolean = false
+  ): Future[Indexed[Seq[Key]]] = {
+    val req = mkreq(
+      http.Method.Get,
+      s"$kvPrefix$path",
+      "index" -> blockingIndex,
+      "dc" -> datacenter,
+      "recurse" -> Some(recurse.toString)
```
adleong: I think this should just be `"recurse" -> recurse`

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -95,15 +121,27 @@ class KvApi(
...
     datacenter: Option[String] = None,
     cas: Option[String] = None,
+    recurse: Option[Boolean] = None,
     retry: Boolean = false
   ): Future[Boolean] = {
     val req = mkreq(
       http.Method.Delete,
       s"$kvPrefix$path",
       "cas" -> cas,
+      "recurse" -> Some(recurse.toString),
```
adleong: ditto

consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
```diff
@@ -9,10 +9,13 @@ import io.buoyant.test.{Awaits, Exceptions}
 import org.scalatest.FunSuite
 
 class KvApiTest extends FunSuite with Awaits with Exceptions {
-  val listBuf = Buf.Utf8("""["foo/bar/", "foo/baz/"]""")
   val getBuf = Buf.Utf8("""foobar""")
```
adleong: I think you need to un-delete all the `list` and `get` tests in this file

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
Ashald: Ah, when I will learn? %) Thanks!

consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
Ashald: Done

adleong:
:star: looks good!

Ashald:
I see a failure on CircleCI but it failed due to dependency resolution error. Can you please re-run the build?

Ashald:
Thank you!

PR #587, opened 2016-08-18
Accept '_' in hostnames in RewritingNamers

Ashald:
Although it looks like that underscore is not supposed to
appear in hostnames in reality it does in some cases. Assuming that
practicality beats purity.


klingerf:
⭐ Works for me.

olix0r:
lgtm

Ashald:
Thanks guys!

PR #588, opened 2016-08-18
addr and resolve apis should wait until Addrs are non-pending

adleong:
# Problem

If address resolution takes some time, the `addr` and `resolve` endpoints will return an address in the pending state.

# Solution

Wait until the address is non-pending before returning.

Reported by @Ashald


olix0r:
lgtm, but a question:

What if the name doesn't become pending for a very long time?  Does it make sense to accept a `wait=N` query param that specifies how long to wait for a pending message?  if it times out, we just return the pending response.  This allows the consumer to specify their tolerance explicitly.  I get that this can be implemented in timeouts, but doesn't seem onerous to be a part of the API, either.

TIOLI.

Ashald:
Checked - works like a charm!

PR #591, opened 2016-08-21
Log which telemeters are in use at startup

olix0r:
Furthermore, do not provide default stats and trace receivers.  When providing defaults, it's easy to think you're overriding _statsReceiver_, for instance, when you should be overriding _stats_.  By removing default values, implementations must explicitly satisfy the receivers in order for the module to compile.


adleong:
:star: 👷

PR #593, opened 2016-08-22
Add io.l5d.default telemeter

olix0r:
The default telemeter use finagle's DefaultStatsReceiver and/or DefaultTracer.

Either of these may be disabled explicitly.

Allows the preservation of /admin/metrics.json when using another stats telemeter.


telemetry/core/src/main/scala/io/buoyant/telemetry/default.scala
```diff
@@ -0,0 +1,30 @@
+package io.buoyant.telemetry
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.Stack
+import com.twitter.finagle.stats.{DefaultStatsReceiver, NullStatsReceiver}
+import com.twitter.finagle.tracing.{DefaultTracer, NullTracer}
+
+class DefaultInitializer extends TelemeterInitializer {
```
klingerf: Similar comment here as in #594 -- for consistency with other initializers, I'd capitalize the filename.  Would also consider calling it DefaultTelemetryInitializer.scala, similar to DefaultInterpreterInitializer.scala.

olix0r: You're right, but it's even crazier -- there are some initializers in [FooConfig.scala](https://github.com/BuoyantIO/linkerd/blob/master/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala#L14).  The issue arises because the style is _match TypeName with TypeName.scala_, but it doesn't make sense to split these things out into separate files because they're so tightly coupled -- and we can't use classes namespaced in companion objects because it messes with the reflection/object loading needed to do plugins & json loading.

Then there's a _newer_ part of the scala style guide that says files with multiple classes should be given lowercase names... Which is what I am (implicitly) proposing here.

Good catch, though.  We should standardize.  I've found the *Initializer.scala naming to be confusing when I'm looking for the non-initializer class.  I'll move these onto that convention for the time being though

adleong:
:star: 🏬

klingerf:
⭐  🔭  Thanks for updating.

PR #594, opened 2016-08-22
Introduce the io.l5d.tracelog telemeter

olix0r:
The io.l5d.tracelog telemeter logs a selection of trace records to a logger.

Intended to replace the debugTrace flag on tracers.


linkerd/examples/acceptance-test.yaml
```diff
@@ -8,7 +8,11 @@ tracers:
   host: zipkincollector
   port: 9410
   sampleRate: 0.02
-  debugTrace: true
+
+telemetry:
+- kind: io.l5d.tracelog
+  sampleRate: 0.2
+  leve: TRACE
```
klingerf: Typo here: "leve" => "level"

adleong:
:star: 🍭

telemetry/tracelog/src/main/scala/io/buoyant/telemetry/tracelog.scala
```diff
@@ -0,0 +1,46 @@
+package io.buoyant.telemetry
+
+import com.twitter.finagle.Stack
+import com.twitter.finagle.buoyant.Sampler
+import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.finagle.tracing.{Trace, Record, TraceId, Tracer}
+import com.twitter.logging.{Level, Logger}
+
+class TracelogInitializer extends TelemeterInitializer {
```
klingerf: It might be more consistent with other initializers to call this file TracelogInitializer.scala, instead of tracelog.scala.

telemetry/tracelog/src/main/scala/io/buoyant/telemetry/TracelogInitializer.scala
```diff
@@ -0,0 +1,46 @@
...
+}
+
+class TracelogTracer(logger: Logger, level: Level, sample: Sampler) extends Tracer {
+
+  // This tracer doesn't influence downstream tracing.
+  def sampleTrace(id: TraceId): Option[Boolean] = None
+
+  def record(record: Record): Unit = {
+    val id = record.traceId
+    if (id.sampled.getOrElse(sample(id.traceId.toLong))) {
```
klingerf: Question about this -- I think that if the linkerd config is setup to use a tracer (such as io.l5d.zipkin), then id.sampled will always be set to `Some(true)` or `Some(false)` before reaching this check, at which point the configured sample rate for this telemeter will always be ignored.  Is that the intended behavior?

olix0r: Yeah, you're probably right that this shouldn't look at `id.sampled` at all...

PR #595, opened 2016-08-23
[issue-589] Consul Namer allow health-check status to be considered

JustinVenus:
My scala is pretty rusty, but this is how I approached a solution for issue-589.  I chose to use [/v1/health/service/<name>](https://www.consul.io/docs/agent/http/health.html#health_service) which returns a json payload like ...
```javascript
[
  {
    "Node": {
      "Node": "foobar",
      "Address": "10.1.10.12",
      "TaggedAddresses": {
        "lan": "10.1.10.12",
        "wan": "10.1.10.12"
      }
    },
    "Service": {
      "ID": "redis",
      "Service": "redis",
      "Tags": null,
      "Address": "10.1.10.12",
      "Port": 8000
    },
    "Checks": [
      {
        "Node": "foobar",
        "CheckID": "service:redis",
        "Name": "Service 'redis' check",
        "Status": "passing",
        "Notes": "",
        "Output": "",
        "ServiceID": "redis",
        "ServiceName": "redis"
      },
      {
        "Node": "foobar",
        "CheckID": "serfHealth",
        "Name": "Serf Health Status",
        "Status": "passing",
        "Notes": "",
        "Output": "",
        "ServiceID": "",
        "ServiceName": ""
      }
    ]
  }
]
```

Tests on my machine.
```
> consul/test
[info] AgentApiTest:
[info] - localAgent returns LocalAgent instance
Aug 23, 2016 10:40:10 AM io.buoyant.consul.v1.BaseApi$$anonfun$1 applyOrElse
ERROR: retrying consul catalog request on error java.lang.Exception: I have no idea who to talk to
Aug 23, 2016 10:40:10 AM io.buoyant.consul.v1.BaseApi$$anonfun$1 applyOrElse
ERROR: retrying consul catalog request on error java.lang.Exception: I have no idea who to talk to
[info] CatalogApiTest:
[info] - datacenters endpoint returns a seq of datacenter names
[info] - serviceNodes endpoint returns a seq of ServiceNodes
[info] - serviceMap endpoint returns a map of serviceNames to tags
[info] - blocking index returned from one call can be used to set index on subsequent calls
[info] - propagates client failures
[info] - makes infinite retry attempts on retry = true
[info] - reports invalid datacenter as an unexpected response
[info] KvApiTest:
[info] - list returns an indexed seq of key names
[info] - list reports wrong paths with an error
[info] - get returns an indexed value
[info] - get uses raw values
[info] - get reports wrong paths with an error
[info] - multiGet returns an indexed seq of values
[info] - multiGet by default is non-recurse
[info] - multiGet with recurse set to true adds a recurse parameter
[info] - multiGet reports wrong paths with an error
[info] - put returns true on success
[info] - put returns false on failure
[info] - put cas flag
[info] - delete returns true on success
[info] - delete returns false on failure
[info] - delete cas flag
[info] - delete by default is non-recurse
[info] - delete with recurse set to true adds a recurse parameter
[info] - blocking index returned from one call can be used to set index on subsequent calls
[info] - propagates client failures
[info] - makes infinite retry attempts on retry = true
[info] - reports invalid datacenter as an unexpected response
[info] - get reports permission errors
[info] Run completed in 799 milliseconds.
[info] Total number of tests run: 30
[info] Suites: completed 3, aborted 0
[info] Tests: succeeded 30, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[success] Total time: 3 s, completed Aug 23, 2016 10:40:11 AM
```


adleong:
As discussed in https://github.com/BuoyantIO/linkerd/issues/589, I think we'd like this to be a separate namer from the CatalogNamer (though it looks like it will be almost identical) so that we can get a health-based namer merged before validating that it is a suitable replacement for the catalog namer in all cases.

Also, I think the passing of `passing` should be configurable so that we can choose to respect consul's health checks or not.

JustinVenus:
As luck would have it '?passing=false' does not have the effect one would hope for.  Even so, why would anyone want down services (that are failing health checks) to be considered as valid endpoints?

adleong:
I think if you omit the `?passing` parameter entirely, the response won't be filtered to passing services.  As for why you would want that, you may not want to treat consul as the authority on which services are healthy.

JustinVenus:
I see your reasoning with ```'?passing'```, and in that case I think it makes sense to use the CatalogAPI in that case.  How do you feel about adding a paramenter ```useHeathCheck: Option[Boolean] = Some(False)``` to [case class ConsulConfig](https://github.com/BuoyantIO/linkerd/blob/master/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala#L33) so I can just switch out [CatalogApi for HealthApi](https://github.com/BuoyantIO/linkerd/blob/master/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala#L73) based on user configured preference.  I ask b/c I feel implementing an entire namer for such a small change is not worth the effort.

JustinVenus:
This is what I'm considering for HealthApi.

```patch
diff --git a/consul/src/main/scala/io/buoyant/consul/v1/HealthApi.scala b/consul/src/main/scala/io/buoyant/consul/v1/HealthApi.scala
new file mode 100644
index 0000000..2235433
--- /dev/null
+++ b/consul/src/main/scala/io/buoyant/consul/v1/HealthApi.scala
@@ -0,0 +1,79 @@
+package io.buoyant.consul.v1
+
+import com.twitter.util.{Duration, Future}
+
+object HealthApi {
+  def apply(c: Client): HealthApi = new HealthApi(c, s"/$versionString")
+}
+
+class HealthApi(
+  override val client: Client,
+  override val uriPrefix: String,
+  override val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
+  override val stats: StatsReceiver = DefaultStatsReceiver
+) extends CatalogApi(
+  client,
+  uriPrefix,
+  backoffs,
+  stats
+) {
+
+  val healthPrefix = s"$uriPrefix/health"
+
+  // https://www.consul.io/docs/agent/http/health.html#health_service
+  override def serviceNodes(
+    serviceName: String,
+    datacenter: Option[String] = None,
+    tag: Option[String] = None,
+    retry: Boolean = false
+  ): Future[Indexed[Seq[ServiceNode]]] = {
+    val req = mkreq(
+      http.Method.Get,
+      s"$healthPrefix/service/$serviceName",
+      "dc" -> datacenter,
+      "tag" -> tag,
+      "passing" -> Some("true")
+    )
+    executeJson[Seq[ServiceHealth]](req, retry).map(
+      indexed => {
+        val service = indexed.value.flatten(_.Service).head
+        val node = indexed.value.flatten(_.Node).head
+        val result = List(
+          ServiceNode(
+            node.Node,
+            service.Address,
+            service.ID,
+            service.Service,
+            service.Tags,
+            service.Address,
+            service.Port
+          )
+        )
+        Indexed[Seq[ServiceNode]](result, indexed.index)
+      }
+    )
+  }
+}
+
+case class Node(
+  Node: Option[String]
+)
+
+case class Service_(
+  ID: Option[String],
+  Service: Option[String],
+  Tags: Option[Seq[String]],
+  Address: Option[String],
+  Port: Option[Int]
+)
+
+case class ServiceHealth(
+  Node: Option[Node],
+  Service: Option[Service_]
+)
diff --git a/consul/src/test/scala/io/buoyant/consul/v1/HealthApiTest.scala b/consul/src/test/scala/io/buoyant/consul/v1/HealthApiTest.scala
new file mode 100644
index 0000000..e4c8829
--- /dev/null
+++ b/consul/src/test/scala/io/buoyant/consul/v1/HealthApiTest.scala
@@ -0,0 +1,33 @@
+package io.buoyant.consul.v1
+
+import com.twitter.io.Buf
+
+class HealthApiTest extends FunSuite with Awaits {
+  val nodesBuf = Buf.Utf8("""[{"Node":{"Node":"Sarahs-MBP-2","Address":"192.168.1.37"}},{"Service": {"Service":"hosted_web","Tags":["master"],"Port":8084, "Address":""}}]""")
+  var lastUri = ""
+
+  def stubService(buf: Buf) = Service.mk[Request, Response] { req =>
+    val rsp = Response()
+    rsp.setContentTypeJson()
+    rsp.content = buf
+    lastUri = req.uri
+    Future.value(rsp)
+  }
+
+  test("serviceNodes endpoint returns a seq of ServiceNodes") {
+    val service = stubService(nodesBuf)
+
+    val response = await(HealthApi(service).serviceNodes("hosted_web")).value
+    assert(response.size == 1)
+    assert(response.head.ServiceName == Some("hosted_web"))
+    assert(response.head.Node == Some("Sarahs-MBP-2"))
+    assert(response.head.ServiceAddress == Some(""))
+    assert(response.head.ServicePort == Some(8084))
+  }
+}
```

adleong:
@JustinVenus that sounds reasonable to me.  In that case I think we would rename `CatalogNamer` to `ConsulNamer` since it can be backed by either the CatalogApi or the HealthApi.

JustinVenus:
@adleong I think this meets the requirements that we discussed.  Please advise if i should add any other members to ```trait ConsulApi extends BaseApi```.  Also sorry git chose to treat the rename as ```delete + add```.

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
```diff
@@ -0,0 +1,158 @@
...
+  ): Future[Indexed[Seq[ServiceNode]]] = {
+    val req = mkreq(
+      http.Method.Get,
+      s"$healthPrefix/service/$serviceName",
+      "index" -> blockingIndex,
+      "dc" -> datacenter,
+      "tag" -> tag,
+      "passing" -> Some("true")
+    )
+    executeJson[Seq[ServiceHealth]](req, retry).map(
```
adleong: You've got the right idea, but I don't think this code is correct.  Have you tested it against consul?

I think you want to map over the seq of HealthNodes and map them to ServiceNodes.  I think you're also missing the Address property in the Node case class.

JustinVenus: I have an unexpected response exception when I run locally.  I'm pretty sure it's the same reason you've pointed out.  I'll send a follow up when I get a chance to dig in.

JustinVenus: @adleong this now works as I expect.  @Ashald would you like to confirm if this change works as you'd expect?  Just add useHealthCheck to your consul namer once you've built my branch.

```yaml
namers:
- kind: io.l5d.consul
  experimental: true
  includeTag: true
  useHealthCheck: true
```

JustinVenus:
@adleong @Ashald I've uploaded [linkerd-0.7.3-SNAPSHOT-exe](https://www.dropbox.com/s/p45l5gvgels8y8t/linkerd-0.7.3-SNAPSHOT-exec?dl=0) to dropbox if you'd like to verify the same binary `sha256 below`.  I'm deploying this version to my test environment.

```bash
shasum -a 256 /Users/jvenus/Projects/linkerd/linkerd/target/scala-2.11/linkerd-0.7.3-SNAPSHOT-exec
4a5cffe054b3e988bbb173c356ca13d2915d3d1334d6a96456e054ffa0eec5ca  /Users/jvenus/Projects/linkerd/linkerd/target/scala-2.11/linkerd-0.7.3-SNAPSHOT-exec
```

Ashald:
@JustinVenus I'll give it a try later today

Ashald:
Hm, just noticed 1 issue: looks like it's not possible to rely on `catalog api` but do respect the `maintenance` health check, right? Or in other words, filter out only those services that are in maintenance mode.

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
adleong: style nit:
for `map`, we prefer this style:
```
executeJson[Seq[ServiceHealth]](req, retry).map { indexed =>
  ...
}
```

JustinVenus:
@Ashald  I do not see a way to use the consul catalog api to filter out anything for us.  From my very quick reading of the consul health api, it seems we can get the `entire`or the `passing` status of a service group.  It appears that we'd have to filter unhealthy out ourselves.

Fwiw, I am happy to accept deferring the knowledge of what is in maintenance to something else.  I personally don't feel like linkerd should care about the health of consul provided services when `useHealthCheck` is explicitly enabled.

JustinVenus:
To follow up on the `maintenance checking` I think it would be relatively easy to implement a `case Class Checks` and do something with the result when we are mapping the other case classes to ServiceNodes.

However, I think a larger discussion about what do do with the result would need to be had.
I'm mostly interested in getting failing nodes out of linkerd's consideration.  Zipkin in my environment shows me significant time spent, retrying to find working services (when consul returns unhealthy services).

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
```diff
@@ -0,0 +1,158 @@
...
+      indexed => {
+        val result = (indexed.value.flatten(_.Service), indexed.value.flatten(_.Node))
```
adleong: I would write this as:
```
indexed.value.map { health =>
  val service = health.flatMap(_.Service)
  val node = health.flatMap(_.Node)
  SerivceNode(
    node.Node,
    ...
```

adleong:
This looks really good to me, aside from a couple of small style nits.  Could you also update `linkerd/docs/namer.md` and `CHANGES.md`?

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
JustinVenus: Yeah that does read much clearer.

JustinVenus:
@adleong thanks for the feedback.  I will make the changes and update this pull request.

JustinVenus:
@adleong next steps?

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
adleong: Ah, I see.  What about

```
indexed.value.map { health =>
  val service = health.Service
  val node = health.Node
  SerivceNode(
    node.flatMap(_.Node),
    node.flatMap(_.Address),
    service.flatMap(_.ID),
    ...
```

adleong:
Hi @JustinVenus, sorry I missed your response to my comment.  I've replied above.

Ashald:
@JustinVenus I verified your custom build and it works as expected

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
JustinVenus: @adleong This is what I tried.

```scala
    executeJson[Seq[ServiceHealth]](req, retry).map { indexed =>
      val result = indexed.value.map { health =>
        val service = health.Service
        val node = health.Node
        ServiceNode(
          node.flatMap(_.Node),
          node.flatMap(_.Address),
          service.flatMap(_.ID),
          service.flatMap(_.Service),
          service.flatMap(_.Tags),
          service.flatMap(_.Address),
          service.flatMap(_.Port)
        )
      }
      Indexed[Seq[ServiceNode]](result, indexed.index)
    }
  }
}
```

This is the failed unit.

```
[info] HealthApiTest:
[info] - serviceNodes endpoint returns a seq of ServiceNodes *** FAILED ***
[info]   List(ServiceNode(Some(Sarahs-MBP-2),Some(192.168.1.37),None,None,None,None,None), ServiceNode(None,None,None,Some(hosted_web),Some(List(master)),Some(),Some(8084))) had size 2 instead of expected size 1 (HealthApiTest.scala:27)
Sep 01, 2016 3:24:50 PM io.buoyant.consul.v1.BaseApi$$anonfun$1 applyOrElse
```

JustinVenus:
@Ashald awesome, thanks for trying it out.

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
adleong: Hmmm... I think that unit test may have incorrect json.  It has:

```"""[{"Node":{"Node":"Sarahs-MBP-2","Address":"192.168.1.37"}},{"Service": {"Service":"hosted_web","Tags":["master"],"Port":8084, "Address":""}}]"""```

The Node object and are Service object are in different objects.  Shouldn't this be:

```
"""[{"Node":{"Node":"Sarahs-MBP-2","Address":"192.168.1.37"},"Service": {"Service":"hosted_web","Tags":["master"],"Port":8084, "Address":""}}]"""```

JustinVenus: @adleong good eyes, I couldn't for the life of me see that.  (even w/ jq).  I'll post a follow up.

JustinVenus:
@adleong I think this is ready to go.

CHANGES.md
```diff
@@ -14,6 +14,7 @@
   * Add `setHost` parameter for Consul CatalogNamer to set `authority` metadata 
   * Add auth `token` parameter to Consul Namer & Dtab Store
   * Add `datacenter` parameter to Consul Dtab Store
+  * Add `useHealthCheck` parameter to Consul Namer #589
```
adleong: This was not part of the 0.7.4 release, can you move this to the "current" section (0.x.x)

linkerd/docs/namer.md
```diff
@@ -147,6 +148,18 @@ The Consul namer is configured with kind `io.l5d.consul`, and these parameters:
...
 
+For example:
+```yaml
+namers:
+- kind: io.l5d.consul
+  experimental: true
+  host: 127.0.0.1
+  port: 2181
+  includeTag: true
+  useHealthCheck: true
```
adleong: There is already an example a few lines up in this section.  Maybe just add the `useHealthCheck` property to that example?

adleong:
Thanks for making those fixes, @JustinVenus.  I just had a couple of requests about the docs and then I think we can merge.

JustinVenus:
@adleong documentation updates have been made.

adleong:
:star: Awesome!  Thanks for all your hard work on this, @JustinVenus

olix0r:
:star: this looks really good. Very happy to merge this ;)

PR #597, opened 2016-08-23
Bump default test timeout to 2 seconds

adleong:


olix0r:
:star: should we just bump the default in Awaits.scala?

adleong:
that's what I did

olix0r:
😑  ...very clever...

PR #601, opened 2016-08-23
Chunked requests must not be considered retryable

### LINKED ISSUE #599: Retried HTTP requests drop chunked bodies
When HTTP requests are retried, it appears that they lose their body when using `Transfer-Encoding: chunked`. 🕳

The _initial_ request (before retries) is passed to the upstream server correctly by linkerd.

```
PUT / HTTP/1.1
Host: thing
Transfer-Encoding: chunked
Accept: */*
Content-Type: application/x-www-form-urlencoded
L5d-Ctx-Trace: U1nuHsM0tDjlY4pt28pW0eVjim3bylbRAAAAAAAAAAI=
L5d-Dst-Concrete: /#/io.l5d.fs/thing
L5d-Dst-Logical: /http/1.1/PUT/thing
L5d-Reqid: e5638a6ddbca56d1
User-Agent: curl/7.43.0

1a
Lorem ipsum dolor sit amet
0
```

But the retry loses the body:

```
PUT / HTTP/1.1
Host: thing
Transfer-Encoding: chunked
Accept: */*
Content-Type: application/x-www-form-urlencoded
L5d-Ctx-Trace: /D3iLPS/al7lY4pt28pW0eVjim3bylbRAAAAAAAAAAI=
L5d-Dst-Concrete: /#/io.l5d.fs/thing
L5d-Dst-Logical: /http/1.1/PUT/thing
L5d-Reqid: e5638a6ddbca56d1
User-Agent: curl/7.43.0

0
```

I've put together [a simple server](https://gist.github.com/obeattie/00ee3ba4ada7fe693ef8672ec76a10e4) as a test case to show this behaviour. It rejects valid requests as 500, triggering the retry when linkerd is configured with the `io.l5d.retryableIdempotent5XX` response classifier.

This does _not_ happen however when `chunked` encoding isn't used. In that case, the retry has the correct body.

olix0r:
When an HTTP router is configured with the Idempotent repsonse classifier, it
may consider a request with a body (e.g. PUTs) as retryable.  However, chunked
bodies may not be replayed, since the stream is not preserved by linkerd.

To fix this, we only treat a request as retryable when it does not have a
chunked body.

Fixes #599

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
@@ -15,7 +15,10 @@ object ResponseClassifiers {
 
     case class ByMethod(methods: Set[Method]) {
       def withMethods(other: Set[Method]): ByMethod = copy(methods ++ other)
-      def unapply(req: Request): Boolean = methods.contains(req.method)
+
+      // Chunked requests are not considered to be retryable
+      def unapply(req: Request): Boolean =
+        methods.contains(req.method)
     }
 
     /** Matches read-only requests */
@@ -77,7 +80,8 @@ object ResponseClassifiers {
    */
   val RetryableIdempotentFailures: ResponseClassifier =
     ResponseClassifier.named("RetryableIdempotentFailures") {
-      case ReqRep(Requests.Idempotent(), RetryableResult()) => ResponseClass.RetryableFailure
+      case ReqRep(req@Requests.Idempotent(), RetryableResult()) if !req.isChunked =>
+        ResponseClass.RetryableFailure
     }
 
   /**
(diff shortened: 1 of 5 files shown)


olix0r:
needs user documentation

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -15,7 +15,10 @@ object ResponseClassifiers {
 
     case class ByMethod(methods: Set[Method]) {
       def withMethods(other: Set[Method]): ByMethod = copy(methods ++ other)
-      def unapply(req: Request): Boolean = methods.contains(req.method)
+
+      // Chunked requests are not considered to be retryable
+      def unapply(req: Request): Boolean =
+        methods.contains(req.method)
```
olix0r: eh, revert this change

adleong:
I think this is a fundamental restriction that should be enforced at a higher level than individual response classifiers.  What about wrapping the configured response classifier with a `ChunkedRequestsCannotBeRetried` classifier:
```
case (req,rep) if req.isChunked =>
  val classification = underlying(req,rep)
  if (classification == RetyableFailure) NonRetryableFailure
  else classificaiton
case (req,rep) => underlying(req, rep)
```

adleong:
:star: 🚰

Issue #523, opened 2016-07-07
dtab playground UI issues


wmorgan:
Bundling a bunch of UI problems/suggestions together into one issue:
- [x] After editing the dtab a couple times, keyboard input get slow. As in, there's a noticeable delay between typing stuff in and seeing it appear in the textarea. Weird!
- [ ] Saving a new dtab in the textarea should clear the resolution displayed (since it’s now different), or maybe better should re-run it. Right now it just keeps the old result so you end up with invalid state if you're not careful.
- [ ] With namerd, the dtab rules from namerd are not editable, but it's not clear why. I think there should be an indication of where these rules are coming from.
- [ ] The textarea should probably be the same size as the path textfield at the top. It appears to be full width right now which looks a bit bad.
- [ ] The UI placement on this page odd. Cancel and save are on top of the textarea, and in fact if namerd is in use, they're on top of the namerd rules, far away from the text area. These buttons should be right under the dtab textarea.
- [ ] I think there should be a reset button (revert to base dtab) in addition to save and cancel.
- [ ] I still don't like styling of rules as button-looking things that aren't actually buttons. I'd much prefer to remove the border, and align the rules around the central " => ".
- [x] Finally, if you put in an invalid dtab, you get a little popup that just says "Bad Request". It should say something better than that!

wmorgan:
Forgot to say, this is all against the UI in 0.7.0.

wmorgan:
Some of these are fixed. Assigned to @rmars for triage.

rmars:
Most of these are now addressed. Filed #1321 for the point about styling the rules.

PR #603, opened 2016-08-24
Fix dtab viewer slowness

adleong:
Fixes item 1 of #523 

# Problem

Each time the dtab save button is clicked, a call to render registers a new input listener on the dtab-input box.  Over time as you edit and save the dtab, the number of listeners increases and the dtab-input gets slower and slower.

# Solution

Register the input listener once instead of on each call to render.

# Result

dtab-input doesn't get slower over time.


rmars:
:star: :art: good find!

klingerf:
⭐ Looks good to me.

olix0r:
lgtm

PR #607, opened 2016-08-26
NameTree transformers

adleong:
You may now specify a list of transformers in any `interpreter` block.  These transformers can perform transformations on the NameTree that is returned from the NameInterpreter.  One use case is to filter the NameTree on the incoming interpreter to only use local addresses.


Ashald:
Looks like you never run out of neat features! Huge like - we would use it!

interpreter/per-host/src/main/resources/META-INF/services/io.buoyant.namer.TransformerInitializer
```diff
@@ -0,0 +1,2 @@
+io.buoyant.transformer.perHost.PortTransformerInitializer
+io.buoyant.transformer.perHost.LocalhostTransformerInitializer
```
olix0r: prefer sorted

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/LocalhostTransformer.scala
```diff
@@ -0,0 +1,43 @@
...
+import java.net.InetAddress
+
+/**
+  * The localhost transformer filters the list of addresses down to only
+  * addresses that have the same IP address as localhost.  This can be used by
+  * an incoming router to only route traffic to local destinations.
+  */
+class LocalhostTransformer extends DelegatingNameTreeTransformer {
+
+  private[this] val localhost = InetAddress.getLocalHost.getAddress
```
olix0r: It's probably worth documenting explicitly how "local" is determined.  From the docs:

> Returns the address of the local host. This is achieved by retrieving the name of the host from the system, then resolving that name into an InetAddress.

It's unclear what this line does:
- when the host has no IP
- when the host's name isn't resolvable
- when the host has multiple IPs

Furthermore, in some cases (kubernetes), we may need to do additional work to determine locally-routable IPs.

I think I'd prefer that the class be configured with a `Seq[InetAddress]` (there can be multiple local addresses), and allow this to be configured.  Our default logic should probably fall back to getLocalHost.getAddress.

Perhaps this class should take an `isLocal: InetAddress => Boolean` argument so that this behavior may be easily swapped?

adleong: Our ability to handle those edge cases depends on InetAddress' ability to handle those edge cases.  In some of those situations, this transformation may not make sense or be applicable.  

Yes, I will be adding a k8s-local transformer in a future review that does the same thing but changes out the `isLocal` logic for a /24 subnet comparison.  No need to prematurely refactor.

olix0r: tioli: per discussion, maybe worth elaboring on the intended use of this class in the docstring.

linkerd/docs/transformer.md
```diff
@@ -0,0 +1,41 @@
...
+--- | ------------- | -----------
+kind | _required_ | One of the transformer kinds listed below.
+
+## Localhost
+
+kind: `io.l5d.localhost`
+
+The localhost transformer filters the list of addresses down to only addresses
+that have the same IP address as localhost.  This can be used by an incoming
+router to only route traffic to local destinations.
```
olix0r: I think we should try to explain the localhost-determination logic here.  (Resolve the local hostname in dns, non updating.)

namer/core/src/main/scala/io/buoyant/namer/FilteringTransformer.scala
```diff
@@ -0,0 +1,41 @@
+package io.buoyant.namer
+
+import com.twitter.finagle.Name.Bound
+import com.twitter.finagle.{Addr, Address, NameTree}
+import com.twitter.util.Activity
+import java.net.InetSocketAddress
+
+trait FilteringTransformer extends DelegatingNameTreeTransformer {
```
olix0r: docstring?  how is this intended to be used

namer/core/src/main/scala/io/buoyant/namer/NameTreeTransformer.scala
```diff
@@ -0,0 +1,33 @@
+package io.buoyant.namer
+
+import com.twitter.finagle.Name.Bound
+import com.twitter.finagle.naming.NameInterpreter
+import com.twitter.finagle._
+import com.twitter.util.Activity
+
+trait NameTreeTransformer {
```
olix0r: a few /** comments */ would probably help here

olix0r:
:star: the approach looks good to me! just some comments on documentation

PR #608, opened 2016-08-26
Improve display of Neg, Empty, Fail in delegator UI

### LINKED ISSUE #604: Delegate API / Namerd UI doesn't handle /$/fail (!) properly
Having a namespace `default`:
```
/env#/foo-123=>/$/inet/127.0.0.1/80;
/env#/foo-456=>~;
/service/foo=>/env#/foo-123 | !;
/service/foo=>/env#/foo-456 | !;
```
expected behavior to see `neg` for `/service/foo`.

The `resolve` endpoint handles that just right:
```
$ curl -s "$NAMERD_URL/api/1/resolve/default?path=/service/foo"
{"type":"neg"}
```
but not the Namerd UI (which is presumably powered by `delegate` API endpoint):
screen shot 2016-08-24 at 18 36 05

adleong:
Fixes #604 

# Problem

Neg, empty, or fail DelegationTrees get collapsed into their delegation parent which makes the display confusing.

# Solution

Don't collapse neg, empty, or fail DelegationTrees.  Also use ~, $, and ! as their display names instead of the parent path.

(See #604 for screenshot)

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
@@ -92,15 +92,11 @@ object DelegateApiHandler {
   ))
   sealed trait JsonDelegateTree
   object JsonDelegateTree {
-    case class Empty(dentry: Option[Dentry]) extends JsonDelegateTree {
-      val path = "$"
-    }
+    case class Empty(path: Path, dentry: Option[Dentry]) extends JsonDelegateTree
     case class Fail(dentry: Option[Dentry]) extends JsonDelegateTree {
       val path = "!"
     }
-    case class Neg(dentry: Option[Dentry]) extends JsonDelegateTree {
-      val path = "~"
-    }
+    case class Neg(path: Path, dentry: Option[Dentry]) extends JsonDelegateTree
     case class Exception(path: Path, dentry: Option[Dentry], message: String) extends JsonDelegateTree
     case class Delegate(path: Path, dentry: Option[Dentry], delegate: JsonDelegateTree) extends JsonDelegateTree
     case class Leaf(path: Path, dentry: Option[Dentry], bound: Bound) extends JsonDelegateTree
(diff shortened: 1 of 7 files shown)


olix0r:
This looks good, though I'd prefer that we add some test cases to illustrate that it works properly.

adleong:
I had to walk this back a bit to preserve the existing behavior of Neg.  Neg is slightly overloaded in that it can arise when there are no matching dentries or when a dentry explicitly maps to Neg.  

screen shot 2016-08-26 at 3 24 21 pm

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -93,7 +93,9 @@ object DelegateApiHandler {
   sealed trait JsonDelegateTree
   object JsonDelegateTree {
     case class Empty(path: Path, dentry: Option[Dentry]) extends JsonDelegateTree
-    case class Fail(path: Path, dentry: Option[Dentry]) extends JsonDelegateTree
+    case class Fail(dentry: Option[Dentry]) extends JsonDelegateTree {
+      val path = "!"
```
olix0r: How do you feel about using the descriptive paths like `/$/fail` instead?  I find these symbols to be really hard to keep track of mentally.

adleong:
A bit more special casing to get both `!` and `/$/fail` working as well as `~` when explicitly present on the right hand side of a dentry.

screen shot 2016-08-26 at 5 09 17 pm

screen shot 2016-08-26 at 5 36 51 pm

Ashald:
👍

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -93,24 +93,32 @@ object DelegateApiHandler {
...
       case DelegateTree.Fail(p, d) =>
-        Future.value(JsonDelegateTree.Fail(p, mkDentry(d)))
+        val path = if (p == null) null
+        else if (p.startsWith(fail)) p.show
+        else "!"
+        Future.value(JsonDelegateTree.Fail(path, mkDentry(d)))
       case DelegateTree.Neg(p, d) =>
-        Future.value(JsonDelegateTree.Neg(p, mkDentry(d)))
+        val path = if (p == null) null
+        else if (p.isEmpty) "~"
```
olix0r: have a slight preference for `"/$/neg"` over `~`, etc...

definitely not a blocker

olix0r:
lgtm

adleong:
```
scala> Namer.global.lookup(Path.read("/$/fail")).sample
res6: com.twitter.finagle.NameTree[com.twitter.finagle.Name] = Fail

scala> Namer.global.lookup(Path.read("/$/neg")).sample
java.lang.ClassNotFoundException: neg
```

rmars:
lgtm!

PR #609, opened 2016-08-26
Improve dtab playground error message

adleong:
If the Path or Dtab is invalid, display a better error message than "Bad Request"

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
@@ -17,17 +17,6 @@ import io.buoyant.namer._
 
 object DelegateApiHandler {
 
-  private object PathStr {
-    def unapply(p: String) = Try(Path.read(p)).toOption
-  }
-
-  private object DtabStr {
-    def unapply(d: String) = if (d == null)
-      Some(Dtab.empty)
-    else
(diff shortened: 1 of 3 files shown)


gtcampbell:
:star: code looks good; what would it take to add a test for this?

rmars:
lgtm!

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -177,9 +166,11 @@ object DelegateApiHandler {
     def writeBuf[T](t: T): Buf = Buf.ByteArray.Owned(mapper.writeValueAsBytes(t))
   }
 
-  def getDelegateRsp(dtab: String, path: String, delegator: Delegator): Future[Response] =
-    (dtab, path) match {
-      case (DtabStr(d), PathStr(p)) =>
+  def getDelegateRsp(dtab: String, path: String, delegator: Delegator): Future[Response] = {
+    val dtabTry = if (dtab == null) Return(Dtab.empty) else Try(Dtab.read(dtab))
```
olix0r: might be a little robust with something like:
```scala
dtab match {
  case null | "" => Return(Dtab.empty)
  case dtab => Try(Dtab.read(dtab))
}
```

olix0r:
lgtm

PR #612, opened 2016-08-30
Add tag to authority in Consul namer

Ashald:
When `includeTag` mode is makes sense to include tag in `authority` meta field as it would match Consul's behavior.


Ashald:
This is a follow-up for #552 that we noticed after starting using it to route our services - @olix0r / @adleong what do you think?

adleong:
:star: makes sense to me

Ashald:
I'm glad you liked it. :)

Ashald:
I hope this will get into next release? :)

olix0r:
lgtm

Ashald:
Yay! Thanks!

PR #614, opened 2016-08-31
Move doc examples' placement

esbie:
Problem:
https://linkerd.io/doc/0.7.4/linkerd/config has lots of good documentation
but is somewhat difficult to navigate.

Solution:
Use a new three column layout (nav, doc, & example) to display linkerd
documentation.

Why this change:
The new layout looks better when example code is placed before the
paragraph that describes it.

(this is the kind of diff git/github doesn't do especially well with, apologies)

### DIFF
--- a/linkerd/docs/announcer.md
+++ b/linkerd/docs/announcer.md
@@ -1,7 +1,5 @@
 # Announcers
 
-*(for the [announcers](config.md#announcers) key)*
-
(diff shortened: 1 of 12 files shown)


adleong:
:star: 📃 looks great!  the namerd docs should get the same treatment too

esbie:
OK great once I've got linkerd in a working state, I'll do namerd next

PR #615, opened 2016-08-31
Make retry filter do not retry on Future.cancel

Ashald:
Previously it wasn't possible to cancel request when retry filter is enable as it will consider `FutureCancelledException` as an error that should be re-tried:
```
E 0831 20:24:25.121 THREAD12: retrying consul catalog request on error Failure(The future was cancelled with Future.cancel, flags=0x02)
```

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
@@ -7,7 +7,7 @@ import com.twitter.finagle.param.HighResTimer
 import com.twitter.finagle.service.{RetryBudget, RetryFilter, RetryPolicy}
 import com.twitter.finagle.stats.StatsReceiver
 import com.twitter.finagle.tracing.Trace
-import com.twitter.finagle.{Filter, http}
+import com.twitter.finagle.{Failure, Filter, http}
 import com.twitter.io.Buf
 import com.twitter.util._
 import io.buoyant.consul.log
@@ -27,8 +27,10 @@ trait BaseApi extends Closable {
     RetryPolicy.backoff(backoffs) {
       // We will assume 5xx are retryable, everything else is not for now
       case (_, Return(rep)) => rep.status.code >= 500 && rep.status.code < 600
+      // Don't retry on interruption
+      case (_, Throw(e: Failure)) if e.isFlagged(Failure.Interrupted) => false
       case (_, Throw(NonFatal(ex))) =>
-        log.error(s"retrying consul catalog request on error $ex")
+        log.error(s"retrying consul request on error $ex")
         true
     },
     HighResTimer.Default,
--- a/consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
+++ b/consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
@@ -1,8 +1,7 @@
 package io.buoyant.consul.v1
 
-import com.twitter.conversions.time._
-import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Request, Response}
+import com.twitter.finagle.{Failure, Service}
 import com.twitter.io.Buf
 import com.twitter.util.Future
 import io.buoyant.test.{Awaits, Exceptions}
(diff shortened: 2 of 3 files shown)


consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -27,8 +27,11 @@ trait BaseApi extends Closable {
     RetryPolicy.backoff(backoffs) {
       // We will assume 5xx are retryable, everything else is not for now
       case (_, Return(rep)) => rep.status.code >= 500 && rep.status.code < 600
+      // Looks like it could be either of 2 depending on where FutureCancelledException occurred
+      case (_, Throw(e: Failure)) if e.isFlagged(Failure.Interrupted) => false
+      case (_, Throw(e: FutureCancelledException)) => false
```
olix0r: It seems like a bug for FutureCancelledException to be thrown ever... (Future.cancel is deprecated).  Any ideas where that's coming from?

adleong: from a `Var.async` we return a closable:
```
      Closable.make { _ =>
        running = false
        pending.raise(new FutureCancelledException)
        Future.Unit
      }
```
Is there a more idiomatic way to cancel a future?

olix0r: We should probably return a Failure with Interrupted set -- I think Failure is the new idiomatic way to do this

Ashald: From the source code:
```
@deprecated("Provided for API compatibility; use raise() instead.", "6.0.0")
  def cancel() { raise(new FutureCancelledException) }
```

So although `Future.cancel` is deprecated the `FutureCancelledException` is not.

> We should probably return a Failure with Interrupted set -- I think Failure is the new idiomatic way to do this

Can you please explain this in ore details please? How we can influence currently running future this way?

olix0r: I think the most idiomatic thing to do is that we do somethign like `future.raise(Failure("consul observation released", Failure.Interrupted))` instead of `future.cancel()`

Basically, I'd prefer to use failures to signal this information instead of a list of exception types.

Ashald: Do you want me to update the way we cancel futures within this PR? Or should I just adjust this PR so retry filter will handle Interrupted failures properly and then submit another PR changing the way we canceling futures?

olix0r: I think it's fine to fix the exception site and the retry handling together in this PR.  Thanks!

Ashald: @adleong Do we want to add `retry = true` in `ConsulDtabStore` to all api calls within this PR as well?

adleong: Hmmm yeah, I think that makes sense for `list` and `observe` (i.e. the methods that return Activities).  Probably not for the methods that return `Future`

Ashald: Deal. ;)

Ashald:
@olix0r / @adleong please take a look at updated code.

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -27,8 +27,10 @@ trait BaseApi extends Closable {
     RetryPolicy.backoff(backoffs) {
       // We will assume 5xx are retryable, everything else is not for now
       case (_, Return(rep)) => rep.status.code >= 500 && rep.status.code < 600
+      // Looks like it could be either of 2 depending on where FutureCancelledException occurred
```
adleong: I don't understand this comment.  I think it should be something like `Don't retry interrupted failures` or something like that

consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
```diff
@@ -212,6 +211,19 @@ class KvApiTest extends FunSuite with Awaits with Exceptions {
     assert(result.value == List("foo/bar/", "foo/baz/"))
   }
 
+  test("doesn't retry on FutureCancelledException with retry = true") {
```
adleong: update test name

adleong:
Looks good, just a couple of clean up items.  Thanks for doing this!

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
Ashald: That was an old comment from previous version of the code. :/ Will fix, thanks.

consul/src/test/scala/io/buoyant/consul/v1/KvApiTest.scala
Ashald: Ooops. :)

Ashald:
Cleanup - done. :)

adleong:
:star: thanks!

olix0r:
lgtm!

PR #616, opened 2016-08-31
Make namers available globally

### LINKED ISSUE #596: Make configured namers available to plugins
linkerd plugins that talk to external services (identifiers, for example) need some way to discover those services.  Make the configured namers available to the plugin so that they can address external services using those namers (e.g. `/#/io.l5d.consul`)

adleong:
Fixes #596 

Plugins that accept a dst may now make use of configured namers.  For examplem, the fs namer could be used to locate namerd.

```
namers:
- kind: io.l5d.fs
  rootDir: linkerd/examples/io.l5d.fs
routers:
- protocol: http
  interpreter:
    kind: io.l5d.namerd
    dst: /#/io.l5d.fs/namerd
```


olix0r:
:star: this looks fine, though I'm having some trouble understanding exactly what this means...

I think this means that namers could potentially be configured to resolve through themselves?  What happens with this sort of configuration? Does it fail gracefully or does the process melt down? (I'd expect the latter).

We have an interpreter plugin interface -- should these interpreters be accessible from modules?

If we do this in linkerd's main, we may also want to do this for namerd's main...

adleong:
My understanding is that linkerd doesn't actually use the global NameInterpreter for anything.  The only place the global NameInterpreter would be used is for clients created manually in plugin code (e.g. to connect to the k8s api or the consul api etc.)  

So yes, if a namer plugin were to manually construct a client with it's own prefix as a dst, it would probably blow up.

Yes, we should probably do this for namerd as well.

olix0r:
I think it's slightly trickier than that--since telemeters are passed to namers, telemeters could be configured using namers in ways that cause things to blow up.  So, a person writing the config has to have some pretty intimate knowledge of how things are wired under the hood and what is allowable and what isn't.

I like the intent of the change, but I'm not sure how to provide this functionality without making the configuration more fragile.

adleong:
I did some testing:

* using a configured namer in a tracer: **WORKS**
  * even though the tracers are passed to the namers, the tracer doesn't start issuing requests until the global NameInterpreter has been set.  so this actually works, though you might run into issues if you're trying to send requests right on startup
* using a namer in a different namer: **WORKS**
  * discovered the k8s api for the k8s namer using the fs namer
* using a namer in itself: **SPECTACULAR STACK-OVERFLOW CRASH**

olix0r:
Can we guard against self-referential naming?  Since we don't do actually allow you to set Dtab.base on the process (right? i hope not...), all names in a config are necessarily concrete -- so we can detect `/#/$prefix` and prevent this?

adleong:
I think this would have to but up to the individual namer.  For example, if the io.l5d.foo namer takes a dst as a config parameter, it should validate that the dst doesn't start with it's own prefix.  Currently none of our namers accept dsts so this isn't an issue yet.

olix0r:
lgtm! probably warrants some docs (or at least a note in CHANGES)

PR #619, opened 2016-09-01
Handle unexpected k8s watch termination

obeattie:
Previously, if a watch response was terminated by a connection close, the reconnection logic became unreachable as the stream of events was never "terminated." This change ensures the stream ends by swallowing exceptions, and ensures the reconnection logic is run. ⚡ 

Fixes BuoyantIO/linkerd#617

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -8,6 +8,7 @@ import com.twitter.finagle.util.DefaultTimer
 import com.twitter.util._
 import io.buoyant.k8s.v1._
 import io.buoyant.namer.EnumeratingNamer
+
 import scala.collection.mutable
 
 class EndpointsNamer(
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -8,7 +8,6 @@ import com.twitter.finagle.util.DefaultTimer
 import com.twitter.util._
 import io.buoyant.k8s.v1._
 import io.buoyant.namer.EnumeratingNamer
-
 import scala.collection.mutable
 
 class EndpointsNamer(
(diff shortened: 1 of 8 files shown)


olix0r:
:star:⚡  Huge thanks for fixing this, @obeattie! Not a blocker for this change, but we should try to get a test case that covers this as well.

olix0r:
This change breaks some tests:
```
org.scalatest.exceptions.TestFailedException: Expected exception com.twitter.io.Reader$ReaderDiscarded to be thrown, but org.scalatest.exceptions.TestFailedException was thrown.
```

It sounds like only some exceptions should be ignored..

adleong:
Here's a unit test for this https://github.com/BuoyantIO/linkerd/tree/alex/k8s-watch-test

obeattie:
Yep, I just saw the test failure. Will fix. 🔧

adleong:
I wonder if that test that expects a ReaderDiscarded exception is describing the behavior we even want.  In the case where the reader is discarded, I think we want to simply terminate the AsyncStream, not have it contain an exception.  This way, reconnect logic (or whatever else) can be appended to it.  

@olix0r and @obeattie: what do you think about changing this back to swallowing all (maybe non-fatal) exceptions and changing that test to expect that the `AsyncStream` is empty instead of expecting a ReaderDiscarded?

obeattie:
Yep, I agree. I've changed that behaviour, and the tests too. I do notice a lot of k8s-related pollution in the test log now though… not sure if related. 🤔

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -8,6 +8,7 @@ import com.twitter.finagle.util.DefaultTimer
 import com.twitter.util._
 import io.buoyant.k8s.v1._
 import io.buoyant.namer.EnumeratingNamer
+
```
adleong: I think buoyant style is just to have all imports alphabetically sorted without any breaks like this

adleong:
:star: :100: 🎆

PR #621, opened 2016-09-02
Update config.md examples and navigation

esbie:
Problem:
Many links in buoyant.io/linkerd are broken, since in this site
layout all markdown files are combined into a single html file.

Solution:
Move the links to use anchors. Unfortunately they will no longer
link to each other in the docs folder. But maybe that is ok.

I also added an example to every section. I haven't used every
feature listed, so it'd be great if folks could find any errors
I may have made.

Also updates the response classifier example to be on the router
block instead of the client block.

### DIFF
--- a/linkerd/docs/config.md
+++ b/linkerd/docs/config.md
@@ -1,6 +1,8 @@
 # Introduction
 
->The most minimal linkerd configuration looks something like the following, which forwards all requests on `localhost:8080` to `localhost:8888`
+> A minimal linkerd configuration example, which forwards all requests on `localhost:8080` to `localhost:8888`
+
+
 
 ```yaml
 routers:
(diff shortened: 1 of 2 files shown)


linkerd/docs/config.md
```diff
@@ -222,13 +265,19 @@ Naming and service discovery are configured via the `namers` section of the
...
+- kind: io.l5d.zipkin
+  sampleRate: 0.02
+```
+
 Requests that are routed by linkerd are also traceable using Finagle's built-in
 tracing instrumentation. Trace data can be exported from a linkerd process by
 configuring tracers via a top-level `tracers` section:
 
-* *tracers* -- An array of [tracer](tracer.md) objects.
+* *tracers* -- An array of [tracer](#tracers10) objects.
```
esbie: this is apparently how slate deals with header name collisions

linkerd/docs/config.md
```diff
@@ -159,16 +167,27 @@ Each router must be configured as an object with the following params:
...
+```yaml
+servers:
+- port: 8080
+  ip: 0.0.0.0
+  tls:
+    certPath: /foo/cert.pem
+    keyPath: /foo/key.pem
+  maxConcurrentRequests: 1000
+  announce:
+    - /#/io.l5d.fs/web
```
adleong: Nothing wrong with this but we don't currently have an announcer that can announce `/#/io.l5d.fs` names.  So far we just have one for `/#/io.l5d.serversets`.

adleong:
:star: it's a bit of a bummer that links will break when viewing the docs on github (https://github.com/BuoyantIO/linkerd/blob/master/linkerd/docs/config.md).  But I guess probably no one does that anyway. 👍

PR #623, opened 2016-09-02
Keep Dtab obervation activities open

Ashald:
For better performance: resolve endpoint goes from 300 qps to 14k qps.

Many thanks to @adleong !

### DIFF
--- a/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
+++ b/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
@@ -343,8 +343,8 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
       case _ => bindingCache.get(NsPath(ns, path)).activity
     }
 
-  case class NsPath(ns: Ns, path: Path)
-  case class NameActClose(activity: Activity[NameTree[Name.Bound]], closable: Closable)
+  private[this] case class NsPath(ns: Ns, path: Path)
+  private[this] case class NameActClose(activity: Activity[NameTree[Name.Bound]], closable: Closable)
 
   private[this] val bindingCache = CacheBuilder.newBuilder()
     .maximumSize(bindingCacheSize)
@@ -382,7 +382,7 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
     }
   }
 
-  case class AddrActClose(activity: Activity[Addr], closable: Closable)
+  private[this] case class AddrActClose(activity: Activity[Addr], closable: Closable)
 
   private[this] val addrCache = CacheBuilder.newBuilder()
     .maximumSize(addrCacheSize)
--- a/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
+++ b/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
@@ -343,24 +343,23 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
       case _ => bindingCache.get(NsPath(ns, path)).activity
     }
 
-
   case class NsPath(ns: Ns, path: Path)
-  case class ActClose(activity: Activity[NameTree[Name.Bound]], closable: Closable)
+  case class NameActClose(activity: Activity[NameTree[Name.Bound]], closable: Closable)
 
   private[this] val bindingCache = CacheBuilder.newBuilder()
     .maximumSize(bindingCacheSize)
-    .removalListener(new RemovalListener[NsPath, ActClose] {
-      override def onRemoval(notification: RemovalNotification[NsPath, ActClose]): Unit = {
+    .removalListener(new RemovalListener[NsPath, NameActClose] {
+      override def onRemoval(notification: RemovalNotification[NsPath, NameActClose]): Unit = {
         val _ = notification.getValue.closable.close()
       }
     })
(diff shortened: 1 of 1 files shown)


namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -341,15 +341,28 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
   private[this] def getBind(ns: String, path: Path, extraDtab: Option[String]): Activity[NameTree[Name.Bound]] =
     extraDtab match {
       case Some(dtab) => delegate(ns).bind(Dtab.read(dtab), path)
-      case _ => bindingCache.get((ns, path))
+      case _ => bindingCache.get((ns, path))._1
     }
 
+  type NsPath = (String, Path)
```
adleong: what about case classes instead of type aliases for both of these?  that way the fields can be referred to by name instead of by `._1`

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -341,15 +341,28 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
...
+
   private[this] val bindingCache = CacheBuilder.newBuilder()
     .maximumSize(bindingCacheSize)
-    .build[(String, Path), Activity[NameTree[Name.Bound]]](
-      new CacheLoader[(String, Path), Activity[NameTree[Name.Bound]]] {
-        override def load(key: (String, Path)): Activity[NameTree[Bound]] =
-          delegate(key._1).bind(Dtab.empty, key._2)
+    .removalListener(new RemovalListener[NsPath, ActClose] {
+      override def onRemoval(notification: RemovalNotification[(String, Path), (Activity[NameTree[Bound]], Closable)]): Unit = {
+        if (notification.wasEvicted()) {
```
adleong: I think we want to close the observation for any kind of removal, not just evictions

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -341,15 +341,28 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
...
+          val _ = notification.getValue._2.close()
+        }
+      }
+    })
+    .build[NsPath, ActClose](
+      new CacheLoader[NsPath, ActClose] {
+        override def load(key: NsPath): ActClose = {
+          val act = delegate(key._1).bind(Dtab.empty, key._2)
+          val closable = act.run.changes.respond(_ => ())
```
adleong: maybe add a comment here explaining that this is a handle to keep the activity observed as long as it's in the cache.

Ashald:
Updated PR according to comments.

Ashald:
Tests failed but the failure looks unrelated. :/

Ashald:
C'mon, CircleCI, tests are 💯 . :)

adleong:
This is awesome, thanks for taking this on, @Ashald.  Do you think the Addr Cache should get the same treatment as the Bind Cache?

Ashald:
Hm, in theory - yes. In practice - let me try to implement it and see the performance impact.
Do you want that as addition to this PR or as a new one?

adleong:
If it makes sense to do then we might as well do it at the same time.

Ashald:
Yeah, this gives 20K QPS almost constantly. I'll update PR over weekend - hope you won't do release today. :)

The only question - how to name those case classes? :) `NameActClsoe` and `AddrActClos`? :)

Ashald:
Actually was able to do that while riding in metro. :) How does it look to you?

adleong:
:star: Looks great!

Ashald:
Yaaay! :)

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -341,15 +340,27 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
   private[this] def getBind(ns: String, path: Path, extraDtab: Option[String]): Activity[NameTree[Name.Bound]] =
     extraDtab match {
       case Some(dtab) => delegate(ns).bind(Dtab.read(dtab), path)
-      case _ => bindingCache.get((ns, path))
+      case _ => bindingCache.get(NsPath(ns, path)).activity
     }
 
+  case class NsPath(ns: Ns, path: Path)
+  case class NameActClose(activity: Activity[NameTree[Name.Bound]], closable: Closable)
```
olix0r: can these be `private`?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlService.scala
```diff
@@ -371,22 +382,33 @@ class HttpControlService(storage: DtabStore, delegate: Ns => NameInterpreter, na
     }
   }
 
+  case class AddrActClose(activity: Activity[Addr], closable: Closable)
```
olix0r: private?

olix0r:
lgtm! though, i'd prefer that we use the least permissive visibility for intermediate types.

Ashald:
@olix0r does it look better now? :)

olix0r:
thanks, @Ashald

PR #624, opened 2016-09-03
Use l5d-dst-concrete as request name

adleong:
If linkerd receives a request with the l5d-dst-concrete header set, use that as the request's name.

This can be useful in setups where a request passes through multiple linkerds (such as linkerd-to-linkerd) and the dst assigned by the first linkerd should be used by the second.

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/Identifier.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/Identifier.scala
@@ -1,0 +1,16 @@
+package io.buoyant.router.http
+
+import com.twitter.finagle.buoyant.Dst
+import com.twitter.finagle.http.Request
+import com.twitter.finagle.{Dtab, Path}
+import com.twitter.util.{Future, Try}
+
+object Identifier {
+  def dstConcrete(req: Request, baseDtab: () => Dtab): Option[Future[(Dst, Request)]] = {
+    req.headerMap.get("l5d-dst-concrete").flatMap { dst =>
+      Try(Path.read(dst)).toOption
+    }.map { dst =>
+      Future.value((Dst.Path(dst, baseDtab()), req))
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


router/http/src/main/scala/io/buoyant/router/http/MethodAndHostIdentifier.scala
```diff
@@ -24,21 +24,23 @@ case class MethodAndHostIdentifier(
...
-    case http.Version.Http10 =>
-      Future.value(
-        (mkPath(Path.Utf8("1.0", req.method.toString) ++ suffix(req)), req)
-      )
-
-    case http.Version.Http11 =>
-      req.host match {
-        case Some(host) if host.nonEmpty =>
+  def apply(req: http.Request): Future[(Dst, http.Request)] =
+    Identifier.dstConcrete(req, baseDtab).getOrElse {
```
olix0r: This seems a little surprising to me. If I specify a prefix on the identifier, I'd expect this prefix to apply to all names provided by the identifier.

router/http/src/main/scala/io/buoyant/router/http/Identifier.scala
```diff
@@ -0,0 +1,16 @@
+package io.buoyant.router.http
+
+import com.twitter.finagle.buoyant.Dst
+import com.twitter.finagle.http.Request
+import com.twitter.finagle.{Dtab, Path}
+import com.twitter.util.{Future, Try}
+
+object Identifier {
+  def dstConcrete(req: Request, baseDtab: () => Dtab): Option[Future[(Dst, Request)]] = {
+    req.headerMap.get("l5d-dst-concrete").flatMap { dst =>
```
olix0r: but this isn't actually concrete, is it? It's going to be applied through a dtab.

olix0r: Also, we should clear this header if we read it.

olix0r:
It's definitely my preference that we not add this behavior to the existing identifiers -- it makes it harder to reason about them, as it allows for the production of highly irregular names, it's unclear what `consume` means, etc...

I'd prefer that we be able to express composed identifiers like:

```
identifier:
  kind: list
  identifiers:
  - kind: header
    header: my-special-header # override default
    prefix: /h
  - kind: path
    prefix: /p
```

This allows the identifiers to stay to doing one thing well.

adleong:
Closed in favor of https://github.com/BuoyantIO/linkerd/pull/638

PR #622, opened 2016-09-02
Handle Interruption Failures for Consul Namer/DtabStore

Ashald:
This prevents messages like:
```
E 0902 17:02:01.067 THREAD18 TraceId:4fa6e8ae5cc43780: Exception propagated to the root monitor!
Failure(Consul observation released, flags=0x02)...
```

PR #625, opened 2016-09-03
Simplify future interruption in ConsulDtabStore

Ashald:
As interruption failures are handled properly now there is no more need in
`@volatile var` - a follow-up after #622


olix0r:
The running flags are necessary.  Future cancellation is advisory.  If the following state arises, this would spin forever:
1. Open var observation, starting a request
2. Receive response
3. Close observation while in handler

Future cancellation doesn't actively stop a thread from executing. It simply notifies the underlying Promise that the future does not need to be satisfied.  The flags are necessary to ensure that work eventually stops.

Ashald:
But throwing an exception is essential the same, isn't it? Also, given the fact the default "timeout" (in Consul API long-polling) is either 5 or 10 mins, it's better to close connections earlier rather than later so we won't accumulate lots of "slowly-dying zombie connections".

Ashald:
And if not - we probably can wrap handler into `try-catch`.

olix0r:
Nope.  Consider what happens if the future is already satisfied when the cancellation happens.  We're executing code in a thread with the satisfied value.  That thread is not killed.  There's no way to force an exception to be raised in that case.

It's idiomatic (i.e. in finagle) to use guards like this.

PR #629, opened 2016-09-07
Useful toString for ConsulApiError trait

Ashald:
It gives log messages like:
```
  retrying consul request on error Forbidden(403: rpc error: ACL not found)
```
rather than
```
  retrying consul request on error io.buoyant.consul.v1.package$Forbidden
```


adleong:
:star: seems reasonable

Ashald:
Yaaay! Thanks! :)

PR #630, opened 2016-09-07
Graceful shutdown

erdody:
Goal: add a way to handle termination signals and gracefully close routers, announcers, etc.

There's still a pre-existing issue (reproducible if you use /admin/shutdown) that prevents closing log lines from showing up.


olix0r:
Thanks for the patch, @erdody!

We had tried to do something like this in twitter-server, but it turns out that it's extremely unreliable. VM shutdown hooks run in a dedicated thread that is run as/after all other threads are killed.  Because finagle relies on netty thread pools, shutdown hooks don't allow for any sort of reliable cleanup action on clients or servers.  The /admin/shutdown admin endpoint is unfortunately the _only_ way to ensure graceful termination.

I'm not vehemently opposed to this, but I don't think it's going to work as you hope.

erdody:
Thanks @olix0r for the context.
Can you elaborate on what exactly is unreliable? Is it the triggering of the hook or the execution?
My main goal is to have a way to cleanly shutdown on a SIGTERM signal. Of course, SIGKILL won't be captured and there's not much you can or want to do about it. 
Specific example: I want to use linkerd as an in-container sidecar. On shutdown, init frameworks (s6-overlay / phusion) will kill all side process and wait for a graceful termination. I can write a specific shutdown script that calls /admin/shutdown and then waits for the process to die, but I
'd rather move that logic to the process.

olix0r:
I can try to elaborate, though I'm not 

Here's the commit where we took addShutdownHook out of com.twitter.app.App: https://github.com/twitter/util/commit/f71efa3f5a699c1c42d2a8bb33ee36caddc0dec8

(This is a few years ago, so I'm trying to shake the cobwebs off and rationalize why we did this).

As I recall, the root of the issue is that we wanted to simply log some information at shutdown, but even logging facilities may not be available from the context of a shutdown thread (presumably because some other shutdown hook is racing to tear down that resource).

I don't want to stand on FUD to block a feature that might improve things, though.  We should test your branch a bit more to see where it encounters problems, if it all.  If it _doesn't_ cause problems, this may even be something worth feeding back to twitter/util.

olix0r:
I've mocked out an App:
```scala

object ShutterDowner extends App {
  private[this] val log = Logger.get("ShutterDowner")

  def main() {
    val client = Http.newService("/$/inet/linkerd.io/80")
    val server = Http.serve("127.1:8123", Service.mk[http.Request, http.Response] { req =>
      client(req).map { rsp =>
        rsp.headerMap("awesome") = "yep"
        rsp
      }
    })
    closeOnExit(server)
    closeOnExit(client)

    sys.addShutdownHook {
      log.info("closing")
      Await.result(close(defaultCloseGracePeriod))
      log.info("closed")
    }

    log.info("running")
    Await.result(server)
    log.info("completed")
  }

}
```

Then I put some load on the service:
```
:; while curl -vso /dev/null http://localhost:8123 ; do ; done
``` 

Here are two runs:

```
:; java -cp test.jar io.buoyant.ShutterDowner
SNAPSHOT.jar io.buoyant.ShutterDowner        
Sep 08, 2016 11:54:49 PM com.twitter.finagle.Init$$anonfun$1 apply$mcV$sp
INFO: Finagle version 6.36.0 (rev=7efeb4cc2babd7c99731090fa76ad960627fce14) built at 20160708-090553
Sep 08, 2016 11:54:50 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[inet] = com.twitter.finagle.InetResolver(com.twitter.finagle.InetResolver@5443d039)
Sep 08, 2016 11:54:50 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[fixedinet] = com.twitter.finagle.FixedInetResolver(com.twitter.finagle.FixedInetResolver@3d1848cc)
Sep 08, 2016 11:54:50 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[neg] = com.twitter.finagle.NegResolver$(com.twitter.finagle.NegResolver$@7dda48d9)
Sep 08, 2016 11:54:50 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[nil] = com.twitter.finagle.NilResolver$(com.twitter.finagle.NilResolver$@6e4566f1)
Sep 08, 2016 11:54:50 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[fail] = com.twitter.finagle.FailResolver$(com.twitter.finagle.FailResolver$@4b6e2263)
Sep 08, 2016 11:54:50 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[flag] = com.twitter.server.FlagResolver(com.twitter.server.FlagResolver@344f4dea)
Sep 08, 2016 11:54:50 PM io.buoyant.ShutterDowner$ main
INFO: running
^C%
:;                                                                                                                                                                                                                                             ```
```
:; java -cp test.jar io.buoyant.ShutterDowner
Sep 08, 2016 11:55:09 PM com.twitter.finagle.Init$$anonfun$1 apply$mcV$sp
INFO: Finagle version 6.36.0 (rev=7efeb4cc2babd7c99731090fa76ad960627fce14) built at 20160708-090553
Sep 08, 2016 11:55:10 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[inet] = com.twitter.finagle.InetResolver(com.twitter.finagle.InetResolver@5443d039)
Sep 08, 2016 11:55:10 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[fixedinet] = com.twitter.finagle.FixedInetResolver(com.twitter.finagle.FixedInetResolver@3d1848cc)
Sep 08, 2016 11:55:10 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[neg] = com.twitter.finagle.NegResolver$(com.twitter.finagle.NegResolver$@7dda48d9)
Sep 08, 2016 11:55:10 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[nil] = com.twitter.finagle.NilResolver$(com.twitter.finagle.NilResolver$@6e4566f1)
Sep 08, 2016 11:55:10 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[fail] = com.twitter.finagle.FailResolver$(com.twitter.finagle.FailResolver$@4b6e2263)
Sep 08, 2016 11:55:10 PM com.twitter.finagle.BaseResolver$$anonfun$resolvers$1 apply
INFO: Resolver[flag] = com.twitter.server.FlagResolver(com.twitter.server.FlagResolver@344f4dea)
Sep 08, 2016 11:55:11 PM io.buoyant.ShutterDowner$ main
INFO: running
^CSep 08, 2016 11:55:31 PM io.buoyant.ShutterDowner$$anonfun$main$1 apply$mcV$sp
INFO: closing
```

So, we're not even guaranteed to see the `closing` message -- I'm unsure this is just because logging is being lost or because the code isn't even running.

So, shutdown hooks definitely introduce some level of unpredictability...

**Comment by erdody** [comment]
Thanks @olix0r for the detailed reply.
I noticed the logging problems, which is what I mention in my original comment. After further investigation I could find the root cause: java.util.logging.LogManager (which is what twitter logging uses) adds a shutdown hook to close handlers (reported in many places, e.g. [this bug](https://bugs.openjdk.java.net/browse/JDK-8161253)).
Since all shutdown hooks are run in parallel, there's a race condition and other hooks log lines might be ignored. The closing code is otherwise executed.

There are many workarounds to the logging problem, but I agree, even if we solve this specific issue there could be other misbehaving libraries (e.g. log4j2 seems to have a similar problem and I saw com.twitter.finagle.zipkin.core.TracerCache also using shutdown hooks, which might not be a major issue).

One thing I just successfully tried is using sun.misc.Signal.handle, which is run before all shutdown hooks. 
Something like:

```scala
  def main() {
    Signal.handle(new Signal("INT"), shutdownHandler)
    Signal.handle(new Signal("TERM"), shutdownHandler)
```
```scala
  private def shutdownHandler(): SignalHandler = new SignalHandler() {
    def handle(sig: Signal): Unit = {
      log.info("Closing all ...")
      Await.result(close(Duration.fromSeconds(graceFlag())))
      log.info("All closed.")
    }
  }
```

Of course it has the downside of not being officially supported which means it can go away (unlikely) and it might not work in all OSes. 
I still think it's better than just dying when the OS kindly asks the process to gracefully terminate :)
What do you think about this option?

olix0r:
I like the signal handler approach! Let's do it.

I think I'd prefer that the grace value come from the `admin` configuration block than from a flag -- we're trying to make the linkerd configs as self-contained as possible and not rely on a mix of flags and config options.  If this seems inconvenient for your use case, it's not a hard requirement.

erdody:
Good, I just changed to SignalHandler and moved parameter to admin section in the config file.

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -1,11 +1,14 @@
...
+import com.twitter.util.{Await, Duration}
+import io.buoyant.admin.{AdminInitializer, App}
 import io.buoyant.linkerd.Linker.LinkerConfig
 import io.buoyant.linkerd.admin.LinkerdAdmin
 import io.buoyant.linkerd.{Build, Linker}
 import java.io.File
+
+import sun.misc.{Signal, SignalHandler}
+
 import scala.io.Source
```
olix0r: omit empty lines, and sort

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -76,4 +82,27 @@ object Linkerd extends App {
...
+   *   <li>LogManager uses a shutdown hook which makes nothing to be logged during shutdown
+   *   <li>TracerCache uses a shutdown hook to flush
+   * </ul>
+   */
+  private def registerTerminationSignalHandler(shutdownGraceMs: Int): Unit = {
+    val shutdownHandler = new SignalHandler {
+      override def handle(sig: Signal): Unit = {
+        log.info("Closing linkerd ...")
+        Await.result(close(Duration.fromMilliseconds(shutdownGraceMs)))
+        log.info("linkerd closed.")
```
olix0r: I'd prefer to omit this. Doesn't add much information

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -76,4 +82,27 @@ object Linkerd extends App {
...
+   * Shutdown hook is not used because it has, at least, the following problems:
+   * <ul>
+   *   <li>LogManager uses a shutdown hook which makes nothing to be logged during shutdown
+   *   <li>TracerCache uses a shutdown hook to flush
+   * </ul>
+   */
+  private def registerTerminationSignalHandler(shutdownGraceMs: Int): Unit = {
+    val shutdownHandler = new SignalHandler {
+      override def handle(sig: Signal): Unit = {
+        log.info("Closing linkerd ...")
```
olix0r: I'd prefer that this said something like: `Received SIGNALNAME. Shutting down.`

olix0r:
This is starting to look really good!  Left a few comments.

erdody:
Thanks, just pushed the updates.

admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
```diff
@@ -2,4 +2,4 @@ package io.buoyant.admin
 
 import io.buoyant.config.types.Port
 
-case class AdminConfig(port: Port = Port(9990))
+case class AdminConfig(port: Port = Port(9990), shutdownGraceMs: Int = 10000)
```
adleong: I would make `shutdownGraceMs` optional by making it an `Option[Int]`.  In `Linkerd.scala` we can apply a default value and convert to a `Duration`.

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -76,4 +80,26 @@ object Linkerd extends App {
...
+    val shutdownHandler = new SignalHandler {
+      override def handle(sig: Signal): Unit = {
+        log.info("Received %s. Shutting down ...", sig)
+        Await.result(close(Duration.fromMilliseconds(shutdownGraceMs)))
+      }
+    }
+
+    Signal.handle(new Signal("INT"), shutdownHandler)
+    Signal.handle(new Signal("TERM"), shutdownHandler)
+    ()
```
adleong: I don't think this is needed

erdody: If I don't add that, sbt linkerd/assembly complains about:
```
[error] /Users/diego/Repo/linkerd-medallia/linkerd/main/src/main/scala/io/buoyant/Linkerd.scala:104: discarded non-Unit value
[error]     Signal.handle(new Signal("TERM"), shutdownHandler)
[error]                  ^
[error] one error found
```

One alternative is to add a val _ = ...

adleong: Ah, I see.  `val _ =` is my personal preference to explicitly indicate value discard.

admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
erdody: Sure, I was trying to emulate what you do with port. Any reason why you want to it differently in this case?

olix0r:
@erdody Can you please sign our [CLA](https://buoyant.io/cla)? Thanks!

admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
olix0r: Just a note to answer your question -- AdminConfig.port is probably wrong (throws NPEs) as it is here. It should probably be an `Option[Port]`.  Don't worry about fixing it here, I have follow branches that clean up admin initialization.

erdody:
@olix0r CLA signed, finally. I just synced with master.

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -1,11 +1,14 @@
...
-import com.twitter.util.{Await, Awaitable, Closable, CloseAwaitably, Future, Return, Throw, Time}
+import com.twitter.util._
 import io.buoyant.admin.App
 import io.buoyant.linkerd.admin.LinkerdAdmin
 import io.buoyant.telemetry.CommonMetricsTelemeter
 import java.io.File
+
+import sun.misc.{Signal, SignalHandler}
+
 import scala.io.Source
```
olix0r: please alphasort with no spaces:

```scala
import java.io.File
import scala.io.Source
import sun.misc....
```

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -104,4 +110,25 @@ object Main extends App {
...
+   * <ul>
+   *   <li>LogManager uses a shutdown hook which makes nothing to be logged during shutdown
+   *   <li>TracerCache uses a shutdown hook to flush
+   * </ul>
+   */
+  private def registerTerminationSignalHandler(shutdownGraceMs: Option[Int]): Unit = {
+    val shutdownHandler = new SignalHandler {
+      override def handle(sig: Signal): Unit = {
+        log.info("Received %s. Shutting down ...", sig)
+        val closeTimeOut = shutdownGraceMs.map(Duration.fromMilliseconds(_)).getOrElse(Duration.fromSeconds(10))
```
olix0r: tioli: const-ify this default somewhere so this is
```scala
shutdownGraceMs.map(Duration.fromMilliseconds(_)).getOrElse(DefaultShutdownGrace)
```

erdody: Added to the Main class

olix0r:
:star: fantastic! I'll merge this once it passes CI.

PR #631, opened 2016-09-07
ConsulNamer reorganization

olix0r:
* Split ConsulNamer into several separate classes.  It was a lot in one file, especially the caching layers.
* Fix cache `index` thread safety
* Eliminate redundant `localAgent()` lookups; use retries for localAgent lookups.
* Eliminate lookup-time checks for tag naming.  Instead, choose the appropriate namer at configuration-time.
* Use the most restrictive access modifiers to help reason about scope.
* Reduce allocations where possible.

This change is staged in several commits.  It may be more helpful to review them in order:
```
commit 25cbb6849aeb4c5dc09e33b3cb79a1ec55b98e74
Author: Oliver Gould <ver@buoyant.io>
Date:   Wed Sep 7 03:05:48 2016 +0000

    Move SvcCache and DcCache into their own files.
    
    The consul namer is a little difficult to reason about as one big intermingled
    class. We begin to simplify this by moving DcCache and SvcCache into their own
    package-local classes.
```

```
commit 4b86b9dd69dc673716a2320696c8b53b2c587fb2
Author: Oliver Gould <ver@buoyant.io>
Date:   Wed Sep 7 04:28:21 2016 +0000

    Stop issuing redundant local agent requests.
    
    Previously, the domain was resolved from consul on every service addition.
    Instead, this should be modeled as a single activity that is
    
    Furthermore, on DcCache updates, the dc cache was updated for each individual
    service addition and removal.  Instead, we batch the DC update to be only once
    per API response.
    
    Furthermore, access & update of `index` variables was not thread safe for
    visibility. `index` should ve volatile.
    
    Finally, some minor allocation improvements were made:
    - val'ing some anonymous functions, where convenient
    - Only build address metadata for bound addresses
```

```
commit 495af8bcc885b8daeba88a812e7f55f89f4ee7d8
Author: Oliver Gould <ver@buoyant.io>
Date:   Wed Sep 7 05:27:57 2016 +0000

    Split ConsulNamer into Tagged, Untagged, and Base.
    
    The ConsulNamer doesn't need to determine whether a tag should be included on
    each and every lookup. Instead, we can determine this once at
    configuration-time and construct a Namer that is statically configured to
    produce service keys.
    
    ConsulNamer.Base has all of the common logic shared between namer
    implementations.
```

```
commit 6c24defcfdc74b0e9ea51b88b0871a98d67f8b8a
Author: Oliver Gould <ver@buoyant.io>
Date:   Wed Sep 7 06:26:23 2016 +0000

    ConsulNamer.Base => LookupCache
```

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
@@ -1,0 +1,79 @@
+package io.buoyant.namer.consul
+
+import com.twitter.finagle._
+import com.twitter.util._
+import io.buoyant.consul.v1
+
+/**
+ * A helper supporting service resolution in consul, caching results
+ * and watching for new updates.
+ */
+private[consul] class LookupCache(
+  consulApi: v1.ConsulApi,
+  agentApi: v1.AgentApi,
+  setHost: Boolean = false
+) {
+
+  def apply(
+    dcName: String,
+    svcKey: SvcKey,
+    id: Path,
+    residual: Path
+  ): Activity[NameTree[Name]] = {
+    log.debug("consul lookup: %s %s", id.show)
+    watchDc(dcName).map { services =>
+      services.get(svcKey) match {
+        case None =>
+          log.debug("consul dc %s service %s missing", dcName, svcKey)
+          NameTree.Neg
+
+        case Some(addr) =>
+          log.debug("consul ns %s service %s found + %s", dcName, svcKey, residual.show)
+          NameTree.Leaf(Name.Bound(addr, id, residual))
+      }
+    }
+  }
+
+  private[this] lazy val domain: Activity[Option[String]] =
+    if (setHost) {
+      Activity.future(agentApi.localAgent(retry = true)).map { la =>
+        val dom = la.Config.flatMap(_.Domain).getOrElse("consul")
+        Some(dom.stripPrefix(".").stripSuffix("."))
+      }
+    } else Activity.value(None)
+
+  protected[this] def watchDc(dc: String): Activity[Map[SvcKey, Var[Addr]]] =
+    domain.flatMap { domain =>
+      Activity(Dc.get(dc, domain).services).map(_.mapValues(_.addrs))
+    }
+
+  /**
+   * Contains all cached responses from the Consul API
+   */
+  private[this] object Dc {
+    private[this] val activity: ActUp[Map[String, DcCache]] =
+      Var(Activity.Pending)
+
+    /**
+     * Returns existing datacenter cache with that name
+     * or creates a new one
+     */
+    def get(name: String, domain: Option[String]): DcCache =
+      synchronized {
+        activity.sample() match {
+          case Activity.Ok(snap) => snap.getOrElse(name, mkAndUpdate(snap, name, domain))
+          case _ => mkAndUpdate(Map.empty, name, domain)
+        }
+      }
+
+    private[this] def mkAndUpdate(
+      cache: Map[String, DcCache],
+      name: String,
+      domain: Option[String]
+    ): DcCache = {
+      val dc = new DcCache(consulApi, name, domain)
+      activity() = Activity.Ok(cache + (name -> dc))
+      dc
+    }
+  }
+}
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
@@ -3,7 +3,7 @@ package io.buoyant.namer.consul
 import com.twitter.finagle._
 import com.twitter.util._
 import io.buoyant.consul._
-import io.buoyant.consul.v1.{ServiceNode, UnexpectedResponse}
+import io.buoyant.consul.v1.{Indexed, ServiceNode, UnexpectedResponse}
 import io.buoyant.namer.Metadata
 
 private[consul] case class SvcKey(name: String, tag: Option[String]) {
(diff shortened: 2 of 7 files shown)


olix0r:
cc @Ashald

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
```diff
@@ -0,0 +1,98 @@
...
+      case (_, Some(ip), Some(port)) if !ip.isEmpty => Try(Address(ip, port)).toOption
+      case (Some(ip), _, Some(port)) if !ip.isEmpty => Try(Address(ip, port)).toOption
+      case _ => None
+    }
+  }
+
+  private[this] val handleUnexpected: PartialFunction[Throwable, Unit] = {
+    case e: ChannelClosedException =>
+      // XXX why doesn't this clear out the addr?
+      log.error(s"""lost consul connection while querying for $datacenter/$key updates""")
```
olix0r: open question

Ashald: IMO it should stay as is - there is no evidence that requested service is unavailable - Consul agent may be just being restarted.
OTOH maybe it's better to put this into `retryHandler`?

Ashald:
Wow, this is amazing! I'll take a more detailed look a bit later today. One of things that, frankly, I didn't like from the very beginning is using a `var index` in caches - _IMHO_ a "recursive future" (as in DtabStore) serves such purpose better.

Also I have a small guess that there might be another issue in caches (although I'm not 100% sure). I did number of load tests on Friday involving CatalogNamer and left `namerd` running over the weekend - on Tuesday we noticed few 10Ks of open connections reported by Nagios on that host. As I understand, neither DcCache nor SvcCache are explicitly closing "watch process" - IIRC Alex mentioned something similar on Friday. I wonder whether fetching can be done with closable `Var.async` and those caches can be done in similar way to what we have in `HttpControlService` - CacheBuilder-based cache that keeps observations open but stopes them upon eviction.

Just few ideas.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
```diff
@@ -0,0 +1,98 @@
...
+  private[this] def run(): Future[Unit] =
+    mkRequest().flatMap(update).handle(handleUnexpected)
+
+  def update(nodes: Seq[v1.ServiceNode]): Future[Unit] = {
+    val addr = nodes.flatMap(serviceNodeToAddr) match {
+      case addrs if addrs.isEmpty => Addr.Neg
+      case addrs =>
+        val meta = domain match {
+          case None => Addr.Metadata.empty
+          case Some(domain) =>
```
Ashald: For my own reference: do you care about name overlapping in situations like this one? I mean the ""outer `Option` called `domain` and the inner `domain` value as a `String`. I saw such pattern number of times through linkerd codebase and used to consider this as error-prone pattern but maybe in Scala world it's not the case. :)

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
olix0r: I think if we hit this code is only reached when the loop has completed and there's nothing here that resumes looping, so this probably doesn't behave as expected. I've changed this in a followup branch.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
olix0r: This is common and intentional throughout Finagle (and while we don't use Finagle's style as dogma, I find it's convenient for the projects to stay aligned).

When we have box types like Option, it's preferred to simply use a name like `domain` rather than `domainOpt` or `maybeDomain` etc.  Because we're just unboxing the value, it's fine to shadow the name because it signals the intent of "just unboxing this thing".  I suppose it'd also be fine to use a temp val like `d` here.

olix0r:
@Ashald hah! i already converted SvcCache to SvcAddr in a followup branch: https://github.com/BuoyantIO/linkerd/tree/ver%2Fconsul-lazy-services.

I think it's possible that we're currently leaking request loops; and we're certainly eagerly starting requests for services we possibly don't care about.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
Ashald: By saying loop you mean:
```
private[this] def run(): Future[Unit] =
  mkRequest().flatMap(update).handle(handleUnexpected)
```
?
For me it looks like it will brake on 1st unexpected error: next cycle is supposed to be triggered within `update` while `flatMap` executes given function only when the original future succeeds. Or am I wrong?

olix0r: right, so if `mkRequest()` throws a `ChannelClosedException`, `handleUnexpected` will log a message but no further requests will be issued.

I think this is a moot point, though, since retries are handled elsewhere and this exception should never be encountered here.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
Ashald: Thanks for explanations!

Ashald:
Great, I have few questions/comments about it as well - please cc me on the PR once it's ready. :)

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcCache.scala
Ashald: Although such "loop break" can occur on any error that is not considered re-triable, right? For instance `403` HTTP response... My concern here is that as long as we're not caching and persisting activities we don't care much as failure-recovery would happen on its own. But if those activities are cached then it's a bit more tricky. One of the possible solutions that I see is to implement "cache eviction" upon unexpected failure.

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -0,0 +1,79 @@
...
+          NameTree.Neg
+
+        case Some(addr) =>
+          log.debug("consul ns %s service %s found + %s", dcName, svcKey, residual.show)
+          NameTree.Leaf(Name.Bound(addr, id, residual))
+      }
+    }
+  }
+
+  private[this] lazy val domain: Activity[Option[String]] =
```
adleong: should this return a `Future` instead of an `Activity`?

olix0r: We'd just wrap that future in an Activity below to compose it in watchDc, so I don't know that it really makes a difference

adleong:
:star: looks like a good refactor to me.  a refactor of this size probably warrants some testing... do we have a good plan for that?

olix0r:
My test plan was "politely ask @Ashald to try this branch overnight..." Any other ideas?

Ashald:
Ha-ha, deal. :) Let me know when code is ready for a test-drive - I'll make a build and try it in our environment.

olix0r:
Since it seems that this is basically uncontroversial, I'm going to address error handling in the follow-up  #632

olix0r:
Merged in https://github.com/BuoyantIO/linkerd/commit/944a2f421f8710fcb9f835b6b46e8bbbbb32b295

PR #632, opened 2016-09-07
Lazy, managable consul resolution

olix0r:
Currently, both the DcCache and SvcCache have clunky lifecycle semantics -- services are resolved eagerly, and teardown logic is not robust enough to cancel ongoing work.

SvcCache becomes SvcAddr, exposing an `apply()` that returns a `Var[Addr]`. This var is computed lazily so that service observations are initiated on-demand and so that observations may be released later (i.e. when a client is closed).

DcCache becomes DcServices, exposing an `apply()` that returns an `Activity[Map[SvcKey, Var[Addr]]]`. This var is computed lazily so that datacenter observatoins are initiated on-demand and so that observations may be released later.

This reduces state in the consul observation layer. Furthermore, we now have explicit lifecycle management for these observations.

Depends on #631


olix0r:
attn @Ashald

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
```diff
@@ -0,0 +1,96 @@
...
+          case Throw(e@v1.UnexpectedResponse(rsp)) if rsp.status == Status.NotFound =>
+            // We assume this means the datacenter doesn't exist, so
+            // we just treat it as empty.  This will never be updated
+            // though.  If we care about this, we should schedule
+            // recovery with e.g.:
+            //   Future.sleep(t).before(loop(index0))
+            state() = Activity.Ok(Map.empty)
+            Future.exception(e)
+
+          case Throw(e) =>
```
adleong: probably worth noting that this is only reachable if getServices encounters a non-retryable error

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
```diff
@@ -0,0 +1,96 @@
...
+            state() = Activity.Failed(e)
+            Future.exception(e)
+
+          case Return(v1.Indexed(_, None)) =>
+            val e = NoIndexException
```
adleong: Just to confirm my understanding: we have to fail out here because without an index, we can't long poll?  What situation would cause this to happen?  Is recovery just impossible?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -42,23 +45,22 @@ private[consul] class LookupCache(
...
-    domain.flatMap { domain =>
-      Activity(Dc.get(dc, domain).services).map(_.mapValues(_.addrs))
-    }
-
   /**
    * Contains all cached responses from the Consul API
    */
   private[this] object Dc {
-    private[this] val activity: ActUp[Map[String, DcCache]] =
+    type Services = Activity[Map[SvcKey, Var[Addr]]]
```
adleong: I'm not sure if this type alias makes things more readable or less

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
olix0r: As far as I can tell, this is exceptional.  AFAIU, the consul api should always return an index.  This is more of a "we're not really talking to consul are we" situation.

I think we can't retry here because we'll just busy loop just issuing the same request over and over again.  Unless this is dictated as part of the consul API, think we should bail here.

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
```diff
@@ -0,0 +1,96 @@
...
+            cache.keys.foreach { k =>
+              if (!keys(k)) log.debug("consul deleted: %s", k)
+            }
+
+            val updated = keys.map { k =>
+              val svc = cache.get(k) match {
+                case Some(svc) => svc
+                case None =>
+                  log.debug("consul added: %s", k)
+                  SvcAddr(consulApi, name, k, domain)
```
adleong: Took me a couple of reads to get the point here:
The Var[Addr]s are created eagerly, but since they're Var.asyncs, they aren't started until observed.

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
Ashald: IIRC, 403 is considered non-retryable right now...

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
```diff
@@ -0,0 +1,96 @@
...
+        retry = true
+      ).map(toServices)
+
+    val states = Var.async[Activity.State[Map[SvcKey, Var[Addr]]]](Activity.Pending) { state =>
+      @volatile var stopped: Boolean = false
+
+      def loop(index0: Option[String], cache: Map[SvcKey, Var[Addr]]): Future[Unit] =
+        if (stopped) Future.Unit
+        else getServices(index0).transform {
+          case Throw(e@v1.UnexpectedResponse(rsp)) if rsp.status == Status.NotFound =>
```
Ashald: Wouldn't it be easier to look for `v1.NotFound` here instead?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -0,0 +1,104 @@
...
+    // the service for updates.
+    Var.async[Addr](Addr.Pending) { state =>
+      @volatile var stopped: Boolean = false
+
+      def loop(index0: Option[String]): Future[Unit] =
+        if (stopped) Future.Unit
+        else getAddresses(index0).flatMap {
+          case v1.Indexed(_, None) =>
+            // If consul doesn't return an index, we're in bad shape.
+            Future.exception(NoIndexException)
```
adleong: Should we set state to `Addr.Failed`?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -0,0 +1,104 @@
...
+      val addrs = nodes.flatMap(serviceNodeToAddr).toSet
+      v1.Indexed(addrs, idx)
+  }
+
+  /**
+   * Prefer service IPs to node IPs. Invalid addresses are ignored.
+   */
+  private val serviceNodeToAddr: v1.ServiceNode => Traversable[Address] = { n =>
+    (n.Address, n.ServiceAddress, n.ServicePort) match {
+      case (_, Some(ip), Some(port)) if !ip.isEmpty => Try(Address(ip, port)).toOption
```
Ashald: I'm just curious what `Try(...).toOption` gives us here compared to `Some`

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -0,0 +1,104 @@
...
+        else getAddresses(index0).flatMap {
+          case v1.Indexed(_, None) =>
+            // If consul doesn't return an index, we're in bad shape.
+            Future.exception(NoIndexException)
+
+          case v1.Indexed(addrs, index1) =>
+            val addr = addrs match {
+              case addrs if addrs.isEmpty => Addr.Neg
+              case addrs =>
+                val meta = domain match {
```
adleong: maybe factor this out into a method that produces a metadata from a service key?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
olix0r: it handles exceptions constructing Address (i.e. in case the IP or port is invalid)

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
olix0r: this is done below in pending -- but this is different from the other file, so perhaps better to just make this a big transform block like in DcServices

adleong: Or make DcServices use a final handle block.  Either way, I don't see a reason to not be consistent.

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
adleong: This might come up if consul was behind a reverse-proxy that served an error page if consul was having issues... or something like that.  Conceivably we could retry with backoff until indexes come back but... probably not worth it.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -0,0 +1,104 @@
...
+        retry = true
+      ).map(indexedToAddresses)
+
+    // Start by fetching the service immediately, and then long-poll
+    // the service for updates.
+    Var.async[Addr](Addr.Pending) { state =>
+      @volatile var stopped: Boolean = false
+
+      def loop(index0: Option[String]): Future[Unit] =
+        if (stopped) Future.Unit
```
Ashald: wouldn't it be more 'idiomatically correct' to do:
```
try { ... } catch { e: Failure if e.isFlagged(Failure.Interrupted) => Future.Unit }
```
? `@volatile var stopped` seems redundant for me. :/ Probably minor stuff.

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
Ashald: Also, it's wrong to expect `404` in such situation (as per comment below) because:
```
$ curl -i "http://localhost:8500/v1/catalog/services?dc=WRONG"
HTTP/1.1 500 Internal Server Error
X-Consul-Index: 0
X-Consul-Knownleader: false
X-Consul-Lastcontact: 0
Date: Wed, 07 Sep 2016 20:09:43 GMT
Content-Length: 32
Content-Type: text/plain; charset=utf-8

rpc error: No path to datacenter
```

Ashald: Which actually will be considered a retryable error - a valid behavior for those cases where connectivity between datacenters is briefly interrupted.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
olix0r: As discussed in #625, it's necessary to use guards to ensure that work stops eventually.

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
olix0r: Alright, though this means that the activity will never be satisfied for invalid datacenters...  I'll remove this error handling.  If we want to have a special-case handler for invalid dc, we should raise some dedicated exception type and handle it here.

Ashald: Yeah, unfortunately it doesn't look like there is a way to distinguish between non-existing datacenter and temporarily disconnected one. It's not clear whether `/v1/catalog/datacenters` lists only datacenters availbale right now or any "ever seen". I think it worth an investigation so maybe put a TODO for now?..

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
olix0r: What would you expect the namer to do if it gets a 403?

Ashald: I'd expect it to retry:
1) in our use-case that usually means misconfigured ACLs
2) you might see 403 when ACL datacenter is disconnected from current one

Although I can imagine situation where people might use ACLs with other purpose and in that case retry on 403 doesn't make sense. Another configuration flag?.. :)

olix0r: As discussed, we're going to fail the activity for now when any non-automatically-retryable exception is encountered.  We can revisit this (in which case, the ConsulApi should be modified to provide clearer exception types).

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
olix0r: Added a comment to this effect

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -35,42 +35,38 @@ private[consul] object SvcAddr {
...
-                  case None => Addr.Metadata.empty
-                  case Some(domain) =>
-                    val authority = key.tag match {
-                      case Some(tag) => s"$tag.${key.name}.service.$datacenter.$domain"
-                      case None => s"${key.name}.service.$datacenter.$domain"
-                    }
-                    Addr.Metadata(Metadata.authority -> authority)
-                }
-                Addr.Bound(addrs.toSet, meta)
+              case addrs => Addr.Bound(addrs, mkMeta(key, datacenter, domain))
```
adleong: use the previously computed `meta` val?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -35,42 +35,38 @@ private[consul] object SvcAddr {
         retry = true
       ).map(indexedToAddresses)
 
+    lazy val meta = mkMeta(key, datacenter, domain)
```
adleong: why lazy?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
olix0r: doh, yeah.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
olix0r: intent was that it may not actually be used if there are no endpoints.

adleong: seems pretty cheap and lazy vals make things so much harder to reason about.  probably not a big deal either way in this case but I'd shy away from lazy vals when possible on principle

olix0r:
@Ashald tested this branch in his environment and did not detect any regressions.

adleong:
:star: 🚢

PR #634, opened 2016-09-07
Disable endpoint probation by default

olix0r:
Probation can be hard to debug in environments like kubernetes and mesos.  Disable it by default to reduce unexpected behavior.


klingerf:
⭐ 

Would also recommend updating `linkerd/core/src/test/scala/io/buoyant/linkerd/LoadBalancerTest.scala` to set `enableProbation: true`, since the test is now testing setting the default value.

adleong:
:star: 🚭

klingerf:
⭐  Thanks for updating!

PR #636, opened 2016-09-07
Allow routers to be configured with a list of identifiers

adleong:
Allow routers to be configured with a list of identifiers.  If an identifier cannot assign a dest to a request, it falls back to the next one in the list.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -14,8 +14,9 @@ import com.twitter.finagle.service.Retries
 import com.twitter.finagle.{Path, Stack}
 import com.twitter.util.Future
 import io.buoyant.linkerd.protocol.http.{AccessLogger, ResponseClassifiers, RewriteHostHeader}
-import io.buoyant.router.RoutingFactory.{UnidentifiedRequest, IdentifiedRequest}
+import io.buoyant.router.RoutingFactory.{RequestIdentification, IdentifiedRequest, UnidentifiedRequest}
 import io.buoyant.router.{Http, RoutingFactory}
+import scala.collection.JavaConverters._
 
 class HttpInitializer extends ProtocolInitializer.Simple {
   val name = "http"
@@ -97,8 +98,7 @@ class HttpIdentifierConfigDeserializer extends JsonDeserializer[Option[Seq[HttpI
     val codec = p.getCodec
     codec.readTree[TreeNode](p) match {
       case n: JsonNode if n.isArray =>
-        import scala.collection.JavaConversions._
-        Some(n.toList.map(codec.treeToValue(_, classOf[HttpIdentifierConfig])))
+        Some(n.asScala.toList.map(codec.treeToValue(_, classOf[HttpIdentifierConfig])))
       case node => Some(Seq(codec.treeToValue(node, classOf[HttpIdentifierConfig])))
     }
   }
@@ -133,17 +133,24 @@ case class HttpConfig(
   override val protocol: ProtocolInitializer = HttpInitializer
 
   @JsonIgnore
-  private[this] val combinedIdentifier = identifier.map { identifierConfigs =>
-    Http.param.HttpIdentifier { (path, dtab) =>
-      identifierConfigs.map(_.newIdentifier(path, dtab))
-        .reduceLeft { (id0, id1) =>
-          { req =>
-            id0(req).flatMap {
-              case identified: IdentifiedRequest[Request] => Future.value(identified)
-              case unidentified: UnidentifiedRequest[Request] => id1(req)
-            }
-          }
+  private[this] val combinedIdentifier = {
+    // use the first identifier that yields an IdentifiedRequest
+    def combine(identifiers: Seq[RoutingFactory.Identifier[Request]]): RoutingFactory.Identifier[Request] = { req =>
+      identifiers.foldLeft(HttpConfig.NilIdentification) { (identification, next) =>
+        identification.flatMap {
+          // the request has already been identified, just use that
+          case identified: IdentifiedRequest[Request] => Future.value(identified)
+          // the request has not yet been identified, try the next identifier
+          case unidentified: UnidentifiedRequest[Request] => next(req)
         }
+      }
+    }
+
+    identifier.map { identifierConfigs =>
+      Http.param.HttpIdentifier { (prefix, dtab) =>
+        val identifiers = identifierConfigs.map(_.newIdentifier(prefix, dtab))
+        combine(identifiers)
+      }
     }
   }
 
@@ -159,4 +166,9 @@ case class HttpConfig(
     .maybeWith(streamingEnabled.map(hparam.Streaming(_)))
     .maybeWith(compressionLevel.map(hparam.CompressionLevel(_)))
 
+}
+
+object HttpConfig {
+  val NilIdentification: Future[RequestIdentification[Request]] =
+    Future.value(new UnidentifiedRequest("no identifiers"))
 }
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -2,13 +2,17 @@ package io.buoyant.linkerd
 package protocol
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.core.{JsonParser, TreeNode}
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.{DeserializationContext, JsonDeserializer, JsonNode}
 import com.twitter.conversions.storage._
-import com.twitter.finagle.{Path, Stack}
 import com.twitter.finagle.Http.{param => hparam}
-import com.twitter.finagle.buoyant.linkerd.{DelayedRelease, Headers, HttpTraceInitializer, HttpEngine}
+import com.twitter.finagle.buoyant.linkerd.{DelayedRelease, Headers, HttpEngine, HttpTraceInitializer}
 import com.twitter.finagle.client.{AddrMetadataExtraction, StackClient}
 import com.twitter.finagle.service.Retries
-import io.buoyant.linkerd.protocol.http.{RewriteHostHeader, AccessLogger, ResponseClassifiers}
+import com.twitter.finagle.{Path, Stack}
+import io.buoyant.linkerd.protocol.http.{AccessLogger, ResponseClassifiers, RewriteHostHeader}
+import io.buoyant.router.http.UnidentifiableRequestException
 import io.buoyant.router.{Http, RoutingFactory}
 
 class HttpInitializer extends ProtocolInitializer.Simple {
@@ -85,9 +89,23 @@ case class HttpServerConfig(
   }
 }
 
+class HttpIdentifierConfigDeserializer extends JsonDeserializer[Option[Seq[HttpIdentifierConfig]]] {
+  override def deserialize(p: JsonParser, ctxt: DeserializationContext): Option[Seq[HttpIdentifierConfig]] = {
+    val codec = p.getCodec
+    codec.readTree[TreeNode](p) match {
+      case n: JsonNode if n.isArray =>
+        import scala.collection.JavaConversions._
+        Some(n.toList.map(codec.treeToValue(_, classOf[HttpIdentifierConfig])))
+      case node => Some(Seq(codec.treeToValue(node, classOf[HttpIdentifierConfig])))
+    }
+  }
+
+  override def getNullValue(ctxt: DeserializationContext): Option[Seq[HttpIdentifierConfig]] = None
+}
+
 case class HttpConfig(
   httpAccessLog: Option[String],
-  identifier: Option[HttpIdentifierConfig],
+  @JsonDeserialize(using = classOf[HttpIdentifierConfigDeserializer]) identifier: Option[Seq[HttpIdentifierConfig]],
   maxChunkKB: Option[Int],
   maxHeadersKB: Option[Int],
   maxInitialLineKB: Option[Int],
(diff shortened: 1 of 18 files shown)


olix0r:
A question worth considering while we're here:

exceptions tend to be much costlier code paths than other graceful returns (i.e. stack tracing, etc).

We're admitting now that an unidentifiable request isn't so exceptional: we expect that identifiers won't always identify requests.

What if we made Identifiers return an ADT like:
```scala
sealed trait RequestIdentification
case class IdentifiedRequest(dst: Dst, request: Request) extends RequestIdentification
object UnidentifiedRequest extends RequestIdentification
```

This makes the interface much clearer, though it's an API breakage (worth it imo)

adleong:
`UnidentifierRequestException` mixes in `NoStackTrace`, but I agree with you that it's not the best to do control flow this way.  

And API break would be ideal, (either to ADT as your suggest or just plain Option) but I was hesitant because we know there are identifier plugins in the wild that would need to be updated.

adleong:
Still think the API break is worth it? 😉

olix0r:
It's my preference that we push forward with the API break and communicate the change.  I think this puts the Identifier in a much better place moving forward.

I have a slight preference for an ADT over an Option:
- Option[Tuple] screams for want of a proper type
- We could conceivable support additional states (idk what, just thinking)
- We could conceivably extend IdentifiedRequest with additional information (on that note, we probably should avoid case classes here -- case classes cause all field additions to be API breakages (on unapply).

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -3,22 +3,26 @@ package io.buoyant.router
...
 
-  case class UnknownDst[Req](request: Req, cause: Throwable)
-    extends Exception(s"Unknown destination: $request / ${cause.getMessage}", cause)
+  case class UnknownDst[Req](request: Req, reason: String)
+    extends Exception(s"Unknown destination: $request / $reason")
     with NoStackTrace
 
+  sealed trait RequestIdentification[Req]
+  class IdentifiedRequest[Req](val dst: Dst, val request: Req) extends RequestIdentification[Req]
+  class UnidentifiedRequest[Req](val reason: String, val request: Req) extends RequestIdentification[Req]
```
olix0r: I'm wary of having Unidentified include a request -- this seems to indicate that failed identification can/should be able to modify requests which seems like something we don't want? (yes yes i know about our awful http types).

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -3,22 +3,26 @@ package io.buoyant.router
...
 
 object RoutingFactory {
   val role = Stack.Role("RoutingFactory")
   val description = "Performs per-request name binding"
 
-  case class UnknownDst[Req](request: Req, cause: Throwable)
-    extends Exception(s"Unknown destination: $request / ${cause.getMessage}", cause)
+  case class UnknownDst[Req](request: Req, reason: String)
+    extends Exception(s"Unknown destination: $request / $reason")
     with NoStackTrace
```
olix0r: what's the motivation for keeping this exception around?

olix0r: nevermind, i see now.  Maybe a comment to indicate how it's used?

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
olix0r: this ADT probably deserves some `/** commentary */`

adleong: You're right, this is confusing because request is in `IdentifiedRequest` and `UnidentifiedRequest` for different reasons.  In `IdentifiedRequest`, the request is part of the output of identification.  In `UnidentifiedRequest` it's there for informational purposes.  I think we can remove it since the original request should be available in any place where you're getting an `UnidentifiedRequest`.

olix0r: tioli: you can provide unapply helpers for these (and we'll just never update them... or have to break API again). This is, for instance, how the Name ADT works: the Name types are classes that have companion objects with unapply

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -97,19 +101,24 @@ class RoutingFactory[Req, Rsp](
       }
 
       for {
-        (dst, req1) <- getDst(req0).rescue {
-          case e: Throwable =>
-            Annotations.Failure.Identification.record(e)
-            Future.exception(UnknownDst(req0, e))
+        identified <- getDst(req0).transform {
```
olix0r: tioli: if you have unapply, you can sugar this with `IdentifiedRequest(dst, req) <- getDst(req0).transform { ...`

router/http/src/main/scala/io/buoyant/router/http/MethodAndHostIdentifier.scala
```diff
@@ -24,21 +26,28 @@ case class MethodAndHostIdentifier(
   private[this] def mkPath(path: Path): Dst.Path =
     Dst.Path(prefix ++ path, baseDtab(), Dtab.local)
 
-  def apply(req: http.Request): Future[(Dst, http.Request)] = req.version match {
+  def apply(req: http.Request): Future[RequestIdentification[Request]] = req.version match {
     case http.Version.Http10 =>
       Future.value(
-        (mkPath(Path.Utf8("1.0", req.method.toString) ++ suffix(req)), req)
+        new IdentifiedRequest[Request](mkPath(Path.Utf8("1.0", req.method.toString) ++ suffix(req)), req)
```
olix0r: i'd guess the type annotation isn't needed here.

ALso I'd be inclined to have a `val path = mkPath(...)` so that you can fit this all one line with a bit more clarity

router/thrift/src/main/scala/io/buoyant/router/thrift/Identifier.scala
```diff
@@ -26,8 +27,8 @@ case class Identifier(
     }
   }
 
-  def apply(req: ThriftClientRequest): Future[(Dst, ThriftClientRequest)] =
+  def apply(req: ThriftClientRequest): Future[RequestIdentification[ThriftClientRequest]] =
     Future.value(
-      (Dst.Path(name ++ suffix(req), dtab(), Dtab.local), req)
+      new IdentifiedRequest[ThriftClientRequest](Dst.Path(name ++ suffix(req), dtab(), Dtab.local), req)
```
olix0r: similarly, i'd be inclined to capture `val dst` separately

router/http/src/main/scala/io/buoyant/router/http/MethodAndHostIdentifier.scala
```diff
@@ -1,9 +1,11 @@
 package io.buoyant.router.http
 
+import com.twitter.finagle.http.Request
 import com.twitter.finagle.{Dtab, Path, http}
 import com.twitter.finagle.buoyant.Dst
```
olix0r: sorting is weird here

olix0r: might consider just refering to http.Request in the code, since we refer to http.Version (or import Version, etc)

router/mux/src/main/scala/io/buoyant/router/Mux.scala
```diff
@@ -31,9 +32,9 @@ object Mux extends Router[Request, Response] with Server[Request, Response] {
       prefix: Path = Path.empty,
       dtab: () => Dtab = () => Dtab.base
     ) extends RoutingFactory.Identifier[Request] {
-      def apply(req: Request): Future[(Dst, Request)] =
+      def apply(req: Request): Future[RequestIdentification[Request]] =
         Future.value(
-          (Dst.Path(prefix ++ req.destination, dtab(), Dtab.local), req)
+          new IdentifiedRequest[Request](Dst.Path(prefix ++ req.destination, dtab(), Dtab.local), req)
```
olix0r: [Request] is probably inferrable

router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -12,16 +14,17 @@ case class PathIdentifier(
...
       case path if path.size >= segments =>
         if (consume) {
           req.uri = req.uri.split("/").drop(segments + 1).mkString("/", "/", "")
         }
+        val dst = Dst.Path(prefix ++ Path.Utf8(path.take(segments): _*), baseDtab(), Dtab.local)
+        Future.value(new IdentifiedRequest[Request](dst, req))
+      case _ =>
         Future.value(
-          (Dst.Path(prefix ++ Path.Utf8(path.take(segments): _*), baseDtab(), Dtab.local), req)
+          new UnidentifiedRequest("not enough segments in path")
```
olix0r: If we don't actually do any string formatting here, we can just make this exception static

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -3,22 +3,46 @@ package io.buoyant.router
...
   /**
    * An Identifier determines a [[com.twitter.finagle.buoyant.Dst
    * destination]] for `Req`-typed requests.
    */
-  type Identifier[Req] = Req => Future[(Dst, Req)]
+  type Identifier[Req] = Req => Future[RequestIdentification[Req]]
+
+  /** The result of attempting to identify a request. */
+  sealed trait RequestIdentification[Req]
+  /**
```
olix0r: i'd leave a little breathing room between the trait and the next comment

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -3,22 +3,46 @@ package io.buoyant.router
...
+   * @param dst The destination of the request.
+   * @param request The request to send to the destination.  This allows
+   *                identifiers to effectively mutate requests that they
+   *                identify.
+   */
+  class IdentifiedRequest[Req](val dst: Dst, val request: Req) extends RequestIdentification[Req]
+
+  object IdentifiedRequest {
+    def unapply[Req](identified: IdentifiedRequest[Req]): Option[(Dst, Req)] =
+      Some(identified.dst, identified.request)
```
olix0r: Surprised this works... I'd expect to have to do Some((a, b)) or Some(a -> b)... but if it compiles...

router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -1,9 +1,11 @@
 package io.buoyant.router.http
 
+import com.twitter.finagle.http.Request
 import com.twitter.finagle.{Dtab, Path, http}
 import com.twitter.finagle.buoyant.Dst
```
olix0r: sording is weird

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -3,22 +3,46 @@ package io.buoyant.router
...
+  }
+
+  /**
+   * This indicates that the identifier could not assign a destination to the
+   * request.
+   */
+  class UnidentifiedRequest[Req](val reason: String) extends RequestIdentification[Req]
```
olix0r: Another thought:

since this ADT is under RoutingFactory, it will be cumbersome to write Java identifiers (they'll have to use $s or, we'll have to provide `RequestIdentifications` compat builders

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -112,9 +133,24 @@ case class HttpConfig(
...
+  private[this] val combinedIdentifier = identifier.map { identifierConfigs =>
+    Http.param.HttpIdentifier { (path, dtab) =>
+      identifierConfigs.map(_.newIdentifier(path, dtab))
+        .reduceLeft { (id0, id1) =>
+          { req =>
+            id0(req).flatMap {
+              case identified: IdentifiedRequest[Request] => Future.value(identified)
+              case unidentified: UnidentifiedRequest[Request] => id1(req)
+            }
+          }
```
olix0r: we may want to break this into a few functions so it's easier to read

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -85,9 +91,24 @@ case class HttpServerConfig(
   }
 }
 
+// Cribbed from https://gist.github.com/Aivean/6bb90e3942f3bf966608
+class HttpIdentifierConfigDeserializer extends JsonDeserializer[Option[Seq[HttpIdentifierConfig]]] {
+  override def deserialize(p: JsonParser, ctxt: DeserializationContext): Option[Seq[HttpIdentifierConfig]] = {
+    val codec = p.getCodec
+    codec.readTree[TreeNode](p) match {
+      case n: JsonNode if n.isArray =>
+        import scala.collection.JavaConversions._
```
olix0r: fwiw JavaConverters is the recommended replacement for JavaConversions.  Supports `n.asScala`

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
olix0r: I think if we just had a `def compositeIdentifiers(id0, id1)`, this might read a lot clearer

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -124,3 +167,8 @@ case class HttpConfig(
     .maybeWith(compressionLevel.map(hparam.CompressionLevel(_)))
 
 }
+
+object HttpConfig {
+  val NilIdentification: Future[RequestIdentification[Request]] =
```
olix0r: can we make this package private to indicate it's for internal use?

olix0r: or even just `private` to the file

olix0r:
:star: my name is oliver and i endorse this message.

[Approved with PullApprove](https://pullapprove.com/BuoyantIO/linkerd/pull-request/636/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=BuoyantIO/linkerd)

PR #638, opened 2016-09-08
Add header identifier

### LINKED ISSUE #373: 'header' http identifier
The 'basic' http identifier is fairly rich, which is good; but overkill for many configurations. Write a 'header' http identifier that works as follows:
- accepts a _dstPrefix_, default: `/`
- accepts a _header_, default: `l5d-dst-id`
- reads this header from requests:
  - if it's a valid path append it to dstPrefix
  - if it's not a valid path, add it to dstPrefix as a path component

This can easily be used to support the `Host` header (whereas the _basic_ identifier includes methods, etc).  Furthermore, this allows applications that can't easily override Host (for instance HTTP libraries that only take a URL and forcefully set _host_) to use another header.  This also will be helpful for proxy applications that want to explicitly set a downstream i.e. from a residual.

adleong:
Fixes #373

With this identifier, HTTP requests are turned into names based only on the value of an HTTP header.  If the header value is a valid path, that path is used.  Otherwise, the header value is converted to a path with one path segment.

### DIFF
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -191,7 +191,7 @@ routers:
 
 Key | Default Value | Description
 --- | ------------- | -----------
-header | `l5d-dst-concrete` | The name of the HTTP header to use
+header | `l5d-name` | The name of the HTTP header to use
 
 #### Namer Path Parameters:
 
@@ -204,7 +204,7 @@ header | `l5d-dst-concrete` | The name of the HTTP header to use
 Key | Default Value | Description
 --- | ------------- | -----------
 dstPrefix | `http` | The `dstPrefix` as set in the routers block.
-headerValue | N/A | The value of the HTTP header as a path, if it is a valid path.  The value of the HTTP header as a single path segment, otherwise.
+headerValue | N/A | The value of the HTTP header as a path.
 
 <a name="http-engines"></a>
 ## HTTP Engines
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -165,6 +165,47 @@ Key | Default Value | Description
 dstPrefix | `http` | The `dstPrefix` as set in the routers block.
 urlPath | N/A | A path from the URL whose number of segments is set in the identifier block.
 
+<a name="header-identifier"></a>
+### Header Identifier
+
+kind: `io.l5d.header`
+
+With this identifier, HTTP requests are turned into names based only on the
+value of an HTTP header.  If the header value is a valid path, that path is
+used.  Otherwise, the header value is converted to a path with one path segment.
+
+#### Namer Configuration:
+
+> With this configuration, the value of the `my-header` HTTP header will be used
+as the logical name.
+
+```yaml
+routers:
+- protocol: http
+  identifier:
+    kind: io.l5d.header
+    header: my-header
+  servers:
+  - port: 5000
+```
+
+Key | Default Value | Description
+--- | ------------- | -----------
+header | `l5d-dst-concrete` | The name of the HTTP header to use
+
+#### Namer Path Parameters:
+
+> Dtab Path Format
+
+```
+  / dstPrefix [/ *headerValue ]
+```
+
+Key | Default Value | Description
+--- | ------------- | -----------
+dstPrefix | `http` | The `dstPrefix` as set in the routers block.
+headerValue | N/A | The value of the HTTP header as a path, if it is a valid path.  The value of the HTTP header as a single path segment, otherwise.
+
 <a name="http-engines"></a>
 ## HTTP Engines
 
--- a/linkerd/protocol/http/src/main/resources/META-INF/services/io.buoyant.linkerd.IdentifierInitializer
+++ b/linkerd/protocol/http/src/main/resources/META-INF/services/io.buoyant.linkerd.IdentifierInitializer
@@ -1,2 +1,3 @@
 io.buoyant.linkerd.protocol.http.MethodAndHostIdentifierInitializer
 io.buoyant.linkerd.protocol.http.PathIdentifierInitializer
+io.buoyant.linkerd.protocol.http.HeaderIdentifierInitializer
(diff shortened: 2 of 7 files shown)


router/http/src/main/scala/io/buoyant/router/http/HeaderIdentifier.scala
```diff
@@ -0,0 +1,24 @@
...
+  baseDtab: () => Dtab = () => Dtab.base
+) extends RoutingFactory.Identifier[Request] {
+
+  def apply(req: Request): Future[(Dst, Request)] = {
+    req.headerMap.get(header) match {
+      case Some(value) =>
+        val path = Try(Path.read(value)).getOrElse(Path.Utf8(value))
+        Future.value(Dst.Path(prefix ++ path, baseDtab(), Dtab.local), req)
+      case None =>
+        Future.exception(new IllegalArgumentException(s"$header header is absent"))
```
adleong: If https://github.com/BuoyantIO/linkerd/pull/636 merges first, update this to UnidentifiableRequestException

linkerd/docs/protocol-http.md
```diff
@@ -165,6 +165,47 @@ Key | Default Value | Description
...
+  identifier:
+    kind: io.l5d.header
+    header: my-header
+  servers:
+  - port: 5000
+```
+
+Key | Default Value | Description
+--- | ------------- | -----------
+header | `l5d-dst-concrete` | The name of the HTTP header to use
```
olix0r: I'm find this naming to be unclear. Elsewhere, we describe pre-delegation names (i.e. Dst.Path) as _logical_ and post-delegation names (i.e. Dst.Id) as _concrete_.  In this context, the value isn't really a concrete name.  We should try to reconcile this.
- `l5d-dst-name`?
- `l5d-dst-path?
- ?

linkerd/protocol/http/src/main/resources/META-INF/services/io.buoyant.linkerd.IdentifierInitializer
```diff
@@ -1,2 +1,3 @@
 io.buoyant.linkerd.protocol.http.MethodAndHostIdentifierInitializer
 io.buoyant.linkerd.protocol.http.PathIdentifierInitializer
+io.buoyant.linkerd.protocol.http.HeaderIdentifierInitializer
```
olix0r: alpha sort, por favor

linkerd/docs/protocol-http.md
adleong: The motivation is the linkerd-to-linkerd case.  The outgoing linkerd writes the concrete name into the `l5d-dst-concrete` header.  The incoming linkerd uses this identifier to read that header and then forwards to a local destination.  

We can change the default value to be something else, but I think we need to set this to `l5d-dst-concete` for the linkerd-to-linkerd setup.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HeaderIdentifierConfigTest.scala
```diff
@@ -0,0 +1,31 @@
...
+  test("parse config") {
+    val yaml = s"""
+                  |kind: io.l5d.header
+                  |header: my-header
+      """.stripMargin
+
+    val mapper = Parser.objectMapper(yaml, Iterable(Seq(HeaderIdentifierInitializer)))
+    val config = mapper.readValue[HttpIdentifierConfig](yaml).asInstanceOf[HeaderIdentifierConfig]
+    assert(config.header == Some("my-header"))
+  }
```
olix0r: add a test with the default header?

router/http/src/main/scala/io/buoyant/router/http/HeaderIdentifier.scala
```diff
@@ -0,0 +1,26 @@
...
+case class HeaderIdentifier(
+  prefix: Path,
+  header: String,
+  baseDtab: () => Dtab = () => Dtab.base
+) extends RoutingFactory.Identifier[Request] {
+
+  def apply(req: Request): Future[RequestIdentification[Request]] = {
+    req.headerMap.get(header) match {
+      case Some(value) =>
+        val path = Try(Path.read(value)).getOrElse(Path.Utf8(value))
```
olix0r: Hmmmm....  I suspect that this fallback logic will yield to some unexpected behavior.

```
scala> com.twitter.finagle.Path.Utf8("how/r/u").show
res0: String = /\x68\x6f\x77\x2f\x72\x2f\x75
```

It may be best to fail if it's not a parseable path.

linkerd/docs/protocol-http.md
adleong: More details here: https://github.com/BuoyantIO/linkerd/issues/659

router/http/src/main/scala/io/buoyant/router/http/HeaderIdentifier.scala
adleong: I'm fine with that, I was just trying to follow the spec in #373: 

"if it's not a valid path, add it to dstPrefix as a path component"

linkerd/docs/protocol-http.md
olix0r: Yeah, I understand that there are configurations where it makes sense to use prior concrete name (or the prior residual) to influence next-hop routing.  It seems like a confusing default, though.

router/http/src/main/scala/io/buoyant/router/http/HeaderIdentifier.scala
olix0r: Who would ever come up with such an awful idea?

We could support this IFF the value contains only showable path chars...  Not sure it's worth it? Up to you.

linkerd/docs/protocol-http.md
adleong: Ok I agree it's a confusing default.  What do you suggest?  How about just `l5d-dst`?

olix0r: That's not bad.  Now that I look at the [header docs](https://buoyant.io/slate-linkerd/0.7.5/linkerd/#user-headers), is it confusing to reuse the l5d-dst prefix?

> The informational headers linkerd emits on outgoing requests.
> ...
> Applications are not required to forward these headers on downstream requests.

Perhaps `l5d-name` or `l5d-target`?

Names are hard.

olix0r:
lgtm!  Do you think we should document `l5d-name` in the http headers documentation?

PR #643, opened 2016-09-08
Measure basic information about the consul namer.

olix0r:
This should help diagnose consul behavior.

Furthermore, fix a malformed consul client naming (was doing Path.toString
instead of Path.show).

Finally, LinkerConfig should scope namer stats explicitly rather than rely on
the namer to do the right scoping.

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
@@ -69,7 +69,7 @@ case class ConsulConfig(
 
     val service = Http.client
       .withParams(Http.client.params ++ params)
-      .withLabel(prefix.show)
+      .withLabel(prefix.show.stripPrefix("/"))
       .withMonitor(interruptionMonitor)
       .withTracer(NullTracer)
       .filtered(filters)
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
@@ -2,7 +2,7 @@ package io.buoyant.namer.consul
 
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.twitter.finagle.http.{Request, Response}
-import com.twitter.finagle.param.Label
+import com.twitter.finagle.param
 import com.twitter.finagle.tracing.NullTracer
 import com.twitter.finagle.{Failure, Filter, Http, Namer, Path, Stack}
 import com.twitter.util.Monitor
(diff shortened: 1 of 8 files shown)


namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
```diff
@@ -68,9 +68,9 @@ case class ConsulConfig(
     }
 
     val service = Http.client
-      .withMonitor(interruptionMonitor)
       .withParams(Http.client.params ++ params)
-      .configured(Label("namer" + prefix))
+      .withLabel(prefix.show)
```
olix0r: (1) withParams comes first so that withMonitor overrides any default monitors in the params.

(2) this should probably be `prefix.show.stripPrefix("/")`

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -11,24 +12,31 @@ import io.buoyant.consul.v1
...
     id: Path,
     residual: Path
   ): Activity[NameTree[Name]] = {
     log.debug("consul lookup: %s %s", id.show)
+
     Dc.watch(dc).map { services =>
       services.get(key) match {
         case None =>
           log.debug("consul dc %s service %s missing", dc, key)
+          negs.incr()
```
adleong: These stats are potentially confusing.  Since they're in the map block of an activity, they will get incremented for each listener registered to this activity.  Therefore the counter represents something like "number of negs returned" rather than "total number of negs that exist" or something that one might expect.

olix0r: good point

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
```diff
@@ -60,19 +86,22 @@ private[consul] object DcServices {
                 case Some(svc) => svc
                 case None =>
                   log.debug("consul added: %s", k)
-                  SvcAddr(consulApi, name, k, domain)
+                  stats.adds.incr()
+                  SvcAddr(consulApi, name, k, domain, stats.service)
```
adleong: should the service stats be scoped by service name rather than all together?

adleong:
:star: 📐

namer/consul/src/main/scala/io/buoyant/namer/consul/DcServices.scala
olix0r: Prefer to start conservative with number of metrics to start.  What we have here should be enough to do relative debugging ("huge spike in service changes", etc).  If this ends up being too opaque, we can always add more stats.  My gut says this is Good Enough.

PR #649, opened 2016-09-09
Simplify ProxyRewriteFilter to parse URIs conditionally

olix0r:
Problem

Currently, ProxyRewriteFilter parses every URI that traverses HTTP routers to
determine whether it is absolute.  However, this can blow up on technically
illegal urls (like "resource//").

Solution

I think we can be both more permissive and more efficient (though malformed
URIs may cause problems elsewhere in linkerd, depending on the configuration).
To fix this, we see if the URI includes `://`, indicating that it includes a
scheme. If it does include a scheme, we parse the URI to ensure it's absolute
and to rewrite the request. If the URI doesn't include a scheme, we just let it
pass though.


adleong:
:star: 🚤

PR #651, opened 2016-09-10
Order server and router shutdown to prevent exceptions

olix0r:
Problem

When shutting down linkerd under load, some exceptions are thrown (and requests failed).

Solution

During graceful shutdown, we must close things from the "top" (server) to the
"bottom" (client connections) so that we stop accepting requests from clients
before harvesting a router's resources.

linkerd's Main now has one big closeOnExit block that ensures proper shutdown
ordering: servers, routers, telemeters/admin.

Furthermore, we ensure that the DstBindingFactory tears down its caches in the
proper order: path, tree, bound, client.

Now everything seems copacetic during shutdown.

### DIFF
--- a/linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
@@ -73,13 +73,11 @@ object Linkerd extends App {
           }
         }
 
-        closeOnExit(Closable.make { deadline =>
-          Closable.all(routers: _*).close(deadline).before {
-            val telems = Closable.all(telemeters: _*)
-            Closable.sequence(telems, adminInitializer.adminHttpServer).close(deadline)
-          }
-        })
-
+        closeOnExit(Closable.sequence(
+          Closable.all(routers: _*),
+          Closable.all(telemeters: _*),
+          adminInitializer.adminHttpServer
+        ))
         Await.all(routers: _*)
         Await.all(telemeters: _*)
         Await.result(adminInitializer.adminHttpServer)
--- a/linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
@@ -1,7 +1,7 @@
 package io.buoyant
 
-import com.twitter.util.Await
-import io.buoyant.admin.{App, AdminInitializer}
+import com.twitter.util.{Await, Closable, CloseAwaitably, Time}
+import io.buoyant.admin.{AdminInitializer, App}
 import io.buoyant.linkerd.Linker.LinkerConfig
 import io.buoyant.linkerd.admin.LinkerdAdmin
 import io.buoyant.linkerd.{Build, Linker}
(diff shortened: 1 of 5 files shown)


olix0r:
related to #630

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -73,7 +52,77 @@ object Linkerd extends App {
...
+    announcers: Seq[(Path, Announcer)],
+    server: Server.Initializer,
+    name: Path
+  ): Closable =
+    announcers.filter { case (pfx, _) => name.startsWith(pfx) } match {
+      case Seq() =>
+        log.warning("no announcer found for %s", name.show)
+        Closable.nop
+
+      case announcers =>
```
adleong: TIOLI: I know shadowing is a fairly common practice, but I think it's especially dangerous when the shadowed variables have the same type.  Here `announcers` refers to both the filtered and unfiltered seq of announcers, depending on the scope.

olix0r: yeah, the intent is that you _can't_ use the unfiltered one in the block. you have to use the right one. that said, i'll rename the first to `announcers0`

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -73,7 +52,77 @@ object Linkerd extends App {
...
+    }
+
+  private def announce(
+    prefix: Path,
+    announcer: Announcer,
+    server: Server.Initializer,
+    name: Path
+  ): Closable = {
+    log.info("announcing %s as %s to %s", server.addr, name.show, announcer.scheme)
+    val f = announcer.announce(server.addr, name.drop(prefix.size))
```
adleong: should `f` be named more descriptively? `announcement`?

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -73,7 +52,77 @@ object Linkerd extends App {
...
+
+    // If we close before the announcer registers, we
+    // cancel the registration. If we close after
+    // registration, we close it.
+    //
+    // XXX we should change the announcer API to return
+    // an Announcement synchronously (that may actually
+    // wrap some asynchronous announcement logic).
+    val closeRef = new AtomicReference[Closable](Closable.make { _ =>
+      f.raise(Failure("closed").flagged(Failure.Interrupted))
```
adleong: This seems unreliable since cancelations are advisory.  Maybe closing of announcement is best-effort anyway so this is okay...?  But allowing the announcement to complete and then closing seems more reliable to me.

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
olix0r: Since the scope is so narrow (just this function), I don't find it any clearer to call it `pending`, but I can if you find it confusing.

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
olix0r: I think we want to signal cancellation in case the announcer is out to lunch / etc.  I added a var that signals to pending-handling whether cancellation has already occurred.   It's not 1000% safe, but I think it's basically as good as we can dod.

linkerd/main/src/main/scala/io/buoyant/Linkerd.scala
```diff
@@ -73,7 +52,83 @@ object Linkerd extends App {
...
+      deadline = Some(d)
+      f.raise(Failure("closed").flagged(Failure.Interrupted))
+      Future.Unit
+    })
+    f.respond {
+      case Throw(_) => closeRef.set(Closable.nop)
+      case Return(announced) =>
+        deadline match {
+          case None => closeRef.set(announced)
+          case Some(d) => announced.close(d)
```
adleong: I think you need to signal value discard here

adleong:
:star: Looks good other than the value discard warning.

PR #652, opened 2016-09-11
Change Announcer API to return a Closable.

olix0r:
It's more convenient (for the purposes of linkerd's execution) for the
Announcer API to just return a Closable. The FutureAnnouncer helper bridges
Finagle's Future API to linkerd's immediate API.

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Announcer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Announcer.scala
@@ -1,11 +1,41 @@
 package io.buoyant.linkerd
 
-import com.twitter.finagle.{Announcement, Path}
-import com.twitter.util.Future
+import com.twitter.finagle.{Announcement, Failure, Path}
+import com.twitter.util.{Closable, Future, Return, Throw, Time}
+import java.util.concurrent.atomic.AtomicReference
 import java.net.InetSocketAddress
 
 abstract class Announcer {
-  val scheme: String
+  def scheme: String
+  def announce(addr: InetSocketAddress, name: Path): Closable
+}
 
-  def announce(addr: InetSocketAddress, name: Path): Future[Announcement]
+abstract class FutureAnnouncer extends Announcer {
+
+  protected def announceAsync(addr: InetSocketAddress, name: Path): Future[Announcement]
+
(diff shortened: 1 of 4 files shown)


olix0r:
based on #651.

linkerd/core/src/main/scala/io/buoyant/linkerd/Announcer.scala
```diff
@@ -1,11 +1,30 @@
...
-import com.twitter.util.Future
+import com.twitter.finagle.{Announcement, Failure, Path}
+import com.twitter.util.{Closable, Future, Return, Throw}
+import java.util.concurrent.atomic.AtomicReference
 import java.net.InetSocketAddress
 
 abstract class Announcer {
-  val scheme: String
+  def scheme: String
+  def apply(addr: InetSocketAddress, name: Path): Closable
```
adleong: Calling this `announce` instead of `apply` seems more clear and explicit.

linkerd/core/src/main/scala/io/buoyant/linkerd/Announcer.scala
```diff
@@ -1,11 +1,30 @@
...
+}
+
+abstract class FutureAnnouncer extends Announcer {
+  protected def future(addr: InetSocketAddress, name: Path): Future[Announcement]
 
-  def announce(addr: InetSocketAddress, name: Path): Future[Announcement]
+  final def apply(addr: InetSocketAddress, name: Path): Closable = {
+    val f = future(addr, name)
+    // If closing before the announcer registers, cancel the
```
adleong: I have `raise`d concerns about this in the parent branch.

linkerd/core/src/main/scala/io/buoyant/linkerd/Announcer.scala
adleong: I actually prefer your second example over your first.  I think `apply` is overused in scala.  But that's just, like, my opinion, man.  FFTI.

olix0r: I don't feel strongly one way or the other. will revert

adleong:
:star: 🚿

PR #653, opened 2016-09-11
Improve admin and Namerd initialization

olix0r:
Admin server initialization is far too verbose. Tidy up admin server
initialization, reducing log level on admin endpoint logging.


olix0r:
based on #651.

adleong:
What's the net effect here?

* Endpoint list at startup moved from INFO to DEBUG level
* Exceptions during admin initialization no longer print stack traces?
* Slight refactor without any change of behavior
* Anything else?

olix0r:
- Endpoint list at startup moved from INFO to DEBUG
- the admin server Monitor still prints stack traces but uses the logger instead of our own version of stacktrace printing.
- AdminInitializer is just a function instead of a mutable/stateful class
- Namerd shutdown ordering is fixed.

adleong:
:star: 🔇

olix0r:
:star: doubleship?

PR #654, opened 2016-09-11
Rename io.buoyant.Linkerd to io.buoyant.linkerd.Main

olix0r:
Consistent with namerd. Makes sense in general.


adleong:
:star: ✈️

PR #655, opened 2016-09-11
Require configuration for tracing.

olix0r:
Problem

Falling back to use DefaultTracer causes log noise for most configurations.

Solution

Do not fall back to DefaultTracer. Instead, if no telemeter is configured, we
provide a default telemeter that has tracing disabled.

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
@@ -102,11 +102,10 @@ object Linker {
       // telemeters provide implementations that do not use the
       // default Metrics registry, linker stats may be missing from
       // /admin/metrics.json
-      val stats = telemeters.collect { case t if !t.stats.isNull => t.stats } match {
-        case Nil => NullStatsReceiver
-        case receivers =>
-          for (r <- receivers) log.info("stats: %s", r)
-          BroadcastStatsReceiver(receivers)
+      val stats = {
+        val receivers = telemeters.collect { case t if !t.stats.isNull => t.stats }
+        for (r <- receivers) log.info("stats: %s", r)
+        BroadcastStatsReceiver(receivers)
       }
 
       // Similarly, tracers may be provided by telemeters OR by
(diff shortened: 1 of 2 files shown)


linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -88,44 +88,48 @@ object Linker {
...
-            DefaultStatsReceiver
-          case receivers =>
-            for (r <- receivers) log.info(s"Using stats receiver: $r")
-            BroadcastStatsReceiver(receivers)
+      val stats = {
+        val receivers = telemeters.collect { case t if !t.stats.isNull => t.stats } match {
+          case Nil => NullStatsReceiver
+          case receivers => BroadcastStatsReceiver(receivers)
         }
+        for (t <- tracers) log.info("stats: %s", t)
```
adleong: should this be `r <- receivers`?

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -88,44 +88,48 @@ object Linker {
...
-      val stats =
-        telemeters.getOrElse(Nil).collect { case t if !t.stats.isNull => t.stats } match {
-          case Nil =>
-            log.info(s"Using default stats receiver")
-            DefaultStatsReceiver
-          case receivers =>
-            for (r <- receivers) log.info(s"Using stats receiver: $r")
-            BroadcastStatsReceiver(receivers)
+      val stats = {
+        val receivers = telemeters.collect { case t if !t.stats.isNull => t.stats } match {
```
adleong: should this name be singular instead of plural?

adleong:
Let me see if I've got this straight:

# Old

* No telemeters => Default tracer + default stats receiver
* Some telemeters but none that provide stats => default stats receiver
* Some telemeters but none that provide tracers => default tracer

# New 

* No telemeters => Null tracer + default stats receiver
* Some telemeters but none that provide stats => default stats receiver
* Some telemeters but none that provide tracers => null tracer

What even is the default tracer?  Is it ever useful to us?

olix0r:
The default tracer is "whatever was loaded on the classpath".  At twitter, this is finagle-zipkin, configured to write to localhost:SCRIBEPORT.  I don't think linkerd should maintain this behavior.

You're right with the one exception that the new behavior is:
- Some telemeters but none that provide stats => null stats receiver

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
olix0r: ah yeah, i thought i fixed this (probably in another branch)

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
olix0r: correct

adleong:
:star: 👡

PR #656, opened 2016-09-11
Remove `debugTrace` from tracer configuration.

olix0r:
With the introduction of the `io.l5d.tracelog` telemeter, there is no longer
any need to support the `debugTrace` flag on tracers.  This flag was a bit of a
wart, behaving differently from all other configuration options by configuring
a global parameter.


adleong:
:star: 🍔

PR #657, opened 2016-09-12
Log more info on retry within Consul API

Ashald:
Transforms:
```
retrying consul request on error UnexpectedResponse(500: rpc error: No path to datacenter)
```
(and not only :) ) into
```
retrying consul request 'GET /v1/health/service/foobar?index=3497381&dc=wrong-dc&tag=&passing=true' on error UnexpectedResponse(500: rpc error: No path to datacenter)
```

which makes debugging much more efficient


olix0r:
lgtm! thanks

Issue #665, opened 2016-09-13
Allow setting consistency mode in Consul Namer and Dtab Store


Ashald:
As per https://www.consul.io/docs/agent/http.html, there are 3 consistency modes in Consul:
1) default
2) consistent
3) stale

It looks like that using default consistency mode with high rate of HTTP requests can lead to extra leader re-elections (https://groups.google.com/forum/#!msg/consul-tool/NrRRxem0Q_g/H9GFHRH3GQAJ)

The proposal is to add an extra option to allow tuning consistency mode for namer and dtab store.

I see 2 possible options:
1)
```
consistency: default/consistent/stale
```
2)
```
allowStale: true/false
```

JustinVenus:
@Ashald if I get to vote I'd prefer option `1` and nice find on the leader election behavior was beginning to think I was seeing a consul bug.

Ashald:
@JustinVenus yeah, that's exactly how it's being implemented - please check out #666

olix0r:
Fixed by #666

PR #666, opened 2016-09-13
Allow setting consistency mode in Consul Namer and Dtab Store

Ashald:
xref #665

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
@@ -49,9 +49,18 @@ trait BaseApi extends Closable {
   private[v1] def mkreq(
     method: http.Method,
     path: String,
+    consistency: Option[ConsistencyMode],
     optParams: (String, Option[String])*
   ): http.Request = {
-    val params = optParams.collect { case (k, Some(v)) => (k, v) }
+    val consistencyMode = consistency.flatMap {
+      case ConsistencyMode.Consistent =>
+        Some("consistent" -> Some(true.toString))
+      case ConsistencyMode.Stale =>
+        Some("stale" -> Some(true.toString))
+      case ConsistencyMode.Default =>
+        None
+    }
+    val params = (consistencyMode ++ optParams).collect { case (k, Some(v)) => (k, v) }.toSeq
     val req = http.Request(path, params: _*)
     req.method = method
     req
--- a/consul/src/main/scala/io/buoyant/consul/v1/ConsistencyMode.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/ConsistencyMode.scala
@@ -1,0 +1,37 @@
+package io.buoyant.consul.v1
+
+import com.fasterxml.jackson.core.{JsonGenerator, JsonParser}
+import com.fasterxml.jackson.databind.{DeserializationContext, SerializerProvider}
+import io.buoyant.config.{ConfigDeserializer, ConfigSerializer}
+
+sealed trait ConsistencyMode
+
+object ConsistencyMode {
+
+  case object Default extends ConsistencyMode
+  case object Stale extends ConsistencyMode
+  case object Consistent extends ConsistencyMode
+
+}
+
+class ConsistencyModeDeserializer extends ConfigDeserializer[ConsistencyMode] {
+
+  override def deserialize(jp: JsonParser, ctxt: DeserializationContext): ConsistencyMode =
+    catchMappingException(ctxt) {
+      _parseString(jp, ctxt) match {
+        case "default" => ConsistencyMode.Default
+        case "stale" => ConsistencyMode.Stale
+        case "consistent" => ConsistencyMode.Consistent
+        case unknown =>
+          throw new IllegalArgumentException(s"Illegal Consul consistency mode level: $unknown")
+      }
+    }
+}
+
+class ConsistencyModeSerializer extends ConfigSerializer[ConsistencyMode] {
+  override def serialize(
+    mode: ConsistencyMode,
+    jgen: JsonGenerator,
+    provider: SerializerProvider
+  ): Unit = jgen.writeString(mode.toString.toLowerCase) // case objects get good toString methods
+}
--- a/project/LinkerdBuild.scala
+++ b/project/LinkerdBuild.scala
@@ -12,7 +12,13 @@ object LinkerdBuild extends Base {
   val Bundle = config("bundle") extend Minimal
   val Dcos = config("dcos") extend Bundle
 
+  val configCore = projectDir("config")
+    .withTwitterLibs(Deps.finagle("core"))
+    .withLibs(Deps.jackson)
+    .withLib(Deps.jacksonYaml)
+
   val consul = projectDir("consul")
+    .dependsOn(configCore)
     .withTwitterLib(Deps.finagle("http"))
     .withLibs(Deps.jackson)
     .withTests()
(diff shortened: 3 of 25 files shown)


consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -49,9 +49,15 @@ trait BaseApi extends Closable {
   private[v1] def mkreq(
     method: http.Method,
     path: String,
+    consistency: Option[ConsistencyMode],
     optParams: (String, Option[String])*
   ): http.Request = {
-    val params = optParams.collect { case (k, Some(v)) => (k, v) }
+    val consistencyMode = consistency match {
```
adleong: I think this reads better as
```
val consistencyMode = consistency.flatMap {
  case ConsistencyMode.Constant =>
    Some("consistent" -> Some(true.toString))
  case ConsistencyMode.Stale =>
    Some("stale" -> Some(true.toString))
  case ConsistencyMode.Default =>
    None
}
val params = (consistencyMode ++ optParams).collect { ...
```

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -49,9 +49,15 @@ trait BaseApi extends Closable {
...
+      case Some(ConsistencyMode.Consistent) => ("consistent", Some(true.toString))
+      case Some(ConsistencyMode.Stale) => ("stale", Some(true.toString))
+      case _ => ("default", None)
```
adleong: This is weird because the "default" string is never used, it just gets filtered out.  See above suggestion.

consul/src/main/scala/io/buoyant/consul/v1/ConsistencyMode.scala
```diff
@@ -0,0 +1,37 @@
...
+      }
+    }
+}
+
+class ConsistencyModeSerializer extends ConfigSerializer[ConsistencyMode] {
+  override def serialize(
+    mode: ConsistencyMode,
+    jgen: JsonGenerator,
+    provider: SerializerProvider
+  ): Unit = jgen.writeString(mode.toString.toLowerCase) // is it OK ro rely on `toString`...?
```
adleong: I think this is fine, case objects get good `toString` methods.

project/LinkerdBuild.scala
```diff
@@ -12,7 +12,13 @@ object LinkerdBuild extends Base {
   val Bundle = config("bundle") extend Minimal
   val Dcos = config("dcos") extend Bundle
 
+  val configCore = projectDir("config")
+    .withTwitterLibs(Deps.finagle("core"))
+    .withLibs(Deps.jackson)
+    .withLib(Deps.jacksonYaml)
+
   val consul = projectDir("consul")
+    .dependsOn(configCore)
```
adleong: @olix0r what do you think about this?

olix0r: I don't _love_ that consul is picking up extra deps, but its probably the least worst option for the time being.

consul/src/main/scala/io/buoyant/consul/v1/ConsistencyMode.scala
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.databind.{DeserializationContext, SerializerProvider}
+import io.buoyant.config.{ConfigDeserializer, ConfigSerializer}
+
+sealed trait ConsistencyMode
+
+object ConsistencyMode {
+
+  case object Default extends ConsistencyMode
+  case object Stale extends ConsistencyMode
+  case object Consistent extends ConsistencyMode
```
olix0r: we tend to just use `object` to `case object` -- it doesn't add anything valuable

adleong: case object adds a better `toString` FWIW

olix0r: after reading up, it seems like the only real value of `case object` is that it generates a "pretty" `toString`.  not a big deal one way or the other.

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
Ashald: Cool, thanks!

consul/src/main/scala/io/buoyant/consul/v1/ConsistencyMode.scala
Ashald: Great, thanks! I'll update the comment with explanation (about good `toString` then). Also as we rely on `toString` to operate in some way - maybe it worth a test?..

project/LinkerdBuild.scala
Ashald: If you will happen to have a better idea - please let me know, not huge fan of current situation as well. :/

Ashald:
Thanks for comments, I'll do adjustments you proposed. How it looks like in general to you? Is it "good enough" so I can proceed with introducing consistency for dtab store? If yes, do you want a single `consistency` parameter for it or 2 separate - for read and for write?

Ashald:
BTW, love PR # :D

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
Ashald: Hm, so with this we have `consistencyMode: Option[(String, Option[String])]` - what would be the idiomatic way of using it on the following line where we have `val params = (consistencyMode +: optParams).collect { case (k, Some(v)) => (k, v) }`? Add `getOrElse`?

consul/src/main/scala/io/buoyant/consul/v1/ConsistencyMode.scala
olix0r: I think `case object` is fine.

olix0r:
I think it's fine to move forward with DtabStore support consistency.   I think operationally speaking, if I were running consul, I'd want to ensure that writes are consistent but reads do not have to be.

Ashald:
So, 2 independent options then? I mean, no "smart" behavior like "if write consistency is not set then it's same as read consistency"?

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
adleong: I think `(consistencyMode ++ optParams).collect { case (k, Some(v)) => (k, v) }` should work since both `consistencyMode` and `optParams` are iterables.

Ashald: `(consistencyMode.toSeq ++ optParams)` worked - it was failing to compile without `toSeq` :/

Ashald: Hm, I think `(consistencyMode ++ optParams).collect { case (k, Some(v)) => (k, v) }.toSeq` would be better in such case...

Ashald:
2nd part is done as well - `readConsistencyMode` & `writeConsistencyMode` for consul dtab store. Please let me know how it looks to you guys?

adleong:
:star: This looks great, @Ashald!  Have you noticed that adjusting the consistency level helps reduce the amount of leader election?  If it makes a big difference on leader election and therefore on linkerd memory pressure, we could consider changing the defaults.

olix0r:
+1 I would support changing the default read consistency

Ashald:
Actually I haven't tried it yet in our environment. I can give it a try but I cannot replace all `namerd` instances in a datacenter with dev build. :( Our security guys are pretty strict :)
But without that I don't think we will see a difference. So I see 2 options:
1) We replace default value for reads to `stale` since it's just makes sense
2) We release it as is, I'm switching to `stale` to see the impact and if it's as great as expected then we change the default in next release

adleong:
I'd say let's release as-is.  Once we have some data regarding how this behaves in prod we can consider changing the default in the next release.  Does that sound good?

Ashald:
Yeah, sounds great to me!

Ashald:
I rebased the PR over latest master - waiting for it to be merged so I can submit another about `failFast`.

Ashald:
Thanks!

PR #667, opened 2016-09-14
Alter Admin initialization to admit plugin-defined handlers

### LINKED ISSUE #98: remove vestigal `/admin` UI
Functionality is now in /, and things have changed sufficiently that this endpoint doesn't even do anything, just shows a empty UI.

olix0r:
#### Problem

There's no mechanism for plugins to install endpoints onto the admin service.
There should be.

#### Solution

Introduce the `Admin.WithHandlers` trait, which allows admin initialization to
detect modules that expose handlers.

Refactor Admin initialization to compose handlers during initialization.
Previously, admin classes were subtyped to extend functionality. Now, admin
endpoints are concatenated during initialization so that an arbitrary number of
endpoints may be installed without comabting the type system.

Furthermore, the `Namerd` and `Linker` types now have
configured-but-not-yet-initialized `Admin` values (rather than `AdminConfig`),
so that Admin initialization is congruent with other subsystems.

Finally, remove unused/unwanted parts of admin interface to fix #98.


admin/src/main/scala/io/buoyant/admin/Admin.scala
```diff
@@ -1,65 +1,112 @@
...
     "/admin/per_host_metrics.json" -> HttpMuxer
   )
 
-  def allRoutes = twitterServerRoutes ++ metricsRoutes
+  /** Generate an index of the provided handlers */
+  private def indexHandlers(handlers: Handlers): Handlers = {
+    val paths = handlers.map { case (p, _) => p }.sorted.distinct
+    val index = new IndexTxtHandler(paths)
+    Seq(
+      "/admin/index.txt" -> index,
```
adleong: What are these text index routes for?

admin/src/main/scala/io/buoyant/admin/Admin.scala
```diff
@@ -1,65 +1,112 @@
...
+  }
+
+  def getHandlers(obj: Any): Handlers =
+    obj match {
+      case wh: WithHandlers => wh.adminHandlers
+      case _ => Nil
+    }
+
+  def collectHandlers(objs: Seq[Any]): Handlers =
+    objs.flatMap(getHandlers(_))
```
olix0r: not required, but it signals (to me) that there's a function allocation here.  Got this convention from finagle.

admin/src/main/scala/io/buoyant/admin/Admin.scala
olix0r: just so you have a way to list all of the admin endpoints without guessing or reading the code (a problem i personally have a lot)

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -13,38 +14,51 @@ import io.buoyant.linkerd.Linker
...
-    linker.routers.find(_.label == label).map { router =>
-      router.params[DstBindingFactory.Namer].interpreter
-    }.getOrElse(NameInterpreter)
-
-  override def allRoutes = super.allRoutes ++ linkerdAdminRoutes
+  def apply(lc: Linker.LinkerConfig, linker: Linker): Admin.Handlers =
+    static ++ config(lc) ++
+      boundNames(linker.namers.map { case (_, n) => n }) ++
+      delegator(linker.routers) ++
+      Admin.collectHandlers(linker.namers ++ linker.routers ++ linker.telemeters)
```
adleong: what about handlers on non-top-level plugins (like identifiers, for example)?

olix0r: TBD.  we'll have to plumb this through into the router, presumably

adleong:
:star: 🉐 

[Approved with PullApprove](https://pullapprove.com/BuoyantIO/linkerd/pull-request/667/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=BuoyantIO/linkerd)

olix0r:
no js changes, rolling forward with this

PR #669, opened 2016-09-14
k8s transformers

### LINKED ISSUE #660: support linkerd-to-linkerd per-host deployments in k8s
Deploying 1 linkerd per-node instead of 1 per-pod can be far more resource efficient when tens or hundreds of pods run on each node.  Unfortunately, this presents several challenges that must be overcome to have a functioning linkerd-to-linkerd setup:

1. Services must be able to locate the local linkerd so that they can send outgoing traffic to it.  A service can request the set of linkerd endpoints from the a linkerd daemonset service api.  Since each node is a distinct /24 subnet, the service can then pick the linkerd endpoint on it's /24 subnet and send outgoing traffic to that address.

2. linkerd must be able to determine which hosts to which to send traffic.  i.e. which hosts have instances of the target service running on them.  https://github.com/BuoyantIO/linkerd/issues/658

3. When receiving incoming traffic, linkerd must be able to locate the local service to which to route the traffic. https://github.com/BuoyantIO/linkerd/issues/659

adleong:
Fixes #660

Adds support for linkerd-to-linkerd deployments in k8s when linkerd is deployed as a daemonset.  This is accomplished by adding two NameTree transformers: `io.l5d.daemonset` and `io.l5d.localnode`.  For linkerd-to-linkerd deployments, one should add the localnode transformer to the incoming router and the daemonset transformer (configured with the linkerd daemonset) to the outgoing router.

## daemonset

The DaemonSetTransformer maps each address in the NameTree to a member of a
given daemon set that is on the same /24 subnet.  Since each k8s node is its own
/24 subnet, the result is that each address is mapped to the member of the
daemon set that is running on the same node.  This can be used to redirect
traffic to a reverse-proxy (e.g. linkerd) that runs as a daemonset.

This transformer assumes that there is a Kubernetes service for the daemonset
which can be used to find all pods in the daemonset.

## localnode 

The localnode transformer filters the list of addresses down to only addresses
that are on the same /24 subnet as localhost.  Since each k8s node is its own
/24 subnet, the result is that only addresses on the local node are used.

This transformer does not have any configuration properties but it does require
the `POD_IP` environment variable be set with the localhost IP address.  This is
most easily done with the
[Kubernetes downward API](http://kubernetes.io/docs/user-guide/downward-api/).


interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformerInitializer.scala
```diff
@@ -0,0 +1,39 @@
...
+import com.twitter.finagle.param.Label
+import com.twitter.finagle.{NameTree, Path}
+import io.buoyant.config.types.Port
+import io.buoyant.k8s.v1.Api
+import io.buoyant.k8s.{ClientConfig, EndpointsNamer}
+import io.buoyant.namer.{NameTreeTransformer, TransformerConfig, TransformerInitializer}
+
+class DaemonSetTransformerInitializer extends TransformerInitializer {
+  val configClass = classOf[DaemonSetTransformerConfig]
+  override val configId = "io.l5d.daemonset"
```
olix0r: perhaps `io.l5d.k8s.daemonset`?  I think _k8s_ should be in here somewhere.

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  override protected def transform(tree: NameTree[Bound]): Activity[NameTree[Bound]] = {
+    daemonSet.map { daemonSet =>
+      val daemons = flatten(daemonSet.eval.toSet.flatten)
+      tree.map(mapBound(_, daemons))
+    }
+  }
+
+  /** Smoosh together all of the bound addresses into a single Var */
```
olix0r: good use of "smoosh" 👍

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/LocalNodeTransformerInitializer.scala
```diff
@@ -0,0 +1,20 @@
+package io.buoyant.transformer.k8s
+
+import io.buoyant.namer.{NameTreeTransformer, TransformerConfig, TransformerInitializer}
+import java.net.InetAddress
+
+class LocalNodeTransformerInitializer extends TransformerInitializer {
+  val configClass = classOf[LocalNodeTransformerConfig]
+  override val configId = "io.l5d.localnode"
```
olix0r: similarly, if this is k8s-oriented, would like to have _k8s_ somewhere in the plugin name.

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * The DaemonSetTransformer maps each address in the NameTree to a member of
+ * a given daemon set that is on the same /24 subnet.  Since each k8s node
+ * is its own /24 subnet, the result is that each address is mapped to the
+ * member of the daemon set that is running on the same node.  This can be used
+ * to redirect traffic to a reverse-proxy that runs as a daemon set.
+ */
+class DaemonSetTransformer(daemonSet: Activity[NameTree[Bound]]) extends DelegatingNameTreeTransformer {
+
+  override protected def transformDelegate(tree: DelegateTree[Bound]): Activity[DelegateTree[Bound]] = {
```
olix0r: I'd probably wrap these lines, they feel a bit long

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  /** Smoosh together all of the bound addresses into a single Var */
+  private[this] def flatten(bounds: Set[Name.Bound]): Var[Addr] = {
+    Var.collect(bounds.map(_.addr)).map { addrs =>
+      val collectedAddresses = addrs.flatMap {
+        case Addr.Bound(addresses, _) => addresses
+        case _ => Set.empty[Address]
+      }
```
olix0r: tioli: could this just be:
```scala
addrs.collect { case Addr.Bound(sas, _) => sas }.flatten
```

(i don't think i have a preference, though this one seems more concise)

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+      Addr.Bound(collectedAddresses)
```
olix0r: Hmm, I'm pretty sure that we shouldn't just drop metadata here?

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+import com.twitter.util.{Var, Activity}
+import io.buoyant.namer.{DelegateTree, DelegatingNameTreeTransformer}
+
+/**
+ * The DaemonSetTransformer maps each address in the NameTree to a member of
+ * a given daemon set that is on the same /24 subnet.  Since each k8s node
+ * is its own /24 subnet, the result is that each address is mapped to the
+ * member of the daemon set that is running on the same node.  This can be used
+ * to redirect traffic to a reverse-proxy that runs as a daemon set.
+ */
```
olix0r: this comment is a little unclear -- there are multiple name trees.  What is daemonSet and how is it used?

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  /**
+   * Return a new Bound with the address replaced by a member of the daemon
+   * set on the same /24 subnet
+   */
+  private[this] def mapBound(bound: Name.Bound, daemons: Var[Addr]): Name.Bound = {
+    val vaddr = Var.collect(List(bound.addr, daemons)).map {
+      case List(Addr.Bound(addrs, meta), Addr.Bound(daemon, _)) =>
```
olix0r: Is it okay to drop daemon metadata? It's hard for me to answer that question in the context of this class alone...

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  private[this] def shareSubnet(a1: Address, a2: Address): Boolean = {
+    (a1, a2) match {
+      case (Address.Inet(addr1, _), Address.Inet(addr2, _)) =>
+        val b1 = addr1.getAddress.getAddress
+        val b2 = addr2.getAddress.getAddress
+        // examine only the first 3 bytes (24 bits)
+        b1.zip(b2).take(3).forall(tupleEqual)
+      case _ => false
+    }
```
olix0r: this is run.... once per name update? is that correct?

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/LocalNodeTransformerInitializer.scala
```diff
@@ -0,0 +1,20 @@
...
+}
+
+class LocalNodeTransformerConfig extends TransformerConfig {
+
+  override def mk: NameTreeTransformer = {
```
olix0r: i prefer parens for `mk()` since it reads imperative and not something like `toTransformer`  (also, from the interface point of view, it may be side-effecty).

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
olix0r: I guess more generally, i might be inclined to do this comparison manually/procedurally just because it seems unfortunate that we have to allocate to do an IP comparison.

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+      case List(addr, _) => addr
+    }
+    Name.Bound(vaddr, bound.id, bound.path)
+  }
+
+  private[this] def shareSubnet(a1: Address, a2: Address): Boolean = {
+    (a1, a2) match {
+      case (Address.Inet(addr1, _), Address.Inet(addr2, _)) =>
+        val b1 = addr1.getAddress.getAddress
+        val b2 = addr2.getAddress.getAddress
```
olix0r: do we need to ensure that these are _ip4_ addresses? (perhaps match against java.netInet4Address?)

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/LocalNodeTransformerInitializer.scala
olix0r: i realize this is nowhere near uniform throughout the codebase. we can sort this out later

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformerInitializer.scala
```diff
@@ -0,0 +1,39 @@
...
+  assert(port != null, "io.l5d.daemonset: port property is required")
+
+  override def host: Option[String] = k8sHost
+  override def portNum: Option[Int] = k8sPort.map(_.port)
+
+  override def mk: NameTreeTransformer = {
+    val client = mkClient(Params.empty)
+    def mkNs(ns: String) = Api(client.configured(Label("daemonsetTransformer"))
+      .newService(dst))
+      .withNamespace(ns)
```
olix0r: I forget... are we required to build a new client for each namespace?

I remember there being a bug that made us do this, but I thought we fixed it.  Perhaps not.

I would expect this to be 
```scala
    val client = mkClient(Stack.Params.empty)
      .configured(Label("daemonsetTransformer"))
      .newService(dst)
    def mkNs(ns: String) = Api(client).withNamespace(ns)
```

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
olix0r: ok. now that i've wrapped my mind around this, i think this is right.

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
```diff
@@ -0,0 +1,73 @@
...
+import io.buoyant.namer.{DelegateTree, DelegatingNameTreeTransformer}
+
+/**
+ * The DaemonSetTransformer maps each address in the NameTree to a member of
+ * a given daemon set that is on the same /24 subnet.  Since each k8s node
+ * is its own /24 subnet, the result is that each address is mapped to the
+ * member of the daemon set that is running on the same node.  This can be used
+ * to redirect traffic to a reverse-proxy that runs as a daemon set.
+ */
+class DaemonSetTransformer(daemonSet: Activity[NameTree[Bound]]) extends DelegatingNameTreeTransformer {
```
olix0r: It occurs to me that this class is in no way kubernetes-specific (only the /24-ness of the subnet match is really kubernetes specific).  If we write a generic `shareSubnet(resolved: Address, routable: Address, subnetMask: Int)`, this is a generic way to route to a single linkerd per subnet!

This is very cool and probably useful in a variety of network configurations.

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/LocalNodeTransformer.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  private[this] val localAddress = local.getAddress
+
+  override protected val isLocal: Address => Boolean = {
+    case Address.Inet(addr, meta) =>
+      val bytes = addr.getAddress.getAddress
+      // examine only the first 3 bytes (24 bits)
+      localAddress.zip(bytes).take(3).forall(tupleEqual)
+    case address => true
+  }
```
olix0r: similarly, this just a stone's throw from being useful for arbitrary subnet masks

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
olix0r: Here's a sketched up change that introduces an arbitrary subnet gateway transformer, usable from the k8s transformers: https://github.com/BuoyantIO/linkerd/pull/671#pullrequestreview-183396, but also probably useful on its own.  We can consider this in a followup change.

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformer.scala
olix0r: understood now, dropping daemon metadata is ok

olix0r:
On reread, I think the only blockers on this are:
- put _k8s_ in the module names
- do we want to gate these with an experimental flag? don't feel strongly, but the k8s namer is still experimental...

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformerInitializer.scala
adleong: cribbed from `io.buoyant.namer.k8s`.  if fixed we can fix it in both places.

PR #670, opened 2016-09-14
Replace the io.l5d.default telemeter with io.l5d.commonMetrics

olix0r:
Problem

The DefaultStatsReceiver just loads things from the classpath. We want to _set_
the DefaultStatsReceiver to be the stats receiver configured by telemeters, and
setting the DefaultStatsReceiver to the DefaultStatsReceiver crashes the
program.  Furthermore, this makes it cleaner to add metrics endpoints to the
proper telemeter.

Solution

Introduce the io.l5d.commonMetrics telemeter, which wraps the
com.twitter.common.metrics library for configuration in linkerd.

This telemeter is used as the default telemter by the linkerd-main package.

Furthermore, metrics export endpoints are now exposed by the
io.l5d.commonMetrics telemeter. To satisfy this, namerd has been altered to
build against a default telemeter, though telemeters are not yet exposed in the
configuration.


adleong:
:star: I didn't try it out, but this looks good to me 🔭

PR #674, opened 2016-09-17
Add support for experimental protocols.

olix0r:
Problem

Namers and DtabStores may require that they are configured with `experimental:
true`. There's no mechanism for protocols to be explicitly marked as experimental.

Solution

ProtocolInitializers now expose a `experimentalRequired` attribute, indicating
whether routers must be configured with `experimental: true`.


adleong:
:star: 🔬

PR #677, opened 2016-09-17
Remove all headers listed in the connection header

### LINKED ISSUE #676: Linkerd should strip all headers listed in the `Connection` header
`StripConnectionHeader` is removing only the `Connection` Header. 
According to https://www.mnot.net/blog/2011/07/11/what_proxies_must_do not only `Connection` 
but all the headers listed in the `Connection` header should be removed.

Actually to be complete all the Hop-By-Hop headers should be removed both from the Requests 
and the Responses.

fantayeneh:
Fixed #676.

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/StripConnectionHeader.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/StripConnectionHeader.scala
@@ -7,13 +7,13 @@ import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack}
 object StripConnectionHeader {
 
   /**
-   * Removes the 'Connection' header from (i.e. downstream) requests.
+   * Removes the 'Connection' and header and any header it list from (i.e. downstream) requests.
    */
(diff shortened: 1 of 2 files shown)


router/http/src/main/scala/io/buoyant/router/http/StripConnectionHeader.scala
```diff
@@ -1,26 +1,31 @@
...
-
   /**
    * Removes the 'Connection' header from (i.e. downstream) requests.
    */
   object filter extends SimpleFilter[Request, Response] {
     def apply(req: Request, svc: Service[Request, Response]) = {
-      req.headerMap.remove(Key)
+      val headersListedInConnection: Seq[String] = req.headerMap.remove(Connection) match {
+        case Some(s) => s.split(",").map(_.trim).filter(_.nonEmpty)
+        case None => Seq()
```
olix0r: prefer `Nil` over `Seq()` -- Seq incurs an allocation and Nil does not.

olix0r:
Thanks for contributing to linkerd!

Issue #678, opened 2016-09-17
Linkerd should strip all Hop by Hop Headers 


fantayeneh:
Linkerd should strip out all  Hop-by-Hop headers from Requests and also from Responses.
https://www.mnot.net/blog/2011/07/11/what_proxies_must_do

The following HTTP/1.1 headers are hop-by-hop headers:
- Connection
- Keep-Alive
- Proxy-Authenticate
- Proxy-Authorization
- TE
- Trailers
- Transfer-Encoding
- Upgrade

olix0r:
We already strip Proxy-Authenticate, Proxy-Authorization and Proxy-Connection headers in ProxyRewriteFilter.  But i fully endorse stripping the rest.

Perhaps it's best to rename StripConnectionHeader to StripHopByHopHeader, and put all of the header-removal logic (including Proxy-* headers), into a single filter.

fantayeneh:
Yes, that was exactly what i would have done. :)

PR #679, opened 2016-09-18
Remove hop by hop headers

fantayeneh:
Addresses #678


router/http/src/main/scala/io/buoyant/router/http/StripHopByHopHeadersFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+        case h@Connection =>
+          val headersListedInConnection: Seq[String] = msg.headerMap.remove(h) match {
+            case Some(s) => s.split(",").map(_.trim).filter(_.nonEmpty)
+            case None => Nil
+          }
+          headersListedInConnection.foreach(msg.headerMap.remove(_))
+        case h => msg.headerMap.remove(h)
+      }
+    }
+  }
```
olix0r: Since this is going to run on _all_ requests that traverse http routers, we want to be as conservative as possible about allocation.  And, unfortunately, things like `list.foreach { _ => ... }` incur an allocation _for the anonymous function!_

We can prevent this allocation by "val-ing up" the anonymous function: 
```scala

def scrub(msg: Message): Unit =
  Headers.foreach(scrubHeader)

private[this] val scrubHeader: String => Unit = {
  case h@Connection => ...
  case h => msg.headerMap.remove(h)
}
```

This allows us to only allocate _scrubHeader_ once and not on each invocation of _scrub_.

router/http/src/main/scala/io/buoyant/router/http/StripHopByHopHeadersFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Removes all Hop-by-Hop headers and any header listed in `Connection` header from requests.
+   */
+  object filter extends SimpleFilter[Request, Response] {
+    def apply(req: Request, svc: Service[Request, Response]) = {
+      HopByHopHeaders.scrub(req)
+      svc(req) map { resp =>
```
olix0r: nit: prefer dotted `svc(req).map {`

router/http/src/main/scala/io/buoyant/router/http/StripHopByHopHeadersFilter.scala
```diff
@@ -0,0 +1,55 @@
...
+        HopByHopHeaders.scrub(resp)
+        resp
+      }
+    }
```
olix0r: oh, we probably want to val-up the response function as well.

```scala
svc(req).map(scrubResponse)

private[this] val scrubResponse: Response => Response = { resp =>
  HopByHopHeaders.scrub(resp)
  resp
}

router/http/src/main/scala/io/buoyant/router/http/StripHopByHopHeadersFilter.scala
fantayeneh: Uhm, How you "val-ing up" if `msg` need to be passed around

olix0r: D'oh, of course you're right.  I'd be inclined to remove the foreach loop and just enumerate the headers in the function.

olix0r:
:star:

PR #686, opened 2016-09-20
Add logging for Consul 5XX retries

Ashald:
As simple as subj says. Bounced the idea with @adleong in Slack before creating this PR.

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
@@ -26,11 +26,13 @@ trait BaseApi extends Closable {
   private[this] val infiniteRetryFilter = new RetryFilter[http.Request, http.Response](
     RetryPolicy.backoff(backoffs) {
       // We will assume 5xx are retryable, everything else is not for now
-      case (_, Return(rep)) => rep.status.code >= 500 && rep.status.code < 600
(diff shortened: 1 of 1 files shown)


consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -26,11 +26,17 @@ trait BaseApi extends Closable {
   private[this] val infiniteRetryFilter = new RetryFilter[http.Request, http.Response](
     RetryPolicy.backoff(backoffs) {
       // We will assume 5xx are retryable, everything else is not for now
-      case (_, Return(rep)) => rep.status.code >= 500 && rep.status.code < 600
+      case (req, Return(rep)) if rep.status.code >= 500 && rep.status.code < 600 =>
+        log.error(
+          s"Retrying Consul request '${req.method} ${req.uri}' on '${rep.status.code}' response"
```
adleong: I'm torn on if we should include the `rep.contentString` as well here.  `ConsulApiError` does so maybe we should include that here as well.

Ashald: Yeah, that's great idea!

Ashald: How do you like it now?

adleong: 👌

PR #683, opened 2016-09-19
Insert via header

fantayeneh:
Addresses #681

PR #687, opened 2016-09-20
Bump to 0.8.0

adleong:
## 0.8.0

* Allow routers to be configured with a list of identifiers.  If an identifier
  cannot assign a dest to a request, it falls back to the next one in the list.
  * **Breaking Change**: Identifier plugins must now return a
    `RequestIdentification` object.
* Consul improvements:
  * Improve performance by only watching services as necessary and tearing
    down watches when they are no longer needed.
  * Add `consistencyMode` option to `io.l5d.consul` namer
  * Add `readConsistencyMode` and `writeConsistencyMode` options to `io.l5d.consul` dtab storage
  * Consul Namerd/DtabStore: `failFast` and `failureAccrual` is now
    disabled by default but can be enabled with the `failFast` option
* Improve shutdown ordering to facilitate graceful shutdown.
  * Gracefully shutdown on SIGINT and SIGTERM.
* Require tracer configuration instead of falling back to
  defaults, reducing logging noise.
* The `debugTrace` tracer configuration flag has been removed in favor
  of the `io.l5d.tracelog` telemeter.
* Add `io.l5d.header` identifier for naming requests based on an HTTP header
* Lowercase `Host` header value in `io.l5d.methodAndHost` identifier
* Introduce transformers for post-processing the set of addresses returned by
  an interpreter.
  * Add k8s transformers to support linkerd-to-linkerd deployments when linkerd
    is deployed as a k8s daemonset.
* Remove hop-by-hop headers for better HTTP proxy compliance

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,15 +8,16 @@
   * Improve performance by only watching services as necessary and tearing
     down watches when they are no longer needed.
   * Add `consistencyMode` option to `io.l5d.consul` namer
-  * Add `readConsistencyMode` and `writeConsistencyMode` options to `io.l5d.consul` dtab storage
+  * Add `readConsistencyMode` and `writeConsistencyMode` options to
+    `io.l5d.consul` dtab storage
   * Consul Namerd/DtabStore: `failFast` and `failureAccrual` is now
     disabled by default but can be enabled with the `failFast` option
 * Improve shutdown ordering to facilitate graceful shutdown.
   * Gracefully shutdown on SIGINT and SIGTERM.
 * Require tracer configuration instead of falling back to
   defaults, reducing logging noise.
-* The `debugTrace` tracer configuration flag has been removed in favor
-  of the `io.l5d.tracelog` telemeter.
+* **Breaking Change**: The `debugTrace` tracer configuration flag has been
+  removed in favor of the `io.l5d.tracelog` telemeter.
 * Add `io.l5d.header` identifier for naming requests based on an HTTP header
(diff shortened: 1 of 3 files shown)


olix0r:
It would be Nice (not required) to get #683 into this as well if the timing works out

CHANGES.md
```diff
@@ -1,10 +1,18 @@
...
+    down watches when they are no longer needed.
+  * Add `consistencyMode` option to `io.l5d.consul` namer
+  * Add `readConsistencyMode` and `writeConsistencyMode` options to `io.l5d.consul` dtab storage
+  * Consul Namerd/DtabStore: `failFast` and `failureAccrual` is now
+    disabled by default but can be enabled with the `failFast` option
 * Improve shutdown ordering to facilitate graceful shutdown.
+  * Gracefully shutdown on SIGINT and SIGTERM.
 * Require tracer configuration instead of falling back to
   defaults, reducing logging noise.
 * The `debugTrace` tracer configuration flag has been removed in favor
```
klingerf: Maybe this should be `**Breaking Change**` as well

olix0r: I don't think it breaks anything. It's an improvement. SIGINT and SIGTERM would shut down the process before, just not before draining its queues.  In other words, I don't think we would have bumped minor version for this change alone.

klingerf: Sorry for the confusion -- was commenting on line 18.  I blame github snippets.

olix0r: D'oh. Of course you were. My eyes were drawn to the green line.

PR #693, opened 2016-09-22
Add an `ip` option to admin configuration

Ashald:
Allow users to limit admin access to i.e. localhost.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## 0.x.y
 
 * Removed unused TLS options from the k8s storage plugin config
+* Add an `ip` option to admin configuration so that access to the admin server may be constrained
 
 ## 0.8.1
 
--- a/admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
+++ b/admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
@@ -1,15 +1,18 @@
 package io.buoyant.admin
 
 import io.buoyant.config.types.Port
-import java.net.InetSocketAddress
+import java.net.{InetAddress, InetSocketAddress}
 
 case class AdminConfig(
-  port: Option[Port],
-  shutdownGraceMs: Option[Int]
+  ip: Option[InetAddress] = None,
+  port: Option[Port] = None,
+  shutdownGraceMs: Option[Int] = None
 ) {
 
-  def mk(defaultPort: Port): Admin = {
-    val addr = new InetSocketAddress(port.getOrElse(defaultPort).port)
+  def mk(defaultAddr: InetSocketAddress): Admin = {
+    val adminIp = ip.getOrElse(defaultAddr.getAddress)
+    val adminPort = port.map(_.port).getOrElse(defaultAddr.getPort)
+    val addr = new InetSocketAddress(adminIp, adminPort)
     new Admin(addr)
   }
 }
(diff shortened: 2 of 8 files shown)


admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
```diff
@@ -1,15 +1,18 @@
 package io.buoyant.admin
 
 import io.buoyant.config.types.Port
-import java.net.InetSocketAddress
+import java.net.{InetAddress, InetSocketAddress}
 
 case class AdminConfig(
+  ip: Option[InetAddress],
   port: Option[Port],
```
olix0r: We have a HostAndPort type.  Perhaps it's preferable to use that?

Ashald: Let me take a look...

Ashald: How would you like field called? `endpoint`?

adleong: I think if we went the HostAndPort route then we would get rid of the `AdminConfig` class and then in `LinkerConfig` the `admin` field would simply be of type `HostAndPort`.

I could go either way on doing that or leaving it as is.

Ashald: Yeah, for backward compatibility we either need to get rid of `AdminConfig` class and replace it with `HostAndPort` or stick to current implementation.

IMHO, it might be worth sticking to current implementation as `AdminConfig` might need extra config parameters later. And we always can refactor it and do what you suggested.

Please let me know what you think and what should I do.

olix0r: We're breaking backwards compatibility by changing the defaults, though.  Either we shouldn't change the default, or I think we should go ahead and break compat.

I think an admin section makes sense, regardless, as I think we want to be able to add additional configuration to the admin subystem.  For instance, we may want to make the behavior of the shutdown and gc endpoints configurable...

adleong: sounds like we all agree 😄

admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
```diff
@@ -1,15 +1,18 @@
...
   shutdownGraceMs: Option[Int]
 ) {
 
-  def mk(defaultPort: Port): Admin = {
-    val addr = new InetSocketAddress(port.getOrElse(defaultPort).port)
+  def mk(defaultPort: Port, defaultIp: InetAddress): Admin = {
```
olix0r: can we just have `def mk(defaultAddr: InetSocketAddress)`?

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 ## 0.x.y
 
 * Removed unused TLS options from the k8s storage plugin config
+* An `ip` option in admin config to set ip to listen on
```
olix0r: Prefer that this start with a verb. Something like:
> Add an `ip` option to `admin` configuration so that access to the admin server may be constrained

Ashald: Thanks! Will fix

Ashald:
Also rebased brunch on `master`.

olix0r:
doh, looks like you're out of sync with master again ;/

Ashald:
Will rebase once I'll get home in few hours ;)

Ashald:
Another attempt ;)

Ashald:
Meeerge, I neeed somebody meeerge... me... :)

olix0r:
Merged https://github.com/BuoyantIO/linkerd/commit/d2dc194e584d06cca03fcc5822e774a041eced60

PR #694, opened 2016-09-22
Remove unused options in k8s storage config

adleong:
These options are unused and can be removed.  One should point this at a locally running `kubectl proxy` to securely talk to the API.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## 0.x.y
+
+* Removed unused TLS options from the k8s storage plugin config
+
 ## 0.8.1
 
 * Fix missing data on the linkerd admin dashboard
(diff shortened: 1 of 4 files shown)


klingerf:
⭐  Looks good. Mind adding something to CHANGES.md as well, in case anybody is surprised that these flags went away?

Also, not part of your change, but while you're in here, it looks like there's a config issue in the namerd example k8s config -- it's missing the `experimental: true` flag for k8s storage.  Maybe add that as well?

adleong:
Hmmm... I don't see where the missing flag is supposed to go.  I see `experimental: true` on line 118.  Am I blind?

klingerf:
Ah, sorry, was not specific in my last comment.  I think we need to add the flag here:

https://github.com/BuoyantIO/linkerd/blob/master/namerd/examples/k8s.yaml#L3

PR #697, opened 2016-09-23
Add insert rewriting namer

### LINKED ISSUE #695: Create a rewriting namer to allow appending path segments
It is impossible to append segments to the end of a path using dentries.  Create a rewriting namer that does this.  

This would allow paths like `/host/foo` to be rewritten to `/$/inet/foo/80`.

adleong:
Fixes #695 

A rewriting namer that accepts names in the form /<index>/<segment>/<rest*>
and inserts <segment> at position <index> in <rest*>.  For example,
 ```
/2/foo/zero/one/two/three
```
will insert foo at position 2, resulting in:
```
/zero/one/foo/two/three
```

screen shot 2016-09-23 at 3 25 29 pm

### DIFF
--- a/router/http/src/main/scala/io/buoyant/http/namer.scala
+++ b/router/http/src/main/scala/io/buoyant/http/namer.scala
@@ -1,5 +1,6 @@
 package io.buoyant.http
 
+import com.twitter.finagle.http.path.Integer
 import com.twitter.finagle.{Name, NameTree, Namer, Path, Service, ServiceNamer}
 import com.twitter.finagle.http._
 import com.twitter.util.{Activity, Future, Try}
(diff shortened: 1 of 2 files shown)


olix0r:
This is cool! It seems useful, but I want to be careful about adding something like this. This sort of thing makes dtabs even more complex, so we should be sure it's worth introducing such a generic utility like this.

adleong:
I agree that this is a powerful and dangerous tool for when we need to break the assumption that paths are hierarchical.  Let's keep this in our back pocket until a need arises.

Ashald:
FWIW, I'm about to release an OSS lib with a set of rewriting namers inspired by what exists in Linkerd now. If you fill hesitant about adding such namers to linkerd core we can put them into that lib and advertise it to whoever needs them (of course that lib can be sued as a Linkerd plugin). And later we can see whether there is a demand and whether it's worth adding them to the core.

olix0r:
closing for now, reopen as needed

PR #701, opened 2016-09-27
Document the namerd http API

adleong:


namerd/docs/interface.md
```diff
@@ -40,9 +40,243 @@ addrCacheInactive | `100` | The size of the address inactive cache.
...
+exist, or 409 if the If-Match header is provded and does not match the current
+dtab version.
+
+Parameter | Type | Description
+--------- | ---- | -----------
+namespace | path | The dtab namespace to update.
+Content-Type | header | The content type of the provided dtab.
+If-Match | header | If provided, the update will only be applied if the If-Match
+                    header matches the current dtab version.  This can be used
+                    to prevent conflicting updates.
```
esbie: this formatting is nice here but breaks the markdown table

namerd/docs/interface.md
```diff
@@ -40,9 +40,243 @@ addrCacheInactive | `100` | The size of the address inactive cache.
...
 kind: `io.l5d.httpController`
 
-A read-write HTTP interface to the [storage](#storage).
+The HTTP controller provides APIs for reading and writing dtabs, as well as for
+viewing how names are resolved.
 
 Key | Default Value | Description
 --- | ------------- | -----------
 ip | loopback | The local IP address on which to serve the namer interface.
 port | `4180` | The port number on which to serve the namer interface.
```
adleong: Do you think we should?  It doesn't seem necessary to me but I'm open to it.

namerd/docs/interface.md
```diff
@@ -40,9 +40,243 @@ addrCacheInactive | `100` | The size of the address inactive cache.
...
+
+Returns the requested dtab.  The dtab version is returned in the Etag response
+header.  Returns status code 404 if the dtab namespace does not exist.
+
+Parameter | Type | Description
+--------- | ---- | -----------
+namespace | path | The dtab namespace to retrieve.
+watch | uri | If true, updates are returned as a streaming response.
+
+Content-types: application/json, application/dtab
```
adleong: Yeah, Accept header is how to select which content-type you want.  My understanding is that's standard practice for http?  Should I call it out anyway?

namerd/docs/interface.md
```diff
@@ -40,9 +40,243 @@ addrCacheInactive | `100` | The size of the address inactive cache.
...
+[{"prefix":"/http/1.1/GET","dst":"/#/io.l5d.fs"}]
+```
+
+Returns the requested dtab.  The dtab version is returned in the Etag response
+header.  Returns status code 404 if the dtab namespace does not exist.
+
+Parameter | Type | Description
+--------- | ---- | -----------
+namespace | path | The dtab namespace to retrieve.
+watch | uri | If true, updates are returned as a streaming response.
```
adleong: I think I'll hold off for now, just because it either make things inconsistent if I only apply it to the new tables or it will make the diff a mess if I apply it everywhere.  It might be worth doing in a followup though.

adleong:
@esbie given that the sample requests sometimes contain bodies and/or headers, do you have a preferred format for the requests?  curl command?  raw HTTP request?

I'm also not sure of a good way to describe the response formats that isn't extremely verbose.

esbie:
@adleong I like using a curl command. I agree that response formats are not always that easy to document. Internally we did some format documentation here: https://github.com/BuoyantIO/helium/pull/1225/files#diff-0f426376b964e609c5c4618eb8bed3b2R41
But I think it's also fine to punt on it for now.

Issue #684, opened 2016-09-20
Consul: reference 'local' datacenter


Ashald:
A proposal to add a feature that would allow to request services from 'local' datacenter (datacenter the Consul API host belongs to) without specifying its name. It would allow us to make linkerd/namerd configs static regardless of the datacenter they're deployed to.

A proposed way to do that to add an extra option like:
```
localDatacenterMoniker: _local
```
so that then used in namer like `/#/io.l5d.consul/_local/foobar` it would mean to take the service from the "current datacenter"

Ashald:
I think it can be closed now as the PR was merged.

PR #702, opened 2016-09-28
Consul Local Datacenter Moniker

Ashald:
I thought about doing PoC for #684 and ended up with a PR.

Please let me know what do you think. Not sure I like the parameter name and doc for it...

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -205,7 +205,7 @@ failFast | `false` | If `false`, disable fail fast and failure accrual for Consu
 Key | Required | Description
 --- | -------- | -----------
 prefix | yes | Tells linkerd to resolve the request path using the consul namer.
-datacenter | yes | The Consul datacenter to use for this request.
+datacenter | yes | The Consul datacenter to use for this request. It can have a value `.local` (otherwise invalid datacenter name from Consul's perspective) in order to reference a datacenter of the agent namer is connected to.
 tag | yes if includeTag is `true` | The Consul tag to use for this request.
 serviceName | yes | The Consul service name to use for this request.
 
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
@@ -17,6 +17,8 @@ private[consul] class LookupCache(
   stats: StatsReceiver = NullStatsReceiver
 ) {
 
+  private[this] val localDcMoniker = ".local"
+
   private[this] val lookupCounter = stats.counter("lookups")
 
   def apply(
@@ -28,7 +30,7 @@ private[consul] class LookupCache(
     log.debug("consul lookup: %s %s", id.show)
     lookupCounter.incr()
 
-    Dc.watch(dc).map { services =>
+    resolveDc(dc).flatMap(Dc.watch).map { services =>
       services.get(key) match {
         case None =>
           log.debug("consul dc %s service %s missing", dc, key)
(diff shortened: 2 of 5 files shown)


namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -41,13 +42,21 @@ private[consul] class LookupCache(
...
-  // Note that this activity is never recomputed.  We simply do a
+  private[this] def resolveDc(Datacenter: String): Activity[String] = localDc match {
+    case Some(Datacenter) => config.map(_.flatMap(_.Datacenter).getOrElse(Datacenter))
+    case _ => Activity.value(Datacenter)
+  }
+
+  // Note that this 2 activities are never recomputed.  We simply do a
   // lookup for the domain and then wrap it an activity for
   // convenience.
+  private[this] lazy val config: Activity[Option[v1.Config]] =
```
adleong: can we name this agentConfig or localAgentConfig?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -41,13 +42,21 @@ private[consul] class LookupCache(
     }
   }
 
-  // Note that this activity is never recomputed.  We simply do a
+  private[this] def resolveDc(Datacenter: String): Activity[String] = localDc match {
+    case Some(Datacenter) => config.map(_.flatMap(_.Datacenter).getOrElse(Datacenter))
```
adleong: lowercase parameter name please.  more idiomatic would be

```
if (localDc.contains(datacenter)) {
   ...
} else {
  ...
}
```

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -41,13 +42,21 @@ private[consul] class LookupCache(
     }
   }
 
-  // Note that this activity is never recomputed.  We simply do a
+  private[this] def resolveDc(Datacenter: String): Activity[String] = localDc match {
```
adleong: I found this pretty confusing because `localDc` isn't actually the local dc, it's the moniker.

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
adleong: it may clarify things to add a third activity called `localDc` which is `config.map(_.flatMap(_.Datacenter))`

linkerd/docs/namer.md
```diff
@@ -187,6 +187,7 @@ token | no authentication | The auth token to use when making API calls.
 setHost | `false` | If `true`, HTTP requests resolved by Consul will have their Host header overwritten to `${serviceName}.service.${datacenter}.${domain}`. `$domain` is fetched from Consul.
 consistencyMode | `default` | Select between [Consul API consistency modes](https://www.consul.io/docs/agent/http.html) such as `default`, `stale` and `consistent`. 
 failFast | `false` | If `false`, disable fail fast and failure accrual for Consul client. Keep it `false` when using a local agent but change it to `true` when talking directly to an HA Consul API 
+localDcMoniker | none | If set, can be used to reference a datacenter of the agent the namer is connected to. E.g., if connected to an agent in datacenter `dc1` given `localDcMoniker: _local` name `/#/io.l5d.consul/_local/myservice` is equal to `/#/io.l5d.consul/dc1/myservice`
```
adleong: Do you think this really needs to be configurable?  Could we just use `.local`?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
Ashald: Should I rename it to `localDcMoniker` or `localDcAlias`?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
Ashald: I used upper-case to make it stable identifier so I can use it in match but it's really not needed with the approach you proposed - thanks

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
Ashald: Sure

Ashald: I thought about that but wasn't sure where we need an extra activity just for that... But if you say that it will clarify things - then sure, will do!

linkerd/docs/namer.md
Ashald: Well, I don't mind hardcoding `.local` (as I don't see how that could be a valid DC name in Consul) - I though you won't like that. :) So, should I remove this option then and update the doc?

Issue #280, opened 2016-04-13
Add a test that validates that the examples configs can be parsed and run


adleong:
We should have a test that linkerd can start with each of the configs in linkerd/examples and a test that namerd can start with each of the configs in namerd/examples.  This helps to protect us from our examples getting out of date.

adleong:
Fixed in #696

PR #706, opened 2016-09-29
Fix units in docs & types in tests for HTTP router

Ashald:
1) Docs are giving ambiguous idea about the proper way to configure http
router request/response size limits
2) HttpConfigTest/"parse config" test uses same config as in docs and
actually doesn't work as expected, maybe related to #280


Ashald:
Branch was out of sync with master so I updated it just in case.

Ashald:
Meh, `MuxEndToEndTest` failed again. Can you please re-run tests on circleci please?

Issue #709, opened 2016-10-03
Linkerd does not try the fallback (or second) Namer, when Health API returns empty list of addresses.


ManishMaheshwari:
**Configuration used:**

```
namers:
- kind: io.l5d.consul
  experimental: true
  host: 127.0.0.1
  port: 8500
  useHealthCheck: true
  failFast: true  
  includeTag: false
routers:
- protocol: http
  httpAccessLog: logs/access.log
  label: int
  dstPrefix: /http
  baseDtab: |
    /consul =>  /#/io.l5d.consul/dc1 | /$/inet/load-balancer-dc1/80;
    /host => /consul;
    /http/1.1/* => /host; 
```

**Given:** That dtab is configured with alternate namers
**When:** The Health API is configured to return only passing instances
**And:** When there are no passing instances to return by the Health API.
**Expectation:** Linkerd must update the binding cache with NameTree.Neg, and fall back to the second namer configured in dtab.
**Observation:** Linkerd does not fallback to the second namer, and continues to responds the client with HTTP 502:

```
No hosts are available for /http/1.1/GET/someservice, Dtab.base=[/consul=>/#/io.l5d.consul/dc1 | /$/inet/load-balancer-dc1/80;/host=>/consul;/http/1.1/*=>/host], Dtab.local=[]. Remote Info: Not Available
```

adleong:
Fixed by #744

PR #710, opened 2016-10-03
Resolving #709 Linkerd does not try the fallback Namer

ManishMaheshwari:
Resolving #709 Linkerd does not try the fallback Namer, when Health API returns empty list.

The cache is updated to use NameTree.Neg, in case there are no addresses available in HealthApi response.


adleong:
Hi @ManishMaheshwari!  This is a good idea but I think the implementation may be a bit more complex.  Instead of using `.sample()` we should use `map` to make sure that the NameTree always has the correct value.  Unfortunately, this means that the NameTree will need to update every time the Addr Var changes, which is not good.  

Ideally, we would like to set the Addr to Addr.Neg when the Health API returns an empty list and linkerd should fall back to alternates in the NameTree.  However, I'm not sure if that's how Addr.Neg is handled today.  I've filed https://github.com/BuoyantIO/linkerd/issues/712 to investigate.

adleong:
Closing this in favor of something like https://github.com/linkerd/linkerd/pull/744

Issue #708, opened 2016-10-02
Linkerd removes tailing slash '/' on request


leozc:
Hello I found out that  linkerd kills the trailing ‘/‘ with io.l5d.path as identifier.

For example - this is a request went through linerd routing:
```curl -v r.foo.com:4140/ou_route_v11/iad/staging/nc/api/v2/item/6351359/ 
```
And I recevied
```
$ nc -l 9999
GET /api/v2/item/6351359 HTTP/1.1
Host: r.foo.com:4140
User-Agent: curl/7.43.0
Accept: */*
content-length: 0
l5d-dst-logical: /http/ou_route_v11/iad/staging/nc
Via: 1.1 linkerd
l5d-dst-concrete: /#/io.l5d.consul/dc1/staging/nc
l5d-ctx-trace: gC27uDZpcgddEpRa8TQSll0SlFrxNBKWAAAAAAAAAAA=
l5d-reqid: 5d12945af1341296
````

Where the trailing '/' is missing.

Why this can cause problem?
The framework (Django) of this endpoint expected a trailing ‘/‘ e.g. ```api/v2/item/6351359/ ```, and hence issued a 301 redirect client to ```api/v2/item/6351359/ ```, again, it won't work, since the subsequent request still failed for the same reason.

Linkerd config
```
admin:
  port: 9990

routers:
- protocol: http
  httpAccessLog: access.log
  identifier:
    kind: io.l5d.path
    segments: 4
    consume: true

  interpreter:
    kind: io.l5d.namerd
    dst: /$/inet/namerd/4100
  servers:
  - port: 4140
    ip: 0.0.0.0
    engine:
      kind: netty4 # need additional test with netty3 vs netty4
  client:
    engine:
      kind: netty3
  responseClassifier:
    kind: io.l5d.retryableRead5XX
```

Namerd config
```
admin:
  port: 9991

namers:
- kind: io.l5d.consul
  host: consul
  experimental: true
  port: 8500
  includeTag: true
storage:
  kind: io.l5d.consul
  host: consul
  experimental: true
  port: 8500
interfaces:
- kind: io.l5d.thriftNameInterpreter
  ip: 0.0.0.0
  port: 4100
- kind: io.l5d.httpController
  ip: 0.0.0.0
  port: 4180
```

leozc:
A test to replicate the error

`  test("preserve trailing slash") {
    val identifier = PathIdentifier(Path.Utf8("http"), 2, consume = true)
[...]

PR #711, opened 2016-10-03
preserving tailing slash in io.l5d.path

leozc:
A fix for #708. 
Plus two unit tests.

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
@@ -21,7 +21,13 @@ case class PathIdentifier(
     req.path.split("/").drop(1) match {
       case path if path.size >= segments =>
         if (consume) {
-          req.uri = req.uri.split("/").drop(segments + 1).mkString("/", "/", "")
+          req.uri = req.uri.split("/").drop(segments + 1) match {
+            case Array() => "/"
+            case x => {
+              val endWithSlash = if (req.path.endsWith("/")) "/" else ""
+              x.mkString("/", "/", endWithSlash)
+            }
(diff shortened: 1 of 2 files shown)


router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -21,7 +21,13 @@ case class PathIdentifier(
     req.path.split("/").drop(1) match {
       case path if path.size >= segments =>
         if (consume) {
-          req.uri = req.uri.split("/").drop(segments + 1).mkString("/", "/", "")
+          req.uri = req.uri.split("/").drop(segments + 1) match {
+            case Array() => "/"
+            case x => {
```
adleong: style nit: no braces for cases

router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -21,7 +21,13 @@ case class PathIdentifier(
...
+              val endWithSlash = if (req.path.endsWith("/")) "/" else ""
```
adleong: this might be more clear if named `trailingSlash` since it's not a boolean

router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
leozc: fixed

router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
leozc: opps, fixed

PR #714, opened 2016-10-03
Serverside TLS support for Netty4 servers

olix0r:
Netty4 servers cannot be configured with TLS, since server TLS configuration
parameters only configure Netty3's TLS subsystem.

Now, server TLS configuration influences netty4 servers as well as netty3
servers. Furthermore, servers may now be configured with a list of supported
ciphers.

Furthermore, the linkerd-tls package now depends on a static boringssl
implementation so that linkerd ships with a functional cross-platform SSL
provider.

NOTE: netty4 servers require that keys be encoded in the PKCS#8 format.

Reported-by: @moderation

### DIFF
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -9,7 +9,7 @@ unitTests() {
 # we can't compute coverage on e2e tests because it
 # conflicts/overwrites unit coverage for some modules.
 e2eTests() {
-  ./sbt e2e:test integration:compile
+  ./sbt e2e:test integration:compile linkerd-protocol-http/integration:test
 }
 
 case "${CIRCLE_NODE_TOTAL:-1}" in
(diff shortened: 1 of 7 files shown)


adleong:
Cool.  What about clientside TLS with netty4?  Does that already work or is that TODO?

olix0r:
Client-side TLS already works (as validated by integration tests)

moderation:
Woohoo. Works with a bunch of `com.twitter.finagle.UnknownChannelException: Response entity too large: DefaultHttpResponse(decodeResult: success, version: HTTP/1.1)` errors, but it does work. Hopefully the Netty 4 stuff improves as Finagle integration matures. Thanks @olix0r

PR #715, opened 2016-10-03
Provide a default log4j configurations with linkerd and namerd

olix0r:
Currently, no log4j configuration is provided, and so linkerd emits a warning
at startup about the missing configuration.

We now provide a default log4j configuration that emits logs on the standard
error stream of linkerd and namerd.

Furthermore, init scripts now support a LOCAL_JVM_OPTIONS environment variable
to support extending JVM parameters (i.e. to specify an external
log4j.properties file).

Outputs in the format:
> DEBUG 1003 23:18:50.068 main: -Dio.netty.processId: 71323 (auto-detected)
> DEBUG 1003 23:18:50.076 main: Loopback interface: lo0 (lo0, 127.0.0.1)
> DEBUG 1003 23:18:50.077 main: /proc/sys/net/core/somaxconn: 128 (non-existent)

which isn't _quite_ the same as that of finagle, but it's close enough given the effort needed to make them identical.

### DIFF
--- a/linkerd/main/src/main/resources/log4j.properties
+++ b/linkerd/main/src/main/resources/log4j.properties
@@ -1,0 +1,5 @@
+log4j.rootLogger=INFO, stderr
+log4j.appender.stderr=org.apache.log4j.ConsoleAppender
+log4j.appender.stderr.Target=System.err
+log4j.appender.stderr.layout=org.apache.log4j.PatternLayout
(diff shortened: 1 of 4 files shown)


olix0r:
com.twitter.util.events.sinkEnabled controls whether finagle emits diagnostic events through this API.  In our testing, we've found it negatively impacts performance (and I happened to discover this on the same branch that I had made logging changes, so just lumped it in).

-log.level still works.  This log4j config is really there to control netty's logging.  AFAICT it has no impact on finagle's logging.

PR #723, opened 2016-10-08
Make Identifiers generically, lazily composable

olix0r:
Previously, HttpConfig had logic for composing Identifiers. This is moved into the RoutingFactory so it may be used across protocols.

Furthermore, identifier composition was eager, identifying each request through all identifiers and then checking the results in order.  Now, identifiers are checked serially. This change is made under the assumption that Identifiers tend to be fast but incur allocation.

### DIFF
--- a/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
+++ b/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
@@ -16,6 +16,42 @@ object RoutingFactory {
    */
   type Identifier[Req] = Req => Future[RequestIdentification[Req]]
 
+  object Identifier {
+
+    def compose[Req](hd: Identifier[Req], tl: Identifier[Req]*): Identifier[Req] =
+      (req: Req) => fold(req, hd, tl)
+
+    /**
+     * Apply each identifier in order until the request is identified.
+     *
+     * Identifier list must not be empty.
+     */
+    def compose[Req](all: Seq[Identifier[Req]]): Identifier[Req] = all match {
+      case Nil => throw new IllegalArgumentException("empty identifier list")
+      case Seq(identifier) => identifier
+      case Seq(hd, tl@_*) => (req: Req) => fold(req, hd, tl)
+    }
+
+    /**
+     * Apply each identifier to the request in order until the request
+     * is identified or identifiers are exhausted.
+     */
+    private[this] def fold[Req](
+      req: Req,
+      hd: Identifier[Req],
+      tl: Seq[Identifier[Req]]
+    ): Future[RequestIdentification[Req]] =
+      tl match {
+        case Nil => hd(req)
+        case Seq(nextHd, nextTl@_*) =>
+          hd(req).flatMap {
+            case id: IdentifiedRequest[Req] => Future.value(id)
+            case _: UnidentifiedRequest[Req] => fold(req, nextHd, nextTl)
+          }
+      }
+
+  }
+
   /** The result of attempting to identify a request. */
   sealed trait RequestIdentification[Req]
 
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -133,24 +133,9 @@ case class HttpConfig(
   override val protocol: ProtocolInitializer = HttpInitializer
 
   @JsonIgnore
-  private[this] val combinedIdentifier = {
-    // use the first identifier that yields an IdentifiedRequest
-    def combine(identifiers: Seq[RoutingFactory.Identifier[Request]]): RoutingFactory.Identifier[Request] = { req =>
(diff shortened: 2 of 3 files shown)


adleong:
Are you sure that the previous implementation, based on `foldLeft` identified requests eagerly?  By my reading, that checked the identifiers serially until an identification was found.

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -15,6 +15,40 @@ object RoutingFactory {
    * destination]] for `Req`-typed requests.
    */
   type Identifier[Req] = Req => Future[RequestIdentification[Req]]
+  object Identifier {
+    private def fail[Req]: Future[RequestIdentification[Req]] =
```
adleong: why not just make this a lazy val?

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -15,6 +15,40 @@ object RoutingFactory {
...
+      def apply(req: Req): Future[RequestIdentification[Req]] = {
+        def iter(all: Seq[Identifier[Req]]): Future[RequestIdentification[Req]] =
+          all match {
+            case Seq(hd) => hd(req)
+            case Seq(hd, tl@_*) =>
+              hd(req).flatMap {
+                case id: IdentifiedRequest[Req] => Future.value(id)
+                case _: UnidentifiedRequest[Req] => iter(tl)
+              }
+            case Nil => fail[Req] // unreachable, tho!
```
adleong: TIOLI: you could encode this unreachability in the type system by having this class take `(hd: Identifier[Req], tl: Identifier[Req]*)` instead of `(all: Seq[Identifier[Req]])`

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -15,6 +15,40 @@ object RoutingFactory {
...
+      case Seq(identifier) => identifier
+      case all => new Composed(all)
+    }
+
+    def compose[Req](hd: Identifier[Req], tl: Identifier[Req]*): Identifier[Req] =
+      compose(hd +: tl)
+
+    private[this] class Composed[Req](all: Seq[Identifier[Req]]) extends Identifier[Req] {
+      def apply(req: Req): Future[RequestIdentification[Req]] = {
+        def iter(all: Seq[Identifier[Req]]): Future[RequestIdentification[Req]] =
```
adleong: isn't this just essentially a fold?

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
olix0r: it's parameterized on [Req]

olix0r:
You're right that it's not as eager as I thought:
```scala
      identifiers.foldLeft(HttpConfig.NilIdentification) { (identification, next) =>
        identification.flatMap {
          // the request has already been identified, just use that
          case identified: IdentifiedRequest[Request] => Future.value(identified)
          // the request has not yet been identified, try the next identifier
          case unidentified: UnidentifiedRequest[Request] => next(req)
        }
      }
```

This _does_ eagerly create the future chain (i.e. we call f.flatMap for all identifiers), though identifiers aren't actually invoked until the future is satisfied and unidentified.

I find the combination of Seq.foldLeft and Future to be a bit mind-bendy.  This implementation definitely isn't bad, though I do find the recursive approach to be a it clearer (or, more idiomatic for future looping/iteration)

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -16,6 +16,42 @@ object RoutingFactory {
    */
   type Identifier[Req] = Req => Future[RequestIdentification[Req]]
 
+  object Identifier {
+
+    def compose[Req](hd: Identifier[Req], tl: Identifier[Req]*): Identifier[Req] =
+      compose(hd +: tl)
```
adleong: would be more direct to go straight to `(req: Req) => fold(req, hd, tl)` here

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -16,6 +16,42 @@ object RoutingFactory {
...
+      case Nil => throw new IllegalArgumentException("empty identifier list")
+      case Seq(identifier) => identifier
+      case Seq(hd, tl@_*) => (req: Req) => fold(req, hd, tl)
+    }
+
+    /**
+     * Apply each identifier to the request in order until the request
+     * is identified or identifiers are exhausted.
+     */
+    private[this] def fold[Req](
```
adleong: could add the `@tailrec` annotation I think

olix0r: not with future.flatMap -- it's still "safe" in that this won't blow the stack, but tailrec and futures are not friends.

PR #729, opened 2016-10-10
Add k8s-external namer

### LINKED ISSUE #726: Add a namer for k8s ingress services
Add a new namer that accepts names of the form:

```
/<namespace>/<port name>/<service name>
```

which resolves to the external load balancer ip for the given k8s service object.  This is useful for linkerd instances which run outside of k8s but need to route requests to services in k8s.

adleong:
Fixes #726

The [Kubernetes](https://k8s.io/) External namer looks up the IP of the external
load balancer for the given service on the given port.  This can be used by
linkerd instances running outside of k8s to route to services running in k8s.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/Ns.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Ns.scala
@@ -1,0 +1,93 @@
+package io.buoyant.k8s
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.service.Backoff
+import com.twitter.finagle.tracing.Trace
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.util._
+
+abstract class Ns[O <: KubeObject: Manifest, W <: Watch[O]: Manifest, L <: KubeList[O]: Manifest, Cache](
+  backoff: Stream[Duration] = Backoff.exponentialJittered(10.milliseconds, 10.seconds),
+  timer: Timer = DefaultTimer.twitter
+) {
+  // note that caches must be updated with synchronized
+  private[this] val caches = Var[Map[String, Cache]](Map.empty[String, Cache])
+  // XXX once a namespace is watched, it is watched forever.
+  private[this] var _watches = Map.empty[String, Activity[Closable]]
+  def watches = _watches
+
+  /**
+   * Returns an Activity backed by a Future.  The resultant Activity is pending until the
+   * original future is satisfied.  When the Future is successful, the Activity becomes
+   * an Activity.Ok with a fixed value from the Future.  If the Future fails, the Activity
+   * becomes an Activity.Failed and the Future is retried with the given backoff schedule.
+   * Therefore, the legal state transitions are:
+   *
+   * Pending -> Ok
+   * Pending -> Failed
+   * Failed -> Failed
+   * Failed -> Ok
+   */
+  private[this] def retryToActivity[T](go: => Future[T]): Activity[T] = {
+    val state = Var[Activity.State[T]](Activity.Pending)
+    _retryToActivity(backoff, state)(go)
+    Activity(state)
+  }
+
+  private[this] def _retryToActivity[T](
+    remainingBackoff: Stream[Duration],
+    state: Var[Activity.State[T]] with Updatable[Activity.State[T]] = Var[Activity.State[T]](Activity.Pending)
+  )(go: => Future[T]): Unit = {
+    val _ = go.respond {
+      case Return(t) =>
+        state() = Activity.Ok(t)
+      case Throw(e) =>
+        state() = Activity.Failed(e)
+        remainingBackoff match {
+          case delay #:: rest =>
+            val _ = Future.sleep(delay)(timer).onSuccess { _ => _retryToActivity(rest, state)(go) }
+          case Stream.Empty =>
+        }
+    }
+  }
+
+  protected def mkResource(name: String): NsListResource[O, W, L]
+
+  protected def mkCache(name: String): Cache
+
+  protected def initialize(cache: Cache, list: L): Unit
+
+  protected def update(cache: Cache)(event: W): Unit
+
+  def get(name: String): Cache = synchronized {
+    caches.sample.get(name) match {
+      case Some(ns) => ns
+      case None =>
+        val ns = mkCache(name)
+        val closable = retryToActivity { watch(name, ns) }
+        _watches += (name -> closable)
+        caches() = caches.sample + (name -> ns)
+        ns
+    }
+  }
+
+  val namespaces: Var[Set[String]] = caches.map(_.keySet)
+
+  private[this] def watch(namespace: String, cache: Cache): Future[Closable] = {
+    val resource = mkResource(namespace)
+    Trace.letClear {
+      log.info("k8s initializing %s", namespace)
+      resource.get().map { list =>
+        initialize(cache, list)
+        val (updates, closable) = resource.watch(
+          resourceVersion = list.metadata.flatMap(_.resourceVersion)
+        )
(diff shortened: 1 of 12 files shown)


k8s/src/main/scala/io/buoyant/k8s/Ns.scala
```diff
@@ -0,0 +1,93 @@
...
+
+abstract class Ns[O <: KubeObject: Manifest, W <: Watch[O]: Manifest, L <: KubeList[O]: Manifest, Cache](
+  backoff: Stream[Duration] = Backoff.exponentialJittered(10.milliseconds, 10.seconds),
+  timer: Timer = DefaultTimer.twitter
+) {
+  // note that caches must be updated with synchronized
+  private[this] val caches = Var[Map[String, Cache]](Map.empty[String, Cache])
+  // XXX once a namespace is watched, it is watched forever.
+  private[this] var _watches = Map.empty[String, Activity[Closable]]
+  def watches = _watches
```
olix0r: nit: would be nice to annotate the type since this is basically a public interface now

olix0r: does this need to be public?

olix0r: does it even need to exist?

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -0,0 +1,174 @@
...
+      Activity.value(NameTree.Neg)
+  }
+
+  private[this] def getPort(service: Service, portName: String): Option[Int] =
+    for {
+      spec <- service.spec
+      port <- spec.ports.find(_.name == portName)
+    } yield port.port
+
+  val serviceNs = new Ns[Service, ServiceWatch, ServiceList, ServiceCache](backoff, timer) {
```
olix0r: does this need to be public?

also, prefer type annotation like:
```scala
val serviceNs: Ns[Service, ServiceWatch, ServiceList, ServiceCache] = new Ns(backoff, timer {
```

I care less about this if it's not public.

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  def update(watch: ServiceWatch): Unit = synchronized {
+    watch match {
+      case ServiceAdded(service) =>
+        for {
+          meta <- service.metadata
+          name <- meta.name
+        } {
+          log.info("k8s added service: %s", name)
```
olix0r: i imagine this logging (and elsewhere) would be more useful if it included the namespace?

k8s/src/main/scala/io/buoyant/k8s/v1.scala
```diff
@@ -15,43 +15,74 @@ package object v1 {
...
+  }
+
+  @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+  @JsonSubTypes(Array(
+    new JsonSubTypes.Type(value = classOf[ServiceAdded], name = "ADDED"),
+    new JsonSubTypes.Type(value = classOf[ServiceModified], name = "MODIFIED"),
+    new JsonSubTypes.Type(value = classOf[ServiceDeleted], name = "DELETED"),
+    new JsonSubTypes.Type(value = classOf[ServiceError], name = "ERROR")
+  ))
+  sealed trait ServiceWatch extends Watch[Service]
```
olix0r: tioli: i know this naming comes from the prior api, but might it be more appropriate to s/Watch/Event/ ?

k8s/src/main/scala/io/buoyant/k8s/Ns.scala
```diff
@@ -0,0 +1,93 @@
...
+import com.twitter.finagle.tracing.Trace
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.util._
+
+abstract class Ns[O <: KubeObject: Manifest, W <: Watch[O]: Manifest, L <: KubeList[O]: Manifest, Cache](
+  backoff: Stream[Duration] = Backoff.exponentialJittered(10.milliseconds, 10.seconds),
+  timer: Timer = DefaultTimer.twitter
+) {
+  // note that caches must be updated with synchronized
+  private[this] val caches = Var[Map[String, Cache]](Map.empty[String, Cache])
```
olix0r: It seems like we actually have a pretty well-defined for interface for Cache.  Does it make more sense to just have NsCache and ServiceCache implement a common interface that handles initialize, update, etc?

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -0,0 +1,174 @@
...
+          update() = None
+      }
+    }
+  }
+
+  private[this]type VarUp[T] = Var[T] with Updatable[T]
+
+  private[this] var cache = Map.empty[String, VarUp[Map[String, Address]]]
+
+  private[this] def extractPorts(service: Service): Map[String, Address] = {
```
olix0r: tioli: there are a few stateless private helpers in this file.  You could move them into a companion object to more clearly differentiate static-helper and stateful-updatey logic

namer/k8s/src/main/scala/io/buoyant/namer/k8s/K8sExternalInitializer.scala
```diff
@@ -0,0 +1,54 @@
...
+ * namers:
+ * - kind: io.l5d.k8s-external
+ *   exerimental: true
+ *   host: localhost
+ *   port: 8001
+ * </pre>
+ */
+class K8sExternalInitializer extends NamerInitializer {
+  val configClass = classOf[K8sExternalConfig]
+  override def configId = "io.l5d.k8s-external"
```
olix0r: this is the first/only dashed name in our kinds...  Perhaps io.l5d.k8s.external?

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -0,0 +1,174 @@
...
+    ): Unit = cache.initialize(list)
+  }
+}
+
+class ServiceCache(namespace: String) {
+
+  def get(serviceName: String, portName: String): Var[Option[Var[Address]]] = synchronized {
+    // we call this unstable because every change to the Address will cause
+    // the entire Var[Option[Address]] to update.
+    val unstable = cache.get(serviceName) match {
```
olix0r: perhaps clearer with a specific type annotation on unstable... just because this is all a little subtle

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -0,0 +1,174 @@
...
+      @volatile var addr: VarUp[Address] = null
+
+      unstable.changes.respond {
+        case Some(address) if addr == null =>
+          // Address created
+          addr = Var(address)
+          update() = Some(addr)
+        case Some(address) =>
+          // Address modified
+          addr() = address
```
olix0r: tioli:
```scala
case  Some(address) =>
  addr match {
    case null =>
      addr = Var(address)
      update() = Some(addr)
    case addr =>
      addr() = address
  }
```

i could go either way on this, just trying to be clear about the handling of null

k8s/src/main/scala/io/buoyant/k8s/Ns.scala
adleong: It's unused.  I think it's here so that one day we can release observations.

PR #733, opened 2016-10-11
Modify RetryFilter to use a counter for retries stat

klingerf:
## Problem

Finagle's [RetryFilter](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/service/RetryFilter.scala) uses a histogram to track the number of retries per request.  That's not immediately comparable to other stats -- such as requests, failures, and success -- that are tracked with a counter.

## Solution

Copy Finagle's RetryFilter into router-core, and update it to use a counter instead of a stat.

With this change in place, I now see stats in the following format, for example:

```
  "rt/incoming/dst/path/http/1.1/GET/books/pending": 0,
  "rt/incoming/dst/path/http/1.1/GET/books/requests": 581,
  "rt/incoming/dst/path/http/1.1/GET/books/retries": 74,
  "rt/incoming/dst/path/http/1.1/GET/books/retries/budget_exhausted": 0,
  "rt/incoming/dst/path/http/1.1/GET/books/success": 581,
```

### DIFF
--- a/router/core/src/main/scala/com/twitter/finagle/buoyant/RetryFilter.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/buoyant/RetryFilter.scala
@@ -1,0 +1,82 @@
+package com.twitter.finagle.buoyant
+
+import com.twitter.conversions.time._
+import com.twitter.finagle.service.{RetryBudget, RetryPolicy}
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.tracing.Trace
+import com.twitter.finagle.{Filter, Service}
+import com.twitter.util._
+
+/**
+ * Cribbed from:
+ * https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/service/RetryFilter.scala
+ *
+ * Copied here and modified in order to change the type of stats exported by the filter.
+ */
+class RetryFilter[Req, Rep](
+  retryPolicy: RetryPolicy[(Req, Try[Rep])],
+  timer: Timer,
+  statsReceiver: StatsReceiver,
+  retryBudget: RetryBudget
+)
+  extends Filter[Req, Rep, Req, Rep] {
+
+  def this(
+    retryPolicy: RetryPolicy[(Req, Try[Rep])],
+    timer: Timer,
+    statsReceiver: StatsReceiver
+  ) = this(
+    retryPolicy,
+    timer,
+    statsReceiver,
+    RetryBudget()
+  )
+
+  private[this] val retriesStat = statsReceiver.counter("retries")
+
+  private[this] val budgetExhausted =
+    statsReceiver.scope("retries").counter("budget_exhausted")
+
+  @inline
+  private[this] def schedule(d: Duration)(f: => Future[Rep]) = {
+    if (d > 0.seconds) {
+      val promise = new Promise[Rep]
+      timer.schedule(Time.now + d) {
+        promise.become(f)
+      }
+      promise
+    } else f
+  }
+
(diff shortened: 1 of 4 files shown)


router/core/src/main/scala/com/twitter/finagle/buoyant/RetryFilter.scala
```diff
@@ -0,0 +1,82 @@
...
+              Trace.record("finagle.retry")
+              dispatch(req, service, nextPolicy, count + 1)
+            }
+          } else {
+            budgetExhausted.incr()
+            retriesStat.incr(count)
+            svcRep
+          }
+        case None =>
+          retriesStat.incr(count)
```
olix0r: I think we should simplify this now that we're not actually tracking a Stat -- can we just do a retries.incr() when an actual retry is attempted (i.e in the schedule block above)?  This will make it so that the counter is updated instantaneously, not only after retries have been exhausted.

Also we should probably rename retriesStat to retriesCounter or something...

adleong:
(discussed in-person)
I'm hesitant to remove this histogram because it seems useful to know how retries are distributed across requests.  I suggested that perhaps we could expose the current retry budget's balance as a more direct indication of how much budget is currently used.

klingerf:
Looks like retry budgets are per router, so tracking the budgets as a gauge doesn't give us the number of retries per path.  I've reintroduced the original retries histogram, and added a separate counter for tracking total retries per path.

router/core/src/main/scala/io/buoyant/router/ClassifiedRetries.scala
```diff
@@ -1,7 +1,8 @@
 package io.buoyant.router
 
 import com.twitter.finagle.{Stack, Stackable, ServiceFactory, param}
-import com.twitter.finagle.service._
+import com.twitter.finagle.buoyant.RetryFilter
```
adleong: I think this can be reverted

klingerf: Hmm, if I revert this then nothing will be using the RetryFilter class that I'm adding in com.twitter.finagle.buoyant.

router/core/src/main/scala/com/twitter/finagle/buoyant/RetryFilter.scala
```diff
@@ -0,0 +1,85 @@
...
+    statsReceiver,
+    RetryBudget()
+  )
+
+  private[this] val retriesStat = statsReceiver.stat("retries")
+
+  private[this] val totalRetries = statsReceiver.scope("retries").counter("total")
+
+  private[this] val budgetExhausted =
+    statsReceiver.scope("retries").counter("budget_exhausted")
```
olix0r: It feels odd to have these scoped like this so that we have
```
foo/retries.count
foo/retries/total
```

Perhaps scope the stat as `statsReceiver.scope("retries").stat("per_request")` so that they end up scoped similarly (also this makes it clearer what the stat is)

klingerf: I'm a bit torn on this change. As currently written, our modifications to the stats exported by this filter are additive -- somebody using our filter instead of Finagle's would get all of the original Finagle stats plus one more.  If we rescope the retries stat though, that won't be the case.  Happy to make the change, just wanted to mention why it's written as it is.

olix0r: That's a good point!

However, we already change scoping of finagle stats substantially, so I think it's better for our stuff to be internally consistent than finagle-compatible.

PR #742, opened 2016-10-13
Add rinet namer

adleong:
An alternate form of the inet namer that expects port before host.

`/$/io.buoyant.http.rinet/<port>/<host>/rest*` is rewritten to `/$/inet/<host>/<port>/rest*`

This is useful for dtabs that address services on a fixed port e.g.

```
/host => /$/io.buoyant.http.rinet/80;
/http/1.1/* => /host;
```

### DIFF
--- a/router/core/src/main/scala/io/buoyant/rinet.scala
+++ b/router/core/src/main/scala/io/buoyant/rinet.scala
@@ -1,0 +1,22 @@
+package io.buoyant
+
+import com.twitter.finagle._
+import com.twitter.util.Activity
+
+/**
+ * An alternate form of the inet namer that expects port before host.
+ *
+ *   /$/io.buoyant.rinet/<port>/<host>
+ *
+ * is equivalent to
+ *
+ *   /$/inet/<host>/<port>
+ */
+class rinet extends Namer {
+  override def lookup(path: Path): Activity[NameTree[Name]] = path.take(2) match {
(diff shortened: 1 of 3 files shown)


olix0r:
What does this have to do with http?

adleong:
Nothing, I just instinctively dumped it here.  I'll move it to router-core.

router/core/src/main/scala/io/buoyant/rinet.scala
```diff
@@ -0,0 +1,22 @@
...
+ *   /$/inet/<host>/<port>
+ */
+class rinet extends Namer {
+  override def lookup(path: Path): Activity[NameTree[Name]] = path.take(2) match {
+    case Path.Utf8(port, host) =>
+      Activity.value(NameTree.Leaf(Resolver.eval(s"inet!$host:$port")))
+    case _ =>
+      Activity.value(NameTree.Neg)
+  }
+}
```
olix0r: Alternatively, consider just making this a rewriting namer that rewrites to `Path.Utf8("$", "inet", host, port)` -- the idea being that if the implementation of /$/inet changes, we'll get it for free.

adleong: This would require moving `RewritingNamer` up into router-core.  Not a big deal but I think we can do this later.

PR #744, opened 2016-10-13
Add NameTree dedupping

### LINKED ISSUE #712: linkerd should fall back on Addr.Neg
If an Addr becomes Addr.Neg, routing should fall back to any available alternates in the NameTree.

adleong:
Fixes #712 
# Problem

When a namer's address set becomes empty, it is desirable to set the NameTree to Neg so that fallback can occur.  Unfortunately, this requires mapping the NameTree Activity off of the Addr Var which causes NameTree updates on every Address update.
# Solution

Add dedupping after the call to `Namer.bind` in the name interpreters so that NameTree updates that don't change the NameTree don't propagate.
# Result

Namers can safely map NameTree Activities off of the Addr Var without worrying about spurious updates to the NameTree.  
# Also
- Updated the consul namer to set the NameTree to Neg when the Address set is empty.
- Took the opportunity to move RichActivity into `io.buoyant.namer`

Most of the changes are a result of that move.  Interesting changes relating to the dedupping can be found in:
- `namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala`
- `namer/core/src/main/scala/io/buoyant/namer/ConfiguredDtabNamer.scala`
- `namer/core/src/main/scala/io/buoyant/namer/DefaultInterpreterInitializer.scala`


olix0r:
I'm curious if, for instance, we can prepare an analogous change in Finagle.  If Twitter can exercise this in production, it would be a pretty strong signal.

rnorth:
It seems that this issue affects the filesystem namer as well - e.g. if a name file is empty, we're seeing the same behaviour as for Consul (mentioned below). It seems counterintuitive. Does the FS namer need to be adjusted as well?

Note that if the name file is _absent_ it works as we'd expect.

cc @jakubka

From [Slack](https://linkerd.slack.com/archives/linkerd/p1488216271919164):
> does the alternate (`|`) routing take Consul healthchecks into account? We’re trying to do something similar to achieve failover to another region if the local region’s instances are down.
> Our linkerd config looks a bit like this:
```namers:
- kind: io.l5d.consul
  host: consul
  useHealthCheck: true

routers:
- protocol: http
  dtab: |
      /svc      =>  /#/io.l5d.consul/.local | /$/inet/other-dc-ingress-linkerd/4140 ;
```
>The Consul healthchecks seem to be doing the right thing for selecting a healthy node within the same DC. We were expecting that the alternate routing dtab above would direct requests to the other DC ingress point if all of the local instances failed their healthchecks. However, actually we’re getting an error back (`No hosts are available`).
>FWIW everything seems to work OK when no instances of the service are registered with Consul - it’s just when there are registered but unhealthy instances that things start to behave unexpectedly.
>Are we doing something wrong, or is there a better way to do what we’re trying to do here?

adleong:
This has been sitting idle for a long time, but I really would like to get it in.  Let me see if I can come up with a testing plan for this branch.

adleong:
Huge thanks to @Ashald for testing this!  As far as I can tell, there are no performance regressions here and I'd like to move forward with this change.  cc/ @olix0r

Ashald:
Yeah, we didn't notice any changes in behavior/resource utilization over ~24 hours running this code and the upstream in parallel. Happy to help! ;)

Issue #731, opened 2016-10-11
h2 stats equivalence


olix0r:
the h2 codec has has a hodgepodge of stats, some of which aren't even instrumented properly.

Make sure we're monitoring the right things and that, at the very least, the admin dashboard works with h2 routers.

olix0r:
stream stats added in https://github.com/BuoyantIO/linkerd/pull/784

adleong:
Done as of the 1.1.3 release

PR #746, opened 2016-10-14
Introduce h2 protocol support

olix0r:
The `h2` protocol provides HTTP/2 routing for linkerd. This protocol is
currently marked as _experimental_, requiring that users opt-in for this
not-yet-production-ready feature.

The h2 protocol may be used with or without TLS.  Plaintext h2 servers support
both HTTP Upgrade and H2 Prior Knowledge requests.  Plaintext h2 clients only
support Prior Knowledge requests.


router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4ClientStreamTransport.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  @volatile private[this] var isRequestFinished, isResponseFinished = false
+  def isClosed = isRequestFinished && isResponseFinished
+
+  private[this] val closeP = new Promise[Unit]
+  def onClose: Future[Unit] = closeP
+
+  private[this] def setRequestFinished(): Unit = {
+    isRequestFinished = true
+    if (isClosed) { closeP.setDone(); () }
```
adleong: semi-colons???

adleong: is this to avoid valid-discard warning?  elsewhere we do
```
val _ = closeP.setDone()
```

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * These types are only intended to satisfy session layer concerned
+ * and are not intended to provide a rish programming interface.
+ * However, it is hoped that other types may be built upon / adapated
+ * to satisfy a broader set of use cases.
+ */
+sealed trait Message {
+  def headers: Headers
+  def data: Stream
+  def dup(): Message
```
olix0r: 'copy' conflicts with case classes

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.twitter.io.Buf
+import com.twitter.util.Future
+
+object H2Transport {
+
+  /**
+   * A codec-agnostic interface supporting writes of H2 messages to a transport.
+   */
+  trait Writer {
+    def writeAll(id: Int, msg: Message): Future[Future[Unit]]
```
adleong: or even better: a case class with 2 named `Future[Unit]` properties

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4ClientStreamTransport.scala
olix0r: yeah i ended up doing `; ()` in this branch because it ends up being a lot less visually heavy -- I can keep one-liners as one liners (whereas having `val _ = ..` feels too cumbersome for a one-liner) and it also throws off vertical alignment in a bunch of cases, e.g.:
```scala
foo.bar()
foo.bah()
val _ = foo.baz()
```
vs
```scala
foo.bar()
foo.bah()
foo.baz(); ()
```

I think the semicolon sends the same sort of _i'm doing something intentional here_ warning.  Not super attached to this, but I do feel it made some parts of this read quite a bit nicer.

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Status.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  val Ok = Status(200)
+  val Created = Status(201)
+  val Accepted = Status(202)
+  val NonAuthoritativeInformation = Status(203)
+  val NoContent = Status(204)
+  val ResetContent = Status(205)
+  val PartialContent = Status(206)
+  val MultiStatus = Status(207)
+  val Cowabunga = Status(222)
```
olix0r: whimsy

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/HeaderPathIdentifier.scala
```diff
@@ -0,0 +1,108 @@
...
+        throw new IllegalArgumentException("segments must be >= 1")
+
+      case Some(segments) =>
+        lazy val tooShort = {
+          val msg = s"Path must have at least $segments segment(s)"
+          Future.value(new UnidentifiedRequest[Request](msg))
+        }
+        (req: Request) => reqPath(req) match {
+          case p if p.size < segments => tooShort
+          case p => identified(p, req)
```
olix0r: yep

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/HeaderPathIdentifier.scala
```diff
@@ -0,0 +1,108 @@
...
+) extends Identifier[Request] {
+  import HeaderPathIdentifier._
+
+  private[this] val identifyPath: Request => Future[RequestIdentification[Request]] =
+    segments match {
+      case Some(segments) if segments < 1 =>
+        throw new IllegalArgumentException("segments must be >= 1")
+
+      case Some(segments) =>
+        lazy val tooShort = {
```
olix0r: it uses `segments`:
```scala
      case Some(segments) =>
        lazy val tooShort = {
          val msg = s"Path must have at least $segments segment(s)"
          Future.value(new UnidentifiedRequest[Request](msg))
        }
```

adleong: oh.  variable shadowing rears its ugly head :)

segments doesn't depend on the request at all though, so I feel like this could be rejiggered.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/HeaderPathIdentifier.scala
```diff
@@ -0,0 +1,108 @@
...
+        (req: Request) => reqPath(req) match {
+          case p if p.size < segments => tooShort
+          case p => identified(p, req)
+        }
+
+      case None =>
+        (req: Request) => identified(reqPath(req), req)
+    }
+
+  private[this] def reqPath(req: Request): Path =
```
olix0r: Yeah, sure. I'm skeptical about that feature in general. But if we want to add it, it would go here.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/HeaderPathIdentifier.scala
```diff
@@ -30,11 +30,18 @@ class HeaderPathIdentifier(
...
         (req: Request) => reqPath(req) match {
           case p if p.size < segments => tooShort
-          case p => identified(p, req)
+          case p => identified(p.take(segments), req)
         }
 
       case None =>
-        (req: Request) => identified(reqPath(req), req)
+        lazy val noPath = {
+          val msg = s"Missing destination path"
```
adleong: doesn't technically need interpolation

dpetersen:
So I tried this out today (thanks to @klingerf building me a Docker image with this code). I'm running k8s, I have a Rails app speaking gRPC to a golang backend. There's a linkerd sidecar in each pod, with the Rails-to-golang communication happening linker-to-linker. None of this is using TLS right now. Everything looks good and works as expected (save for missing stats in the admin, but I see #731). Metadata is coming through just fine.

I was unable to test whether linkerd respects the h2 authority header, since I can't figure out how to wrangle the ruby gRPC client into setting that header for me. Hopefully I'll figure it out at some point.

I did notice that the `httpAccessLog` setting isn't available for the `h2` router, which I didn't expect. It's pretty handy to point that to `/dev/stdout` when you're first setting things up.

Thanks for the work on this, it's looking great!

PR #753, opened 2016-10-17
Factor AsyncQueueStreamer from Netty4Stream.

olix0r:
Netty4Stream implements a generic pattern for implementing a stream on top of
an AsyncQueue of input frames.  We factor out the common pieces of queuing and
accumulation in an input-agnostic way so that other stream implementations are
not required to implement the same state management.

Furthermore, a default Stream implementation may now be constructed with
`Stream.apply()`.


olix0r:
Merged https://github.com/BuoyantIO/linkerd/commit/b0f422286c431d2d00731ebe97ade3a617c9bc53

PR #755, opened 2016-10-17
[namerd:747] Namerd does not recognize namer io.l5d.zkLeader

JustinVenus:
Namerd wasn't requiring Namer.zkLeader and as a result one cannot use zkLeader as a valid namer w/ namerd.


JustinVenus:
Fixes https://github.com/BuoyantIO/linkerd/issues/747

PR #757, opened 2016-10-18
announce all routable addresses when config uses meta-address

JustinVenus:
This change should resolve https://github.com/BuoyantIO/linkerd/issues/748.

When an announcer is configured to server the `meta-address=0.0.0.0` the following should occur
```sh
I 1018 21:14:50.537 THREAD1: serving http on /0.0.0.0:4140
I 1018 21:14:50.556 THREAD1: announcing /10.111.254.195:4140 as /#/io.l5d.serversets/prod/linkerd to zk-serversets
I 1018 21:14:50.670 THREAD1: announcing /10.111.255.50:4140 as /#/io.l5d.serversets/prod/linkerd to zk-serversets
I 1018 21:14:50.673 THREAD1: initialized
I 1018 21:14:50.723 THREAD22: Set group member ID to member_0000000014
I 1018 21:14:50.747 THREAD22: Set group member ID to member_0000000015
``` 

When an announcer is configured to serve a `concrete-address=10.111.254.195` the following should occur
```sh
I 1018 21:15:12.897 THREAD1: serving http admin on /0.0.0.0:9990
I 1018 21:15:12.907 THREAD1: serving http on /10.111.254.195:4140
I 1018 21:15:12.918 THREAD1: announcing /10.111.254.195:4140 as /#/io.l5d.serversets/prod/linkerd to zk-serversets
I 1018 21:15:13.024 THREAD1: initialized
I 1018 21:15:13.087 THREAD22: Set group member ID to member_0000000016
```

### DIFF
--- a/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
@@ -2,10 +2,12 @@ package io.buoyant.linkerd
 
 import com.twitter.finagle.Path
 import com.twitter.util._
+import collection.JavaConversions.enumerationAsScalaIterator
 import io.buoyant.admin.App
 import io.buoyant.linkerd.admin.LinkerdAdmin
 import io.buoyant.telemetry.CommonMetricsTelemeter
 import java.io.File
+import java.net.{InetSocketAddress, NetworkInterface}
 import scala.io.Source
 import sun.misc.{Signal, SignalHandler}
 
@@ -95,17 +97,33 @@ object Main extends App {
     server: Server.Initializer,
     name: Path
   ): Closable = {
+    val addrs = if (server.ip.getHostAddress == "0.0.0.0") {
+      val a = for {
+        interface <- NetworkInterface.getNetworkInterfaces
+        if interface.isUp
+        inet <- interface.getInetAddresses
+        if !inet.isLoopbackAddress
+      } yield new InetSocketAddress(inet.getHostAddress, server.port)
+      a.toSeq
+    } else {
+      Seq(server.addr)
+    }
+
     announcers0.filter { case (pfx, _) => name.startsWith(pfx) } match {
       case Nil =>
         log.warning("no announcer found for %s", name.show)
         Closable.nop
 
       case announcers =>
-        val closers = announcers.map {
+        val closers = announcers.flatMap {
           case (prefix, announcer) =>
-            log.info("announcing %s as %s to %s", server.addr, name.show, announcer.scheme)
-            announcer.announce(server.addr, name.drop(prefix.size))
-        }
+            for {
+              addr <- addrs
+            } yield {
+              log.info("announcing %s as %s to %s", addr, name.show, announcer.scheme)
+              announcer.announce(addr, name.drop(prefix.size))
+            }
+          }
         Closable.all(closers: _*)
     }
   }


linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -95,17 +97,36 @@ object Main extends App {
     server: Server.Initializer,
     name: Path
   ): Closable = {
+    def addrs: Seq[InetSocketAddress] = {
```
adleong: just a few style nits:
* addrs can be a val and doesn't need a type annotation
* omit unnessary braces for single expression bodies
* omit empty `()` for non-side-effecting methods
* use == for string equality
* that double for-loop can be simplified into a single

something like this

```
    val addrs = if (server.ip.getHostAddress == "0.0.0.0" ) {
      val addresses = for {
        interface <- NetworkInterface.getNetworkInterfaces
        if interface.isUp
        inet <- interface.getInetAddresses
        if !inet.isLoopbackAddress
      } yield new InetSocketAddress(inet.getHostAddress, server.port)
      addresses.toSeq
    } else {
      Seq(server.addr)
    }
```

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -95,17 +97,36 @@ object Main extends App {
...
+    }
+
     announcers0.filter { case (pfx, _) => name.startsWith(pfx) } match {
       case Nil =>
         log.warning("no announcer found for %s", name.show)
         Closable.nop
 
       case announcers =>
-        val closers = announcers.map {
+        val closers: Seq[Closable] = announcers.map {
```
adleong: don't need the type annotation here

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -95,17 +97,36 @@ object Main extends App {
...
-            log.info("announcing %s as %s to %s", server.addr, name.show, announcer.scheme)
-            announcer.announce(server.addr, name.drop(prefix.size))
-        }
+            for {
+              addr <- addrs
+            } yield {
+              log.info("announcing %s as %s to %s", addr, name.show, announcer.scheme)
+              announcer.announce(addr, name.drop(prefix.size))
+            }
+          }.flatten
```
adleong: you can do `announcers.flatMap { ... }` instead of `announcers.map { ... }.flatten`

JustinVenus:
@adleong Thanks for the feedback ... I didn't know I could flatten multiple sequence comprehensions into one 😄

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
JustinVenus: fixed

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
JustinVenus: fixed

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -95,17 +97,35 @@ object Main extends App {
     server: Server.Initializer,
     name: Path
   ): Closable = {
+    val addrs = {
```
adleong: this is just a single expression so you don't need braces here

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -1,11 +1,13 @@
 package io.buoyant.linkerd
 
+import collection.JavaConversions._
```
adleong: is this still needed?

JustinVenus: yes .. I can look at focusing on just the imports needed as opposed to blinding importing everything.

NetworkInterface.getNetworkInterfaces()  returns an Enumeration<NetworkInterface> and interface.getInetAddresses() returns an Enumeration<InetAddress>.  I get a compiler error w/o this import when attempting to loop over either Enumeration.

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
JustinVenus: sure enough.

PR #772, opened 2016-10-25
Adding Exponential Jittered Backoff to Marathon Api Polling

### LINKED ISSUE #171: Add jitter to marathon api polling
As mentioned in https://github.com/BuoyantIO/linkerd/pull/169 so that we don't send a bunch of requests simultaneously.

satybald:
Not sure if I choose constants(start and end duration) correctly. 

Fixes #171


satybald:
@olix0r thanks for the warm welcome. I don't quite get your first question ;( Can you please rephrase a bit?

I think request can fail due to validation errors or invalid marathon configuration. In the case of validation errors, it will get some 4xx error. But in the case of invalid configuration, it gets timeouts exception. Am I on the right track?

satybald:
@olix0r I guess you're referring to "request storm", when clients simultenisly start polling marathon API. Added possible implementation for that.

adleong:
Thanks for working on this!  A couple of comments:

* AppIdNamer accepts the ttl by-value (not by-name) which means that your `getTtl` and `getRandomJitterMs` will be called once and that fixed value will be used each time.  Consider modifying AppIdNamer to accept a `Stream[Duration]` instead of just a `Duration`.
* Rather than using 10% jitter, consider making the jitter amount configurable.  For example, a user may want to specify 3000ms ttl with 50ms jitter.

satybald:
@adleong thank for suggestions. I added a stream based computation of TTL with jiitter as you recommended.

Should I make `getJitteredNextTtl` method a package level to cover it with the tests?

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -70,15 +70,15 @@ class AppIdNamer(
             state() = Activity.Ok(apps)
             Trace.recordBinary("marathon.apps", apps.map(_.show).mkString(","))
             if (!stopped) {
-              pending = Future.sleep(ttl).onSuccess(_ => loop())
+              pending = Future.sleep(ttl.head).onSuccess(_ => loop())
```
adleong: It looks like we're just always using `ttl.head` which will only use the first value of the stream over and over.  I think we probably want to modify `loop` to accept a stream and sleep based on the head of the stream and pass the tail of the stream into the recursive call to `loop`.

namer/marathon/src/main/scala/io/buoyant/namer/marathon/MarathonInitializer.scala
```diff
@@ -49,7 +52,18 @@ case class MarathonConfig(
     case None => 80
   }
   private[this] def getUriPrefix = uriPrefix.getOrElse("")
-  private[this] def getTtl = ttlMs.getOrElse(5000).millis
+
+  @JsonIgnore
+  private[this] def getJitteredNextTtl(ttlMs: Int, jitterMsConf: Int): Stream[Duration] = {
```
adleong: I think you could simplify this to something like `Stream.continually(ttl + (scala.util.Random.nextDouble() * 2 - 1) * jitter)`

satybald: thanks @adleong for review. I'll work on in on the weekends.

satybald:
@adleong I added necessary changes && resolved merge conflicts. Could you please look at a PR?

namer/marathon/src/main/scala/io/buoyant/namer/marathon/MarathonInitializer.scala
```diff
@@ -64,7 +66,17 @@ case class MarathonConfig(
     case None => 80
   }
   private[this] def getUriPrefix = uriPrefix.getOrElse("")
-  private[this] def getTtl = ttlMs.getOrElse(5000).millis
+
+  @JsonIgnore
+  private[this] def getJitteredTtl(ttl: Int, jitterConf: Int): Stream[Duration] = {
```
adleong: feel free to make this package private if you want to be able to test it (`private[marathon]`)

namer/marathon/src/main/scala/io/buoyant/namer/marathon/MarathonInitializer.scala
```diff
@@ -64,7 +66,17 @@ case class MarathonConfig(
...
+    require(ttl > jitterConf, "TTL should be greater than jitter")
+    val jitter = (ttl + (scala.util.Random.nextDouble() * 2 - 1) * jitterConf).toInt
```
adleong: since this is a `val` a value will be computed once and then used for every position in the stream.  I think this should be a `def` so that each position is recomputed.

namer/marathon/src/test/scala/io/buoyant/namer/marathon/MarathonTest.scala
```diff
@@ -38,17 +39,19 @@ class MarathonTest extends FunSuite {
     assert(marathon.uriPrefix.contains("/marathon"))
     assert(marathon._prefix.contains(Path.read("/io.l5d.marathon")))
     assert(marathon.ttlMs.contains(300))
+    assert(marathon.jitterMs.contains(50))
```
adleong: Would probably also be useful to grab a handful (maybe a few hundred?) of items off the stream and test that they're not all the same, that they all lie between 250 and 350, etc.

esbie:
@satybald are you still interested in moving this PR forward?

satybald:
@esbie yes, but I don't have enough time for that ;(

adleong:
Closed due to inactivity.

PR #773, opened 2016-10-26
Add const and replace transformers

### LINKED ISSUE #770: Add io.l5d.replace transformer
Add a transformer called io.l5d.replace that replaces each leaf with a new bound tree.

```
kind: io.l5d.replace
path: /$/inet/google.com/80
```

This transformer replaces each bound name with the bound tree resulting from `/$/inet/google.com/80`.

adleong:
Fixes #770


namer/core/src/main/scala/io/buoyant/namer/ConstTransformer.scala
```diff
@@ -0,0 +1,14 @@
+package io.buoyant.namer
+
+import com.twitter.finagle.Name.Bound
+import com.twitter.finagle.{NameTree, Namer, Path}
+import com.twitter.util.Activity
+
+/** Bind the given path and use that instead of the tree. */
+class ConstTransformer(path: Path) extends NameTreeTransformer {
+
+  private[this] val bound = Namer.global.bind(NameTree.Leaf(path))
```
olix0r: this should be lazy, since name resolution can be side-effecty, do networked things.  That shouldn't start at instantiation.

olix0r: also, re: use of Namer.global -- does this mean that configured namers can't be used?  should the transformer be configured with a `Path => Activity[NameTree[Name]` or something?

adleong: This should have been NameInterpreter.global; I've updated.  We set NameInterpreter.global so this will be able to use configured namers.

PR #779, opened 2016-10-27
Fix typo in HTTP/2 Indentifiers documentation

klingerf:
Looking at the code, the `header` param is required when configuring the `io.l5d.headerToken` h2 identifier, but the documentation had it listed as optional with a default.  Updating the documentation to match.

### DIFF
--- a/linkerd/docs/protocol-h2.md
+++ b/linkerd/docs/protocol-h2.md
@@ -94,8 +94,7 @@ kind | _required_ | The name of an identifier plugin, such as [`io.l5d.headerTok
 kind: `io.l5d.headerToken`.
 
 With this identifier, requests are turned into logical names using the
-value of the named header. By default, the `:authority` pseudo-header
-is used to provide host-based routing.
+value of the named header.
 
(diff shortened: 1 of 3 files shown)


adleong:
By my reading the Header stack param has default value of `Headers.Authority`.  So I think the written description is correct but the table needs to be updated to indicate that the default is `:authority`, not `:path`.

klingerf:
@adleong Hmm, I see that now too.  When trying to figure this out I was looking at this line specifically:

https://github.com/BuoyantIO/linkerd/pull/779/files#diff-6bfd786381a12996131585664726929eR38

I don't see how that takes into account the default value, since the headers are read directly off the request object.  But it's entirely possible that I'm just missing it.

adleong:
I think what that line is saying is that if the specified header is not present on the request, then fail identification.  So if no header name is specified in the config, it will default to :authority and then if no :authority header is present on the request, it will fail identification.

klingerf:
@adleong Ah, yep, you're totally write. Have updated accordingly.

PR #786, opened 2016-10-30
Fix TLS for h2

olix0r:
In 59df8f2, the TLS pipeline was broken so that requests are not served and
exceptions are thrown.

This is fixed by reserving a slot in the pipeline into which the framer may be
installed during ALPN initialization.


adleong:
do we have a test for TLS with h2?

olix0r:
not yet.  i'll try to throw something together

olix0r:
615d52d

PR #787, opened 2016-10-31
Admit HTTP/1 upgrade requests in h2 plaintext servers.

### LINKED ISSUE #765: support http1 upgrade requests
In #764, the ServerUpgraderHandler can no longer be used, and so we lose support for HTTP1 Upgrade messages.

olix0r:
In 59df8f2, ServerUpgradeHandler was disabled to accomodate the new
multiplexing regime, requiring that all plaintext clients send a prior
knowledge request.

This change introduces an adapter, io.netty4...Http2FrameCodecServerUpgrader,
that instantiates an Http2ServerUpgradeCodec with a Http2FrameCodec so that the
server properly accepts HTTP/1 upgrade requests in addition to prior-knowledge
requests.

Fixes #765


olix0r:
It's really tricky to test this without supporting upgrade on the clientside (which we don't).  I've done a manual test for now.

PR #792, opened 2016-11-01
add marathonHealth option to Marathon namer

### LINKED ISSUE #704: Marathon: Do not add / remove nodes when marathon-marked "unhealthy" nodes in namer
As discussed on slack.

When scaling, linkerd routes to "unhealthy" nodes resulting in 502 responses. Assuming there are multiple ways to do this (such as retries) but might be good to add a configuration item to the namer to not route to "unhealthy" instances (and indeed remove them from routing).

/cc @olix0r

siggy:
fixes #704

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -183,7 +183,7 @@ experimental | _required_ | Because this namer is still considered experimental,
 host | `localhost` | The Consul host.
 port | `8500` | The Consul port.
 includeTag | `false` | If `true`, read a Consul tag from the path.
-useHealthCheck | `false` | If `true`, rely on Consul health checks.
+useHealthCheck | `false` | If `true`, exclude app instances that are failing Consul health checks. Even if `false`, linkerd's built-in resiliency algorithms will still apply.
 token | no authentication | The auth token to use when making API calls.
 setHost | `false` | If `true`, HTTP requests resolved by Consul will have their Host header overwritten to `${serviceName}.service.${datacenter}.${domain}`. `$domain` is fetched from Consul.
 consistencyMode | `default` | Select between [Consul API consistency modes](https://www.consul.io/docs/agent/http.html) such as `default`, `stale` and `consistent`.
@@ -327,13 +327,14 @@ kind: `io.l5d.marathon`
 
 ```yaml
 namers:
-- kind:         io.l5d.marathon
-  experimental: true
-  prefix:       /#/io.l5d.marathon
-  host:         marathon.mesos
-  port:         80
-  uriPrefix:    /marathon
-  ttlMs:        5000
+- kind:           io.l5d.marathon
+  experimental:   true
+  prefix:         /io.l5d.marathon
+  host:           marathon.mesos
+  port:           80
+  uriPrefix:      /marathon
+  ttlMs:          5000
+  useHealthCheck: false
 ```
 > Then reference the namer in the dtab to use it:
 
@@ -354,7 +355,8 @@ experimental | _required_ | Because this namer is still considered experimental,
 host | `marathon.mesos` | The Marathon master host.
 port | `80` | The Marathon master port.
 uriPrefix | none | The Marathon API prefix. This prefix depends on your Marathon configuration. For example, running Marathon locally, the API is avaiable at `localhost:8080/v2/`, while the default setup on AWS/DCOS is `$(dcos config show core.dcos_url)/marathon/v2/apps`.
-ttlMs | `5000` | The polling interval in milliseconds against the marathon API.
+ttlMs | `5000` | The polling interval in milliseconds against the Marathon API.
+useHealthCheck | `false` | If `true`, exclude app instances that are failing Marathon health checks. Even if `false`, linkerd's built-in resiliency algorithms will still apply.
 
 ### Marathon Path Parameters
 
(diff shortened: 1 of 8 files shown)


linkerd/docs/namer.md
```diff
@@ -354,6 +354,7 @@ host | `marathon.mesos` | The Marathon master host.
 port | `80` | The Marathon master port.
 uriPrefix | none | The Marathon API prefix. This prefix depends on your Marathon configuration. For example, running Marathon locally, the API is avaiable at `localhost:8080/v2/`, while the default setup on AWS/DCOS is `$(dcos config show core.dcos_url)/marathon/v2/apps`.
 ttlMs | `5000` | The polling interval in milliseconds against the marathon API.
+marathonHealth | `false` | Override linkerd's view of instance health with Marathon health checks. In general, setting this to `false` results in more aggressive request-based health detection using [Finagle's Failure Accrual](https://twitter.github.io/finagle/guide/Clients.html#failure-accrual). Setting this to `true` results in more conservative polling of health check endpoints based on [Marathon's health check configuration](https://mesosphere.github.io/marathon/docs/health-checks.html).
```
adleong: I don't think this description is quite right.  This option will just turn on filtering at the service discovery layer such that we only consider endpoints that marathon considers healthy.  Failure accrual still applies to the remaining endpoints as normal.  This is an important distinction because a service can be alive and responding to health checks, but still return errors when servicing requests.

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -27,8 +27,8 @@ object Api {
 
   case class UnexpectedResponse(rsp: http.Response) extends Throwable
 
-  def apply(client: Client, uriPrefix: String): Api =
-    new AppIdApi(client, s"$uriPrefix/$versionString")
+  def apply(client: Client, uriPrefix: String, marathonHealth: Boolean): Api =
```
adleong: `marathonHealth` makes me think that this variable is an indication of marathon health.  Maybe rename to `checkMarathonHealth` or `useMarathonHealth` or something?

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -79,11 +82,21 @@ object Api {
 
   private[this] case class AppRsp(app: Option[App] = None) {
 
-    def toAddresses: Set[Address] =
+    private[this] def healthy(healthCheckResults: Seq[HealthCheckResult]): Boolean =
+      healthCheckResults.forall(
```
adleong: style nit:
```
healthCheckResults.forall { 
  case HealthCheckResult(Some(alive)) if (alive) => true
  case _ => false
}
```

or even

```
healthCheckResults.forAll(_ == HealthCheckResult(Some(true))
```

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -79,11 +82,21 @@ object Api {
...
+          case _ => false
+        }
+      )
+
+    private[v2] def toAddresses(marathonHealth: Boolean): Set[Address] =
       app match {
         case Some(App(_, Some(tasks))) =>
           tasks.collect {
-            case Task(_, Some(host), Some(Seq(port, _*))) =>
+            case Task(_, Some(host), Some(Seq(port, _*)), _) if (!marathonHealth) =>
```
adleong: This is nice.  Better than my filter suggestion.

Style nit: don't need the parens around the if predicate, I don't think.

linkerd/docs/namer.md
```diff
@@ -354,7 +354,7 @@ host | `marathon.mesos` | The Marathon master host.
...
 ttlMs | `5000` | The polling interval in milliseconds against the marathon API.
-marathonHealth | `false` | Override linkerd's view of instance health with Marathon health checks. In general, setting this to `false` results in more aggressive request-based health detection using [Finagle's Failure Accrual](https://twitter.github.io/finagle/guide/Clients.html#failure-accrual). Setting this to `true` results in more conservative polling of health check endpoints based on [Marathon's health check configuration](https://mesosphere.github.io/marathon/docs/health-checks.html).
+checkMarathonHealth | `false` | Setting this to `false` results in more aggressive request-based health detection using [Finagle's Failure Accrual](https://twitter.github.io/finagle/guide/Clients.html#failure-accrual), ignoring marathon health checks. Setting this to `true` limits linkerd's view of instances to those marathon reports as healthy, based on [Marathon's health check configuration](https://mesosphere.github.io/marathon/docs/health-checks.html). In this case failure accrual still applies to the remaining nodes.
```
adleong: I'd be interested to hear what other people think but I don't think we need to talk about failure accrual here.  F.A. isn't more aggressive than marathon health checking, it's just a different thing.

linkerd/docs/namer.md
```diff
@@ -354,6 +355,7 @@ host | `marathon.mesos` | The Marathon master host.
 port | `80` | The Marathon master port.
 uriPrefix | none | The Marathon API prefix. This prefix depends on your Marathon configuration. For example, running Marathon locally, the API is avaiable at `localhost:8080/v2/`, while the default setup on AWS/DCOS is `$(dcos config show core.dcos_url)/marathon/v2/apps`.
 ttlMs | `5000` | The polling interval in milliseconds against the marathon API.
+useHealthCheck | `false` | If `true`, rely on Marathon health checks. If `false`, rely on linkerd's built-in resiliency algorithms.
```
adleong: What about something like:

If `true`, exclude app instances that are failing Marathon health checks.  Even if `false`, linkerd's built-in resiliency algorithms will still apply.

PR #797, opened 2016-11-02
fix piping configs to dcos bootstrap

siggy:
also simplify default dcos dtab


siggy:
@adleong Yup, I thought it best to make the default dtab as generic as possible. It now matches most of the configs in the linkerd-examples repo.

PR #800, opened 2016-11-02
Add domainToPathPfx & subdomainOfPfx to namer.md

### LINKED ISSUE #795: Add domainToPathPfx & subdomainOfPfx to config.md
We use them a fair amount, but there's no explanation of what they do on the docs site

esbie:
Fixes #795


linkerd/docs/namer.md
```diff
@@ -419,3 +419,54 @@ Key | Required | Description
...
+/$/io.buoyant.http.domainToPathPfx/<prefix>/<host>
+```
+
+Rewrites the path's prefix with `<prefix>` first, followed by each subdomain of
+`<host>` separated and in reverse order. This is most useful when using marathon
+application groups, whose ids are represented as domains.
+
+For example,
+`/$/io.buoyant.http.domainToPathPfx/pfx/foo.buoyant.io/resource/name` would be
+rewritten to `pfx/io/buoyant/foo/resource/name`.
```
olix0r: nit: preserve leading slash: `/pfx/io/buoyant/foo/resource/name`

linkerd/docs/namer.md
```diff
@@ -419,3 +419,54 @@ Key | Required | Description
...
+
+> Dtab Path Format
+
+```yaml
+/$/io.buoyant.http.domainToPathPfx/<prefix>/<host>
+```
+
+Rewrites the path's prefix with `<prefix>` first, followed by each subdomain of
+`<host>` separated and in reverse order. This is most useful when using marathon
+application groups, whose ids are represented as domains.
```
olix0r: Is this second sentence true? I thought marathon used slashes.  Either way, I might omit this sentence.  (I see this as useful in a bunch of cases where some portion of an FQDN is used to route -- hardly marathon-centric).

linkerd/docs/namer.md
```diff
@@ -419,3 +419,54 @@ Key | Required | Description
...
+
+> Dtab Path Format
+
+```yaml
+/$/io.buoyant.http.subdomainOfPfx/<domain>/<prefix>/<host>
+```
+
+Rewrites the path's prefix with `<prefix>` first, followed by `<host>` with the
+`<domain>` dropped. This is most useful when using consul, whose dns interface
+has the service's name as the subdomain.
```
olix0r: Similarly, I'd be inclined to leave out the consul references...

linkerd/docs/namer.md
esbie: oops yeah this isn't right... it's not the ids that are domains, it's marathon's dns interface. Regardless I can take it out

linkerd/docs/namer.md
```diff
@@ -419,3 +419,52 @@ Key | Required | Description
 prefix | yes | Tells linkerd to resolve the request path using the marathon namer.
 zkPath | yes | The ZooKeeper path of a leader group. This path can be multiple path segments long. The namer resolves to the address stored in the data of the leader.
 
+<a name="rewritingNamers"></a>
+## Rewriting Namers
+
+In addition to service discovery namers, linkerd supplies a number of utility
+namers. These namers assist in path rewriting when the transformation is more
+complicated than just prefix substitution.
```
adleong: Maybe note that these namers don't have to be included in the `namers:` list in your config file.  Since they are on the classpath and referenced with the `/$/` syntax, they will be automatically available.

esbie: oooooh yes good idea

PR #801, opened 2016-11-03
Make labelSelector configurable for k8s namers.

### LINKED ISSUE #799: Support filtering by label in k8s namer
It'd be cleaner to deploy multiple versions of a service and shift traffic between labels (instead of shifting between uniquely named services)

esbie:
Fixes #799 

Problem:
Folks using the k8s namer may want to route not only on a per-service basis, but also between services that are tagged with different labels. See http://kubernetes.io/docs/user-guide/labels/#label-selectors for more info on k8s labelSelectors.

Solution:
Adds a `labelSelector` config option, similar to the consul namer's `includeTag`, that if specified will read a label selector's value from the path. `labelSelector` is a string used as the label selector's key. This implementation supports only the most basic equality-based label selector requirement. It does not support multiple requirements, nor set-based requirements. It does not support labels without values.

I see the consul namer uses separate tagged and untagged implementations, but that seemed not ideal in the k8s case.... the k8s namer is more complex. Open to code restructuring feedback.

I had only intended to update the k8s namer, but the daemonset and external namers use the same api, so I added the flag to those as well.


interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformerInitializer.scala
```diff
@@ -34,7 +35,7 @@ case class DaemonSetTransformerConfig(
   override def mk(): NameTreeTransformer = {
     val client = mkClient(Params.empty).configured(Label("daemonsetTransformer"))
     def mkNs(ns: String) = Api(client.newService(dst)).withNamespace(ns)
-    val namer = new EndpointsNamer(Path.empty, mkNs)
+    val namer = new EndpointsNamer(Path.empty, labelSelector, mkNs)
```
adleong: I think setting a labelSelector on this EndpointsNamer is going to cause problems.  The only name we ever bind with this namer is the hardcoded name on the next line `/ns/port/service` which does not include a label value.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -24,11 +25,28 @@ class EndpointsNamer(
...
+    case (id@Path.Utf8(nsName, _, _), None) =>
+      val residual = path.drop(variablePrefixLength)
       log.debug("k8s lookup: %s %s", id.show, path.show)
-      endpointNs.get(nsName).services.flatMap { services =>
+      lookupServices(endpointNs.get(nsName, None), id, residual)
+
+    case (id@Path.Utf8(nsName, _, _, labelValue), Some(label)) =>
+      val residual = path.drop(variablePrefixLength)
+      log.debug("k8s lookup: %s %s %s", id.show, label, path.show)
+      lookupServices(endpointNs.get(nsName, Some(s"${label}=${labelValue}")), id.take(PrefixLen), residual)
```
adleong: nit: don't need braces for simple interpolated variables.  `s"$label=$labelValue"`

adleong: the label should be part of the `id`, right?  so should this just be `id` instead of `id.take`?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -24,11 +25,28 @@ class EndpointsNamer(
...
+
+    case (id@Path.Utf8(nsName, portName, serviceName), Some(label)) =>
+      log.debug("k8s lookup: ns %s service %s label value segment missing for label %s", nsName, serviceName, portName, label)
+      Activity.value(NameTree.Neg)
+
+    case _ =>
+      Activity.value(NameTree.Neg)
+  }
+
+  private[this] def lookupServices(cache: NsCache, id: Path, residual: Path): Activity[NameTree[Name]] = id match {
```
adleong: the id may or may not contain a label so I think you need to do `id.take(3)` here

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -50,11 +68,10 @@ class EndpointsNamer(
             }
         }
       }
-
-    case _ =>
-      Activity.value(NameTree.Neg)
   }
 
+  private[this] val variablePrefixLength = if (labelName.isEmpty) PrefixLen else PrefixLen + 1
```
adleong: TIOTLI golf: `PrefixLen + labelName.size`

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -18,16 +18,29 @@ import java.net.InetSocketAddress
...
   private[this] val PrefixLen = 3
+  private[this] val variablePrefixLength = if (labelName.isEmpty) PrefixLen else PrefixLen + 1
 
-  override def lookup(path: Path): Activity[NameTree[Name]] = path.take(PrefixLen) match {
-    case id@Path.Utf8(nsName, portName, serviceName) =>
+  def lookup(path: Path): Activity[NameTree[Name]] = (path.take(variablePrefixLength), labelName) match {
+    case (id@Path.Utf8(nsName, portName, serviceName), None) =>
+      val residual = path.drop(variablePrefixLength)
+      val nameTree = serviceNs.get(nsName, None).get(serviceName, portName).map {
+        case Some(address) =>
```
adleong: since this is duplicated below you could probably factor it out into a `toNameTree` method

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -18,16 +18,29 @@ import java.net.InetSocketAddress
...
+          val bound = address.map(Addr.Bound(_))
+          NameTree.Leaf(Name.Bound(bound, idPrefix ++ id, path.drop(PrefixLen)))
```
adleong: should `path.drop(PrefixLen)` just be `residual`?

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -18,16 +18,29 @@ import java.net.InetSocketAddress
...
+        case None =>
+          NameTree.Neg
+      }
+      Activity(nameTree.map(Activity.Ok(_)))
 
-      val nameTree = serviceNs.get(nsName).get(serviceName, portName).map {
+    case (id@Path.Utf8(nsName, portName, serviceName, labelValue), Some(label)) =>
+      val residual = path.drop(variablePrefixLength)
+      val nameTree = serviceNs.get(nsName, Some(s"${label}=${labelValue}")).get(serviceName, portName).map {
```
adleong: ditto re interpolation variables

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -18,16 +18,29 @@ import java.net.InetSocketAddress
...
         case Some(address) =>
           val bound = address.map(Addr.Bound(_))
           NameTree.Leaf(Name.Bound(bound, idPrefix ++ id, path.drop(PrefixLen)))
```
adleong: residual?

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformerInitializer.scala
esbie: ah, right. I can back out of that change

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
esbie: It just so happens that every place calling this method has done `id.take(3)` before passing the id in, but I can move the take to be part of this method instead

PR #804, opened 2016-11-04
Add hostNetwork option to k8s transformers

adleong:
# Problem

In CNI environments (such as calico), daemonsets must be run with `hostNetwork: true` in order for `hostPort` to work.  This means that the daemonset pods use the host's network and therefore get the host's IP as their pod IP.  This breaks the assumption that pods on the same host will always be on the same /24 subnet and therefore breaks the service mesh transformers (daemonset and localnode).

# Solution

Have the k8s namer record the nodeName of each endpoint in the Address.Metadata.  We then add a `hostNetwork` option to the daemonset and localnode transformers which changes the behavior of these transformers to use the `nodeName` metadata to determine which pods are on the same nodes instead of comparing the /24 subnets of the pod IPs.  This should be more reliable in general, even when `hostNetwork` is not turned on on the daemonset.  However, we hide this behavior behind an option to be safe.

# Result

Service mesh deployments on k8s clusters using CNI work correctly if `hostNetwork` is enabled on the daemonset and on the daemonset transformer and on the localnode transformer.


adleong:
TODO: The nodeName metadata probably doesn't get propagated from namerd back to linkerd.

olix0r:
namerd change ticketed: https://github.com/BuoyantIO/linkerd/issues/805

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformerInitializer.scala
```diff
@@ -31,11 +32,19 @@ case class DaemonSetTransformerConfig(
 
   private[this] val netmask = InetAddress.getByName("255.255.255.0")
 
+  private[this] val nodeLocal: (Address, Address) => Boolean = {
+    case (Address.Inet(_, a), Address.Inet(_, b)) => a.get("nodeName") == b.get("nodeName")
+    case _ => true
+  }
```
olix0r: seems slightly odd for this to be in the config class...

Should we have a MetadataGatewayTransformer?

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/LocalNodeTransformerInitializer.scala
```diff
@@ -1,23 +1,43 @@
...
+  extends TransformerConfig {
 
   override def mk(): NameTreeTransformer = {
-    val ip = sys.env.getOrElse(
-      "POD_IP",
-      throw new IllegalArgumentException("POD_IP env variable must be set to the pod's IP")
-    )
-    val local = InetAddress.getByName(ip)
-    new SubnetLocalTransformer(local, Netmask("255.255.255.0"))
+    if (hostNetwork.contains(true)) {
```
olix0r: tioli: .getOrElse(false) reads slighly clearer to me

interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/LocalNodeTransformerInitializer.scala
```diff
@@ -1,23 +1,43 @@
...
+      val nodeName = sys.env.getOrElse(
+        "NODE_NAME",
+        throw new IllegalArgumentException(
+          "NODE_NAME env variable must be set to the node's name"
+        )
+      )
+      new FilteringNameTreeTransformer {
+        protected val predicate: Address => Boolean = {
+          case a@Address.Inet(_, meta) =>
+            meta.get("nodeName").contains(Some(nodeName))
```
olix0r: we use this string a lot.  Worth constifying? (so that typos are compile errors)  Something like `NodeNameMetaKey`?

adleong:
CI passed before I updated CHANGES.md so I'm just going to merge this.

PR #809, opened 2016-11-08
Add nodeName to namerd addr metadata

### LINKED ISSUE #805: Pass nodeName Address metadata in namerd thrift
Namer's thrift interface requires all address metadata fields to be explicitly enumerated.  Add the nodeName field to the thrift definition and plumb it through.

adleong:
Fixes #805 

This allows namerd to turn on the `hostNetwork` option for the localnode and daemonset transformers.

In this change we also switch the type of the nodeName metadata field from `Option[String]` to `String]`.

### DIFF
--- a/namerd/iface/interpreter-thrift-idl/src/main/thrift/namer.thrift
+++ b/namerd/iface/interpreter-thrift-idl/src/main/thrift/namer.thrift
@@ -82,6 +82,7 @@ struct AddrReq {
 
 struct AddrMeta {
   1: optional string authority // HTTP/1.1 Host or HTTP/2.0 :authority
+  2: optional string nodeName
 }
 
 struct TransportAddress {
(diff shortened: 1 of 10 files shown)


namerd/iface/interpreter-thrift-idl/src/main/thrift/namer.thrift
```diff
@@ -82,6 +82,7 @@ struct AddrReq {
 
 struct AddrMeta {
   1: optional string authority // HTTP/1.1 Host or HTTP/2.0 :authority
+  2: optional string nodeName
```
olix0r: Worth a comment about what the field refers to.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerInterface.scala
```diff
@@ -410,10 +410,17 @@ class ThriftNamerInterface(
...
+    val authority = scalaMeta.get(Metadata.authority).map(_.toString)
+    val nodeName = scalaMeta.get(Metadata.nodeName).map(_.toString)
+
+    if (authority.isDefined || nodeName.isDefined)
+      Some(thrift.AddrMeta(
+        authority = authority,
+        nodeName = nodeName
+      ))
+    else
+      None
```
olix0r: tioli (i like pattern matching ;p):

```scala
(authority, nodeName) match {
  case (None, None) => None
  case (authority, nodeName) =>
    Some(...)
}
```

PR #814, opened 2016-11-16
Identifiers must return a Dst.Path

### LINKED ISSUE #724: Identifiers should not be able to short-circuit the path stack
Currently, Identifiers return a Dst, which means they are responsible for returning Dtabs or may even bypass binding entirely  This complicates the client stack, which assumes (i.e. for retries, tracing etc), that requests traverse each stack in order.

We should change the Identifier interface to return a narrower IdentifiedRequest that includes only a destination Path and a Request.

adleong:
Prevent Identifiers from returning a Dst.Bound and skipping the path stack.

Fixes #724


adleong:
This branch is back from the dead, rebased on master, and hungry for reviews/revenge!

PR #815, opened 2016-11-16
Add static identifier

adleong:
Add a `io.l5d.static` identifier which always assigns the same name to all requests.

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/StaticIdentifier.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/StaticIdentifier.scala
@@ -1,0 +1,20 @@
+package io.buoyant.router.http
+
+import com.twitter.finagle.buoyant.Dst
+import com.twitter.finagle.{Dtab, Path}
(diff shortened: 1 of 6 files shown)


router/http/src/main/scala/io/buoyant/router/http/StaticIdentifier.scala
```diff
@@ -0,0 +1,20 @@
...
+import io.buoyant.router.RoutingFactory.{IdentifiedRequest, RequestIdentification}
+
+class StaticIdentifier(
+  prefix: Path,
+  path: Path,
+  baseDtab: () => Dtab = () => Dtab.base
+) extends RoutingFactory.Identifier[Request] {
+
+  def apply(req: Request): Future[RequestIdentification[Request]] = {
+    val dst = Dst.Path(prefix ++ path, baseDtab(), Dtab.local)
```
olix0r: consider caching `prefix ++ path` on the class (or better yet, just take a path in the constructor) so we don't have to do additional per-request work

PR #821, opened 2016-11-22
Configure jackson so that only public fields are used for config classes

olix0r:
Configuration de/serializers sometimes get confused by non-public members.
This change removes the need for per-class visibility configuration.


namer/marathon/src/main/scala/io/buoyant/namer/marathon/MarathonInitializer.scala
```diff
@@ -50,7 +54,7 @@ case class MarathonConfig(
   useHealthCheck: Option[Boolean]
 ) extends NamerConfig {
 
-  private[this] val log = Logger.get(getClass.getName)
+  import MarathonConfig.log
```
adleong: Less invasive would be to just mark as `@JsonIgnore`

PR #822, opened 2016-11-22
Add `io.l5d.namerd.http` interpreter

adleong:
Add an interpreter that uses namerd's HTTP streaming API.  

This is the first time that we've needed to parse JsonDelegateTrees which required switching from using Manifests to using TypeReferences.  Unfortunately, this is less ergonomic than Manifests or TypeTags since scala won't generate the implicits for us.


admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -126,6 +137,44 @@ object DelegateApiHandler {
...
+      case Leaf(p, d, bound) =>
+        DelegateTree.Leaf(p, d.getOrElse(Dentry.nop), parseBound(bound))
+      case Alt(p, d, alts) => DelegateTree.Alt(p, d.getOrElse(Dentry.nop), alts.map(toDelegateTree): _*)
+      case Union(p, d, weighteds) =>
+        val delegateTreeWeighteds = weighteds.map {
+          case Weighted(w, tree) =>
+            DelegateTree.Weighted(w, toDelegateTree(tree))
+        }
+        DelegateTree.Union(p, d.getOrElse(Dentry.nop), delegateTreeWeighteds: _*)
+      case Transformation(p, name, bound, tree) => DelegateTree.Transformation(p, name, parseBound(bound), toDelegateTree(tree))
```
olix0r: nit: would you mind breaking some of these long lines on `=>` so we keep things generally <100 chars?

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdHttpInterpreterInitializer.scala
```diff
@@ -0,0 +1,100 @@
...
+              HighResTimer.Default,
+              stats,
+              RetryBudget.Infinite
+            )
+            retry.andThen(next)
+          }
+        }
+    }
+
+    val tlsTransformer = new Transformer {
```
olix0r: consistency/clarity: use `Stack.Transformer` here, and don't import `Transformer`

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdHttpInterpreterInitializer.scala
```diff
@@ -0,0 +1,100 @@
...
+
+case class NamerdHttpInterpreterConfig(
+  dst: Option[Path],
+  namespace: Option[String],
+  retry: Option[Retry],
+  tls: Option[ClientTlsConfig]
+) extends InterpreterConfig {
+
+  @JsonIgnore
+  private[this] val log = Logger.get("namerd-http")
```
olix0r: If we're going to special-case the name, should we just move this into `newInterpreter` and use _label_?  Otherwise, getClass.getName

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdHttpInterpreterInitializer.scala
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Construct a namer.
+   */
+  @JsonIgnore
+  def newInterpreter(params: Stack.Params): NameInterpreter = {
+    val name = dst match {
+      case None => throw new IllegalArgumentException("`dst` is a required field")
+      case Some(dst) => Name.Path(dst)
+    }
+    val label = s"namer/${NamerdHttpInterpreterInitializer.configId}"
```
olix0r: this isn't a namer, should that be `interpreter/`?

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdHttpInterpreterInitializer.scala
```diff
@@ -0,0 +1,100 @@
...
+  retry: Option[Retry],
+  tls: Option[ClientTlsConfig]
+) extends InterpreterConfig {
+
+  @JsonIgnore
+  private[this] val log = Logger.get("namerd-http")
+
+  @JsonIgnore
+  val defaultRetry = Retry(5, 10.minutes.inSeconds)
+
```
olix0r: I'd imagine we want this interpreter marked as experimental?

olix0r:
Testing wise -- at the _least_ I think we need an EndToEnd test for this.

Issue #785, opened 2016-10-30
h2 client does not reconnect


olix0r:
Repro:

Run linkerd with a config like:

``` yaml
- protocol: h2
  experimental: true
  baseDtab: |
    /h2/localhost:4142 => /$/inet/127.1/8888;
    /h2 => /srv;
  servers:
  - port: 4142
  identifier:
    kind: io.l5d.headerToken
    header: ":authority"
```

Run nghttpd

```
:; nghttpd --no-tls -v 8888                                                                                                              
IPv4: listen 0.0.0.0:8888
IPv6: listen :::8888
```

Perform request:

```
:; curl --http2 http://localhost:4142/
```

Kill nghttpd

Curl again (should fail)

Restart nghttpd

Curl again.  Should succeed, but fails with:

```
l5d-err: com.twitter.finagle.DroppedWriteException
```

zackangelo:
The error appears to change when building against Finagle 6.40.0: 

```E 1126 20:38:00.044 THREAD680: service failure
com.twitter.finagle.ChannelClosedException: null at remote address: /127.0.0.1:9897 from service: $/inet/127.0.0.1/9897. Remote Info: Upstream Address: Not Available, Upstream Client Id: Not Available, Downstream Address: /127.0.0.1:9897, Downstream Client Id: $/inet/127.0.0.1/9897, Trace Id: fde0d31c44d52c0e.fde0d31c44d52c0e<:fde0d31c44d52c0e
Caused by: java.nio.channels.ClosedChannelException
	at io.netty.channel.AbstractChannel$AbstractUnsafe.write(...)(Unknown Source)```

olix0r:
this was fixed by #826 and released in 0.8.4

PR #826, opened 2016-11-27
[h2] Support stream cancelation

### LINKED ISSUE #728: stream cancellation does not propagate properly
When, for instance, a client of linkerd is streaming a response body and the client terminates the stream prematurely, linkerd continues to process the stream from the server instead of issuing a RST_STREAM.  We should propagate stream cancellation.

olix0r:
### Background

The HTTP2 protocol defines RST_STREAM frames to support request-level cancelation and GO_AWAY frames to signal connection-level interruption. This distinction is especially germane to proxies like linkerd since we want to allow long-lived multiplexed connections between endpoints while also allowing request-level cancelation (to support timeouts/deadlines, etc).

### Problem

linkerd's HTTP2 support isn't at all well-conditioned to request cancelation or connection loss (#728). For instance, if a linkerd instance loses its connection to a remote server, the client-side of an active stream may close, but the upstream-server-side of a stream will be left open indefinitely. Similar behavior is exhibited when a remote explicitly issues a RST_STREAM frame.

Furthermore, clients that disconnect are not properly evicted from load balancers because the ClientDispatcher Service does not alter its `status` after connection loss. (#785)

### Solution

Because resets may occur at any stage of a Stream's lifecycle, several changes are required. As much as possible, we try to rely on finagle primitives like Future-cancelation to achieve the desired behavior.

The `h2.Stream` API has been altered substantially: the `Stream.Reader` and `Stream.Nil` types have been unified as a `Stream`. Stream consumers that care about resets & closures must call `Stream.read()` until a failure is encountered. Stream producers (i.e. `Netty4StreamTransport`) must close or reset the stream appropriately.

`Netty4StreamTransport` has been substantially refactored to more explicitly model the HTTP2 state machine.  It ensures that canceled and failed writes are propagated as resets, and that resets cancel writes and prevent receipt of remote frames.

`Netty4ClientDispatcher` and `Netty4ServerDispatcher` have been substantially simplified, moving a good portion of common stream-multiplexing logic to `Netty4DispatcherBase`. Dispatchers now ensure that a GO_AWAY frame or connection loss actively triggers stream cancelation.

Various dedicated `Reset`, `GoAway`, and `StreamError` types have been introduced to model failures appropriately.

Fixes #728 #785

### DIFF
--- a/router/h2/src/e2e/scala/io/buoyant/router/h2/ConcurrentStreamsEndToEndTest.scala
+++ b/router/h2/src/e2e/scala/io/buoyant/router/h2/ConcurrentStreamsEndToEndTest.scala
@@ -17,7 +17,7 @@ class ConcurrentStreamsEndToEndTest
 
   val FrameSize = 16 * 1024
   val WindowSize = 4 * FrameSize
-  val lengths = Seq(/*1,*/ WindowSize * 2 - WindowSize / 2)
+  val lengths = Seq(1, WindowSize * 2 - WindowSize / 2)
 
   case class Spec(len: Long, frameSize: Int, concurrency: Int)
   val specs = lengths.flatMap { len =>
--- a/router/h2/src/e2e/scala/io/buoyant/router/h2/ConcurrentStreamsEndToEndTest.scala
+++ b/router/h2/src/e2e/scala/io/buoyant/router/h2/ConcurrentStreamsEndToEndTest.scala
@@ -6,6 +6,7 @@ import com.twitter.io.Buf
 import com.twitter.logging.Level
 import com.twitter.util._
 import io.buoyant.test.FunSuite
+import scala.annotation.tailrec
 
 class ConcurrentStreamsEndToEndTest
   extends FunSuite
@@ -16,7 +17,7 @@ class ConcurrentStreamsEndToEndTest
 
   val FrameSize = 16 * 1024
   val WindowSize = 4 * FrameSize
-  val lengths = Seq(1, WindowSize * 2 - WindowSize / 2)
+  val lengths = Seq(/*1,*/ WindowSize * 2 - WindowSize / 2)
 
   case class Spec(len: Long, frameSize: Int, concurrency: Int)
   val specs = lengths.flatMap { len =>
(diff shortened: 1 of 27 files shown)


zackangelo:
Seeing these whenever a linkerd process is SIGINT'd: 

```I 1128 17:25:07.119 THREAD1687: Received SIGINT. Shutting down ...
E 1128 17:25:07.815 THREAD1687: [S L:/127.0.0.1:4144 R:/127.0.0.1:64426] dispatcher failed
Failure(com.twitter.finagle.buoyant.h2.Reset$Cancel$, flags=0x02) with NoSources
Caused by: Reset.Cancel

E 1128 17:25:08.105 THREAD1687: [C L:/127.0.0.1:64437 R:/127.0.0.1:9897] dispatcher failed
Failure(com.twitter.finagle.buoyant.h2.Reset$Cancel$, flags=0x02) with NoSources
Caused by: Reset.Cancel```

**Comment by olix0r** [comment]
thanks for the note, @zackangelo. that error is benign but i'll fix logging

File: router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4H2Listener.scala [hunk 1]
```diff
@@ -53,6 +53,12 @@ object Netty4H2Listener {
         proto match {
           case ApplicationProtocolNames.HTTP_2 =>
             ctx.channel.config.setAutoRead(true)
+            // ctx.pipeline.addBefore(PlaceholderKey, "wtf", new ChannelDuplexHandler {
```
adleong: is this a debug handler?  should this be deleted, or add a comment explaining?

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4H2Writer.scala
```diff
@@ -25,33 +28,10 @@ private[netty4] trait Netty4H2Writer extends H2Transport.Writer {
...
-    }
-
-  override def write(id: Int, data: Frame.Data): Future[Unit] =
-    write(id, data.buf, data.isEnd)
-
-  override def write(id: Int, tlrs: Frame.Trailers): Future[Unit] =
-    write(id, tlrs, true /*eos*/ )
+  override def write(id: Int, f: Frame): Future[Unit] = f match {
+    case data: Frame.Data => write(id, data.buf, data.isEnd)
+    case tlrs: Frame.Trailers => write(id, tlrs, true /*eos*/ )
```
adleong: how about using a named parameter?
`, eos = true)`

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -2,188 +2,560 @@ package com.twitter.finagle.buoyant.h2
...
+              } else admitRemote(data)
+            } else {
+              if (admitFrame(toFrame(data), remote)) true
+              else admitRemote(data)
+            }
+        }
+    }
   }
 
+  /** */
```
adleong: empty doc string

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -2,188 +2,560 @@ package com.twitter.finagle.buoyant.h2
...
   private[this] def toFrame(f: Http2StreamFrame): Frame = f match {
     case f: Http2DataFrame => Netty4Message.Data(f, updateWindow)
     case f: Http2HeadersFrame if f.isEndStream => Netty4Message.Trailers(f.headers)
     case f => throw new IllegalArgumentException(s"invalid stream frame: ${f}")
   }
 
-  private[this] val mapFutureUnit = (_: Any) => Future.Unit
+  private[this] val updateWindow: Int => Future[Unit] =
+    incr => transport.updateWindow(streamId, incr)
```
adleong: tioli: `transport.updateWindow(streamId, _)`

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -0,0 +1,202 @@
...
+ * flow control semantics are built into the Stream--consumers MUST
+ * release each data frame after it has processed its contents.
+ *
+ * Consumers SHOULD call `read()` until it fails (i.e. when the
+ * stream is fullt closed).
+ *
+ * If a consumer cancels a `read()` Future, the stream is reset.
+ */
+trait Stream {
+  override def toString = s"Stream(empty=$isEmpty, onEnd=$onEnd)"
```
olix0r: yep, i suppose i could use .poll to get a slightly nicer string

router/h2/src/e2e/scala/io/buoyant/router/h2/RouterEndToEndTest.scala
```diff
@@ -44,4 +44,132 @@ class RouterEndToEndTest
...
+      val _ = await(dogReqP)
+      assert(!rspF.isDefined)
+
+      dogRspP.setException(Failure("failz").flagged(Failure.Rejected))
+      assert(await(rspF.liftToTry) == Throw(Reset.Refused))
+
+    } finally {
+      try await(client.close().before(dog.server.close()).before(router.close()))
+      catch { case e: Throwable => log.error(e, "closing") }
+      finally setLogLevel(Level.OFF)
```
olix0r: I'd hope so, but honestly the logging stuff gets wonky _all_ the time, so i've taken to being OCD about clearing it out

router/h2/src/e2e/scala/io/buoyant/router/h2/ConcurrentStreamsEndToEndTest.scala
```diff
@@ -16,7 +17,7 @@ class ConcurrentStreamsEndToEndTest
 
   val FrameSize = 16 * 1024
   val WindowSize = 4 * FrameSize
-  val lengths = Seq(1, WindowSize * 2 - WindowSize / 2)
+  val lengths = Seq(/*1,*/ WindowSize * 2 - WindowSize / 2)
```
olix0r: i forgot to reenable the test ;)

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4H2Writer.scala
adleong:
```
alex-helloworld@ def foo(x: Int, y: Int, z: Int) = x + y + z
defined function foo
alex-helloworld@ foo(1,2, z = 3)
res4: Int = 6
```

olix0r: thanks. most of my scala knowledge comes from things i tried once in 2011 ;p

PR #830, opened 2016-11-30
Improve commentary; reuse code in stream transport tests.

olix0r:
Netty4StreamTransport is a bit dense, so this is an initial attempt to clarify
it somewhat.

The directionality in the bidirectional stream was easy to lose track of --
I've renamed `LocalMsg` and `RemoteMsg` to _SendMsg_ and _RecvMsg_
respectively; and the method names `write` and `admitRemote` have been renamed
_send_ and _recv_.

The stream `recv`s `RecvMsg`s from the _remote_ and `send`s `SendMsg`s from the
_local_ side of the stream.

I've tried to add commentary around Netty4StreamTransport's internal state
machine.  In trying to clarify this, I improved the scoping of the `RecvMsg`
promise by removing it from the instance scope and protecting it within the
`RemotePending` remote state.

Finally, I've extracted a bunch of common code from within
Netty4StreamTransportTest so the tests are easier to read.


olix0r:
https://github.com/BuoyantIO/linkerd/commit/9fe3c4d8791d1f7f9779486578ad050bd3c61158

PR #834, opened 2016-12-02
[h2] fail streams with malformed connection headers

### LINKED ISSUE #833: [h2] linkerd does not properly fail malformed HTTP/2 requests
RFC7540 is quite clear about what constitutes a valid message and what to do when an invalid message is encountered:

> 8.1.2.2.  Connection-Specific Header Fields
> 
>    HTTP/2 does not use the Connection header field to indicate
>    connection-specific header fields; in this protocol, connection-
>    specific metadata is conveyed by other means.  An endpoint MUST NOT
>    generate an HTTP/2 message containing connection-specific header
>    fields; any message containing connection-specific header fields MUST
>    be treated as malformed (Section 8.1.2.6).
> 
>    The only exception to this is the TE header field, which MAY be
>    present in an HTTP/2 request; when it is, it MUST NOT contain any
>    value other than "trailers".
> 
>    This means that an intermediary transforming an HTTP/1.x message to
>    HTTP/2 will need to remove any header fields nominated by the
>    Connection header field, along with the Connection header field
>    itself.  Such intermediaries SHOULD also remove other connection-
>    specific header fields, such as Keep-Alive, Proxy-Connection,
>    Transfer-Encoding, and Upgrade, even if they are not nominated by the
>    Connection header field.

Currently, as done in HTTP/1, linkerd simply removes these connection headers and forwards the request, which is in violation of the RFC:

> 8.1.2.6.  Malformed Requests and Responses
> 
>    A malformed request or response is one that is an otherwise valid
>    sequence of HTTP/2 frames but is invalid due to the presence of
>    extraneous frames, prohibited header fields, the absence of mandatory
>    header fields, or the inclusion of uppercase header field names.
> 
>    ...
>
>    Intermediaries that process HTTP requests or responses (i.e., any
>    intermediary not acting as a tunnel) MUST NOT forward a malformed
>    request or response.  Malformed requests or responses that are
>    detected MUST be treated as a stream error (Section 5.4.2) of type
>    PROTOCOL_ERROR.

Furthermore, linkerd removes the `te: trailers` header from requests; but this header is required by, for instance, some gRPC implementations.

linkerd should preserve the `te: trailers` header, and it should flat-out fail requests that have invalid headers.

olix0r:
linkerd removes all connection-specific headers from requests that transit the
router. However, this behavior is not in accordance with RFC7540:

> An endpoint MUST NOT generate an HTTP/2 message containing
> connection-specific header fields; any message containing connection-specific
> header fields MUST be treated as malformed...
>
> The only exception to this is the TE header field, which MAY be present in an
> HTTP/2 request; when it is, it MUST NOT contain any value other than
> "trailers".

> Intermediaries that process HTTP requests or responses (i.e., any
> intermediary not acting as a tunnel) MUST NOT forward a malformed request or
> response.  Malformed requests or responses that are detected MUST be treated
> as a stream error (Section 5.4.2) of type PROTOCOL_ERROR.

To fix this, the stream transport detects connection-specific headers and
resets streams appropriately.  Furthermore, the `te: trailers` header is not
removed from requests, which is required for gRPC compatibility.

Fixes #833


router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -290,33 +290,43 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
     val state = stateRef.get
     log.trace("[%s] admitting %s in %s", prefix, in.name, state)
 
-    def resetRemote(remote: ResettableState, rst: Reset): Boolean =
+    def resetFromRemote(remote: ResettableState, rst: Reset): Boolean =
       if (stateRef.compareAndSet(state, Closed(rst))) {
+        println(s"REMOTE RESET: $rst")
```
adleong: rm or switch to log

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -290,33 +290,43 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
         remote.reset(rst)
         resetP.setException(StreamError.Remote(rst))
         true
       } else false
 
+    def resetFromLocal(remote: ResettableState, rst: Reset): Boolean =
+      if (stateRef.compareAndSet(state, Closed(rst))) {
+        println(s"LOCAL RESET: $rst")
```
adleong: rm or switch to log

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -290,33 +290,43 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
+        true
+      } else false
+
     def recvFrame(f: Frame, remote: RemoteStreaming): Boolean =
       if (remote.offer(f)) {
         statsReceiver.recordRemoteFrame(f)
         true
       } else false
 
+    println(s"IN: $in")
```
adleong: rm or switch to log

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -369,38 +391,50 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
+            }
 
           case LocalClosed(remote@RemotePending()) =>
             val q = new AsyncQueue[Frame]
-            if (stateRef.compareAndSet(state, LocalClosed(RemoteStreaming(q)))) {
-              remote.setMessage(hdrs.headers, Stream(q))
-              true
-            } else recv(hdrs)
+            val msg = mkRecvMsg(hdrs.headers, Stream(q))
+            if (ConnectionHeaders.detect(msg.headers)) {
```
adleong: Possible to DRY up usages of this block?

olix0r: I gave it a shot but it's a little tricky to do without allocating. going to punt for now

PR #835, opened 2016-12-02
[h2] Reset streams on routing error

olix0r:
Rather than lift intermediary errors into the application (5xx), we can simply refuse the entire stream for unroutable requests.  This indicates to the client in a more principled way that absolutely nothing was done to process the request.

Furthermore, the Netty4StreamTransportTest has bee improved both in terms of coverage and readability


adleong:
I think this needs to be re-diffed against latest master

PR #836, opened 2016-12-03
Add top-level logical service stats to each client

olix0r:
Problem

Currently linkerd exposes top-level (latency histogram, and request/success/failure counters) for both the logical ('dst/path') and concrete ('dst/id') scopes. However, these stats expose no information about which logical requests each concrete client is serving.

Solution

Finagle provides a 'Context' primitive for sharing i.e. request-level metadata within a processing scope, without encoding additional information in a protocol-specific request type.

This change introduces two local contexts: ctx.DstPath and ctx.DstBound. This allows for client-level filters to access routing context in a protocol-agnostic fashion.

The ctx.DstPath context is used by a new client module, `PerDstPathStatsFilter`, which annotates metrics like:

- dst/id/$id/dst/path/$path/request_latency_ms...
- dst/id/$id/dst/path/$path/requests
- dst/id/$id/dst/path/$path/success
- dst/id/$id/dst/path/$path/failures...


router/core/src/main/scala/io/buoyant/router/PerDstPathStatsFilter.scala
```diff
@@ -0,0 +1,41 @@
...
+        next: ServiceFactory[Req, Rsp]
+      ): ServiceFactory[Req, Rsp] = _stats match {
+        case param.Stats(statsReceiver) if !statsReceiver.isNull =>
+          val param.ResponseClassifier(classifier) = _classifier
+          val param.ExceptionStatsHandler(handler) = _exceptions
+          val StatsFilter.Param(unit) = _unit
+
+          def mkScopedStatsFilter(path: Path): Filter[Req, Rsp, Req, Rsp] = {
+            val name = path.show.stripPrefix("/")
+            val sr = statsReceiver.scope("dst/path", name)
```
klingerf: TIOLI: I have a very slight preference for scoping this as: `statsReceiver.scope("path", name)` or  `statsReceiver.scope("dst", "path", name)`, to more closely resemble the existing dst path stats.

router/core/src/main/scala/io/buoyant/router/ctx.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.twitter.finagle.context.Contexts
+import com.twitter.util.Local
+
+package object ctx {
+
+  private class SetFilter[T, Req, Rsp](key: Contexts.local.Key[T], value: T)
+    extends SimpleFilter[Req, Rsp] {
+
+    def apply(req: Req, service: Service[Req, Rsp]) = {
+      val save = Local.save()
```
adleong: Is this right?  Doesn't `let` already take care of saving and restoring?  Isn't the thread local referred to by `Local.save` actually different from `Contexts.local`?

router/core/src/main/scala/io/buoyant/router/Router.scala
```diff
@@ -317,13 +321,15 @@ object StackRouter {
     stk.push(ClassifiedRetries.module)
     stk.push(StatsFilter.module)
```
adleong: This provides path level request stats right?  Can this now be removed?

router/core/src/main/scala/io/buoyant/router/ctx.scala
olix0r: honestly, just cargo-culted from finagle, i'll take a deeper look

router/core/src/main/scala/io/buoyant/router/ctx.scala
```diff
@@ -0,0 +1,55 @@
+package io.buoyant.router
+
+import com.twitter.finagle._
+import com.twitter.finagle.buoyant.Dst
+import com.twitter.finagle.context.Contexts
+import com.twitter.util.Local
+
+package object ctx {
```
olix0r: does making it a plain-object change that behavior, though?

adleong: You'd have to capitalize `Ctx`.  😜 

plus package objects are just weird.

olix0r: well, i wanted this to be a package rather than an object, but i was being lazy and wanted to put it in one file without creating a directory...  hence package object.

Moving it into its own directory doesn't really address your concern, though

olix0r: how about i rename it `context` so it seems like less of temp val

router/core/src/main/scala/io/buoyant/router/ctx.scala
olix0r: you're right -- the local-clearing stuff only belongs at the top of the server

router/core/src/main/scala/io/buoyant/router/ctx.scala
adleong: what about moving them into a directory AND importing `DstPath` and `DstBound` directly?  Just trying to follow the principle of least surprise.

olix0r: yeah, went that route

PR #838, opened 2016-12-06
Add localhost, port, daemonset, and localnode namers

adleong:


adleong:
After some in-person discussion we will probably not pursue this.  Doing this kind of thing as a transformer instead of as a namer is appealing because it allows the initial naming to happen in namerd but still allows post-processing to happen in linkerd.  This is useful in the case where you want to resolve an id in namerd and then filter that set of addresses down based on some contextual information available in linkerd.  An example of this is the `localnode` namer/transformer.  In this use case we want to be able to resolve the destination service id in namerd but then filter that down to only endpoints which are node-local to the linkerd.  This is better accomplished using a transformer rather than a namer since the entirety of naming takes place in namerd (if namerd is used at all).

PR #844, opened 2016-12-07
Add headerPath config option to header identifier

### LINKED ISSUE #719: Header identifier that uses header value as a "path component" rather than a "path"
A follow-up on #638 .

The example use-case is HTTP/1.0 clients that (for whatever reason) cannot move to HTTP/1.1 but can set the `Host` header.

After a brief discussion with @olix0r and @adleong in Slack it looks like we agree that we need something like that. The question how exactly we want to make this happen.

There are a few options:
1) an option for `io.l5d.header` identifier to switch it's behavior
2) an extra identifier that does it

Also @olix0r mentioned that with next 'breaking' release of linkerd he'd like to have this as 'default behavior' (guess that applies if we will decide to extend current identifier).

adleong:
Fixes #719 

If headerPath is true (default), parse the header as a full path.
If headerPath is false, parse the header as a single path segment.

### DIFF
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -172,7 +172,8 @@ urlPath | N/A | A path from the URL whose number of segments is set in the ident
 kind: `io.l5d.header`
 
 With this identifier, HTTP requests are turned into names based only on the
-value of an HTTP header.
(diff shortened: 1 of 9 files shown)


olix0r:
In h2 we have two separate identifiers.  I have a mild preference for that approach, but ultimately prefer that it's consistent between protocols, either way.

PR #845, opened 2016-12-07
Localhost transformer uses all local network interfaces

### LINKED ISSUE #831: localhost transformer shouldn't depend on DNS
As discussed in #774, the `io.l5d.localhost` transformer tries to find the local IP with, effectively, `$(dig $(hostname))` -- This is useless when `hostname` isn't resolvable (which actually happens in a variety of cloud configurations, weeeee!).

I propose that we use the same solution introduce din #757 for announcers: enumerate all non-loopback host interfaces and their addresses; and use these values to determine locality.

adleong:
FIxes #831

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## In the next release...
 
+* Fix localhost transformer when used on systems with unresolvable hostname.
+
 ## 0.8.4 2016-12-05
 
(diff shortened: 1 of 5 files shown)


adleong:
I'm having a hard time reproducing #831 in order to validate this fix.  I'm considering just shipping this anyway since the new behavior seems better.  Thoughts?

PR #858, opened 2016-12-13
Upgrade to finagle 6.39.0

adleong:


admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -174,7 +174,7 @@ object DelegateApiHandler {
     private[this] val mapper = new ObjectMapper with ScalaObjectMapper
     mapper.registerModule(DefaultScalaModule)
     mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
-    mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL)
```
olix0r: Is this related to the finagle change? Do we need to change our jackson version to be consistent with finagle?

router/core/src/test/scala/io/buoyant/router/RouterTest.scala
```diff
@@ -120,7 +120,6 @@ class RouterTest extends FunSuite {
         val id: Any = path match {
           case Path.Utf8("0") => Path.Utf8("some", "path")
           case Path.Utf8("1") => "string"
-          case Path.Utf8("2") => null
```
olix0r: What changed in finagle related to this?

adleong: finagle now calls `.hashCode` on ids which means that using `null` for an id is a no-no.

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
adleong: Yep.  Now updated to 2.8.4.

PR #859, opened 2016-12-14
Add client failureAccrual configuration object

### LINKED ISSUE #468: Support configuration for minSuccessRate failure accrual policy
https://twitter.github.io/finagle/guide/Clients.html#failure-accrual

We currently do not expose failure accrual policies via configuration.  We should expose a plugin interface that supports loading/installation of failure accrual factories. Out of the box, we should support both the default behavior (maxConsecutiveFailures -- the number of failures should be configurable) as well as the minimum-success-rate policy.

klingerf:
Adding a new client configuration key -- `failureAccrual` -- which supports configuring a failure accrual policy for all clients created by the configured router. I've exposed failured accrual as a plugin interface, and have provided 3 implementations based on what is provided by finagle: consecutive failures, success rate by count, and success rate by time. Fixes #468.

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
@@ -1,0 +1,37 @@
+package io.buoyant.linkerd
+
+import com.fasterxml.jackson.annotation.{JsonIgnore, JsonTypeInfo}
+import com.twitter.conversions.time._
+import com.twitter.finagle.service.{Backoff, FailureAccrualFactory}
+import com.twitter.finagle.service.exp.FailureAccrualPolicy
+import com.twitter.util.Duration
+import io.buoyant.config.ConfigInitializer
+
+abstract class FailureAccrualInitializer extends ConfigInitializer
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
+trait FailureAccrualConfig {
+  @JsonIgnore
+  def policy: () => FailureAccrualPolicy
+
+  val backoff: Option[BackoffConfig]
+
+  @JsonIgnore
+  val backoffOrDefault =
+    backoff.map(_.mk).getOrElse(FailureAccrualConfig.defaultBackoff)
+}
+
+object FailureAccrualConfig {
+  // Settings here mirror Finagle's FailureAccrualFactory.defaultPolicy, but are provided
+  // in linkerd to make it easier to reason about the default failure accrual settings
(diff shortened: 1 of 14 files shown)


linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
```diff
@@ -0,0 +1,37 @@
...
+import com.twitter.finagle.service.exp.FailureAccrualPolicy
+import com.twitter.util.Duration
+import io.buoyant.config.ConfigInitializer
+
+abstract class FailureAccrualInitializer extends ConfigInitializer
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
+trait FailureAccrualConfig {
+  @JsonIgnore
+  def policy: () => FailureAccrualPolicy
```
adleong: This type seemed odd to me so I dug into FailureAccrualFactory to figure out why.  It turns out that the FailureAccrualPolicy isn't just a policy but also keeps the failure accrual state, which is why you would want to be able to create new ones on demand, which explains this type.  ☄️ :star:

linkerd/failure-accrual/src/main/scala/io/buoyant/linkerd/failureAccrual/SuccessRateWindowedInitializer.scala
```diff
@@ -0,0 +1,23 @@
...
+
+class SuccessRateWindowedInitializer extends FailureAccrualInitializer {
+  val configClass = classOf[SuccessRateWindowedConfig]
+  override def configId = "io.l5d.successRateWindowed"
+}
+
+object SuccessRateWindowedInitializer extends SuccessRateWindowedInitializer
+
+case class SuccessRateWindowedConfig(
+  sr: Double,
```
adleong: is `sr` too cryptic for a config param?  do you think `successRate` would be too verbose?

linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  val backoff: Option[BackoffConfig]
```
klingerf: Awesome, yeah, I figured there had to be a better way to go about it.

linkerd/failure-accrual/src/main/scala/io/buoyant/linkerd/failureAccrual/SuccessRateWindowedInitializer.scala
klingerf: `successRate` works for me -- switching to that in the latest commit.

PR #860, opened 2016-12-14
[h2] Add a generic Trailers constructor

olix0r:
Currently there's no easy way to build a Trailers frame from application code.
(There is a netty-specific builder, but that is and should only be visible
within the netty codec.)

This change introduces a generic `Frame.Trailers(hdrs: Seq[(String, String)])`
constructor that's usable from application code (or a gRPC iplementation ;)

### DIFF
--- a/router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
+++ b/router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
@@ -199,4 +199,24 @@ object Frame {
     final override def isEnd = true
   }
 
+  object Trailers {
+
+    private class Impl(orig: Seq[(String, String)])
+      extends Headers.Impl(orig) with Trailers {
+
+      private[this] val promise = new Promise[Unit]
+      def release(): Future[Unit] = {
+        promise.updateIfEmpty(Return.Unit)
+        Future.Unit
+      }
+      def onRelease: Future[Unit] = promise
+    }
+
+    def apply(pairs: Seq[(String, String)]): Trailers =
+      new Impl(pairs)
+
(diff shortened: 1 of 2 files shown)


router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -199,4 +199,24 @@ object Frame {
...
+        promise.updateIfEmpty(Return.Unit)
+        Future.Unit
+      }
+      def onRelease: Future[Unit] = promise
+    }
+
+    def apply(pairs: Seq[(String, String)]): Trailers =
+      new Impl(pairs)
+
+    def apply(hd: (String, String), tl: (String, String)*): Trailers =
```
adleong: is this to specifically disallow an empty list of headers?  ie `Trailers()`

olix0r: yeah (force of habit). i don't think that this is necessary in this case, though.

olix0r: oh nm, this is why:
```
[error] def apply(pairs: Seq[(String, String)]): com.twitter.finagle.buoyant.h2.Frame.Trailers at line 215 and
[error] def apply(pairs: (String, String)*): com.twitter.finagle.buoyant.h2.Frame.Trailers at line 218
```

force of habit for good

adleong: wild

olix0r:
https://github.com/BuoyantIO/linkerd/commit/b4a1ae3807d855b792e0680c133d3de7723f9a99

PR #862, opened 2016-12-14
Fix incorrect router request graph in dashboard

adleong:
In a pervious change, we switched from using the "request" stat to the "connects" stat for discovering clients from metrics.  I believe this had the consequence of no longer displaying the requests data on the router graph.

This change makes the computation of that data a two step process:
1. Get a list of clients by looking for the "connects" stat
2. Get the "requests" metrics for those clients

Unfortunately, this means that the list of clients must be continuously recomputed in case it changes.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
@@ -8,7 +8,7 @@ var CombinedClientGraph = (function() {
   function currentClients(routerName, metrics) {
     var clientQuery = Query.clientQuery().withRouter(routerName).withMetric("connects").build();
     return _.map(Query.filter(clientQuery, metrics), function(metric) {
-        var match = metric.name.match(clientQuery)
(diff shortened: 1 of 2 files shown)


adleong:
Thanks for the tip, @esbie!  I wasn't able to use `onAddedClients` because it seems like that triggers on clients added to any router so instead I just recompute the clients each time.  Let me know if I'm missing something, though.

admin/src/main/resources/io/buoyant/admin/js/combined_client_graph.js
```diff
@@ -49,7 +51,8 @@ var CombinedClientGraph = (function() {
         // where the first values from /metrics are very large [linkerd#485]
         count++;
       } else {
-        var filteredData = Query.filter(query, data.specific);
+        var clients = _.map(routers.clients(routerName), function(client) { return client.label; });
```
rmars: tioli: you can use the map shorthand here
`_.map(routers.clients(routerName),'label')`

https://lodash.com/docs/4.17.2#map

PR #866, opened 2016-12-15
Transformer ids

adleong:
Transformers will now prepend a prefix to the `id` (concrete name) of the bound names they modify.  This allows consumers of linkerd metrics to distinguish between transformed and un-transformed concrete names.  For example, an outgoing router that uses the fs namer and the port transformer to talk to the incoming linkerd on the same host as the destination would see concrete names like this:
```
/%/io.l5d.port/4141/#/io.l5d.fs/foo
```
The `%` symbol denotes a transformer prefix.  Note that transformer prefixes are never present in dtabs so users don't need to be aware of them.

## Gotchas

The namerd naming interface is a two-step process with a `bind` step and an `addr` step.  When transformers are involved, the `bind` step now returns name trees with bound ids that include the transformer prefix.  Therefore we need to strip the transformer prefix when that id is submitted to the `addr` step.

### Outgoing clients: 
screen shot 2016-12-15 at 12 04 24 pm

### Outgoing delegation:
screen shot 2016-12-15 at 12 04 47 pm

### Incoming clients:
screen shot 2016-12-15 at 12 04 31 pm

### Incoming delegation:
screen shot 2016-12-15 at 12 04 57 pm


namerd/core/src/main/scala/io/buoyant/namerd/package.scala
```diff
@@ -9,4 +11,16 @@ package object namerd {
...
+  val Dollar = Buf.Utf8("$")
+  val NotHashOrDollar: Buf => Boolean =
+    (b: Buf) => !(b == Hash || b == Dollar)
+  val Percent = Path.Utf8("%")
+
+  def stripTransformerPrefix(p: Path): Path =
+    if (p.startsWith(Percent))
+      Path(p.elems.dropWhile(NotHashOrDollar): _*)
+    else
+      p
```
olix0r: I think I'd prefer that we scope this into a `Paths` object and reference as e.g.  `Paths.Percent`, `Paths.stripTransfomerPrefix()?

Also perhaps better to call `Hash`->`ConfiguredNamerPrefix`, `Dollar`->`LoadedNamerPrefix`, `Percent`->`TransformerPrefix`, etc?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/AddrHandler.scala
```diff
@@ -6,7 +6,7 @@ import com.twitter.io.Buf
 import com.twitter.util.Future
 import io.buoyant.admin.names.DelegateApiHandler
 import io.buoyant.admin.names.DelegateApiHandler.{Addr => JsonAddr}
-import io.buoyant.namerd.RichActivity
+import io.buoyant.namerd.{RichActivity, stripTransformerPrefix}
```
olix0r: tioli:

I like to do

```scala
package io.buoyant.namerd
package iface
```

for this sort of thing so that intra-project imports aren't necessary.

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/PortTransformer.scala
```diff
@@ -26,19 +26,27 @@ class PortTransformer(port: Int) extends DelegatingNameTreeTransformer {
...
   override protected def transform(tree: NameTree[Bound]): Activity[NameTree[Bound]] =
     Activity.value(tree.map(mapBound))
 
   override protected def transformDelegate(tree: DelegateTree[Bound]): Activity[DelegateTree[Bound]] =
     Activity.value(tree.flatMap { leaf =>
+      val bound = mapBound(leaf.value)
+      val path = bound.id match {
+        case id: Path => id
+        case _ => leaf.path
+      }
```
olix0r: ... What does this do?  Are we setting the residual path of the transformed name?   And we set it to be only the bound id? should this be (effectively) `bound.id ++ leaf.path`?

adleong: I've factored this common code into the DelegatingNameTreeTransformer companion object and renamed variables for clarity.
Here we are extracting the transformed id from the transformed `Name.Bound`.  

`bound.id` should be equal to `prefix ++ orig.id`.

namer/core/src/main/scala/io/buoyant/namer/TransformerConfig.scala
```diff
@@ -1,12 +1,26 @@
...
+  @JsonIgnore
+  def prefix = TransformerConfig.Percent ++ _prefix.getOrElse(defaultPrefix)
+
+  @JsonIgnore
   def mk(): NameTreeTransformer
 }
 
+object TransformerConfig {
+  val Percent = Path.Utf8("%")
+}
```
olix0r: nit: Should the Paths object actually be in namer-core?

PR #873, opened 2016-12-21
Add RequireJS, add javascript tests

rmars:
## Problem
* The javascript in the admin page has no tests
* Currently we load all the js files by <script> includes in the scala generated html. This is fine as the app is small, but it does leave us with having to touch a scala file to add js, and also doesn't set us up well for adding tests. It's also less clear where a file's dependencies are coming from, and we need to manually order our js loads.

## Solution
* Add RequireJS to the repo
* Add jasmine to the repo (along with karma to run the tests)

I chose RequireJS because I didn't want to add a build/bundle step to the js. Using require allows us to include the app with a one line <script> include, and doesn't require any changes to the way we develop js right now.

Yays:
* We can explicitly declare a module's dependencies
* We can require files for tests
* We don't have to rely on global variables
* We can move around js files at will without having to touch the scala code.
* We use a requirejs eslint module and no longer need as many /* globals */ in each file

Nays:
* [fixed] I'm detecting whether this is namerd or linkerd via a fragile page title test
* Added basic manual routing based on window.location in main.js. I'm sure there's a better way to handle this
* [fixed] Since smoothie.js exports two things (SmoothieChart and TimeSeries) I needed to copy paste the file (so now we have two copies for it) in order to make a separate entry for each export in shims (config is in main.js). There's probably a better way to do this; I didn't want to edit the original file though.

## Changes

I don't think github will deal well with my moves but here's a rundown of the code changes:

* Changed: bugfix in query.js (withServers)
* Added: /spec for specs
  	* test-main.js is the require entrypoint for the tests
	* there aren't very many specs; I'll write more in a follow-up branch. Included two small specs as a start.
* Added main.js as the entry point for the app
* Moved all the files from /js (except for /lib) into /src
        The js files in /js are just copy pasted except for 
	* adding the `define([], function() {})` wrappers on each file
 	* in admin.js, wrapped code into an initialize() function and then exported each function in that file 
	* adding main.js as an entry point into the app
* moved /templates to /js/templates and used text.js to include them via require, removing the need for a series of `$.get()`s
* logger_overrides.js and metrics.js are still included directly from <script> - I didn't think it was useful to add them as they're only used in one place
* added karma to run the tests, added a spec reporter

Since I messed with ALL the javascripts I'm sure I could have introduced bugs. Please let me know if you find errors/anything looks wonky.


admin/src/main/resources/io/buoyant/admin/spec/test-main.js
```diff
@@ -0,0 +1,54 @@
...
+    // If you require sub-dependencies of test files to be loaded as-is (requiring file extension)
+    // then do not normalize the paths
+    var normalizedTestModule = file.replace(/^\/base\/|\.js$/g, '')
+    allTestFiles.push(normalizedTestModule)
+  }
+})
+
+require.config({
+  // Karma serves files under /base, which is the basePath from your config file
+  baseUrl: '/base',
```
rmars: I think this is where karma likes to load its assets from:

http://karma-runner.github.io/1.0/config/files.html
```
By default all assets are served at http://localhost:[PORT]/base/
...
Notice the base in the URL, it is a reference to your basePath. You do not need to replace or provide your own base.
```

So if our basePath was "/foo" karma would serve from "/base/foo" - but I am not sure we want to change the /base part. It loads our assets with that /base/ in the url, e.g. http://localhost:9876/base/js2/lib/lodash.min.js with relative requires.

Looks like I can get around this by doing something like
```
proxies: {
  '/': '/base',
},
```
but we don't rely on the urls anyway... so it doesn't matter if we are using /base. What do you think?

esbie: yeah fine as is thanks

admin/src/main/resources/io/buoyant/admin/js/main.js
```diff
@@ -0,0 +1,62 @@
...
+  'src/delegate',
+  'src/dashboard_delegate'
+], function (
+  $, _,
+  adminPage,
+  dashboard,
+  linkerdDtabPlayground,
+  namerdDtabPlayground
+) {
+  if ($('title').text().indexOf("namerd") !== -1) {
```
rmars: great point. if I use separate entrypoints I don't need a data attr to know we should initialize namerd admin

admin/src/main/resources/io/buoyant/admin/js/src/dashboard.js
```diff
@@ -0,0 +1,77 @@
...
+      $.get("/files/template/router_server.template"),
+      $.get("/files/template/router_client.template"),
+      $.get("/files/template/router_client_container.template"),
+      $.get("/files/template/router_server_container.template"),
+      $.get("/files/template/server_rate_metric.partial.template"),
+      $.get("/files/template/metric.partial.template"),
+      $.get("/files/template/router_summary.template"),
+      $.get("/files/template/process_info.template"),
+      $.get("/files/template/request_totals.template"),
+      $.get("/admin/metrics.json")
```
rmars: You're right! doing this means we can get rid of loading all the templates and passing them down from the router_controller too. I've moved templates/ to js/ templates and gotten rid of most of these $gets.

I would like to stop passing these down, and instead directly include the templates in the modules that they are used. This will involve a little refactoring (the module initializations) so I will make a branch right after this one, so the changes will actually be reflected in the diffs.

admin/src/main/resources/io/buoyant/admin/js/lib/smoothie.js
```diff
@@ -801,4 +802,4 @@
   exports.SmoothieChart = SmoothieChart;
 
 })(typeof exports === 'undefined' ? this : exports);
-
+});
```
esbie: ur my hero ^^

rmars:
thank you for looking that this! 🙏 

agreed! I also just realized I forgot to move references to /files/template and delegate.js and dashboard_delegate.js - doing that now.

Issue #884, opened 2017-01-04
Coverage reporting


olix0r:
scoverage integration has been removed in the ver/grpc-gen branch, because scoverage plugin breaks the grpc-eg project tests: these tests build the grpc-gen binary before testing, and since the scoverage plugin injects itself as a compiler plugin, the produced binary is linked against scoverage, despite the fact that the runtime library is not present for the binary.

It'd be nice to restore coverage reporting.  Figure this out eventually.  (It would also be nice to include e2e and integration tests in coverage reporting, which wasn't possible in scoverage).

Issue #855, opened 2016-12-13
gRPC namerd API


olix0r:
Expose namerd's APIs as a gRPC service so that it's easier to build clients (esp for streaming responses).

olix0r:
dup of #842

PR #885, opened 2017-01-04
Implement gRPC code generation for finagle

olix0r:
This change introduces the _grpc_ aggregate project including:
- The grpc-gen project, providing the io.buoyant.grpc protoc plugin
- The grpc-runtime library, supporting de/serialization and H2 transport
- The grpc-eg project, including a gRPC service that tries to exercise all of
  gRPC's features and an end-to-end test.

This change is not compatible with the 'scoverage' sbt plugin, and so coverage
reporting has been removed from the repo for now. See #884.

Relates to #855.


grpc/eg/src/main/protobuf/eg.proto
```diff
@@ -0,0 +1,50 @@
...
+
+message Req {
+  Enumeration value = 1;
+}
+message Rsp {
+  Message.Enumeration value = 1;
+}
+
+service Eggman {
+  rpc Uplatu (Req) returns (Rsp) {}
```
adleong: I get that U is for unary and S is for stream... but what is plat?

olix0r: The service is called Eggman and the original method was Splat, and then I took it too far.  I think we'll replace this test in the future, so I'm not too worried about it for now.

PR #889, opened 2017-01-05
Clean up template passing in dashboard js

rmars:
Now that we are using require, we don't need to pass down templates all the way from the root.
Instead, let's include the template alongside the module that uses it.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
@@ -1,10 +1,14 @@
 "use strict";
 
 define([
-  'jQuery',
+  'jQuery', 'Handlebars',
   'src/router_client',
-  'src/combined_client_graph'
-], function($, RouterClient, CombinedClientGraph) {
+  'src/combined_client_graph',
+  'text!template/router_client_container.template'
+], function($, Handlebars,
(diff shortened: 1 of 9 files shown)


admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
```diff
@@ -22,7 +26,9 @@ define([
       return numClients < EXPAND_CLIENT_THRESHOLD;
     }
 
-    return function (metricsCollector, routers, $clientEl, routerName, clientTemplate, metricPartial, clientContainerTemplate, colors) {
+    return function (metricsCollector, routers, $clientEl, routerName, colors) {
+      var clientContainerTemplate = Handlebars.compile(routerClientContainerTemplate);
```
esbie: These vars can now be hoisted out of the returned function... although this is only actually useful if we're instantiating these component more than once. (are we?)

Template compilation on the client is usually a very costly operation because of the string manipulation required.

rmars: we only instantiate RouterClients once (in RouterController) - so I think we are safe!

esbie: Awesome 👍

PR #898, opened 2017-01-06
statsd telemeter

### LINKED ISSUE #429: statsd telemeter
Good Evening Linkerd.

We currently have some finagle stats reporters sending to local statsd collectors, it would be really great if we could configure linkerd to do the same.  In the short term it's possible for us to write a script to pull these metrics on an interval but long term it would be really neat if this were configurable.

siggy:
fixes #429

### DIFF
--- a/linkerd/docs/config.md
+++ b/linkerd/docs/config.md
@@ -37,6 +37,19 @@ namers:
 tracers:
 - kind: io.l5d.zipkin
   sampleRate: 0.02
+
+telemetry:
+- kind: io.l5d.commonMetrics
+- kind: io.l5d.statsd
+  experimental: true
+  prefix: linkerd
+  hostname: 127.0.0.1
+  port: 8125
+  gaugeIntervalMs: 10000
+  sampleRate: 0.01
+- kind: io.l5d.tracelog
+  sampleRate: 0.2
+  level: TRACE
 ```
 
 Welcome to the Configuration Reference for linkerd!
@@ -56,6 +69,7 @@ Key | Required | Description
 [admin](#administrative-interface) | no | Configures linkerd's administrative interface.
 [routers](#routers) | yes | Configures linkerd's RPC support for various protocols.
 [namers](#namers-and-service-discovery) | no | Configures linkerd's integration with various service discovery backends.
+[telemetry](#telemetry) | no | Configures linkerd's metrics instrumentation.
 [tracers](#tracers) | no | Configures linkerd's request instrumentation.
 
 
@@ -110,6 +124,28 @@ is governed by the [routing](#router-parameters) configuration.
 Naming and service discovery are configured via the `namers` section of the
 configuration file.  A namer acts on paths that start with `/#` followed by the
 namer's prefix. See [namers](#namers).
+
+### Telemetry
+
+```yaml
+telemetry:
+- kind: io.l5d.commonMetrics
+- kind: io.l5d.statsd
+  experimental: true
+  prefix: linkerd
+  hostname: 127.0.0.1
+  port: 8125
+  gaugeIntervalMs: 10000
+  sampleRate: 0.01
+- kind: io.l5d.tracelog
+  sampleRate: 0.2
+  level: TRACE
+```
+
+A telemeter may receive stats and trace annotations, i.e. to send to a collector
+or export. Telemetry data can be collected and exported from a linkerd process by
+configuring telemeters via a top-level `telemetry` section. See
+[telemetry](#telemetry14).
 
 ### Tracers
 
--- a/linkerd/examples/statsd.yaml
+++ b/linkerd/examples/statsd.yaml
@@ -1,0 +1,26 @@
+# A fairly minimal linkerd configuration with statsd exporting.
+
+admin:
+  port: 9990
+
+telemetry:
+- kind: io.l5d.commonMetrics
+- kind: io.l5d.statsd
+  experimental: true
+  prefix: linkerd
+  hostname: 127.0.0.1
+  port: 8125
+  gaugeIntervalMs: 10000
+  sampleRate: 1.0 # for testing only, setting this to 1.0 will increase linkerd latency
+
+# An example service discovery backend that uses the filesystem to resolve endpoints.
+namers:
+- kind: io.l5d.fs
+  rootDir: linkerd/examples/io.l5d.fs
+
+routers:
+- protocol: http
+  baseDtab: /http/1.1/* => /#/io.l5d.fs
+  servers:
+  - port: 4140
+    ip: 0.0.0.0
--- a/telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
+++ b/telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
@@ -1,0 +1,62 @@
+package io.buoyant.telemetry
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.timgroup.statsd.NonBlockingStatsDClient
+import com.twitter.finagle.Stack
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.logging.Logger
+import io.buoyant.telemetry.statsd.{StatsDStatsReceiver, StatsDTelemeter}
+
+class StatsDInitializer extends TelemeterInitializer {
+  type Config = StatsDConfig
+  val configClass = classOf[StatsDConfig]
+  override val configId = "io.l5d.statsd"
+}
+
+private[telemetry] object StatsDConfig {
+  val DefaultPrefix = "linkerd"
+  val DefaultHostname = "127.0.0.1"
+  val DefaultPort = 8125
+  val DefaultGaugeIntervalMs = 10000 // for gauges
+  val DefaultSampleRate = 0.01d // for counters and timing/histograms
+
+  val MaxQueueSize = 10000
+}
+
+case class StatsDConfig(
+  prefix: Option[String],
+  hostname: Option[String],
+  port: Option[Int],
+  gaugeIntervalMs: Option[Int],
+  sampleRate: Option[Double]
+) extends TelemeterConfig {
+  import StatsDConfig._
+
+  @JsonIgnore override val experimentalRequired = true
+
+  @JsonIgnore private[this] val log = Logger.get("io.l5d.statsd")
+
+  @JsonIgnore private[this] val statsDPrefix = prefix.getOrElse(DefaultPrefix)
+  @JsonIgnore private[this] val statsDHost = hostname.getOrElse(DefaultHostname)
+  @JsonIgnore private[this] val statsDPort = port.getOrElse(DefaultPort)
+  @JsonIgnore private[this] val statsDInterval = gaugeIntervalMs.getOrElse(DefaultGaugeIntervalMs)
+  @JsonIgnore private[this] val statsDSampleRate = sampleRate.getOrElse(DefaultSampleRate)
+
+  @JsonIgnore
+  def mk(params: Stack.Params): StatsDTelemeter = {
+    // initiate a UDP connection at startup time
+    log.info(s"connecting to StatsD at $statsDHost:$statsDPort as $statsDPrefix")
+    val statsDClient = new NonBlockingStatsDClient(
+      statsDPrefix,
+      statsDHost,
+      statsDPort,
+      MaxQueueSize
+    )
+
+    new StatsDTelemeter(
+      new StatsDStatsReceiver(statsDClient, statsDSampleRate),
+      statsDInterval,
+      DefaultTimer.twitter
+    )
+  }
+}
--- a/telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/Metric.scala
+++ b/telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/Metric.scala
@@ -1,0 +1,24 @@
+package io.buoyant.telemetry.statsd
+
+import com.timgroup.statsd.StatsDClient
+import com.twitter.finagle.stats.{Counter => FCounter, Stat => FStat}
+
+private[statsd] object Metric {
+
+  // stats (timing/histograms) only send when Math.random() <= sampleRate
+  class Counter(statsDClient: StatsDClient, name: String, sampleRate: Double) extends FCounter {
+    def incr(delta: Int): Unit = statsDClient.count(name, delta, sampleRate)
+  }
+
+  // gauges simply evaluate on send
+  class Gauge(statsDClient: StatsDClient, name: String, f: => Float) {
+    def send: Unit = statsDClient.recordGaugeValue(name, f)
+  }
+
+  // stats (timing/histograms) only send when Math.random() <= sampleRate
+  class Stat(statsDClient: StatsDClient, name: String, sampleRate: Double) extends FStat {
+    def add(value: Float): Unit =
+      // would prefer `recordHistogramValue`, but that's an extension, supported by Datadog and InfluxDB
+      statsDClient.recordExecutionTime(name, value.toLong, sampleRate)
+  }
+}
--- a/telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
+++ b/telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
@@ -1,0 +1,48 @@
+package io.buoyant.telemetry.statsd
+
+import java.util.concurrent.ConcurrentHashMap
+import com.timgroup.statsd.StatsDClient
+import com.twitter.finagle.stats.{Counter, Stat, StatsReceiverWithCumulativeGauges}
+import scala.collection.JavaConverters._
+
+private[telemetry] object StatsDStatsReceiver {
+  // from https://github.com/researchgate/diamond-linkerd-collector/
+  private[statsd] def mkName(name: Seq[String]): String = {
+    name.mkString("/")
+      .replaceAll("[^/A-Za-z0-9]", "_")
+      .replace("//", "/")
+      .replace("/", ".") // http://graphite.readthedocs.io/en/latest/feeding-carbon.html#step-1-plan-a-naming-hierarchy
+  }
+}
+
+private[telemetry] class StatsDStatsReceiver(
+  statsDClient: StatsDClient,
+  sampleRate: Double
+) extends StatsReceiverWithCumulativeGauges {
+  import StatsDStatsReceiver._
+
+  val repr: AnyRef = this
+
+  private[statsd] def flush(): Unit = {
+    gauges.values.foreach(_.send)
+  }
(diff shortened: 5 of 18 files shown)


telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
```diff
@@ -0,0 +1,56 @@
...
+  private[this] val log = Logger.get("io.l5d.statsd")
+
+  // prefix format is: "prefix_l5d-uuid-process-id"
+  val statsDPrefix = prefix.getOrElse(DefaultPrefix) + "_" + ProcessId
+  val statsDHost = hostname.getOrElse(DefaultHostname)
+  val statsDPort = port.getOrElse(DefaultPort)
+
+  // initiate a UDP connection at startup time
+  log.info(s"connecting to StatsD at $statsDHost:$statsDPort as $statsDPrefix")
+  private[this] val statsDClient = new NonBlockingStatsDClient(
```
adleong: consider moving this inside the mk method so that the client is created when the telemeter is created instead of when the config is parsed.

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
```diff
@@ -0,0 +1,44 @@
...
+private[telemetry] class StatsDStatsReceiver(statsDClient: StatsDClient)
+  extends StatsReceiverWithCumulativeGauges {
+  import StatsDStatsReceiver._
+
+  val repr: AnyRef = this
+
+  private[statsd] def flushGauges(): Unit = gauges.values.foreach(_.send)
+  private[statsd] def close(): Unit = statsDClient.stop()
+
+  private[this] val gauges = new ConcurrentHashMap[String, Metric.Gauge].asScala
```
adleong: Do you know if the thread-safety of ConcurrentHashMap is preserved if you .asScala it?

siggy: Good point, I'll check.

linkerd/docs/config.md
```diff
@@ -37,6 +37,17 @@ namers:
 tracers:
 - kind: io.l5d.zipkin
   sampleRate: 0.02
+
+telemetry:
+- kind: io.l5d.commonMetrics
+- kind: io.l5d.statsd
```
leozc: should we mark some of the kind" "experimental" 
[linkerd doc example](https://linkerd.io/config/0.8.4/linkerd/index.html#informational-response-headers)

> experimental	false	Set this to true to opt-in to experimental XXX support.

telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
leozc: Also it may be worth to catch StatsDClientException, since this exception may be thrown upon initiation according to 
[NonBlockingStatsDClient.java](https://github.com/tim-group/java-statsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java#L63)

leozc: Suggestion: We Should use the [datadog Statsd client](https://github.com/DataDog/java-dogstatsd-client)?

> A statsd client library implemented in Java. Allows for Java applications to easily communicate with statsd.
> This version was originally forked from java-dogstatsd-client and java-statsd-client but it is now the canonical home for the java-dogstatsd-client. Collaborating with the former upstream projects we have now combined efforts to provide a single release.

[DataDog Statsd Client](https://github.com/DataDog/java-dogstatsd-client)
> https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java

// We can then support Datadog tagging (well we are using Datadog) for free.

```
 private static final StatsDClient statsd = new NonBlockingStatsDClient(
   statsDPrefix,                       /* prefix to any stats; may be null or empty string */
   statsDHost,                        /* common case: localhost */
   statsDPort,                         /* port */
    MaxQueueSize,                   /*queuesize*/
   listofTagFromConfig         // String[]    
            /* Datadog extension: Constant tags, always applied, this can be passed down from config optionally*/
  );
 
```

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
siggy: From 
https://github.com/scala/scala/blob/v2.9.2/src/library/scala/collection/JavaConverters.scala#L330
`The returned Scala <code>Iterator</code> is backed by the provided Java <code>Iterator</code>`

Also, the `ConcurrentHashMap` section of https://twitter.github.io/scala_school/concurrency.html seems to imply thread-safety.

telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
```diff
@@ -0,0 +1,56 @@
...
+  hostname: Option[String],
+  port: Option[Int],
+  gaugePeriodMs: Option[Int]
+) extends TelemeterConfig {
+  import StatsDConfig._
+
+  private[this] val log = Logger.get("io.l5d.statsd")
+
+  // prefix format is: "prefix_l5d-uuid-process-id"
+  val statsDPrefix = prefix.getOrElse(DefaultPrefix) + "_" + ProcessId
```
siggy: That's correct. The UUID is intended to distinguish the same stat coming from different linkerd processes. This can be useful if a single instance of a service is unhealthy. To evaluate the overall health of a service, these stats are intended to be rolled up, ignoring linkerd UUID.

Does this make sense in the StatsD world? In testing a StatsD+Graphite setup, I did not see another way to distinguish stats coming from multiple linkerd processes. As a counter-example, Prometheus identifies metric sources by IP address by default.

siggy:
@leozc This PR is already using the datadog Statsd client? Re: Datadog tagging, let me look into it. My intent was to provide a generic statsd exporter.

@adleong: Load testing showed throughput nearly in half, and latency nearly doubled. I believe this was due to sending network calls on every counter increment and histogram event. For Counters, I've modified them to batch up and send periodically, similar to gauges. For Histograms, we now use a sample rate. Using these new defaults, performance change is negligible.

linkerd/docs/config.md
siggy: @adleong, @olix0r: thoughts on an `expermental` flag for telemeters?

telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
leozc: I would suggest leaving this to user, user can always generate machine specific prefix (using sed for example, per instance).

User probably will encode things like this in prefix, by doing `sed` for the config file.

l5d.${stage}.${service}.${instanceId}

leozc:
@siggy 
Re: This PR is already using the datadog Statsd client
Thank you!

Just FYI

Both InfluxDB and Datadog speaks StatsD, and they BOTH have tagging extension. (the tagging is ideal way to solve the problem we discussed above e.g. host, service etc - they can support operations like group by)

Unfortunately, the wire protocol is slightly different (not much).
We can leave the tag support later.

Possible tasks look like
e.g.
1. Adding new option statsd_backend = [default|datadog|influxDB]
2. Adding new option for tags: String[] which only valid for non-default statsd backend
3. Extend the NonBlockingUDP client to emit tags into corresponding wire format based on different backend

siggy:
That's helpful @leozc! Do you know if the datadog tagging is compatible with the influxDB tagging?

telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
olgert: I agree with @leozc, this shouldn't be included in metric name by default, as it makes initial kick-off more complex: one should explicitly configure statsd/telegraf to skip prefix to be able to aggregate stats from multiple instances, which is (at least to us) the main goal of using statsd for linkerd monitoring. To distinguish between instances it is better to use tags. Another thing to consider: it is [not recommended](https://docs.influxdata.com/influxdb/v1.1/concepts/schema_and_data_layout/#don-t-have-too-many-series) to have tags with unlimited number of possible values or encode this data into measurement name, known as hitting series cardinality limit. This will eventually put InfluxDB instance into extremely slow and memory consuming state. In other words, instance id should be if not disabled at all, then at least not included in measurement name/tags by default.

olgert:
@siggy DataDog StatsD and InfluxDB/Telegraf StatsD are not compatible, and I'd better treat them as a separate dialects. I also like the idea @leozc suggested: 
> Adding new option statsd_backend = [default|datadog|influxDB]
where "default" is etsy's original statsd dialect, e.g. encode tags into measurement name.

siggy:
For my reference and yours, some info on tagging...

## Datadog tagging
We're using `java-dogstatsd-client`, adding Datadog tagging should be fairly straightforward. 

http://docs.datadoghq.com/guides/dogstatsd/#datagram-format
`metric.name:value|type|@sample_rate|#tag1:value,tag2`

## InfluxDB tagging
This format appends the tag to the metric name itself. It should still be possible to implement this using the existing library, just a bit more work.

https://www.influxdata.com/getting-started-with-sending-statsd-metrics-to-telegraf-influxdb/
`user.logins,service=payroll,region=us-west:1|c`

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/Metric.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  // counters batch up deltas and flush on send
+  class Counter(statsDClient: StatsDClient, name: String) extends FCounter {
+    val count = new AtomicInteger(0)
+
+    def incr(delta: Int): Unit = {
+      val _ = count.addAndGet(delta)
+    }
+
+    def send: Unit = statsDClient.count(name, count.getAndSet(0))
```
siggy: fwiw i'm considering rolling this particular change back in favor of simple counter sampling.

initially i liked the idea of getting full-fidelity counters via batching, but counter sampling seems to be the statsd way. it will also will make this code simpler and avoid adding state to linkerd.

olix0r: You should also note that histograms are windowed internally to the metrics registry and so it doesn't make sense to write histograms any more frequently than the window (1m)

siggy: I think we're good wrt histograms? The statsd server handles histogram calculation. On our side we simply implement `Stat.add()` to forward the value to statsd.

leozc:
Tags support should be fairly easy to implement:

Datadog version:
https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java

An influxDB client implemented using NonBlockingStatsDClient:
https://github.com/njawalkar/java-influx-statsd-client

Somehow it is worth for the community to merge them together.

linkerd/examples/acceptance-test.yaml
```diff
@@ -10,6 +10,14 @@ tracers:
   sampleRate: 0.02
 
 telemetry:
+- kind: io.l5d.commonMetrics
```
adleong: per https://github.com/BuoyantIO/linkerd/pull/914 this is going away

linkerd/examples/statsd.yaml
```diff
@@ -0,0 +1,32 @@
...
+  sampleRate: 1.0 # for testing only, setting this to 1.0 will increase linkerd latency
+
+# An example service discovery backend that uses the filesystem to resolve endpoints.
+namers:
+- kind: io.l5d.fs
+  rootDir: linkerd/examples/io.l5d.fs
+
+routers:
+- protocol: http
+  baseDtab: |
```
adleong: this is a pretty old-school dtab

what about just `/http/*/* => /#/io.l5d.fs`?

linkerd/examples/statsd.yaml
```diff
@@ -0,0 +1,32 @@
...
+    /host     => /#/io.l5d.fs;
+    /method   => /$/io.buoyant.http.anyMethodPfx/host;
+    /http/1.1 => /method;
+  httpAccessLog: logs/access.log
```
adleong: I think we can omit config options like `httpAccessLog`, `label`, and `dstPrefix` that aren't really relevant to demonstrating statsd

telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
```diff
@@ -0,0 +1,60 @@
...
+case class StatsDConfig(
+  prefix: Option[String],
+  hostname: Option[String],
+  port: Option[Int],
+  gaugeIntervalMs: Option[Int],
+  sampleRate: Option[Double]
+) extends TelemeterConfig {
+  import StatsDConfig._
+
+  override val experimentalRequired = true
```
adleong: @JsonIgnore

telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  private[this] val log = Logger.get("io.l5d.statsd")
```
adleong: @JsonIgnore

telemetry/statsd/src/main/scala/io/buoyant/telemetry/StatsDInitializer.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  val statsDPrefix = prefix.getOrElse(DefaultPrefix)
+  val statsDHost = hostname.getOrElse(DefaultHostname)
+  val statsDPort = port.getOrElse(DefaultPort)
+  val statsDInterval = gaugeIntervalMs.getOrElse(DefaultGaugeIntervalMs)
+  val statsDSampleRate = sampleRate.getOrElse(DefaultSampleRate)
```
adleong: all-a dees, @JsonIgnore

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
```diff
@@ -0,0 +1,49 @@
+package io.buoyant.telemetry.statsd
+
+import java.util.concurrent.ConcurrentHashMap
+import com.timgroup.statsd.StatsDClient
+import com.twitter.finagle.stats.{Counter, Stat, StatsReceiverWithCumulativeGauges}
+import scala.collection.JavaConverters._
+
+private[telemetry] object StatsDStatsReceiver {
+  // from https://github.com/researchgate/diamond-linkerd-collector/
+  private[statsd] def mkName(name: Seq[String]): String = {
```
adleong: it looks like `:` are replaced by `_` below.  Not sure if this covers all invalid characters.  Maybe a whitelist is better than a blacklist?

telemetry/statsd/src/test/scala/io/buoyant/telemetry/statsd/MetricTest.scala
```diff
@@ -0,0 +1,98 @@
...
+import org.scalatest._
+
+class MetricTest extends FunSuite {
+
+  class MockStatsDClient extends StatsDClient {
+
+    var lastName = ""
+    var lastValue = ""
+
+    def count(x$1: String, x$2: Long, x$3: Double, x$4: String*): Unit = {
```
adleong: were these variable names autogenerated by something?  that's probably fine for the unimplemented methods but for the ones where we provide implementations here it might be helpful to have real variable names.

telemetry/statsd/src/test/scala/io/buoyant/telemetry/statsd/MetricTest.scala
```diff
@@ -0,0 +1,98 @@
+package io.buoyant.telemetry.statsd
+
+import com.timgroup.statsd.StatsDClient
+import org.scalatest._
+
+class MetricTest extends FunSuite {
+
+  class MockStatsDClient extends StatsDClient {
```
adleong: Consider moving this to its own file because its big and boilerplatey and it breaks up the flow.

linkerd/examples/statsd.yaml
olix0r: Eh, you want /http/1.1 if you're routing on host...

telemetry/statsd/src/test/scala/io/buoyant/telemetry/statsd/MetricTest.scala
siggy: yup, i grabbed all of it from an error output. i'll move it into its own file and use real var names for the relevant methods.

leozc:
SUPER looking forward for this PR to release!!

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
```diff
@@ -0,0 +1,48 @@
...
+    name.mkString("/")
+      .replaceAll("[^A-Za-z0-9/]", "_")
+      .replace("//", "/")
+      .replace("/", ".") // http://graphite.readthedocs.io/en/latest/feeding-carbon.html#step-1-plan-a-naming-hierarchy
```
olix0r: eh? `.replaceAll("[^A-Za-z0-9/]", "_")` replaces slashes with `_` and then we replace slashes again.  Should replaceAll come after the first replaces?

telemetry/statsd/src/test/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiverTest.scala
```diff
@@ -0,0 +1,20 @@
+package io.buoyant.telemetry.statsd
+
+import com.timgroup.statsd.NoOpStatsDClient
+import org.scalatest._
+
+class StatsDStatsReceiverTest extends FunSuite {
```
olix0r: Can we test name rewriting?

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
siggy: we don't replace slashes, there's a `/` in that regex. after pushing this change i actually considered writing it as [^/A-Za-z0-9]` to make it more obvious for this very reason. will update.

olix0r: doh gotcha

PR #901, opened 2017-01-07
add XFF header into linkerd downstream and access log

leozc:
A prototype adding XFF header to linkerd
1. passing & updating XFF to downstream 
2. add XFF to access log

TODO: How to get the machine IP for linkerd?

```
      class ClientFilter extends SimpleFilter[Request, Response] {
          val currentHopIP = "0.0.0.0" //FIXME:Where to get my IP?

          if (!req.headers.contains(Key))
            req.headers.set(Key, s"""${currentHopIP}""")
          else
            req.headers.set(Key, s"""${req.headers.get(Key)}, ${currentHopIP}""")

          service(req)
        }
      }
```

Ref: https://en.wikipedia.org/wiki/X-Forwarded-For

### DIFF
--- a/linkerd/examples/static_namer.yaml
+++ b/linkerd/examples/static_namer.yaml
@@ -1,6 +1,7 @@
 # Routes 'default' to a single backend on localhost:8080
 routers:
 - protocol: http
+  httpAccessLog: access.log
   baseDtab: |
     /host/default => /$/inet/127.0.0.1/8080;
     /http => /host/default;
--- a/linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
+++ b/linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
@@ -93,9 +93,10 @@ object LinkerdHeaders {
 
         val deadline = new Deadline.ClientFilter
         val dtab = new Dtab.ClientFilter
+        val xff = new XFFInjector.ClientFilter
 
         def make(next: ServiceFactory[Request, Response]) =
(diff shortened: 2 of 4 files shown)


olix0r:
linkerd already sets the [`Forwarded` header](https://github.com/BuoyantIO/linkerd/blob/master/router/http/src/main/scala/io/buoyant/router/http/ForwardedFilter.scala), which [supersedes use of XFF](https://tools.ietf.org/html/rfc7239). We can still consider how to log this, but it was a conscious decision to use the newer standard.

olix0r:
I should note that that this is disabled by default, per RFC7239:

> Due to the sensitive nature of the data passed in this header field (see Sections 8.2 and 8.3), this header field should be turned off by default.

And it appears that this isn't exposed via configuration -- we'll add an option.

leozc:
1. Yes I agree with "Forwarded" header is more appropriate 
2. How does it interact with XFF based header - will linkerD ever log it or update or convert to "Forwarded"
3. Reason for asking is to ensure LinkerD access log has sufficient data in all cases, so that individual service can skip implementing access.log and rely on LinkerD access log.

olix0r:
Forwarded header configuration support: https://github.com/BuoyantIO/linkerd/pull/905

We should probably augment this to rewrite XFF as Forwarded _when it can_.  Note that, as the RFC points out, requests with both X-Forwarded-By and X-Forwarded-For headers cannot be upgraded, but upgrade; but requests with only, for instance, XFF, should be upgradeable.  We should add that as an additional stack filter as a followup to that review.

esbie:
@olix0r are we still interested in moving this PR forward?

leozc:
We don't need this now

PR #903, opened 2017-01-08
subdomainOfPfx should take into account hostname with port number

### LINKED ISSUE #839: Host name matching in io.buoyant.http.subdomainOfPfx fails when a port is included
Example dtab:
```
/svc => /$/nil
/http/1.1/* => /$/io.buoyant.http.subdomainOfPfx/foo.bar/svc
```
Example request
`/http/1.1/GET/baz.foo.bar:9001`
Attempted rewrite step with failure:
```
/http/1.1/GET/baz.foo.bar:9001 => /$/io.buoyant.http.subdomainOfPfx/foo.bar/svc/baz.foo.bar:9001
```
Since `foo.bar` doesn't match `foo.bar:9001` the subdomain splitting never happens.

fantayeneh:
Fixes #839

### DIFF
--- a/namer/core/src/main/scala/io/buoyant/http.scala
+++ b/namer/core/src/main/scala/io/buoyant/http.scala
@@ -14,9 +14,15 @@ private object Match {
   val host = """^[A-Za-z0-9.:_-]+$""".r
   val method = "[A-Z]+".r
 
+  def dropPort(hostname: String): String = {
+    val idx = hostname.indexOf(":")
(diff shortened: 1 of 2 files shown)


wmorgan:
Thanks, @fantayeneh!

olix0r:
Thanks, @fantayeneh. This looks good to me. I'm going to wait a little on pulling this, as I want to think about #904 at the same time.  But this will make the next release.

PR #905, opened 2017-01-09
Support the RFC 7329 `Forwarded` header

### LINKED ISSUE #902: Support RFC7239 `Forwarded` headers
linkerd has a [`ForwardedFilter`](https://github.com/BuoyantIO/linkerd/blob/master/router/http/src/main/scala/io/buoyant/router/http/ForwardedFilter.scala) that is capable of adding the `Forwarded` header, but it's disabled by default (per the RFC) and is not exposed via configuration.  Furthermore, the implementation doesn't quite meet the requirements of the RFC, especially as pertains to address obfuscation.

olix0r:
Problem

linkerd includes a `ForwardedFilter` but (1) it is disabled and the
configuration exposes no way to enable it, and (2) it is not quite
RFC-compliant.

Solution

Support the `addForwardedHeader` configuration object on HTTP servers. By
default, no `Forwarded` header is added to requests. When enabled, a random
string is used for each of the `by` and `for` endpoints. Users may configure
`by` and `for` to be labeled as: an IP, an IP:PORT, a random string, the parent
router's label, or a static label.

Fixes #902


olix0r:
I have a feeling the documentation may need some work still?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/AddForwardedHeaderConfig.scala
```diff
@@ -0,0 +1,121 @@
...
+ * configuration parameters.
+ */
+object AddForwardedHeaderConfig {
+
+  case class Param(config: Option[AddForwardedHeaderConfig])
+  implicit object Param extends Stack.Param[Param] {
+    val default = Param(None)
+  }
+
+  val module: Stackable[ServiceFactory[Request, Response]] =
```
adleong: I wrote a big long comment about how I didn't understand what this was for, and in the process of writing that I figured it out.  (the values for forwarded for and forwarded by need to be computed here since they potentially depend on linkerd related values such as router label.  those values can then be passed down to the http router filter module that actually appends the headers)

So anyway, this is probably non-obvious enough to need some comments.

linkerd/docs/protocol-http.md
```diff
@@ -61,6 +64,59 @@ concurrent usage of linkerd. Changing these parameters may
...
+    kind: static
+    label: linkerd
+```
+
+Kind | Description
+---- |
+ip | A textual IP address like _192.168.1.1_ or _"[2001:db8:cafe::17]"_.
+ip:port | A textual IP:PORT address like _"192.168.1.1:80"_ or _"[2001:db8:cafe::17]:80"_.
+**random** | Generate an obfuscated random label like __6Oq8jJ_.
+router | Uses the router's `label` as an obfuscated static label.
```
olix0r: No and I had code to restrict this before, but I thought that it was simpler to provide a generic implementation than to have to describe the minor difference.

linkerd/docs/protocol-http.md
```diff
@@ -61,6 +64,59 @@ concurrent usage of linkerd. Changing these parameters may
...
+  by:
+    kind: static
+    label: linkerd
+```
+
+Kind | Description
+---- |
+ip | A textual IP address like _192.168.1.1_ or _"[2001:db8:cafe::17]"_.
+ip:port | A textual IP:PORT address like _"192.168.1.1:80"_ or _"[2001:db8:cafe::17]:80"_.
+**random** | Generate an obfuscated random label like __6Oq8jJ_.
```
olix0r: The spec says that, for the purposes of correlation, a random id can be used for the lifetime of the connection, which is how I implemented this in linkerd. I think it's to be used to show the _existence of_ an intermediary and not any additional diagnostic information. We should probably trace these values (and even store them as records?)

router/http/src/main/scala/io/buoyant/router/http/AddForwardedHeader.scala
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  class LabelingProxy(
+    byLabeler: Labeler.By,
+    forLabeler: Labeler.For,
+    underlying: ServiceFactory[Request, Response]
+  ) extends ServiceFactoryProxy(underlying) {
+
+    override def apply(conn: ClientConnection): Future[Service[Request, Response]] = {
+      // The `by` and `for` labels are computed once per
```
adleong: it's actually the labelers that are computed once per connection, right?  This is what allows ObfuscatedRandom.PerRequest to work.

router/http/src/test/scala/io/buoyant/router/http/AddForwardedHeaderTest.scala
```diff
@@ -0,0 +1,123 @@
...
+    val req = mkReq()
+    req.host = "buoyant.pizza"
+    val rsp = await(service("_cowabunga", "unknown", req))
+    val headers =
+      assert(rsp.contentString == "by=_cowabunga;for=unknown;host=buoyant.pizza")
+  }
+
+  test("supports forwarding proto") {
+    val req = mkReq()
+    req.uri = "https://buoyant.pizza/"
```
adleong: I got really excited and then really disappointed

router/http/src/main/scala/io/buoyant/router/http/AddForwardedHeader.scala
olix0r: correct, will update comment

PR #908, opened 2017-01-09
Add javascript tests, fix MsToStringConverter bug

rmars:
## Changes
* Added test fixtures
* Added tests for the rest of utils.js
* Added tests for routers.js
* Fixed bug in twitter-server MsToStringConverter calculations

## Bug in twitter-server js

I noticed negative days appearing in the MsToStringConverter, for example,
a ms value that I expected to be '183d 0h 0m' was being represented as '-183d 18h 11m'.

Here's an example:

```
// 183 days i.e. 183 * 24 * 60 * 60 * 1000 = 15811200000 ms

>> msToStr.convert(15811200000) // 183 days
>> "-183d 18h 11m"
```

Looking at MsToStringConverter.prototype.convert, rounding appears to be the culprit.
The `years` calculation below gives 1 year (since the `toFixed(0)` will
approximate the 0.5 years as 1 year)... however, 183 days should really be 0 years!

```
var years = (ms/this.msInYear).toFixed(0) // years = 1
```

The subsequent statement then subtracts the years, but since the approximation is
1 year but the actual number of milliseconds is ~0.5 year, we get _negative_ ~0.5 year.

```
var days = Math.floor(((ms - this.msInYear*years)/this.msInDay))

>> // running through that with ms = 183 * 24 * 60 * 60 * 1000 gives:
>> 15811200000 - (365 * dayInMs)
>> -15724800000 // negative days - this shouln't be!
```

If the number of years were 0 instead of the rounded up 1, we'd get the correct answer: 183 days.

Similarly with any fractional year over 0.5 - the rounding up causes too many
milliseconds to be subtracted, resulting in a negative date.

For example, a little less than five years (i.e. 5 * 365 days).
This gives '5y -2d 18h', however it should give '4y 364d 0h'.

```
>> 5 * 365 * 24 * 60 * 60 * 1000
>> 157680000000 // a little less than 5 years

>> msToStr.convert(157680000000)
>> "5y -2d 18h" // unfortunately we get negative days

```


admin/src/main/resources/io/buoyant/admin/js/src/utils.js
```diff
@@ -25,7 +25,17 @@ define([
 
   MsToStringConverter.prototype.convert = function(ms) {
     var years = (ms/this.msInYear).toFixed(0)
```
adleong: What about just `Math.floor(ms/this.msInYear)`?

rmars: doh! yep that works!

admin/src/main/resources/io/buoyant/admin/js/spec/RoutersSpec.js
```diff
@@ -0,0 +1,62 @@
...
+        _.each(multiplierClients, function(client) {
+          expect(client.router).toBe("multiplier");
+          expect(client.color).toBeTruthy(); // nonempty string
+          expect(client.prefix.indexOf("rt/multiplier/dst/id")).toBe(0);
+        });
+      });
+
+      it("returns all clients if router is invalid or unspecified", function() {
+        expect(routers.clients("foobar").length).toEqual(10);
+        expect(routers.clients().length).toEqual(10);
```
esbie: We're doing a lot of length checks here and other places... not a requirement but it'd be nice to provide a custom `toHaveLength` matcher

rmars: Good idea... I've added a simple toHaveLength matcher in this file, but may look at just including something (e.g. https://github.com/just-boris/jasmine-collection-matchers/blob/master/src/toHaveLength.js) instead. (Or write more of our own matchers and also have specs for these matchers).

PR #909, opened 2017-01-10
Add recent request log

adleong:
Add a new page to the linkerd admin dashboard that shows a log of recent requests including their source, destination, and logical and concrete names.
screen shot 2017-01-09 at 4 00 28 pm
To enable this feature, all you have to do is add a requestLog telemeter to your config:
```
telemetry:
- kind: io.l5d.commonMetrics
- kind: io.l5d.requestLog
  sampleRate: 1.0
```


telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogAdminHandler.scala
```diff
@@ -0,0 +1,52 @@
...
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets.ISO_8859_1
+
+class RequestLogAdminHandler(requestLog: RequestLogTracer) extends Handler {
+
+  override def apply(req: Request): Future[Response] = {
+
+    val tbody = requestLog.getRequests.map { requestMeta =>
+
+      val encodedLogicalName = URLEncoder.encode(requestMeta.logicalName, ISO_8859_1.toString)
```
leozc: ISO_8859_1 indicates we can only have single byte char in requestMeta, I think it is fine in general.

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
```diff
@@ -0,0 +1,99 @@
...
+
+class RequestLogTracer(sampleRate: Double, capacity: Int) extends Tracer {
+
+  private[this] val sampler = Sampler(sampleRate.toFloat)
+
+  private[this] val cache = CacheBuilder.newBuilder().maximumSize(capacity).build[SpanId, Seq[Record]]()
+
+  override def record(record: Record): Unit = {
+    val traceId = record.traceId.traceId
+    if (sampler(traceId.toLong) && acceptedAnnotations(record.annotation)) synchronized {
```
leozc: How often is this being called?

Maybe we can use lock free structures?
**ConcurrencyMap** has _putIfAbsent_ and by combining with a concurrent linear structure e.g. **ConcurrentLinkedQueue**, then we should be able to remove the synchronized block?

Something like:
`
cache.putIfAbsent(traceId, new ConcurrentLinkedQueue())
cache.get(traceId).add(record)
`

project/LinkerdBuild.scala
```diff
@@ -142,7 +142,10 @@ object LinkerdBuild extends Base {
       .dependsOn(core, Router.core)
       .withTests()
 
-    val all = aggregateDir("telemetry", core, commonMetrics, tracelog)
+    val requestLog = projectDir("telemetry/request-log")
+      .dependsOn(admin, core, Router.core)
```
olix0r: "tracelog" actually logs traces, whereas this module does not. Also, it's not named consistently with e.g. tracelog.  Any ideas for how to name this to more clearly differentiate it?

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
```diff
@@ -0,0 +1,99 @@
...
+import com.twitter.finagle.tracing._
+import com.twitter.util.Time
+import scala.collection.JavaConverters._
+import com.twitter.finagle.buoyant.Sampler
+
+class RequestLogTracer(sampleRate: Double, capacity: Int) extends Tracer {
+
+  private[this] val sampler = Sampler(sampleRate.toFloat)
+
+  private[this] val cache = CacheBuilder.newBuilder().maximumSize(capacity).build[SpanId, Seq[Record]]()
```
olix0r: prefer to split this up:
```scala
private[this] val cache = CacheBuilder.newBuilder
  .maximumSize(capacity)
  .build[SpanId, Seq[Record]]

File: telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala [hunk 5]
```diff
@@ -0,0 +1,99 @@
...
+    annotations.collectFirst {
+      case Annotation.BinaryAnnotation(n, r) if n == name => r.toString
+    }
+  }
+
+  private[this] def mkRequestMetadata(records: Seq[Record]): Seq[RequestMetadata] = {
+
+    val allServerAddrs = records.flatMap { record =>
+      record.annotation match {
+        case Annotation.ServerAddr(addr) => Some((record.traceId, addr))
```
olix0r: tioli, could be simplified:
```scala
records.collect { case Record(id, _, Annotation.ServerAddr(addr), _) => id -> addr }
```

same applies to other flatMaps below

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
```diff
@@ -0,0 +1,99 @@
...
+
+        for {
+          (sourceId, src) <- clientAddrs
+          (serverId, srv) <- serverAddrs
+          (destinationId, dst) <- allServerAddrs
+          if destinationId.spanId != span && destinationId.parentId == span
+          router <- annotationNamed(annotations, "router.label")
+          logicalName <- annotationNamed(annotations, "namer.path")
+          concreteName <- annotationNamed(annotations, "dst.id")
+        } yield {
```
olix0r: There's a lot of logic in here.  What's going on?

I'm confused that this is doing a nested iteration of three lists (clientAddrs, serverAddrs, and allServerAddrs), and then some filter is done on annotations for the given span? Anyway, this needs english. I don't understand it.

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
```diff
@@ -0,0 +1,99 @@
...
+          val timestamp = recs.map(_.timestamp).minBy(_.inMillis)
+          RequestMetadata(timestamp, src.toString, srv.toString, router, logicalName, concreteName, dst.toString)
+        }
+    }.toSeq
+  }
+
+  case class RequestMetadata(
```
olix0r: Is this defined on the instance?  AFAIU, this makes it hard for anything outside of this class to refer to this type (since it's a path-dependent type). Can this be made static?

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  private[this] val cache = CacheBuilder.newBuilder().maximumSize(capacity).build[SpanId, Seq[Record]]()
+
+  override def record(record: Record): Unit = {
+    val traceId = record.traceId.traceId
+    if (sampler(traceId.toLong) && acceptedAnnotations(record.annotation)) synchronized {
+      cache.getIfPresent(traceId) match {
+        case null => cache.put(traceId, Seq(record))
+        case records => cache.put(traceId, records :+ record)
+      }
```
olix0r: This is racey and can lose records

telemetry/tracelog/src/main/scala/io/buoyant/telemetry/TracelogInitializer.scala
```diff
@@ -34,7 +34,9 @@ class TracelogTelemeter(logger: Logger, level: Level, sampler: Sampler) extends
 class TracelogTracer(logger: Logger, level: Level, sample: Sampler) extends Tracer {
 
   // This tracer doesn't influence downstream tracing.
-  def sampleTrace(id: TraceId): Option[Boolean] = None
+  def sampleTrace(id: TraceId): Option[Boolean] = {
+    None
+  }
```
olix0r: ?

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogAdminHandler.scala
```diff
@@ -0,0 +1,52 @@
+package io.buoyant.telemetry.requestLog
+
+import com.twitter.finagle.http.{MediaType, Request, Response, Status}
+import com.twitter.util.Future
+import io.buoyant.admin.Admin.Handler
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets.ISO_8859_1
+
+class RequestLogAdminHandler(requestLog: RequestLogTracer) extends Handler {
```
olix0r: Thinking aloud, it'd be particularly nice if this could return a JSON representation. Not a blocker.

rmars: +1 to this, would be nice to not nave to generate the table here

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
olix0r: Can this be simplified as:

```scala
  val reqs =
        for {
          (destinationId, dst) <- allServerAddrs
          if destinationId.spanId != span && destinationId.parentId == span
          router <- annotationNamed(annotations, "router.label")
          logicalName <- annotationNamed(annotations, "namer.path")
          concreteName <- annotationNamed(annotations, "dst.id")
        } yield {
          for {
            (sourceId, src) <- clientAddrs
          } yield {
            val clientAddrs = ...
            val serverAddrs = ...
            val timestamp = ...
            clientAddrs.flatMap { ca =>
              serverAddrs.map { sa =>
                RequestMetadata(...)
              }
            }
          }
        }
  reqs.flatten
```

Am i correct that this is equivalent? If so, this seems to be preferable, since the top-level loop should do a lot less work?

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
adleong: Good suggestions, thanks!

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminFilter.scala
```diff
@@ -0,0 +1,18 @@
+package io.buoyant.linkerd.admin
+
+import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.finagle.http.{MediaType, Request, Response}
+import com.twitter.util.Future
+
+object AdminFilter extends SimpleFilter[Request, Response] {
```
adleong: Added scaladoc.  Lmk if you think this needs more explanation beyond that.

project/LinkerdBuild.scala
adleong: How about recentRequests?

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
adleong: Slightly refactored and added comments.

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
adleong: Good call.  Moved to top level.

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogTracer.scala
adleong: Fixed per @leozc's suggestion above.

telemetry/request-log/src/main/scala/io/buoyant/telemetry/requestLog/RequestLogAdminHandler.scala
adleong: Gonna punt on this for now.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -53,12 +53,22 @@ object LinkerdAdmin {
...
   )
 
-  def apply(lc: Linker.LinkerConfig, linker: Linker): Admin.Handlers =
+  def apply(lc: Linker.LinkerConfig, linker: Linker): Admin.Handlers = {
+    val extHandlers = Admin.extractHandlers(
+      linker.namers ++
+        linker.routers ++
+        linker.telemeters
+    ).map {
+        case (path, handler) => (path, AdminFilter.andThen(handler))
```
olix0r: nit: weird indentation

bashofmann:
It would also be great to add the time the request took to the log

adleong:
@bashofmann great suggestion!  I'll take a crack at that in a follow up.

Issue #911, opened 2017-01-10
Add an option to ConsulNamer to support stripping domains from service names


akreiling:
We would like to use the following pattern to facilitate service discovery for our application developers when migrating to our new linkerd mesh. I propose adding an option to `ConsulNamer` to strip extraneous domain information during the Host to Consul service lookup. This would be advantageous by allowing our customers to start using services backed by linkerd via configuration changes without forcing any code/client changes to support the necessary header manipulation.

Here's the idea:

1. Create a wildcard domain record for *.service.local that points to linkerd. (Or one could use dnsmasq trickery)
2. Clients issue HTTP requests to `servicename.service.local`
3. Linkerd's ConsulNamer resolves `servicename.service.local` to the Consul service `servicename`
4. Everybody is happy.

olix0r:
Hi @akreiling! Thanks for filing the issue.

I think this can be accomplished without code change by layering the [`/$/io.buoyant.http.subdomainOfPfx`](https://linkerd.io/config/head/linkerd/index.html#rewriting-namers) namer, for instance:
```
/consul => /#/io.l5d.consul/dc1 ;
/host => /$/io.buoyant.http.subdomainOfPfx/service.local/consul ;
/http/1.1/* => /host ;
```

I think this should Just Work for the situation you describe.

akreiling:
thanks -- this works like a charm.
-a

PR #912, opened 2017-01-10
Add `stripDomain` parameter for Consul CatalogNamer to optionally strip domains from Consul service name lookups

akreiling:
Proposed code changes to support #911

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -163,6 +163,7 @@ namers:
   includeTag: true
   useHealthCheck: true
   setHost: true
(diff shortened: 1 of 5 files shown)


wmorgan:
Thanks, Andrew!

akreiling:
I will apologize in advance that I my tests need some help. After opening this, I realized they aren't really testing how/what I wanted. Any advice is welcome.

akreiling:
turns out that this is already supported via `io.buoyant.http.subdomainOfPfx`

thanks!
-a

PR #913, opened 2017-01-10
Introduce io.buoyant.hostportPfx and io.buoyant.porthostPfx namers

### LINKED ISSUE #904: Add a simple host:port namer with no lookup
Could I request a namer than maps host:port directly without any external service lookup.

For example,
- bind `/http/1.1/GET/10.2.2.56/16992` to `10.2.2.56:16992`
- bind `/http/1.1/GET/10.2.3.43/17335` to `10.2.3.43:17335`
etc

The reason for this is so that calling apps can do service lookup themselves and leave linkerd to just proxy the URL.

olix0r:
It's common to have a host string like `web:4434` that needs to be split into
its constituent `host` and `port`.

This change introduces two new auto-loaded namers to support these rewrites.

Fixes #904


olix0r:
RFC: now that I've done this, i'm considering moving these to be configured namers so that we don't have to do the weird prefix-suffix dance (as done by all Pfx namers)

namer/core/src/main/scala/io/buoyant/hostport.scala
```diff
@@ -0,0 +1,73 @@
+package io.buoyant
+
+import com.twitter.finagle.{Name, NameTree, Namer, Path}
+import com.twitter.util.{Activity, Try}
+
+/**
+ * A helper
+ */
+trait RewritingNamer extends Namer {
```
adleong: I don't think you meant to move this here.

namer/core/src/main/scala/io/buoyant/hostport.scala
```diff
@@ -0,0 +1,73 @@
...
+          case None => NameTree.Neg
+        }
+      case _ => NameTree.Neg
+    })
+}
+
+private object HostColonPort {
+  // do very coarse host matching so that we can support dns names,
+  // ipv4, and ipv6 without going crazy.
+  val hostRE = """^[A-Za-z0-9.:_-]+$""".r
```
adleong: unused?

namer/core/src/test/scala/io/buoyant/HostPortTest.scala
```diff
@@ -0,0 +1,30 @@
+package io.buoyant
+
+import com.twitter.finagle._
+import io.buoyant.test.FunSuite
+
+class HostPortTest extends FunSuite {
```
adleong: think its worth adding a test for ipv6 addresses?

olix0r: IPv6 isn't going to work... host:port notation for ipv6 is e.g. `[2001:db8:cafe::17]:8080` and braces are not legal in Paths...

namer/core/src/main/scala/io/buoyant/hostport.scala
olix0r: i did mean to move this here (into the `io.buoyant` package)

namer/core/src/main/scala/io/buoyant/hostport.scala
olix0r: :+1:

namer/core/src/main/scala/io/buoyant/hostport.scala
adleong: ah, I see.  Pretty confusing that it's in hostport.scala though.  How about giving it its own file?

namer/core/src/main/scala/io/buoyant/hostport.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  object PortStr {
+    def unapply(s: String): Option[Int] = {
+      val port = try s.toInt catch { case _: java.lang.NumberFormatException => -1 }
+      if (0 < port && port < math.pow(2, 16)) Some(port) else None
+    }
+  }
+
+  def unapply(s: String): Option[(String, Int)] = s.split(":") match {
```
klingerf: I don't think this `.split` will play well with ipv6 addresses. Mind adding a test?

klingerf: Ah, now I see the other comments on this branch... carry on.

namer/core/src/main/scala/io/buoyant/hostport.scala
```diff
@@ -0,0 +1,73 @@
...
+    case Path.Utf8(pfx, HostColonPort(host, port)) =>
+      Some(Path.Utf8(pfx, host, port.toString) ++ path.drop(2))
+    case _ => None
+  }
+}
+
+/**
+ * A rewriting namer that accepts names in the form:
+ *
+ *   /host:port/pfx/etc
```
adleong: I think the pfx is supposed to be before the host:port segment.

olix0r: Yeah, you're right.  I think I'm leaning towards making these explicitly configured with prefixes, since I find the pfx-namers to be a bit awkward -- though these unconfigured namers are certainly more flexible.  Thoughts?

adleong: This is somewhat similar to other pfx-namers so consistency suggests making this a pfx-namer as well.  But I agree they're somewhat awkward and having both /$ namers and /# namers is perhaps unnecessary complexity.  I wouldn't be opposed to converting all of our /$ namers into configured namers in the future (perhaps in 0.9.0)

PR #917, opened 2017-01-11
[h2] Configure client-side protocol upgrade via TLS

### LINKED ISSUE #760: h2 request fails with REFUSED_STREAM error
I'm trying to configure linkerd to talk to http2.golang.org, which serves HTTP/2 requests with TLS.

I can curl the site (using prior knowledge and disabling certificate validation):

```
$ curl -s --http2 --http2-prior-knowledge -k https://http2.golang.org:443/reqinfo
Method: GET
Protocol: HTTP/2.0
Host: http2.golang.org
...
```

But when I run linkerd with a similar config:

```
routers:
- protocol: h2
  experimental: true
  baseDtab: |
    /h2 => /$/inet/http2.golang.org/443;
  servers:
  - port: 4142
  client:
    tls:
      kind: io.l5d.noValidation
```

Then the curl command hangs:

```
$ curl -s --http2 --http2-prior-knowledge -v http://localhost:4142/reqinfo
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 4142 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 4142 (#0)
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7f882200f000)
> GET /reqinfo HTTP/1.1
> Host: localhost:4142
> User-Agent: curl/7.50.3
> Accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
```

In the linkerd logs, I see:

```
E 1019 19:12:14.938 THREAD78: demuxer
com.twitter.finagle.ChannelClosedException: ChannelException at remote address: http2.golang.org/130.211.116.44:443. Remote Info: Not Available
```

For context, here's the whole log with netty logging enabled:

```
INFO 1019 12:14:43.238 finagle/netty4-3: 
----------------OUTBOUND--------------------
[id: 0xf1580aa6, L:/127.0.0.1:4142 - R:/127.0.0.1:54205] SETTINGS: ack=false, settings={}
------------------------------------
INFO 1019 12:14:43.238 finagle/netty4-3: 
----------------INBOUND--------------------
[id: 0xf1580aa6, L:/127.0.0.1:4142 - R:/127.0.0.1:54205] SETTINGS: ack=false, settings={}
------------------------------------
INFO 1019 12:14:43.238 finagle/netty4-3: 
----------------OUTBOUND--------------------
[id: 0xf1580aa6, L:/127.0.0.1:4142 - R:/127.0.0.1:54205] SETTINGS: ack=true
------------------------------------
INFO 1019 12:14:43.239 finagle/netty4-3: 
----------------INBOUND--------------------
[id: 0xf1580aa6, L:/127.0.0.1:4142 - R:/127.0.0.1:54205] HEADERS: streamId=1, headers=DefaultHttp2Headers[:method: GET, :path: /reqinfo, :scheme: http, :authority: localhost:4142, user-agent: curl/7.50.3, accept: */*], padding=0, endStream=true
------------------------------------
1019 19:14:43.249 fa963b7cb402f926.fa963b7cb402f926<:fa963b7cb402f926] Message(namer.success)
INFO 1019 12:14:43.259 finagle/netty4-3: 
----------------INBOUND--------------------
[id: 0xf1580aa6, L:/127.0.0.1:4142 - R:/127.0.0.1:54205] SETTINGS: ack=true
------------------------------------
INFO 1019 12:14:43.332 finagle/netty4-4: 
----------------OUTBOUND--------------------
[id: 0x0fbea0d1, L:/192.168.1.7:54206 - R:http2.golang.org/130.211.116.44:443] SETTINGS: ack=false, settings={}
------------------------------------
DEBUG 1019 12:14:43.426 finagle/netty4-4: Accepting a server certificate: CN=http2.golang.org, O=Google Inc, L=Mountain View, ST=California, C=US
INFO 1019 12:14:43.491 finagle/netty4-4: 
----------------OUTBOUND--------------------
[id: 0x0fbea0d1, L:/192.168.1.7:54206 - R:http2.golang.org/130.211.116.44:443] HEADERS: streamId=3, headers=DefaultHttp2Headers[:method: GET, :path: /reqinfo, :scheme: http, :authority: localhost:4142, user-agent: curl/7.50.3, accept: */*, l5d-dst-logical: /h2/localhost:4142, via: h2 linkerd, l5d-dst-concrete: /$/inet/http2.golang.org/443, l5d-dst-residual: /localhost:4142], streamDependency=0, weight=16, exclusive=false, padding=0, endStream=true
------------------------------------
DEBUG 1019 12:14:43.493 finagle/netty4-4: [id: 0x0fbea0d1, L:/192.168.1.7:54206 - R:http2.golang.org/130.211.116.44:443] HANDSHAKEN: TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
INFO 1019 12:14:43.554 finagle/netty4-4: 
----------------OUTBOUND--------------------
[id: 0x0fbea0d1, L:/192.168.1.7:54206 - R:http2.golang.org/130.211.116.44:443] GO_AWAY: lastStreamId=0, errorCode=1, length=77, bytes=4669727374207265636569766564206672616d6520776173206e6f742053455454494e47532e204865782064756d7020666f7220666972737420352062797465...
------------------------------------
DEBUG 1019 12:14:43.555 finagle/netty4-4: [id: 0x0fbea0d1, L:/192.168.1.7:54206 - R:http2.golang.org/130.211.116.44:443] Sending GOAWAY failed: lastStreamId '0', errorCode '1', debugData 'First received frame was not SETTINGS. Hex dump for first 5 bytes: 485454502f'. Forcing shutdown of the connection.
javax.net.ssl.SSLException: SSLEngine closed already
    at io.netty.handler.ssl.SslHandler.wrap(...)(Unknown Source)
E 1019 19:14:43.561 THREAD86: demuxer
com.twitter.finagle.ChannelClosedException: ChannelException at remote address: http2.golang.org/130.211.116.44:443. Remote Info: Not Available
```

olix0r:
Problem

Finagle doesn't expose any way for clients to configure ALPN protocol
negotiation. This breaks HTTP/2 clients.

Solution

Introduce new TLS configuration params to be used throughout linkerd: TlsClientPrep.TransportSecurity and TlsClientPrep.Trust.

Now, linkerd TlsClientInitializers are responsible for providing a Stack module that configures the appropriate TlsClientPrep modules on the underlying stack. This module is installed at the bottom of the client stack so it has full access to endpoint parameters.  Below this module, the TlsClientPrep.configureFinagleTls reads TlsClientPrep params and generates finagle-specific params (e.g. `Transport.Tls`).

H2 clients replace the configureFinagleTls with one that disables finagle TLS entirely.  Instead, the Netty4H2Transporter installs a netty channel handler that uses the TlsClientPrep params with Netty4ClientTls.ApplicationProtocols to configure TLS handlers on the netty pipeline.

Automated TLS tests have been run and pass.  Additionally, a bunch of manual testing was performed.

Once Finagle fixes their TLS support, we'll be able to replace the configureFinagleTls module without altering TlsClientInitializers.

Fixes #760


olix0r:
This change is insufficient. linkerd client TLS configurations do not us the netty-compatible config types (due to ca cert path configuration).  TlsConfig is sort of a mess...

olix0r:
This is going to be trickier than it seemed at first:

The initial change works only with the io.l5d.noValidation.  This is because the io.l5d.noValidation happily uses a netty ssl context under the hood.

The other client validators, as constructed by `TlsClientPrep`, use a java SSLContext built with a trust chain from a file. ALPN relies on netty's SslContext, but the rest of linkerd depends on java's SSLContext.

In order to change TlsClientPrep to not use java's SSLContext, we need to replace the client tls initialization for _all of linkerd_ (and not just h2).

Our options are basically:
1. Overhaul all client TLS in linkerd to support an alternative means of configuring trust managers for TLS.  We're going to have to do this eventually when @ryanoneill [moves Finagle onto new APIs](https://github.com/twitter/finagle/issues/580#issuecomment-272006634).
2. Don't use the default TlsClientPrep module system for h2 client configurations.  We could introduce a new config type that configures h2 differently, but this will require a bit of a refactor in linkerd-core to change TlsClientPrep to stop exposing java's SSLEngine, etc.

olix0r:
After some discussion, I think the short-term plan is the following:

* Modify this branch to _fail to load h2 configs_ when unsupported TLS configurations are configured
* h2 will only support io.l5d.noValidation for now
* we'll wait to see what's going to land in the next finagle release before we figure out whether we can move onto the new finagle types or if we have to solve this ourselves (to remove _experimental_ from h2).

olix0r:
As soon as I left the office yesterday, I realized there's a simpler solution (which I've implemented and will push momentarily):
* Move `TlsClientPrep` to a new finagle/core project, usable from finagle/h2.
* `TlsClientPrep` now exposes _two_ stack roles:
  * `TlsClientPrep.role` is responsible for configuring the `TlsClientPrep.TlsEnabled` and  `TlsClientPrep.Trust` params.
  * `TlsClientPrep.role.finagle` is responsible for consuming these params and configuring finagle params on the underlying stack (i.e. `Transport.Tls`, etc).
    * An alternative implementation of the `Tls.ClientPrep.role.finagle` module can be used (or it can be elided entirely) to support ALPN.
    * When finagle changes the TLS config api, we should be able to update this module only without changing TlsClientPrep module implementations.
* Modify `TlsClientInitializer` to expose `tlsClientPrep[Req, Rsp]: Stackable[ServiceFactory[Req, Rsp]]`, which is installed into the `TlsClientPrep.role` stack slot

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientPrep.scala
```diff
@@ -0,0 +1,168 @@
...
+  sealed trait Trust
+  implicit object Trust extends Stack.Param[Trust] {
+    case class Verified(name: String, certs: Seq[X509Certificate]) extends Trust
+    object UnsafeNotVerified extends Trust
+    object NotConfigured extends Trust
+
+    val default = NotConfigured
+  }
+
+  /**
```
adleong: should this scaladoc be on TlsTrustModule?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientPrep.scala
```diff
@@ -0,0 +1,168 @@
...
+      val parameters = Nil
+      def make(params0: Stack.Params, next: Stk[Req, Rsp]) = {
+        val tlsParams = params0 +
+          Transport.Tls(TlsConfig.Disabled) +
+          Transport.TLSClientEngine(None)
+        Stack.Leaf(role, next.make(tlsParams))
+      }
+    }
+
+  def finagleModule[Req, Rsp]: Stkable[Req, Rsp] =
```
adleong: This module could use some scaladoc, I think.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientPrep.scala
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+  /** A module that always disables TLS. */
+  def disable[Req, Rsp]: Stkable[Req, Rsp] =
+    new TlsTrustModule[Req, Rsp] {
+      override val tlsEnabled = TlsEnabled(false)
+      override val trust = Trust.NotConfigured
+    }
+
+  def finagleDisabled[Req, Rsp]: Stkable[Req, Rsp] =
```
adleong: Consider moving this below finagleModule and add a comment indicating that this is an alternative to finagleModule that disables the finagle tls params.

linkerd/tls/src/main/scala/io/buoyant/linkerd/clientTls/BoundPathInitializer.scala
```diff
@@ -21,45 +19,52 @@ class BoundPathInitializer extends TlsClientInitializer {
...
+        val n = names.map(_.prefix).mkString(",")
+        throw new MatcherError(s"Unable to match ${path.show} with available names: $n")
+      case default => default
+    }
 
-      def newEngine(cn: Option[String]): Option[(SocketAddress) => Engine] =
-        cn.map(TlsClientPrep.addrEngine(_, caCertPath))
+  @JsonIgnore
+  override def tlsClientPrep[Req, Rsp]: Stackable[ServiceFactory[Req, Rsp]] =
+    new Stack.Module[ServiceFactory[Req, Rsp]] {
```
adleong: could this extend TlsTrustModule?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/H2.scala
```diff
@@ -18,7 +19,14 @@ object H2 extends Client[Request, Response] with Server[Request, Response] {
   private[this]type Http2FrameTransport = Transport[Http2Frame, Http2Frame]
 
   object Client {
-    val newStack: Stack[ServiceFactory[Request, Response]] = StackClient.newStack
+    val newStackWithoutTlsClientPrep: Stack[ServiceFactory[Request, Response]] =
+      StackClient.newStack[Request, Response]
+
+    val newStack: Stack[ServiceFactory[Request, Response]] = {
```
adleong: is this used?

linkerd/tls/src/main/scala/io/buoyant/linkerd/clientTls/BoundPathInitializer.scala
olix0r: it needs to read params so it has to implement make itself

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/H2.scala
olix0r: newStack is used in finagle...H2.Client(), newStackWithoutTlsClientPrep is used in router.H2.clientStack

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientPrep.scala
```diff
@@ -0,0 +1,203 @@
...
+   * be used at all.
+   */
+  case class TransportSecurity(config: TransportSecurity.Config)
+  implicit object TransportSecurity extends Stack.Param[TransportSecurity] {
+    sealed trait Config
+
+    /** TODO support Protocols, Ciphers, & timeouts */
+    case class Secure() extends Config
+
+    object Insecure extends Config
```
olix0r: Note: I moved TlsEnabled to TransportSecurity, so this can later include tls protocol & cipher configuration.

I moved to wrapper types (rather than making the param an ADT) because it simplifies type inference for params (also this tends to be what we see in finagle)

PR #923, opened 2017-01-12
Ensure that a trace id is set on incoming thrift requests

adleong:
# Problem

If no trace id is set on an incoming thrift request, no trace id is set into the context until the client stack.  This means that records on the server span will have a completely different traceId than the records on the client span.

# Solution

Add a server module that generates a traceId if one does not exist yet.

# Result

For requests that otherwise don't have trace data, a traceId will be generated for the server span and the spanId will be incremented for the clientSpan.

### DIFF
--- a/linkerd/protocol/thrift/src/main/scala/com/twitter/finagle/buoyant/linkerd/ThriftTraceInitializer.scala
+++ b/linkerd/protocol/thrift/src/main/scala/com/twitter/finagle/buoyant/linkerd/ThriftTraceInitializer.scala
@@ -1,0 +1,31 @@
+package com.twitter.finagle.buoyant.linkerd
+
+import com.twitter.finagle._
+import com.twitter.finagle.tracing.{Trace, TraceInitializerFilter, Tracer}
+
+object ThriftTraceInitializer {
+  val role = TraceInitializerFilter.role
+
+  val serverModule: Stackable[ServiceFactory[Array[Byte], Array[Byte]]] =
+    new Stack.Module1[param.Tracer, ServiceFactory[Array[Byte], Array[Byte]]] {
+      val role = ThriftTraceInitializer.role
+      val description = "Ensure that there is a trace id set"
+
+      def make(_tracer: param.Tracer, next: ServiceFactory[Array[Byte], Array[Byte]]) = {
+        val param.Tracer(tracer) = _tracer
+        new ServerFilter(tracer) andThen next
+      }
+    }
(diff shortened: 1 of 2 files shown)


linkerd/protocol/thrift/src/main/scala/com/twitter/finagle/buoyant/linkerd/ThriftTraceInitializer.scala
```diff
@@ -0,0 +1,31 @@
...
+        new ServerFilter(tracer) andThen next
+      }
+    }
+
+  class ServerFilter(tracer: Tracer)
+    extends SimpleFilter[Array[Byte], Array[Byte]] {
+
+    def apply(req: Array[Byte], service: Service[Array[Byte], Array[Byte]]) = {
+      if (!Trace.hasId)
+        Trace.letTracerAndNextId(tracer) {
```
klingerf: Not positive, but I think it's possible that you need to always call `Trace.letTracerAndNextId` here, regardless of whether or not `Trace.hasId`.  Otherwise, you won't advance the span ID until the client span is created, and the server span will share its span ID with the caller's client span.

adleong: My thinking is that if the trace context is loaded from the request, it's at that point that the span ID should be advanced.  If that has already happened by this point we don't want to advance the span ID again.  If it hasn't happened yet then the trace ID generated here will simply be overwritten by the trace loaded from the request.

klingerf: Ok, that makes sense to me too. I guess the question is moot for thrift anyway, since there is no context and no possible way that `Trace.hasId` would be true at this point in the server stack.

adleong: Well, until TTwitter is being used.  But otherwise, yeah.

adleong:
If my understanding is correct, the default `TraceInitializerFilter.role` is slightly broken in the case where you have a server that accepts a protocol that does not propagate trace data.  In this situation a trace id is not generated until the client stack which means that annotations on the server stack don't have a trace id set and use the default trace id.  The result is that the client span is not a child of the server span.

Issue #505, opened 2016-06-24
Address review comments on #499 and #501


rmars:
- Ticket to myself to track addressing of esbie's review comments on #499 and #501
  - [x] #499 - fixed in 51d5636b
  - [x] #501 - esbie fixed this in #959
- [x] Also, deleting the metrics browser/cleaning up the metrics handler might be nice (kl fixed this in #924)
- [x] Consider consolidating dashboard.css and dashboard-shared.css, since they're both now in everything (fixed in 51d5636b)

rmars:
closing this, issues have been addressed

PR #924, opened 2017-01-12
Sort clients and servers on admin dashboard

### LINKED ISSUE #915: admin dashboard: sort servers and clients alphabetically
I think this is a relatively small change, but it would be great to sort the routers and clients that show up on the admin dashboard alphabetically, so that they load in a consistent order each time. Right now I think it's just hash ordering. Here's an example:

screen shot 2017-01-10 at 2 19 18 pm

klingerf:
Sorts clients and servers on the admin dashboard alphabetically by their label. The values are pulled from a map, and without sorting they show up in arbitrary order. After the page loads, as new clients are added they are also inserted alphabetically into the page, and the coloring of the existing clients is redrawn accordingly.

As part of this branch, I'm also sorting routers in the order that they're defined in the config file, and I'm removing the old metrics handler and all of its assets, since that page is no longer accessible in the admin dashboard.

Fixes #915.


rmars:
Looks awesome, thanks for doing this.

I think we'll also need to add something in `router_clients.js` - when new clients show up they'll just be appended to the dom... then the clients won't appear in sorted order.  

So we'd need to clear and reappend all the `client-container`s. This could be expensive if there's a lot of clearing and reappending going on so maybe that's not worth it, since you could always refresh to get a sorted page. (Though maybe we don't want people to have to do this).


example:
add a bunch of clients
screen shot 2017-01-11 at 5 13 45 pm

need to refresh to resort:
screen shot 2017-01-11 at 5 14 39 pm

rmars:
I think we'll also need to change `assignColorsToClients` in `router-clients.js`(my fault for making it order-dependent) - the reordering is causing color mapping issues, were more than 1 client is assigned the same color, even though we have more colors available:

screen shot 2017-01-11 at 5 21 04 pm

adleong:
@rmars is is possible to insert the new client into the correct position in the dom instead of clearing and re-appending?

rmars:
yeah! good point! we can use $.insertBefore and the like to just do inserts

klingerf:
@rmars Thanks for the suggestion! I added client sort-on-update in 7504903c. I also removed a few things that appeared to be unused, but please have a look to make sure I didn't break anything.

adleong:
Does this mean that when new clients are added, existing clients might change color?

klingerf:
@adleong That's correct.

admin/src/main/resources/io/buoyant/admin/js/src/logger-overrides.js
```diff
@@ -12,6 +12,7 @@ $(function() {
   '        <div id="navbar" class="collapse navbar-collapse">' +
   '          <ul class="nav navbar-nav">' +
   '            <li><a href="../delegator">dtab</a></li>' +
+  '            <li><a href="../requests">requests</a></li>' +
```
adleong: ahh, good catch.  thanks for fixing!

admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
```diff
@@ -13,12 +13,13 @@ define([
...
 
     function assignColorsToClients(colors, clients) {
-      var colorIdx = 0;
+      var clientMapping = {};
 
-      return _.reduce(clients, function(clientMapping, client) {
-        clientMapping[client.label] = colors[colorIdx++ % colors.length];
-        return clientMapping;
-      }, {});
+      _(clients).sortBy('label').forEach(function(client, idx) {
```
siggy: there's a `return _.sortBy(clients, 'label');` on line 204 of `routers.js`. is this sorting twice?

admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
```diff
@@ -70,15 +79,15 @@ define([
           routerClient.updateColors(clientToColor);
         });
 
-        var expandClients = shouldExpandClients(addedClients.length + routerClients.length);
+        expandClients = shouldExpandClients(routers.clients(routerName).length);
```
siggy: nit: i see `routers.clients(routerName)` repeated (and also assignned to `clients`), consider DRY up.

admin/src/main/resources/io/buoyant/admin/js/src/logger-overrides.js
siggy: consider a comment here and/or in `AdminHandler.scala` that the nav's should stay in sync.

rmars: (I'm working on that right now as part of #505!)

klingerf:
@adleong Ok, per discussion, have updated the behavior to append new clients and preserve the color of existing clients.

adleong:
👏 👏 👏  thank you, @klingerf!  Sorry for the runaround

PR #925, opened 2017-01-12
Fix path identifier bug when slash precedes uri params

### LINKED ISSUE #910: trailing slash appended after param when using io.l5d.path identifier
When a slash precedes ?foo=bar param another slash is appended as trailing slash to the url.
Example: /foo/bar/?foo=bar is forwarded as /foo/bar/?foo=bar/
This causes problems with parsing the params by upstream service.

adleong:
Fixes #910

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
@@ -24,7 +24,7 @@ case class PathIdentifier(
           req.uri = req.uri.split("/").drop(segments + 1) match {
             case Array() => "/"
             case x =>
-              val trailingSlash = if (req.path.endsWith("/")) "/" else ""
+              val trailingSlash = if (req.uri.endsWith("/")) "/" else ""
               x.mkString("/", "/", trailingSlash)
           }
         }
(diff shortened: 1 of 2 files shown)


router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -24,7 +24,7 @@ case class PathIdentifier(
           req.uri = req.uri.split("/").drop(segments + 1) match {
             case Array() => "/"
             case x =>
-              val trailingSlash = if (req.path.endsWith("/")) "/" else ""
```
leozc: +1

router/http/src/test/scala/io/buoyant/router/http/PathIdentifierTest.scala
```diff
@@ -71,7 +71,8 @@ class PathIdentifierTest extends FunSuite with Awaits with Exceptions {
       "/mysvc/subsvc/path1/" -> ExpParsedPath("/http/mysvc/subsvc", "/path1/"),
       "/mysvc/subsvc/path1" -> ExpParsedPath("/http/mysvc/subsvc", "/path1"),
       "/mysvc/subsvc/path1/path2" -> ExpParsedPath("/http/mysvc/subsvc", "/path1/path2"),
-      "/mysvc/subsvc/path1/path2/" -> ExpParsedPath("/http/mysvc/subsvc", "/path1/path2/")
+      "/mysvc/subsvc/path1/path2/" -> ExpParsedPath("/http/mysvc/subsvc", "/path1/path2/"),
+      "/mysvc/subsvc/path1/?foo=bar" -> ExpParsedPath("/http/mysvc/subsvc", "/path1/?foo=bar")
```
leozc: +1

olix0r: Would you mind adding a test like `/mysvc/subsvc/path1?foo=bar/ -> /path1?foo=bar/`?

PR #926, opened 2017-01-12
Define kind on config objects, set kind visible, pass config in param stack

esbie:
Problem:
I'm writing a telemeter plugin that reads from `LinkerConfig`, but can't access config `kind` properties defined in `JsonTypeInfo`

Solution:
Expose the kind property


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpIdentifierConfig.scala
```diff
@@ -1,12 +1,16 @@
...
+import com.fasterxml.jackson.annotation.{JsonIgnore, JsonProperty, JsonTypeInfo}
 import com.twitter.finagle.http.Request
 import com.twitter.finagle.{Dtab, Path}
 import io.buoyant.router.RoutingFactory.Identifier
 
-@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "kind")
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "kind", visible = true)
 trait HttpIdentifierConfig {
+
+  @JsonProperty("kind")
```
adleong: TIOLI: could create a `Config` trait that defines this var and have each Config class extend it.  That would reduce the boilerplate a bit.

esbie: oh good idea, I'll try that out

config/src/main/scala/io/buoyant/config/Config.scala
```diff
@@ -0,0 +1,9 @@
+package io.buoyant.config
+
+import com.fasterxml.jackson.annotation.{JsonProperty, JsonTypeInfo}
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "kind", visible = true)
```
adleong: Whoa!  Are annotations inherited?  If this works, that's pretty awesome.

There are many more config classes (e.g. RouterConfig and many others) and we should probably be consistent about having them all inherit this.

esbie: OK no problem, I will hunt them down

olix0r: Do all Config types actually have a kind?  clients and servers, for instance, do not.  Perhaps call this PolymorphicConfig or something that indicates when it needs to be used?  It should have a comment that clarifies what it's for, what needs to extend it, etc.

esbie:
Alright, this is ready to go on my side. It's changed a fair bit from the outset, so a quick re-review would be appreciated.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -63,8 +63,8 @@ object LinkerdAdmin {
         linker.routers ++
         linker.telemeters
     ).map {
-      case (path, handler) => (path, AdminFilter.andThen(handler))
-    }
+        case (path, handler) => (path, AdminFilter.andThen(handler))
```
adleong: damn you, scalariform

config/src/main/scala/io/buoyant/config/PolymorphicConfig.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.fasterxml.jackson.annotation.{JsonProperty, JsonTypeInfo}
+
+/**
+ * A trait that defines the property "kind" as both a var on the object
+ *   and as type information for JSON de/serialization.
+ *
+ * Config objects that expect to be subclassed by multiple other configs
+ *   with different "kind" values should extend this trait.
+ */
+@JsonTypeInfo(
```
adleong: I'm actually surprised this works because as far as I can tell, annotations are not inherited.  But it does seem to work and is a great reduction in boilerplate.  Must be some jackson magic going on.  🚀

esbie: This confuses me also  ¯\\\_(ツ)\_/¯

PR #931, opened 2017-01-13
UsageDataTelemeter

esbie:
Adds telemeter package to `io.buoyant.linkerd`
Adds UsageDataTelemeter
Adds `usage.proto` and `Grpc.usage` project


linkerd/docs/telemetry.md
```diff
@@ -103,3 +103,30 @@ Key        | Default Value | Description
...
+```yaml
+telemetry:
+ - kind: io.l5d.commonMetrics
+ - kind: io.l5d.usage
+```
+
+kind `io.l5d.usage`
+
+In order to make improvements and prioritize features, we'd like to gain a
+broader picture of how users run linkerd. This telemeter helps us by gathering anonymized usage data,
```
adleong: It might be nice to explicitly call out certain data that will NOT be collected (labels, dtabs, service addresses, request data of any kind, etc.)

linkerd/docs/telemetry.md
```diff
@@ -103,3 +103,30 @@ Key        | Default Value | Description
...
+sent to buoyant once an hour. The kind of data captured is as follows:
+
+1. How linkerds are configured (The kinds of namers, initializers, identifiers, transformers, protocols, & interpreters used)
+2. What environments are linkerds running in (Operating System, Container orchestration solution),
+3. How do linkerds perform in those contexts (JVM performance, Number of requests served)
+
+See proto for exact message and structure.
```
adleong: Perhaps add a note saying that the exact data that will be transmitted can be viewed on the admin site.

linkerd/docs/telemetry.md
```diff
@@ -103,3 +103,30 @@ Key        | Default Value | Description
...
+
+
+Key | Default Value | Description
+--- | ------------- | -----------
+orgId | empty by default | Optional unique string of your choosing that helps buoyant better understand linkerd behavior across restarts.
```
adleong: This should be a single identifier used for the whole company/organization, right?  Maybe something like "Optional string of your choosing that identifies your organization".

linkerd/docs/telemetry.md
```diff
@@ -103,3 +103,30 @@ Key        | Default Value | Description
...
+
+```yaml
+telemetry:
+ - kind: io.l5d.commonMetrics
+ - kind: io.l5d.usage
+```
+
+kind `io.l5d.usage`
+
+In order to make improvements and prioritize features, we'd like to gain a
```
adleong: Not sure if this is the right place but somewhere we should probably stress that this is opt-in and no data will be transmitted until this telemeter is added to the config.

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+    }
+  }
+
+  def mkUsageMessage(
+    config: Linker.LinkerConfig,
+    orgId: Option[String],
+    metrics: Map[String, Number]
+  ): UsageMessage =
+    UsageMessage(
+      pid = Some(java.util.UUID.randomUUID().toString),
```
adleong: I think the calculation of this pid should be moved up to UsageDataTelemeter so that it's not recalculated for each message.

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+  def mkNamers(config: Linker.LinkerConfig): Seq[String] =
+    config.namers.getOrElse(Seq()).map(_.kind)
+
+  def mkGauges(metrics: Map[String, Number]): Seq[Gauge] =
+    Seq(
+      Gauge(Some("jvm_mem"), metrics.get("jvm/mem/current/used").map(_.doubleValue())),
+      Gauge(Some("jvm/gc/msec"), metrics.get("jvm/mem/current/used").map(_.doubleValue()))
+    )
+
+  val RequestPattern = s"""^*/srv/*/requests$$""".r
```
adleong: I think you're missing character classes before the kleenes.
`s"""^.*/srv/.*/requests$$""".r` maybe?

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+  def mkCounters(metrics: Map[String, Number]): Seq[Counter] =
+    metrics.filter({
+      case (RequestPattern(key), value) => true
```
adleong: I think this extractor pulls out capture groups from the regex, but there aren't any.

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  def mkGauges(metrics: Map[String, Number]): Seq[Gauge] =
+    Seq(
+      Gauge(Some("jvm_mem"), metrics.get("jvm/mem/current/used").map(_.doubleValue())),
+      Gauge(Some("jvm/gc/msec"), metrics.get("jvm/mem/current/used").map(_.doubleValue()))
+    )
+
+  val RequestPattern = s"""^*/srv/*/requests$$""".r
+  def mkCounters(metrics: Map[String, Number]): Seq[Counter] =
+    metrics.filter({
```
adleong: more concise style 
```
metrics.collect {
  case (RequestPattern(key), v) => Counter(...)
}
```

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+      case (RequestPattern(key), value) => true
+      case _ => false
+    }).map({
+      case (k, v: Number) => Counter(Some("srv_requests"), Some(v.longValue()))
+    }).toSeq
+
+  def mkRouters(config: Linker.LinkerConfig): Seq[Router] =
+    config.routers.map(r => {
```
adleong: style nit:
```
config.routers.map { r =>
  ...
}

File: linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala [hunk 10]
```diff
@@ -0,0 +1,201 @@
...
+    service: Service[Request, Response],
+    config: Linker.LinkerConfig,
+    orgId: Option[String],
+    registry: Metrics
+  ) extends Admin.Handler {
+
+    def apply(request: Request): Future[Response] = {
+      val msg: UsageMessage = mkUsageMessage(config, orgId, registry.sample().asScala.toMap)
+      val rsp = Response()
+      rsp.contentType = "application/json"
```
adleong: You can use MediaType.json

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+    }
+  }
+
+  val mapper = new ObjectMapper(new JsonFactory) with ScalaObjectMapper
+  mapper.registerModule(DefaultScalaModule)
+  mapper.setSerializationInclusion(Include.NON_ABSENT)
+  mapper.setVisibility(PropertyAccessor.ALL, Visibility.PUBLIC_ONLY)
+}
+
+class UsageDataTelemeter(
```
adleong: some scaladoc would be nice on this

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+  val stats = NullStatsReceiver
+
+  private[this] val metricsService = Http.client.newService(metricsDst, "usageData")
+  private[this] val started = new AtomicBoolean(false)
+
+  val adminHandlers: Admin.Handlers = Seq(
+    "/admin/metrics/usage" -> new UsageDataHandler(metricsService, config, orgId, registry)
+  )
+
+  def sample(registry: Metrics) = registry.sample().asScala.toMap
```
adleong: explicit return type on public members, please

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+    new Closable with CloseAwaitably {
+      def close(deadline: Time) = closeAwaitably {
+        closer.close(deadline)
+      }
+    }
+  }
+}
+
+case class UsageDataTelemeterConfig(
+  orgId: Option[String]
```
adleong: can we also include a "dryRun" parameter that will let users see what data we would send but without actually sending anything

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+) extends TelemeterConfig {
+
+  @JsonIgnore
+  def mk(params: Stack.Params): UsageDataTelemeter = {
+    val config = params[LinkerConfigStackParam] match {
```
adleong: I think you can just do `params[LinkerConfigStackParam].config`

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -0,0 +1,201 @@
...
+      case LinkerConfigStackParam(c) => c
+      case e => LinkerConfig(None, Seq(), None, None, None)
+    }
+
+    val DefaultPath = "/$/inet/130.211.174.30/80"
+
+    new UsageDataTelemeter(
+      Name(DefaultPath),
```
adleong: might be more idiomatic to do: `Name.bound(Address("130.211.174.30", 80))`

linkerd/telemeter/usage/src/test/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeterTest.scala
```diff
@@ -0,0 +1,29 @@
+package io.buoyant.linkerd.telemeter
+
+import com.twitter.finagle.util.LoadService
+import io.buoyant.config.Parser
+import io.buoyant.linkerd.{TracerConfig, TracerInitializer}
+import io.buoyant.telemetry.{TelemeterConfig, TelemeterInitializer}
+import org.scalatest.FunSuite
+
+class UsageDataTelemeterTest extends FunSuite {
```
adleong: I think UsageDataTelemeter could use an e2e test where you spin up a Service and receive data from it.

adleong:
Could you also add an entry to CHANGES.md for this?

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
esbie: oh good catch

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
olix0r: Does this even need string interpolation?  Isn't this just `"""^.*/srv/.*/requests$""".r`?

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
esbie: I like this idea! I'm worried that someone might infer dryRun to mean "representative of any performance degradation while using the telemeter", which wouldn't be the case if we don't actually make the calls. What do you think, am I being paranoid here?

olix0r: Is there anything specific we're worried about exposing?

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
olix0r: also typical to do:
```scala
val LinkerConfigStackParam(config) = params[LinkerConfig]
```

tioli: Also, fwiw, we typically don't have _StackParam_ in param names. When we want to clarify this, we usually do so by sticking it in a `param` package/object so it's referenced like `param.LinkerConfig`

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
esbie: No, only trying to be as transparent as possible about what's being sent. Currently in order to use `/admin/metrics/usage` to see what's being sent, the telemeter needs to be loaded. But the telemeter also sends an initial payload on load. If we think docs + protobuf file sets enough context then this may not be an issue.

adleong: I have no attachment to the name `dryRun` if we can think of a clearer name.  Given that usage data is sent so infrequently, it shouldn't have any real performance impact.

esbie: @siggy's load test agrees that performance impact is negligible, so let's stick with dryRun

esbie:
I haven't yet figured out a clever place on the admin site to put the payload, so I'll add something in a follow up branch

PR #932, opened 2017-01-17
Introduce io.buoyant.grpc.runtime.EventStream

olix0r:
Problem

It's common to model a Var/Activity over an RPC stream. It's awkward to do this
with the AsyncQueue-backed Stream/Provider because the future-oriented
back-pressure model does not interact well with the Event.respond API.

Solution

Provide a new Stream type, EventStream, that maintains only the most
recently observed Event value. Stream receivers may miss intermediate states,
but the stream should generally behave better when the consumer is slow.


grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/EventStream.scala
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Event values are wrapped by EventStream.Ev to indicate whether a
+ * value is terminal.
+ *
+ * TODO streams are not properly failed on event failure.
+ *
+ * TODO when a stream fails (from the remote), we should eagerly
+ * stop observing the event.
+ */
+class EventStream[+T](events: Event[EventStream.Ev[T]]) extends Stream[T] {
```
adleong: Just wondering aloud if we should name this in a way to indicate it does not reflect all event states.  LossyEventStream?  EventuallyConsistentEventStream?

PR #933, opened 2017-01-17
Rewrite Namer

### LINKED ISSUE #843: Rewriting Namer
Add a namer that takes an existing name and rewrites it according to a template.  This is inspired by the way that the boundPath client TLS plugin works (https://linkerd.io/config/0.8.3/linkerd/index.html#tls-with-bound-path)

```
kind: io.l5d.rewrite
pattern: /http/*/{method}/{host}
name: /s/{host}/{method}
```

The above config will accept names like `/#/io.l5d.rewrite/http/1.1/GET/foo` and resolve them as `/s/foo/GET`.

1. If the name after the `/#/io.l5d.rewrite` prefix does not match the pattern, resolve to Neg.
2. Otherwise, capture segments of the name into template variables (`GET` into `{method}` and `foo` info `{host}` in the above example).
3. Fill template variables in the `name` field and resolve it.

adleong:
Fixes #843 

A namer that completely rewrites a path.  This is useful to do arbitrary reordering of the path segments that is not possible using standard prefix replacement.  

If the name matches the pattern in the config, it will be replaced by the name in the config.  Additionally, any variables in the pattern will capture the value of the matching path segment and may be used in the final name.


namer/core/src/main/scala/io/buoyant/namer/RewritingNamerInitializer.scala
```diff
@@ -0,0 +1,24 @@
...
+
+case class RewritingNamerConfig(
+  pattern: String,
+  name: String
+) extends NamerConfig {
+  assert(pattern != null)
+  assert(name != null)
+
+  @JsonIgnore
+  override def defaultPrefix: Path = Path.read("/io.l5d.rewrite")
```
olix0r: I sort of feel that we should require explicit prefixes for this namer?  Since it seems awkward to ever have `/#/io.l5d.rewrite` be meaningful...

namer/core/src/main/scala/io/buoyant/namer/RewritingNamerInitializer.scala
```diff
@@ -0,0 +1,24 @@
...
+  val configClass = classOf[RewritingNamerConfig]
+}
+
+case class RewritingNamerConfig(
+  pattern: String,
+  name: String
+) extends NamerConfig {
+  assert(pattern != null)
+  assert(name != null)
+
```
olix0r: Should we mark this as experimental? (could go either way)

adleong: Even though we're not 100% on how we feel about this namer, it should be safe for production use from a performance point of view.  I would lean towards not marking it as experimental.

PR #948, opened 2017-01-18
make admin work on arbitrary base paths

### LINKED ISSUE #397: configurable admin url
it may be useful to configure admin to serve on a different url. for example, on DC/OS, admin can run at `dcos-url.com/service/linkerd`. today that page is broken because admin expects assets to be available under root, not under `/service/linkerd`.

this is similar to grafana's `root_url` flag: http://docs.grafana.org/installation/configuration/#root_url

siggy:
fixes #397 

this change makes every url/ajax/href/src in admin relative.

initially i had planned to add a new linkerd config flag specifying the base path. this was problematic for two reasons:
1) the base path would need to be passed from config -> admin -> templates -> js
2) it would prevent linkerd from running with multiple base paths. primary example is DC/OS runs admin on `/service/linkerd/`, but the user can still also access any linkerd in their cluster by browsing directly to the admin port. if `/service/linkerd/` were coded into the config, this would not be possible.

linkerd-dcos

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ project/boot/
 project/plugins/project/
 project/plugins/src_managed/
 http.pid
+npm-debug.log
 .sbt-launch.jar
 .idea
 .iml
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * Add experimental StatsD telemeter
 * Add a recent requests log in the linkerd admin dashboard
+* Admin dashboard now works if served at a non-root url
 
 ## 0.8.5 2017-01-06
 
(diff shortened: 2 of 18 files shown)


adleong:
If I browse to http://localhost:4140/foo (as opposed to http://localhost:4140/foo/ with the trailing slash) I get broken resource links.

siggy:
@adleong that's expected, we require the trailing the slash. this makes the root consistent with all other paths: `/files...`, `/admin...`, etc.

adleong:
That's unfortunate, it would be nice to be able to accept either.  I wonder if in javascript land we can force the trailing slash somehow (edit the window location???)?

That said, this is a strictly better than the current state of affairs.  🚀

PR #951, opened 2017-01-19
Change the internal Delegator API to accept a NameTree

olix0r:
Problem

The Delegator API accepts a `DelegatorTree[Name.Path]`, but this isn't really appropriate: nothing has been delegated yet! This is awkward when constructing an initial DelelgatorTree for requests.

Solution

To fix this, the Delegator API is modified to accept a `NameTree[Name.Path]`, as is this is actually the input we want.  The Thrift and HTTP APIs remain unchanged, though they only permit simple NameTree[Path] names now (which I think is actually all that was really supported before).

Note: NameTree[Path] is accepted so that users may evaluate expressions like `/s/foo | /s/bar`.

### DIFF
--- a/namer/core/src/main/scala/io/buoyant/namer/DelegateTree.scala
+++ b/namer/core/src/main/scala/io/buoyant/namer/DelegateTree.scala
@@ -130,20 +130,20 @@ object DelegateTree {
         DelegateTree.Union(path, dentry, delegates: _*)
     }
 
-  def fromNameTree[T](names: NameTree[T]): DelegateTree[T] =
+  def fromNameTree[T](path: Path, names: NameTree[T]): DelegateTree[T] =
     names match {
-      case NameTree.Empty => DelegateTree.Empty(null, Dentry.nop)
-      case NameTree.Fail => DelegateTree.Fail(null, Dentry.nop)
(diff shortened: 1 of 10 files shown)


namer/core/src/main/scala/io/buoyant/namer/DelegateTree.scala
```diff
@@ -130,20 +130,20 @@ object DelegateTree {
         DelegateTree.Union(path, dentry, delegates: _*)
     }
 
-  def fromNameTree[T](names: NameTree[T]): DelegateTree[T] =
+  def fromNameTree[T](path: Path, names: NameTree[T]): DelegateTree[T] =
```
adleong: This is a pretty unfortunate area of the code.  Any time that you're converting a NameTree to a DelegateTree, the path property is meaningless.  We could set it to Path.empty all the time but DelegateTrees are actually used to render the namerd json API (uggggh) and we need to set path to null to make sure it doesn't show up in the json output.  

What about changing the method signature to:
`def fromNameTree[T](names: NameTree[T], path: Path = Path.empty): DelegateTree[T]`

and most callers can leave path unspecified.  Only the json api handlers need to explicitly pass null in.

PR #954, opened 2017-01-20
verify marathon app state with useHealthCheck

### LINKED ISSUE #941: Marathon namer should return only addresses of tasks in RUNNING state
Marathon namer currently returns IP:port of all tasks that exists for a specific application. 
https://github.com/BuoyantIO/linkerd/blob/master/marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
` def getAddrs(app: Path): Future[Set[Address]] = {
    **val req = http.Request(s"$apiPrefix/apps${app.show}?embed=app.tasks")**
    Trace.letClear(client(req)).flatMap(rspToAddrs(_, useHealthCheck))
  }`

This means that during a rolling upgrade of application, the namer returns bind addresses of STAGING and KILLED tasks as well as RUNNING tasks. Which makes this operation a non-zero downtime operation.

I think that only RUNNING tasks should be returned. Combined with marathon upgrade strategy*, we can reach a zero downtime upgrade without using more sophisticated processes like B/G upgrade where it's not needed.

* `"upgradeStrategy": {
    "minimumHealthCapacity": 1,
    "maximumOverCapacity": 0.5
  },`

siggy:
currently useHealthCheck only looks at a marathon app's
`healthCheckResults` field. this change additionally checks the `state`
field for `TASK_RUNNING`

fixes #941

### DIFF
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
@@ -20,6 +20,7 @@ trait Api {
 
 object Api {
   private[this] val versionString = "v2"
+  private[this] val taskRunning = "TASK_RUNNING"
   private[this] val mapper = new ObjectMapper with ScalaObjectMapper
 
   mapper.registerModule(DefaultScalaModule)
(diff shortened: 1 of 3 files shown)


marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -91,8 +93,8 @@ object Api {
 
   object healthyHostPort {
     def unapply(task: Task): Option[(String, Int)] = {
-      task.healthCheckResults.map(isHealthy) match {
-        case Some(true) => hostPort.unapply(task)
+      isHealthy(task) match {
```
adleong: style: consider if-else instead of match for a simple boolean.

PR #957, opened 2017-01-20
[grpc] Framing-agnostic stream decoding

### LINKED ISSUE #953: gRPC message framing coupled to H2 framing
My first implementation of gRPC message framing was, um, a bit... let's go with _expedient_:
```scala
            case Return(data: h2.Frame.Data) =>
              // TODO framing/buffering
              val msg = codec.decodeBuf(Codec.decodeGrpcFrame(data.buf))
              stream.send(msg).before(data.release()).before(loop())
```

Basically io.buoyant.grpc can only read grpc frames that comprise an entire h2 frame.  This causes issues when an intermediary reframes H2 messages (i.e. combining/splitting grpc frames into/across h2 frames) or when a single grpc message can't fit in a single h2 frame (i.e. larger than 16K). 

This is complicated by the way we implemented flow control in the h2 library:
```scala
/**
 * A single item in a Stream.
 */
sealed trait Frame {

  /**
   * When `isEnd` is true, no further events will be returned on the
   * Stream.
   */
  def isEnd: Boolean

  def onRelease: Future[Unit]
  def release(): Future[Unit]
}
```

We don't currently support the ability to _partially_ release an h2 frame, which is necessary to support a frame containing more than one message. Similarly, when composing a grpc message over multiple h2 frames, an earlier frame shouldn't be released until the entire grpc message is released.

We need to do some sort of offset accounting to figure out when an h2 frame may be released. I'm unsure of whether this belongs as a part of the h2 API or whether this can be layered on i.e. in the gRPC stream codec.

I'm going to try the latter approach first.

olix0r:
Problem

When grpc endpoints receive a stream, the transport assumes each HTTP2 frame has exactly one gRPC message. Transport framing needs to be decoupled entirely from message encoding.

Solution

Introduce the grpc.runtime.DecodingStream[T] type, which implements grpc.runtime.Stream[T], backed by an h2.Stream and a Codec[T]. Accounting is done to ensure that h2 frames are released when all gRPC messages encoded in that frame have been released.

Fixes #953


olix0r:
still need to add some explicit tests for the conditions we're concerned about

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+ * the stream.
+ */
+private[runtime] trait DecodingStream[+T] extends Stream[T] {
+  import DecodingStream._
+
+  protected[this] def frames: h2.Stream
+  protected[this] def decoder: ByteBuffer => T
+
+  // convenience
+  private[this]type Releasable = Stream.Releasable[T]
```
adleong: nit: space before type

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  def apply[T](
+    rsp: h2.Response,
+    decodeF: ByteBuffer => T
+  ): Stream[T] = new DecodingStream[T] {
+    protected[this] val frames: h2.Stream = rsp.stream
+    protected[this] def decoder: ByteBuffer => T = decodeF
+  }
+
+  private sealed trait RecvState
```
adleong: This sealed trait only has one subtype.  Are we planning to add more RecvState states?  Or should we just rename Buffer to RecvState and ditch the sealed trait?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+        } else copy(remaining = remaining - n)
+
+      override def releasable(): (Releaser, Func) =
+        if (remaining == 0) {
+          val (rest, releaseRest) = next.releasable()
+          val doRelease = () => release().before(releaseRest())
+          rest -> doRelease
+        } else {
+          // This frame isn't yet releasable, but we require that this
+          // sub-slice is released before releasing the frame:
```
adleong: I don't understand this part.  If this frame isn't yet releasable, we return a Func that, when invoked, allows this Frame to be released?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  private[this] case class Header(compressed: Boolean, size: Int)
+
+  private[this] val GrpcFrameHeaderSz = Codec.GrpcFrameHeaderSz
+
+  private val hasHeader: ByteBuffer => Boolean =
+    bb => { GrpcFrameHeaderSz <= bb.remaining }
+
+  private val hasMessage: (ByteBuffer, Int) => Boolean =
+    (bb, msgsz) => { msgsz <= bb.remaining }
```
adleong: `msgSz` might be clearer

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+
+  private def decodeHeader(bb0: ByteBuffer): Header = {
+    require(hasHeader(bb0))
+    val bb = bb0.duplicate()
```
adleong: can we grab a slice of the original bb?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  private def decode[T](
+    s: RecvState,
+    decoder: ByteBuffer => T
+  ): (RecvState, Option[Stream.Releasable[T]]) =
+    s match {
+      case RecvState.Buffer(Some(h0), buf, releaser) =>
+        val Buf.ByteBuffer.Owned(bb0) = Buf.ByteBuffer.coerce(buf)
+        decodeMessage(h0, bb0.duplicate(), releaser, decoder)
```
adleong: can we grab a slice of size `h0.size` instead of duplicating the buffer?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+      val (nextReleaser, release) = releaser.consume(h0.size).releasable()
+      val msg = {
+        val msgbb = bb.duplicate()
+        msgbb.limit(end)
+        val msg = decoder(msgbb)
+        Some(Stream.Releasable(msg, release))
+      }
+      bb.position(end)
+      if (hasHeader(bb)) {
+        // And another header buffered...
```
adleong: Could there be multiple messages per frame?  Should this recurse?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
adleong: Oh, I get it.  The doRelease func returned here is associated with the previous message.  That message must be released before we can release this frame.  This stuff is hard to get my head around.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
adleong: I know duplicate doesn't copy the bytes (I presume) but maybe the slice api makes managing limits easier or make the code more readable?  Not sure.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/Stream.scala
```diff
@@ -48,4 +47,53 @@ object Stream {
 
   object Closed extends Throwable
   object Rejected extends Throwable
+
+  /**
+   * A stream backed by a promised stream.
+   */
+  def deferred[T](streamFut: Future[Stream[T]]): Stream[T] =
+    new Stream[T] {
+      // What happens to a dream deferred?
```
adleong: s/dream/stream

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,288 @@
...
+      }
+      updateF.flatMap(_updateBuffer)
+    }
+
+  /**
+   * Update the receive buffer before returning the result
+   * (and i.e. releasing the mutex).
+   */
+  private[this] val _updateBuffer: ((RecvState, Try[Releasable])) => Future[Releasable] = {
+    case (rb, v) =>
```
adleong: what does rb stand for?  Receive buffer?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: this is an extremely annoying emacsism, i'll try to fix

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: that should be rs (RecvState used to be called RecvBuffer)

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: Follow-up change includes a RecvState.Reset

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: Yep... If you have any ideas about how to make this clearer, I'm all ears. This was the least-awful thing I could come up with.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: AIUI, the main difference with slice is that it resets the start of the buffer so that `bb.position == 0` and that it sets the _capacity_ to the _limit_ of the buffer.  I don't think buys us much but I'll take a look.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: There may be multiple messages in the frame, yes.  We could introduce a new RecvState that can hold additional fully-parsed messages, but I think it simpler to just read the next header so that on the next call to recv(), the next message will be read and returned.

olix0r: we could also not even read the next header and just update the state to `RecvState.Buffer(None, ...)`

adleong: Ah, yeah.  I don't see the benefit of reading the next message's header in advance.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: ACTUALLY this is a scalariformism so we're stuck with it for now

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,326 @@
...
+          val releaseSegment = () => { p.setDone(); Future.Unit }
+          rest -> releaseSegment
+        } else {
+          // This frame has been entirely consumed, so return a
+          // release function that releases this frame (after all
+          // prior segments have been released).
+          val (rest, releaseRest) = tail.releasable()
+          val releaseFrame = () => release().before(releaseRest())
+          rest -> releaseFrame
+        }
```
olix0r: This should be changed.  Instead of blocking the final message's release future on prior futures, it should complete immediately -- but the last message to be released of a given frame should cause the flush...

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,382 @@
...
+       * released and, when all have been released, calls `underlying`
+       * to release the frame.
+       */
+      trait Latch {
+        def segmentReleaseFunc(): Func
+        def lastSegmentReleaseFunc(): Func
+      }
+
+      object Latch {
+        private[this] trait State
```
adleong: should this be sealed?

adleong: This trait and it's implementations are unused?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,382 @@
...
+
+    private[this] object FrameReleaser {
+
+      /**
+       * Tracks how many segments of a single frame need to be
+       * released and, when all have been released, calls `underlying`
+       * to release the frame.
+       */
+      trait Latch {
+        def segmentReleaseFunc(): Func
```
adleong: both these methods need docs

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,382 @@
...
+        private[this] case class Releasable(pending: Int) extends State {
+          require(pending >= 0)
+          def decrement = copy(pending = pending - 1)
+        }
+
+        private[this] class Segmented(underlying: Func) extends Latch {
+          @volatile private[this] var canReleaseUnderlying = false
+          @volatile private[this] var pending = 0
+
+          private[this] val releaseSegment: Func = { () =>
```
adleong: If I'm following this correctly, I think this can be broken into it's two codepaths and put directly into `segmentReleaseFunc` and `lastSegmentReleaseFunc`.  I think the `canReleaseUnderyling` variable is unnecessary

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -0,0 +1,382 @@
...
+              pending -= 1
+              if (canReleaseUnderlying && pending == 0) underlying()
+              else Future.Unit
+            }
+          }
+
+          def segmentReleaseFunc(): Func = synchronized {
+            if (canReleaseUnderlying) throw new IllegalStateException("segment after last")
+            pending += 1
+            releaseSegment
```
adleong:
```
{ () => synchronized { 
  if (pending < 1) throw ...
  pending -= 1
  Future.Unit
} }`

File: grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala [hunk 15]
```diff
@@ -0,0 +1,382 @@
...
+          }
+
+          def lastSegmentReleaseFunc(): Func = synchronized {
+            if (canReleaseUnderlying) throw new IllegalStateException("segment after last")
+            canReleaseUnderlying = true
+            pending += 1
+            releaseSegment
```
adleong:
```
{ () => synchronized { 
  if (pending < 1) throw ...
  pending -= 1
  if (pending == 0 ) underlying() else Future.Unit
} }
```

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
olix0r: I don't think that's right. (Should document this):

Let's say that we have an h2 frame containing two messages. If we parse and release the first message before we parse the second message, we should not call underlying.  We can _only_ call underlying after we have created the releasable for all segments in the frame.

Furthermore, these segments may be released out of order.  `release()` of the second message should not be blocked on release of the first message.  Instead, release of the second message should happen instantaneously and release of the first message should release the underlying frame.

adleong: Ok, I think I get it now. It started making more sense when I started renaming `canReleaseUnderlying` to `allSegmentsConsumed` in my head.  I keep thinking that there's some subtle race condition here but I thiiiiiiiiiiiiiiiiink the synchronization protects you from everything I can think of.

PR #959, opened 2017-01-21
Render admin logging page on the client

esbie:
Problem: We struggle to keep style and navbar overrides for the logging page up to date with the rest of the admin pages, resulting in mismatched UI. Also the logging UX is.... quirky, to put it nicely.

Solution: Render the page on the client like we do for the other admin pages.

screen shot 2017-01-20 at 3 26 34 pm


linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
```diff
@@ -74,6 +74,7 @@ object AdminHandler extends HtmlView {
           <title>linkerd admin</title>
```
rmars: in this file, `<li><a href="admin/logging">logging</a></li>` on line 34 needs to be updated

esbie:
Thanks for helping me find the "root" of that issue, @rmars! It works a lot better now.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
klingerf: You can also remove the comment on lines 11-14 in this file. 🎉

PR #965, opened 2017-01-23
grpc-interop server, grpc-finagle codegen bugfixes

stevej:
* Adds an implementation of the official interop service for testing.
* Adds Stream.reset method.
* Fixes grpc-finagle codegen bug when reserved keywords are in protos.


olix0r:
merged in #1024

Issue #967, opened 2017-01-24
Add "use node address" option to consul namer


markeijsermans:
A service definition in consul can have an optional explicit IP address defined along side the host's node address. Currently linkerd does the right thing and defaults to using the service address if defined, and node address otherwise. In some situations it is needed to use the node address regardless.

As and example, we have an in-house routing solution which is similar the "linker to linker" model. 
We use the node address to route the request to the host, then use the service address to route to the required container locally.

Proposal:  add `useNodeAddress` option to `io.l5d.consul` namer. Defaults to `false` for current behavior.  

Code in question:
https://github.com/linkerd/linkerd/blob/0ab3f9f8cc6293893803f55085724b32a81295e6/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala#L116

olix0r:
This sounds reasonable to me. nitpick: are we instructing linkerd to use the node address? Or are we instructing it _not_ to use the service address?  It might be clearer to describe this as `ignoreServiceAddress` or something along those lines.

markeijsermans:
I agree, `ignoreServiceAddress` is clearer with it's intention

PR #968, opened 2017-01-24
Added ignoreServiceAddress option to consul namer 

markeijsermans:
* Added `ignoreServiceAddress` option to consul namer to always get the node address of a service. 
* default  `ignoreServiceAddress: false`
* updated docs, changelog
* updated some tests/comments for other options that were out of sync.
fixes: #967


linkerd/docs/namer.md
```diff
@@ -186,6 +186,7 @@ token | no authentication | The auth token to use when making API calls.
 setHost | `false` | If `true`, HTTP requests resolved by Consul will have their Host header overwritten to `${serviceName}.service.${datacenter}.${domain}`. `$domain` is fetched from Consul.
 consistencyMode | `default` | Select between [Consul API consistency modes](https://www.consul.io/docs/agent/http.html) such as `default`, `stale` and `consistent`.
 failFast | `false` | If `false`, disable fail fast and failure accrual for Consul client. Keep it `false` when using a local agent but change it to `true` when talking directly to an HA Consul API.
+ignoreServiceAddress | `false` | If `false` use the service address if defined and default to the node address. If `true` always use the node address.
```
adleong: I generally prefer boolean options to be stated positively instead of negatively.  What do you think about calling this `useNodeAddress` or reversing it (default to true) and calling it `preferServiceAddress`?

adleong:
@Ashald I'd be interested in what you think of this

linkerd/docs/namer.md
olix0r: If you look at the issue, I suggested this.  `useNodeAddress` is undesirable, since it _always_ may use the node address.  What this flag does is disables use of the service address.

adleong: Thoughts on `preferServiceAddress`?

olix0r: 👍

Ashald: With default to `true` I assume?

markeijsermans:
@Ashald yes, it was added to solve a situation where we have a "linker to linker" type of setup with our existing routing mesh based on consul. #967 explains it a bit better.

Option is now renamed `preferServiceAddress` and defaults to `true`.

Issue #937, opened 2017-01-17
Add namerd client info to admin dashboard


adleong:
If linkerd is configured with the namerd interpreter, add a namerd tab to the admin dashboard with information about the namerd client.  

Potential information to display:
* number of bind watches active
* number of addr watches active
* number of connections to namerd
* list of namerd namespaces and their dtabs

adleong:
Closed by #1060

PR #970, opened 2017-01-25
Add namerd tab to linkerd admin

esbie:
Problem:
When debugging connection problems between namerd and linkerd, there's not a lot of visibility besides `admin/metrics.json`.

Solution:
Display more internal state about configuration and connection with namerd. 

screen shot 2017-01-24 at 3 27 28 pm

This branch also fixes an issue where the active tab was not being highlighted.

The page currently only shows namespaces that this linkerd knows about, not all namespaces available. I'm worried that could be a point of confusion. I think adding all possible namerd namespaces would require implementing a namespace list endpoint for the thrift interpreter. Is that something we're interested in doing?

TODO: client stats are not yet surfaced on the page
Related to #937


linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
```diff
@@ -1,12 +1,11 @@
...
 
-import com.twitter.finagle.http.{MediaType, Request, Response}
-import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.finagle.http.{MediaType, Response}
 import com.twitter.util.Future
 import io.buoyant.admin.HtmlView
 import io.buoyant.linkerd.Build
 
-object AdminHandler extends HtmlView {
+class AdminHandler(namerdNav: Boolean) extends HtmlView {
```
adleong: This seems like a reasonable stopgap fix.  I've ticketed a more long term solution here: https://github.com/linkerd/linkerd/issues/972

siggy: This may be moot given alex's ticket above, but consider just passing `namerdNav` into the `html` method, and generating the navBar using that. This way, `AdminHandler` can go back to being an object. Or, consider moving all the methods that do not depend on state into an object.

esbie: Interesting, I hadn't really considered that option. 

I agree that changing `AdminHandler` to a class was kind of painful. Currently I don't want to pass `namerdNav` into the `html` method because it means each handler using the function would need to know information about namerd and I'd rather keep handlers self contained. 

Keeping `mkResponse` in an object seems reasonable enough.

esbie: hrm, I tried putting `mkResponse` in an object, but any handler using `mkResponse` also at some point uses `html` 🙍

PR #978, opened 2017-01-25
gRPC interop server implementation

stevej:
Also added a README with the relevant github issues so our users can track our interop progress.


grpc/interop/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+# To find all the test cases you can run, use the --help flag.
+`$ ./go-grpc-interop-client --help`
+
+## build and run the grpc-java interop client.
+First, you will need gradle installed. (`brew install gradle` on macOS)
+```
+$ git clone https://github.com/grpc/grpc-java.git
+$ cd grpc-java/interop-testing
+$ ./gradlew installDist -PskipCodegen=true
```
olix0r: These steps don't work as-is.  I think you don't want to cd into `interop-testing`?

stevej: yeah, you're totally right. fixing it.

stevej:
If there's a cleaner way to build to the grpc-go interop client, please let me know. The current method is pretty ugly.

olix0r:
We could potentially dockerize the interop clients so they can be invoked via docker run...

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
```diff
@@ -0,0 +1,92 @@
...
+
+        req.recv().map { outputCallRequest =>
+          outputCallRequest.value.payload.map { payload =>
+            payload.body.map { body =>
+              val chars = Array.fill(body.length) { 0.toByte } // {} specializes us to a Byte array
+              val toSend = Buf.ByteArray.Owned(chars)
+              tx.send(StreamingOutputCallResponse(Some(Payload(None, Some(toSend)))))
+            }
+          }
+        }
```
olix0r: This only requests a single frame from the stream.  I think this should be something like:
```scala
      def fullDuplexCall(reqs: Stream[StreamingOutputCallRequest]): Stream[StreamingOutputCallResponse] = {
        val rsps = Stream[StreamingOutputCallResponse]
        def process(): Future[Unit] =
          reqs.recv().respond {
            case Throw(e) => rsps.close()
            case Return(req) =>
              req.value.payload.flatMap(_.body) match {
                case None => rsps.close()
                case Some(body) =>
                  val chars = Array.fill(body.length) { 0.toByte } // {} specializes us to a Byte array
                  val toSend = Buf.ByteArray.Owned(chars)
                  val msg = StreamingOutputCallResponse(Some(Payload(None, Some(toSend))))
                  rsps.send(msg).before(process())
              }
          }
        process()
        rsps
      }
```

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
```diff
@@ -0,0 +1,92 @@
...
+
+        req.responseparameters.map { params =>
+          val size = params.size.getOrElse(0)
+          val chars = Array.fill(size) { 0.toByte } // {} specializes us to a Byte array
+          val toSend = Buf.ByteArray.Owned(chars)
+          tx.send(StreamingOutputCallResponse(Some(Payload(None, Some(toSend)))))
+        }
+
+        return tx
+      }
```
olix0r: similarly, this doesn't chain the writes (and so ignores backpressure) and doesn't ever close the response stream.  I think the following is correct:
```scala
      def streamingOutputCall(req: StreamingOutputCallRequest): Stream[StreamingOutputCallResponse] = {
        val rsps = Stream[StreamingOutputCallResponse]()

        def process(params: Seq[ResponseParameters]): Future[Unit] = params match {
          case Seq(param, tail@_*) =>
            val size = param.size.getOrElse(0)
            val chars = Array.fill(size) { 0.toByte } // {} specializes us to a Byte array
            val toSend = Buf.ByteArray.Owned(chars)
            val msg = StreamingOutputCallResponse(Some(Payload(None, Some(toSend))))
            rsps.send(msg).before(processSeq(tail))

          case _ => rsps.close()
        }

        process(req.responseparameters)
        rsps
      }
```

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
```diff
@@ -0,0 +1,92 @@
...
+      /**
+       * Returns the aggregated size of input payloads.
+       */
+      def streamingInputCall(req: Stream[StreamingInputCallRequest]): Future[StreamingInputCallResponse] = {
+        var size = 0
+
+        req.recv().map { _.value.payload.map { _.body.map { size += _.length } } }
+
+        Future.value(StreamingInputCallResponse(Some(size)))
+      }
```
olix0r: This won't work, we read from the stream (and update `size) asynchronously, but return a response immediately.  Need something like:

```scala
      def streamingInputCall(reqs: Stream[StreamingInputCallRequest]): Future[StreamingInputCallResponse] = {
        def process(processed: Int): Future[Int] = reqs.recv().transform {
          case Throw(Stream.Closed) => Future.value(processed)
          case Throw(e) => Future.exception(e)
          case Return(req) =>
            val sz = req.value.payload.flatMap(_.body).map(_.length).getOrElse(0)
            process(processed + sz)
        }
        process(0).map { sz => StreamingInputCallResponse(Some(sz)) }
      }
```

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
```diff
@@ -0,0 +1,92 @@
...
+
+      /**
+       * For each ResponseParameter sent, we return a frame in the stream with the requested size.
+       */
+      def streamingOutputCall(req: StreamingOutputCallRequest): Stream[StreamingOutputCallResponse] = {
+        val tx = Stream[StreamingOutputCallResponse]()
+
+        req.responseparameters.map { params =>
```
olix0r: this looks like a codegen bug -- should be `req.responseParameters`

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
stevej: 👍

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
stevej: Filed as #992

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
stevej: This fixes `server_streaming`

grpc/interop/src/main/scala/grpc/testing/ServerMain.scala
stevej: Unfortunately, `client_streaming` is still broken with this but only after a certain size is hit so I suspect it's related to the issue you mentioned in #980.

olix0r:
merged in #1024

PR #989, opened 2017-01-25
round robin load balancer

siggy:
i found this useful during some load testing, where i needed to guarantee that linkerd was sending the same load to different instances.


linkerd/core/src/main/scala/io/buoyant/linkerd/LoadBalancerConfig.scala
```diff
@@ -58,3 +59,8 @@ class Heap extends LoadBalancerConfig {
   @JsonIgnore
   val factory = Balancers.heap()
 }
+
+class RoundRobin extends LoadBalancerConfig {
+  @JsonIgnore
+  val factory = Balancers.roundRobin()
+}
```
olix0r: Shouldn't `maxEffort` be exposed as a config option?

Issue #966, opened 2017-01-24
Calculate counter deltas client side for the admin dashboard


adleong:
The admin dashboard makes calls to `/admin/metrics` every second to get delta values for a subset of counters.  It would simplify the metrics interface to not need to provide this and only provide `/admin/metrics.json` (the full list of metrics without any deltas).  The admin dashboard could store the previous value for counters and calculate diffs on the client side.

The deltas from `/admin/metrics` are not on a strict 1 second window, they refresh only when queried (up to a maximum of once per second).  Therefore, I don't think moving the delta calculation to the client will result in any loss of accuracy since the windowing was already driven by the rate at which the client queried.  

The result is that the server will no longer need to make any assumptions about the rate at which clients are querying.

adleong:
Hey wait a minute.  The fact that `/admin/metrics` only refreshes the deltas when queried probably explains https://github.com/linkerd/linkerd/issues/379.  Moving to client side deltas is probably a more robust solution.

adleong:
@rmars @esbie what do you think about the feasibility/level-of-effort for this?

esbie:
extremely feasible/moderate amount of work (like a couple days worth)

PR #999, opened 2017-01-26
Use MetricsTree as the default metrics registery in linkerd

### LINKED ISSUE #934: Use MetricsTree as linkerd's internal metrics registry
In order to decouple the storage of metrics (i.e. the metrics registry) from the transmission/export of metrics, we should use MetricsTree as the metrics registry for linkerd.  Linkerd will then pass the MetricsTree to the telemeters so that they may process or export those metrics.

This has the additional benefit of preserving the tree structure of the metrics instead of just the flattened structure exposed by the default metrics registry.

adleong:
Fixes #934 

Linkerd will now create a MetricsTree instance and use a
MetricsTreeStatsReciever in addition to any StatsReceivers provided by
telemeters.  This MetricsTree is passed to telemeters as a Stack.Param so that
they may export or expose this data.
Also add a `io.l5d.admin` telemeter which provides the `/admin/metrics.json` endpoint backed by the MetricsTree.

Also changes the behavior of the "default" telemeter.  Instead of adding the default (currently CommonMetrics) if no telemeters are specified, this change will ensure that the list of telemeters ALWAYS includes the default telemeter by adding it if it's not present.  Since the default telemeter is required for the admin dashboard to work, we should ensure that it's always present.  In the future we will switch `io.l5d.admin` to be the default telemeter instead of CommonMetrics, but this is blocked on https://github.com/linkerd/linkerd/issues/966

I realize this change to handling defaults might be controversial and I'm very interested in hearing feedback and thoughts.


project/LinkerdBuild.scala
```diff
@@ -131,11 +131,16 @@ object LinkerdBuild extends Base {
     val core = projectDir("telemetry/core")
       .dependsOn(configCore)
       .withTwitterLib(Deps.finagle("core"))
-      .withTwitterLib(Deps.finagle("stats") % Test)
+      .withTwitterLib(Deps.finagle("stats"))
```
olix0r: Does this pull in common-metrics transitively? Can we safely exclude it?

telemetry/admin/src/main/scala/io/buoyant/telemetry/admin/AdminTelemeter.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  private[this] def flattenMetricsTree(tree: MetricsTree, prefix: String = ""): Seq[(String, Metric)] = {
+    tree.children.foldLeft(Seq(prefix -> tree.metric)) {
+      case (acc, (name, child)) =>
+        if (prefix == "")
+          acc ++ flattenMetricsTree(child, name)
+        else
+          acc ++ flattenMetricsTree(child, s"$prefix/$name")
+    }
+  }
```
olix0r: tioli: I would probably opt to do this procedurally with a `mutable.Buffer[(String, Metric)]` to reduce garbage generated from hitting metrics.json (these things end up being _megabytes_ in production).

telemetry/admin/src/main/scala/io/buoyant/telemetry/admin/AdminTelemeter.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  private[this] val json = new JsonFactory()
+  private[this] def writeJson(out: OutputStream, pretty: Boolean = false): Unit = {
+    val jg = json.createGenerator(out)
+    if (pretty) jg.setPrettyPrinter(new DefaultPrettyPrinter())
+    jg.writeStartObject()
+    val flattened = if (pretty)
+      flattenMetricsTree(metrics).sortBy(_._1)
+    else
+      flattenMetricsTree(metrics)
```
olix0r: style nit: this is awkard to read for me (indentation doesn't make control flow obvious)

preferred:
```scala
val flattened =
  if (pretty) ...
  else ...
```
or
```scala
val flattened =
  if (pretty) {
    ...
  } else {
    ...
  }
```

telemetry/admin/src/main/scala/io/buoyant/telemetry/admin/AdminTelemeter.scala
```diff
@@ -0,0 +1,121 @@
...
+    flattened.foreach {
+      case (name, metric) =>
+        metric match {
+          case c: Counter =>
```
olix0r: i'd opt to just collapse these cases like:
```scala
 case (name, c: Counter) =>
 case (name, g: Gauge) =>
```

CHANGES.md
```diff
@@ -6,6 +6,8 @@
   * Displays routers in the order that they are defined
 * Update Marathon namer to evaluate an app's running state
 * Add `roundRobin` as a load balancer option.
+* Add `io.l5d.admin` telemeter as an eventual replacement for the `io.l5d.commonMetrics`
```
olix0r: I'm not really a fan of the name _io.l5d.admin_ -- "the admin telemeter" doesn't really tell me much about what it is or does.  Is the intent here "the thing that exports stats on the admin page?"  perhaps `...adminMetricsExport`?

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -104,24 +105,33 @@ object Linker {
...
-        }
+      val metrics = MetricsTree()
+
+      val telemeterConfigs = telemetry match {
+        case Some(telemeters) if telemeters.exists(_.kind == "io.l5d.commonMetrics") =>
+          telemeters
+        case Some(telemeters) =>
+          telemeters :+ CommonMetricsConfig()
+        case None =>
+          Seq(CommonMetricsConfig())
```
olix0r: I'm not clear on this.  We always add CommonMetricsConfig?  This doesn't use MetricsTree, though, right? Are jvm stats, etc exported to _both_ common metrics and metrics tree?  What happens if both this and AdminTelemeter are configured?

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -104,24 +105,33 @@ object Linker {
...
-          case t if t.disabled =>
-            val msg = s"The ${t.getClass.getCanonicalName} telemeter is experimental and must be " +
-              "explicitly enabled by setting the `experimental' parameter to `true'."
-            throw new IllegalArgumentException(msg) with NoStackTrace
-          case t => t.mk(Stack.Params.empty + param.LinkerConfig(this))
-        }
+      val metrics = MetricsTree()
+
+      val telemeterConfigs = telemetry match {
+        case Some(telemeters) if telemeters.exists(_.kind == "io.l5d.commonMetrics") =>
```
adleong: I think I'm going to revert this change to the way the default telemeter is handled.  My motivation was that it's really easy to mistakenly not configure a telemeter that provides `/admin/metrics.json` in which case the admin dashboard does not work.  In some sense, the admin dashboard should REQUIRE a telemeter that can provide that endpoint.  But I think what I have here is too weird.  

I'll revert this part and think on it some more.

project/LinkerdBuild.scala
adleong: Even if it is pulled in, it shouldn't be used... right?

telemetry/admin-metrics-export/src/main/scala/io/buoyant/telemetry/admin/AdminMetricsExportTelemeter.scala
```diff
@@ -0,0 +1,123 @@
...
+ * AdminMetricsExportTelemeter provides the /admin/metrics.json admin endpoint, backed by a
+ * MetricsTree.  It does not provide a StatsReciever or Tracer.  Histograms are snapshotted into
+ * summaries at a regular interval and the values served on /admin/metrics.json are taken from the
+ * most recent summary snapshot.  Counter and gauge values are served live.
+ */
+class AdminMetricsExportTelemeter(
+  metrics: MetricsTree,
+  snapshotInterval: Duration,
+  timer: Timer
+) extends Admin.Handler with Telemeter with Admin.WithHandlers {
```
olix0r: What's the thought on making this extend Admin.Handler directly?  It might be nicer to have a scoped member `val handler: Admin.Handler = new Service[...] { ... }`

telemetry/admin-metrics-export/src/test/scala/io/buoyant/telemetry/admin/AdminMetricsExportTelemeterInitializerTest.scala
```diff
@@ -0,0 +1,38 @@
...
+import io.buoyant.config.Parser
+import io.buoyant.telemetry.{TelemeterConfig, TelemeterInitializer}
+import io.buoyant.test.FunSuite
+
+class AdminMetricsExportTelemeterInitializerTest extends FunSuite {
+
+  test("io.l5d.admin telemeter loads") {
+    val yaml =
+      """|kind: io.l5d.admin
+         |""".stripMargin
```
olix0r: wait... how do these tests pass?  I thought we changed the kind name?

adleong: these tests weren't even being run because I forgot to update the build file hahahaha whoooooooops

Issue #972, opened 2017-01-25
Allow plugins to add nav items to the linkerd admin site


adleong:
Admin defines a `WithHandlers` trait that allows plugins to add handlers to the linkerd admin site. Add a similar mechanism for plugins to add nav items to the linkerd site top nav.  This way, nav items for a plugin will only appear if that plugin is loaded.

PR #1001, opened 2017-01-27
Allow plugins to provide admin dashboard nav items

adleong:
Fixes https://github.com/linkerd/linkerd/issues/972

* Add a `WithNav` trait that allows plugins to provide nav items to the admin dashboard.
* Remove the recent requests placeholder
* Refactor Admin handler types to be richer
* Move namerd handler into namerd interpreter plugin
  * Note that this breaks the functionality where multiple interpreters' namespaces can be displayed.  May or may not be a big deal considering that the next iteration of the namerd api should have a namespace list endpoint.


adleong:
I wondered the same thing, actually.  I went with this because it's slightly more flexible.  You could, for example, add nav items that link to external urls, but I'm not sure we'd ever do that.  I could easily go either way on it.

esbie:
Let's leave it as is, we can restrict it later if it's a problem

esbie:
I support making sure relative urls work, but this breaks nav highlighting as mentioned in https://github.com/linkerd/linkerd/pull/970

PR #1003, opened 2017-01-27
[grpc] support stream cancelation and grpc status control in responses

olix0r:
Problem

gRPC response contain a grpc-status and grpc-message trailer fields indicating
the application-level response code (http status codes are ignored). Our grpc
library currently ignores these fields and provides no mechanism for setting
them on responses.

Solution

Introduce the GrpcStatus type, modeling grpc-status codes. Furthermore, grpc
Streams may be `reset` with a status to indicate a failure.

### DIFF
--- a/grpc/eg/src/test/scala/io/buoyant/grpc/EgTest.scala
+++ b/grpc/eg/src/test/scala/io/buoyant/grpc/EgTest.scala
@@ -1,7 +1,6 @@
 package io.buoyant.grpc
 
 import com.google.protobuf.CodedOutputStream
-import com.twitter.conversions.time._
 import com.twitter.finagle.buoyant.H2
 import com.twitter.io.Buf
 import com.twitter.util.{Await, Future, Promise, Throw}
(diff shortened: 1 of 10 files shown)


grpc/eg/src/test/scala/io/buoyant/grpc/EgTest.scala
```diff
@@ -135,11 +134,12 @@ class EgEndToEndTest extends FunSuite {
       val rf2 = rsps.recv()
       assert(!rf2.isDefined)
       await(tx.close())
-      assert(await(rf2.liftToTry) == Throw(Stream.Closed))
+      assert(await(rf2.liftToTry) == Throw(GrpcStatus.Ok()))
     } finally await(h2client.close().before(h2srv.close()))
   }
 
   test("streaming request and unary response") {
+    setLogLevel(com.twitter.logging.Level.ALL)
```
stevej: don't forget to remove this before checkin.

PR #1004, opened 2017-01-27
Localhost transformer should use loopback interface

adleong:
The localhost transformer should consider the addresses from all network interfaces to be local.  Loopback was mistakenly excluded from this which caused the localhost transformer to filter out endpoints bound to 127.0.0.1.


olix0r:
Can you describe this problem a little more (i.e. in the commit message)?

adleong:
updated commit message and description above ^^

PR #1005, opened 2017-01-27
RFC: namerd grpc API

### LINKED ISSUE #842: gRPC API for namerd
Have namerd expose a gRPC API for naming (bind, addr, delegate, etc.)

olix0r:
This branch is not complete, though it is fully functional.

This change introduces the `io.l5d.namerd.grpc` interpreter.  The gRPC interface exposes unary and streaming endpoints.  Streaming endpoints are used to watch dtabs, bound and delegate trees, and addrs by mapping these streams into Vars/Activities, and vice versa.

I've manually tested this locally, and I've also written a small Golang client for this API -- this should move into namerctl in the future.

TODO:
- wait for #1003 to land, so that proper error handling may be done (and error types may be removed from protobuf)

Fixes #842


grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/Stream.scala
```diff
@@ -3,7 +3,9 @@ package io.buoyant.grpc.runtime
 import com.twitter.concurrent.{AsyncMutex, AsyncQueue}
 import com.twitter.finagle.buoyant.h2
 import com.twitter.io.Buf
-import com.twitter.util.{Future, Promise, Return, Throw, Try}
+import com.twitter.util.{Activity, Event, Future, Promise, Return, Throw, Try, Var}
+import java.util.concurrent.atomic.AtomicReference
+import scala.collection
```
olix0r: unused import

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdGrpcInterpreterInitializer.scala
```diff
@@ -0,0 +1,106 @@
...
+        .withParams(H2.client.params ++ params)
+        .transformed(retryTransformer)
+      configureTls(c, tls).newService(name, label)
+    }
+
+    InterpreterClient(new Interpreter.Client(client), namespace.getOrElse("default"))
+  }
+
+  @JsonIgnore
+  private[this] def configureTls[Req, Rsp](client: StackClient[Req, Rsp], tls: Option[ClientTlsConfig]): StackClient[Req, Rsp] =
```
olix0r: is this still necessary?

namerd/iface/grpc/src/main/scala/io/buoyant/namerd/iface/GrpcServerConfig.scala
```diff
@@ -0,0 +1,40 @@
...
+    def kind = GrpcServerConfig.kind
+    def serve() = {
+      val interpreter = grpc.InterpreterServer(store, namers, stats)
+      H2.serve(addr, ServerDispatcher(interpreter))
+    }
+  }
+}
+
+object GrpcServerConfig {
+  val kind = "io.l5d.namerd.grpc"
```
adleong: The other namerd iface plugins don't have namerd in the kind name.  I think we should be consistent.

namerd/iface/grpc/src/main/scala/io/buoyant/namerd/iface/grpc/InterpreterClient.scala
```diff
@@ -0,0 +1,174 @@
...
+    override def bind(dtab: Dtab, path: Path): Activity[NameTree[Name.Bound]] = {
+      val req = mkBindReq(namespace, path, dtab)
+      val run = Var.async[Activity.State[NameTree[Name.Bound]]](Activity.Pending) { state =>
+        val trees = client.streamBoundTree(req)
+
+        @volatile var closed = false
+        def loop(): Future[Unit] =
+          if (closed) Future.Unit
+          else trees.recv().transform {
+            case Throw(e) =>
```
adleong: We should retry the request on a stream failure.  Is that what your TODO alludes to?

namerd/iface/grpc/src/main/scala/io/buoyant/namerd/iface/grpc/InterpreterClient.scala
```diff
@@ -0,0 +1,174 @@
...
+              state() = Activity.Failed(e)
+              Future.exception(e)
+
+            case Return(Stream.Releasable(BoundTreeRsp(ptree), release)) =>
+              ptree match {
+                case None =>
+                case Some(ptree) =>
+                  state() = Activity.Ok(fromProtoBoundNameTree(ptree))
+              }
+              // XXX shouldn't release until this state is changed
```
adleong: can you explain why we shouldn't release right after we've read the value?

namerd/iface/grpc/src/main/scala/io/buoyant/namerd/iface/grpc/InterpreterClient.scala
```diff
@@ -0,0 +1,174 @@
...
+          closed = true
+          f.raise(Failure("closed", Failure.Interrupted))
+          dtabs.reset(GrpcStatus.Ok())
+          Future.Unit
+        }
+      }
+      Activity(run)
+    }
+
+    override def delegate(
```
adleong: these methods all follow the same pattern.  is it possible to DRY this up?

namerd/iface/grpc/src/main/scala/io/buoyant/namerd/iface/grpc/InterpreterClient.scala
olix0r: Releasing needs happen when the memory associated the returned value can be freed.  We're going to need to rely on this to get off of direct pools and onto buffered pools.  Furthermore, it allows for much more explicit provisioning -- if I say that I'll allow M connections with N streams each, each with a window size of W, then I'm able to estimate memory provisioning by M*N*W.  If an application updates the window immediately, they are effectively operating with a window of 2*W.

adleong: Ah, interesting.  This means that the window size must be big enough for 2 values or else you'll get stuck.  Is the window size really meant to represent the full amount of memory holding messages?  Note that memory usage is only 2 messages worth for the short duration from when the message is received to when it is stored in the state.  Should the window size perhaps not include this temporary memory use?

olix0r:
dropping this in favor of https://github.com/linkerd/linkerd/pull/1065

Issue #936, opened 2017-01-17
Add additional client stats to admin dashboard


adleong:
Including load balancer pool status (available/size) and retry budget.

PR #1007, opened 2017-01-27
Hide transformer prefixes on clients in the admin dashboard

adleong:
Fixes https://github.com/linkerd/linkerd/issues/936

# Problem

When transformers are used, the transformer prefixes are displayed as part of the client id in the linkerd admin dashboard.  These ids are very long and unwieldy.

# Soltuion

Hide transformer prefixes initially.  The full id can be displayed by clicking on the client id.


admin/src/main/resources/io/buoyant/admin/js/template/router_client_container.template
```diff
@@ -1,6 +1,6 @@
 <div class="client-container clearfix">
   <div class="router-header-large" style="border-bottom: 2px solid {{clientColor}};">
-    /{{client}}
+    <span class="transformer-prefix client-id" style="display: none">{{prefix}}</span><span class="client-id">{{client}}</span>
```
rmars: a `cursor: pointer` might be nice here to indicate we can click this

adleong:
The wiggle drives me crazy.  No idea what's going on.

Issue #992, opened 2017-01-26
codegen bug in CamelCase generation for member names


stevej:
In the interop project, in the messages proto

```
message StreamingOutputCallRequest {
  PayloadType response_type = 1;
  repeated ResponseParameters response_parameters = 2;
  Payload payload = 3;
  EchoStatus response_status = 7;
}
```

In the generated code
```
case class StreamingOutputCallRequest(
  `responsetype`: Option[grpc.testing.PayloadType.Value],
  `responseparameters`: Seq[grpc.testing.ResponseParameters],
  `payload`: Option[grpc.testing.Payload],
  `responsestatus`: Option[grpc.testing.EchoStatus]
)
```

PR #1010, opened 2017-01-28
[grpc] Fix field name snake-casing

olix0r:
Problem
    
Field names like _destroy_fascism_ should be generated as _destroyFascism_. Currently, our snake-casing doesn't do capitalization properly and so we get _destroyfascism_.
    
Solution
    
Upper-case snake components.

Closes: #992 
Co-Authored-By: @stevej


esbie:
Yay, thanks for fixing this! Can you update the usage telemeter also?

Issue #943, opened 2017-01-18
Change the default identifier to host


adleong:
The current default identifier is `io.l5d.methodAndHost`.  This identifier is problematic because different protocol versions or http methods don't typically signify different logical destinations.  We should change the default identifier to one that simply uses the hostname only.

PR #1015, opened 2017-01-30
Change default HTTP identifier to only use host header

adleong:
Fixes https://github.com/linkerd/linkerd/issues/943

* Change the default `dstPrefix` from the protocol name to `/s`.
* Change the default http identifier from `io.l5d.methodAndHost` to just use the host header.

By default, http requests will now be given logical names like
```
/s/<host>
```


CHANGES.md
```diff
@@ -8,6 +8,8 @@
 * Add `roundRobin` as a load balancer option.
 * Add the ability to route basted on the `dest` request header when using the
   TTwitter Thrift protocol
+* Change the default `dstPrefix` from the protocol name to `/s`
+* Change the default HTTP identifier to simply use the `Host` header
```
olix0r: Please indicate what identifier is used by default now. That was not easy for me to determine (i get it now, io.l5d.header?)

linkerd/docs/protocol-http.md
```diff
@@ -48,7 +28,7 @@ This protocol has additional configuration options on the *routers* block.
 
 Key | Default Value | Description
 --- | ------------- | -----------
-dstPrefix | `http` | A path prefix used by [Http-specific identifiers](#http-1-1-identifiers).
+dstPrefix | `/s` | A path prefix used by [Http-specific identifiers](#http-1-1-identifiers).
 httpAccessLog | none | Sets the access log path.  If not specified, no access log is written.
 identifier | The `methodAndHost` identifier | An identifier or list of identifiers.  See [Http-specific identifiers](#http-1-1-identifiers).
```
olix0r: Default value should be `{kind: io.l5d.header}`

router/http/src/main/scala/io/buoyant/router/http/HeaderIdentifier.scala
```diff
@@ -28,3 +28,8 @@ case class HeaderIdentifier(
     }
   }
 }
+
+object HeaderIdentifier {
+  def default(prefix: Path, baseDtab: () => Dtab = () => Dtab.base): HeaderIdentifier =
+    new HeaderIdentifier(prefix, "Host", false, baseDtab)
```
olix0r: tioli: header key already constified in finagle

CHANGES.md
adleong: sort of... it's the io.l5d.header identifier with "host" as the header.  I think I'll just make "host" the default header for the io.l5d.header identifier which makes this much easier to explain.

adleong: wait nevermind, this is just the `io.l5d.header.token` identifier.

linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala
```diff
@@ -40,9 +40,10 @@ trait Router {
 
   protected def configureServer(s: Server): Server
 
-  def withParams(ps: Stack.Params): Router =
-    _withParams(ps)
-      .withServers(servers.map(configureServer(_)))
+  def withParams(ps: Stack.Params): Router = {
+    val routerWithParams = _withParams(ps)
+    routerWithParams.withServers(routerWithParams.servers.map(routerWithParams.configureServer))
+  }
```
olix0r: this is a bugfix right? not worth mentioning in changes, but worth mentioning in commit

esbie:
My 2c on `/s`: I don't find it approachable. I think it's one of the reasons why the [finagle documentation](http://twitter.github.io/finagle/guide/Names.html#interpreting-paths-with-delegation-tables) is dense and hard to parse. My vote is a token that provides more context, like a word eg `/service` `/router` `/id` `/request` `/go` `/start`

adleong:
Agree that /s is pretty arcane.  But I think we also want something very concise.  /go and /id are interesting.  I'd also consider /name and /dst but those terms might be too overloaded...

olix0r:
let's stay away from /go -- way overloaded.

my preferred are /svc, /srv, /call, /dst, /to.

I think it's somewhat important to have a scheme that is both succinct and allows for extension (i.e. for source-prefixed schemes)

adleong:
Per offline discussion, going to change `/s` to `/svc`

Issue #997, opened 2017-01-26
Consider collapsing client and server tracing spans


klingerf:
Linkerd's current handing of span IDs is different than how spans are handled in most Finagle services. Typically, when a finagle service receives a request, it does not advance the span ID. This means that all annotations added by the server are attached to caller's client span, and the server only advances the span ID when it initiates a new client call. By contrast, Linkerd advances the span ID when it first receives a request, such that the server annotations are added to a child of the caller's client span, instead of the caller's client span itself. This doubles the number of spans that we report to tracing systems.

We should investigate if it makes sense for linkerd to stop advancing the ID on server spans. As part of this, we should also consider changing the RPC name that we attach to the consolidated span. Right now we use the router label for the server span and the HTTP method name for the client span. For the consolidated span, I propose that we use the router label + the dst path for the current request.

klingerf:
Ok, have pushed the [`kl/span-ids`](https://github.com/linkerd/linkerd/compare/kl/span-ids) branch, which collapses span IDs as described in the description for this issue.

For comparison, here's a trace in the Zipkin UI with the existing span behavior:

![screen shot 2017-01-27 at 10 40 56 am](https://cloud.githubusercontent.com/assets/9226/22384367/def09a8c-e482-11e6-8d2e-874438869b3d.png)

And here's the same call pattern with the proposed new behavior:

![screen shot 2017-01-27 at 10 50 31 am](https://cloud.githubusercontent.com/assets/9226/22384380/e8a9f94c-e482-11e6-8221-1e9ec729b3f5.png)

You can see that the number of spans has dropped from 10 to 6, which is the result of the first and last spans being unmodified, and the middle 8 spans being collapsed to 4. I've also changed the span labels to provide more information.

Here's my attempt to describe the spans in the first trace:

```
1  external client          => external linkerd server (server receive)
2  external linkerd client => web service              (client send)
3  web service             => outgoing linkerd server  (server receive)
4  outgoing linkerd client => incoming linkerd server  (client send)
5  outgoing linkerd client => incoming linkerd server  (server receive)
6  incoming linkerd client => books service            (client send)
7  books service           => outgoing linkerd server  (server receive)
8  outgoing linkerd client => incoming linkerd server  (client send)
9  outgoing linkerd client => incoming linkerd server  (server receive)
10 incoming linkerd client => authors service          (client send)
```

Whereas for the second trace, we have:

```
1 external client         => external linkerd server                  (server receive)
[...]

PR #1016, opened 2017-01-30
Use more descriptive client and server span labels

klingerf:
The existing labels are vague. For server spans, start using the router name plus the dst path for the request. For client http spans, use the http method plus the http path. As part of this change, I realized that the server span labels are generated independent of protocol, so I removed the protocol-specific server span labels that we were generating in the thrift protocol. Relates to #997.

Sample trace with the updated labels:

screen shot 2017-01-30 at 12 44 47 pm


viglesiasce:
For posterity here is what I see in StackDriver trace now. Thanks for putting this in!

screen shot 2017-01-31 at 11 01 38 pm

PR #1020, opened 2017-01-31
Add a client-colored border around each client section 

### LINKED ISSUE #918: More delineation in admin dashboard
The admin dashboard represents the following type of tree structure:
```
Routers
+- RouterA
|    +- Clients
|    |    +- Client1
|    |    +- Client2
|    +- Servers
|         +- Server1
+- RouterB
     +- Clients
     |    +- Client3
     |    +- Client4
     +- Servers
          +- Server2
```

However, I find this structure difficult to see when scanning the page due to lack of delineation between sections and small headings.

It's hard to notice where one router begins an another ends.  There is a slightly thicker blue bar separating them but it doesn't really stand out to me and looks pretty similar to the thinner colored bars on the clients:
screen shot 2017-01-11 at 11 18 01 am
A larger router header might help more clearly signal the start of a router section.

Clients are servers are hard to tell apart.  The "Clients" and "Servers" headings are smaller than their children which makes it hard to see how things are grouped.  In the screenshot below the colored horizontal line looks like the beginning of a section that encompasses everything below it until the next horizontal line.  In other words it looks like both the client and server stats are children of a section called "/#/io.l5d.fs/proxy".
screen shot 2017-01-11 at 11 21 09 am

(possibly related to https://github.com/BuoyantIO/linkerd/issues/915)

rmars:
Try to make the admin page's sections clearer:
* delineate client sections with colored boxes
* move the All routers view to the last one in the nav list, and default "/" to the first router (alphabetically)

Fixes #918

screen shot 2017-01-31 at 11 46 22 am

screen shot 2017-01-30 at 11 58 21 am

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/admin.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/admin.js
@@ -19,7 +19,7 @@ define([
 
     return $.get("config.json").done(function(routersRsp) {
       var template = Handlebars.compile(templateRsp);
-      var routers = _.sortBy(routersRsp.routers, "label");
+      var routers = routersRsp.routers;
 
       if (!matches && routers.length > 0) {
         selectRouter(routers[0].label || routers[0].protocol);
--- a/admin/src/main/resources/io/buoyant/admin/js/src/admin.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/admin.js
@@ -19,17 +19,16 @@ define([
 
     return $.get("config.json").done(function(routersRsp) {
       var template = Handlebars.compile(templateRsp);
-      var routers = routersRsp.routers;
+      var routers = _.sortBy(routersRsp.routers, "label");
 
-      //Not every page supports a mult-router view!
-      if(!removeRoutersAllOption) {
-        routers.unshift({label: "all"});
-      } else {
-        if (!matches && routers.length > 0) {
-          selectRouter(routers[0].label || routers[0].protocol);
-        }
+      if (!matches && routers.length > 0) {
+        selectRouter(routers[0].label || routers[0].protocol);
       }
 
+      //Not every page supports a mult-router view!
+      if(!removeRoutersAllOption || routers.length === 0) {
+        routers.push({label: "all"});
+      }
       var routerHtml = routers.map(template);
 
       $(".dropdown-menu").html(routerHtml.join(""));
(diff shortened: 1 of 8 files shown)


admin/src/main/resources/io/buoyant/admin/js/src/admin.js
```diff
@@ -19,17 +19,16 @@ define([
 
     return $.get("config.json").done(function(routersRsp) {
       var template = Handlebars.compile(templateRsp);
-      var routers = routersRsp.routers;
+      var routers = _.sortBy(routersRsp.routers, "label");
```
adleong: prior to sorting here, are the routers in an arbitrary order, or the order they appear in the config file?  Preserving the order from the config file might be nice.

admin/src/main/resources/io/buoyant/admin/js/src/router_client.js
```diff
@@ -169,8 +170,14 @@ define([
 
       function toggleClientDisplay(expand) {
         if (expand) {
+          $contentContainer.css({"border": colorBorder});
+          $headerLine.css("border-bottom", null);
+
           metricsCollector.registerListener(metricsHandler, getDesiredMetrics);
         } else {
+          $contentContainer.css({'border': null});
+          $headerLine.css({'border-bottom': colorBorder});
```
esbie: It's my preference to separate js and css as much as possible. Instead of directly setting the border-bottom here, I'd like to add/remove a class instead. That way css for the class can remain in dashboard.css

rmars: Hmmmm what about passing the colorBorder? do you prefer that we create a class for each shade and then toggle that?

admin/src/main/resources/io/buoyant/admin/js/src/router_client.js
```diff
@@ -169,8 +170,14 @@ define([
 
       function toggleClientDisplay(expand) {
         if (expand) {
+          $contentContainer.css({"border": colorBorder});
```
adleong: The tiniest of nitpicky bike sheds: I think in the mocks the corners are rounded.

admin/src/main/resources/io/buoyant/admin/js/src/router_client.js
esbie: ahhh. I mean generally yes but that would be to big of a refactor

admin/src/main/resources/io/buoyant/admin/js/src/admin.js
rmars: oh! right, I think they are in config order, let me remove this sort

adleong:
wiggle BEGONE! 🕺

admin/src/main/resources/io/buoyant/admin/js/src/router_client.js
rmars: (agree with you on separation of js/css, maybe we can refactor later)

PR #1025, opened 2017-02-01
add usage data telemeter tab to admin

### LINKED ISSUE #974: Display usage data on the admin logging page
Follow-on of #931 
The plugin is vaguely related to logging, so I figure the admin logging tab is a good place to present it.

siggy:
screen shot 2017-02-01 at 1 46 22 pm

test with:
```bash
./sbt linkerd-examples/admin:run
```
and then:
```bash
open http://localhost:9990/usage
open http://localhost:4140/foo/usage
```

fixes #974

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/main-linkerd.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/main-linkerd.js
@@ -29,14 +29,16 @@ require([
   'src/dashboard',
   'src/delegate',
   'src/namerd',
-  'src/logging'
+  'src/logging',
+  'src/usage'
 ], function (
   $, _, bootstrap,
   adminPage,
   dashboard,
   linkerdDtabPlayground,
   namerd,
-  loggingConfig
+  loggingConfig,
+  usage
 ) {
   // poor man's routing
   if (window.location.pathname.endsWith("/delegator")) {
@@ -50,6 +52,10 @@ require([
   } else if (window.location.pathname.endsWith("/logging")) {
     adminPage.initialize().done(function() {
       new loggingConfig();
+    });
+  } else if (window.location.pathname.endsWith("/usage")) {
+    adminPage.initialize().done(function() {
+      new usage();
     });
   } else {
     adminPage.initialize().done(function(routerConfig) {
--- a/admin/src/main/resources/io/buoyant/admin/js/src/usage.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/usage.js
@@ -1,0 +1,21 @@
+"use strict";
+
+define([
+  'jQuery',
+  'Handlebars',
+  'text!template/usage.template',
+], function(
+  $, Handlebars,
+  usageTemplate
+) {
+  return function() {
+    var template = Handlebars.compile(usageTemplate)
+    var $usageContainer = $(".usage-content");
+
+    $.get("admin/metrics/usage").done(function(usageJson) {
+      var view = {usageJson: JSON.stringify(usageJson, null, '  ')};
+      var $templateContainer = $(template(view));
+      $templateContainer.appendTo($usageContainer);
+    });
+  }
+});
--- a/admin/src/main/resources/io/buoyant/admin/js/template/usage.template
+++ b/admin/src/main/resources/io/buoyant/admin/js/template/usage.template
@@ -1,0 +1,21 @@
+<h1 class="title">Usage Data</h1>
+
+<p>
+You are seeing this page because you have enabled the <b><i>io.l5d.usage</i></b> telemeter in your <a href="config.json">config file</a>
+</p>
+
+<p>
+To see the payload endpoint, you can browse to <a href="admin/metrics/usage">/admin/metrics/usage</a>.
+</p>
+
+<p>
+For more information on configuring this telemeter, head over the <a href="https://linkerd.io/config/head/linkerd/index.html#usage">linkerd config reference</a>.
+</p>
+
+<h3>Payload</h3>
+
+<p>
+Below is a snapshot of the payload sent to Buoyant:
+</p>
+
+<pre><code class="language-json">{{usageJson}}</code></pre>
--- a/linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
+++ b/linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
@@ -18,8 +18,7 @@ import com.twitter.finagle.util.DefaultTimer
 import com.twitter.io.Buf
 import com.twitter.logging.Logger
 import com.twitter.util._
-import io.buoyant.admin.Admin
-import io.buoyant.admin.Admin.Handler
+import io.buoyant.admin.Admin.{Handler, NavItem, WithHandlers, WithNavItems}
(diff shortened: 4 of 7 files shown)


admin/src/main/resources/io/buoyant/admin/js/main-linkerd.js
```diff
@@ -51,6 +53,10 @@ require([
     adminPage.initialize().done(function() {
       new loggingConfig();
     });
+  } else if (window.location.pathname.endsWith("/usage")) {
```
siggy: it feels a bit weird to have plugin-specific code here.

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -163,10 +162,13 @@ class UsageDataTelemeter(
   log.info(s"connecting to usageData proxy at $metricsDst")
   val client = Client(metricsService, config, pid, orgId)
 
-  val adminHandlers = Seq(
-    Handler("/admin/metrics/usage", new UsageDataHandler(metricsService, config, pid, orgId, registry))
+  override def adminHandlers: Seq[Handler] = Seq(
+    Handler("/admin/metrics/usage", new UsageDataHandler(metricsService, config, pid, orgId, registry)),
+    Handler("/usage", new UsageDataAdminHandler)
```
siggy: i considered passing `UsageDataHandler` into `UsageDataAdminHandler` and generating the usage json payload server-side, but i think leveraging the templating/js infrastructure on the client-side makes more sense for this use case.

adleong: I think to do this without adding the templates and js to the admin project, you'll need to add them to the usage project and also add handlers to serve them.

siggy:
@adleong believe it or not, i actually wrote my comment above re: `UsageDataHandler` before seeing your comment.

my initial implementation had this all on the server-side, but i found writing so much html in scala to be cumbersome. i noticed the namerd module already uses js and a template. is there a long-term plan for allowing plugins to add tabs and be able to leverage js/templating?

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
siggy: right, i guess my question going forward is: should each project be responsible for serving its own static assets? today admin is the only project doing this.

esbie:
I don't recommend splitting the resources across multiple plugins. It sounds like a nightmare to maintain. If we're squeamish about this admin project we should move the resources into the linkerd/namerd projects themselves or create a separate project solely for assets et al

admin/src/main/resources/io/buoyant/admin/js/main-linkerd.js
esbie: via meat bag conversation: it could be interesting to create a `main-usage.js` file that would be loaded in place of `main-linkerd.js`. Doesn't really solve the "where to serve static assets" problem, but would make this routing less of an eyesore

admin/src/main/resources/io/buoyant/admin/js/src/usage.js
```diff
@@ -0,0 +1,21 @@
...
+  usageTemplate
+) {
+  return function() {
+    var template = Handlebars.compile(usageTemplate)
+    var $usageContainer = $(".usage-content");
+
+    $.get("admin/metrics/usage").done(function(usageJson) {
+      var view = {usageJson: JSON.stringify(usageJson, null, '  ')};
+      var $templateContainer = $(template(view));
+      $templateContainer.appendTo($usageContainer);
```
esbie: not that it matters in any way whatsoever, but could be a little less verbose as 
`$usageContainer.html(template(view))`

adleong:
I feel like the point of a plugin system is to be modular.  Assets or code for a plugin shouldn't go in the core project.  If a plugin has static assets, I feel like it should either have handlers to serve them itself OR we should provide some API for plugins to expose static assets (like they expose handlers and nav items).
Alternatively, we can avoid this entirely by rendering completely serverside and not needing any static assets in the first place.  FWIW, this is what the recent requests plugin does.

siggy:
Agree that plugin systems should be modular, that's what motivated my original comment about it being weird to have plugin-specific assets and code in the admin project. This approach seemed viable because that abstraction was already broken, given the namerd-specific assets and routing code in the admin project which only makes sense when a namerd namer is loaded. I was simply following that pattern.

Going forward, I see 4 options:
1) ship as-is with Usage assets and code in Admin
2) render everything server-side, ala `RecentRequestsAdminHandler`
3) move Usage assets to the Usage project, copy js/require/templating infrastructure from Admin to Usage
4) refactor the admin plugin system to provide templating to plugins

If there are other options I am missing, I'm open to it.

I think 1 and 2 are doable in this review, 3 is probably a bad idea all around, and 4 may be the correct long-term approach. If we do 2, it means admin plugins do not have the same access to assets and templating that the rest of admin does, making frontend development quite limited, and I think should motivate prioritizing 4.

esbie:
I'll throw in nuclear option #5: move js development to a different repo

adleong:
Briefly, what are the downsides to option 3?  It seems intuitive to me that each plugin would contain and serve their own static assets, but I don't know anything about js development so I'm probably missing something.  I took a stab at moving the namerd tab's assets (js file and template) into the namerd interpreter plugin just to see what it would take.  The place I got hung up was that without an explicit list of pages in linkerd-main.js, the namerd function is never invoked.  It feels like the namerd handler should be able to supply its own main somehow though?

I feel like I don't really have a good enough grasp on requireJS (or on how js apps should be structured in general) to know how to solve this.  But I feel like the level of modularity I want should be possible...

adleong:
I took a stab at option 3 here: https://github.com/linkerd/linkerd/pull/1027

image

Is this a bad idea?

esbie:
Yeah it's going to be really painful to do that going forwards once we introduce bundling, minifying, uglifying, transpiling, testing, etc.
We'd have to duplicate these systems in every plugin or house themat the root level of the project

adleong:
Ok, thanks for background.  I have a slight preference for doing this on the server-side (option 2) but I won't block on it.  Going forward I've ticketed https://github.com/linkerd/linkerd/issues/1035 for finding a better solution to this problem.

admin/src/main/resources/io/buoyant/admin/js/template/usage.template
```diff
@@ -0,0 +1,21 @@
+<h1 class="title">Usage Data</h1>
+
+<p>
+You are seeing this page because you have enabled the <b><i>io.l5d.usage</i></b> telemeter in your <a href="config.json">config file</a>
```
adleong: TIOLI: We could display on here whether or not `dryRun` is enabled.  That is, whether data is actually being sent to Buoyant or not.

TIOLI, probably out of scope: We could even make `dryRun` mutable and settable from here.  That way, after seeing the payload, users can turn on sending data with a flick of a switch.

siggy:
closing for now. users can validate usage data by hitting the `/admin/metrics/usage` endpoint.

Issue #973, opened 2017-01-25
Admin dashboard is slow


esbie:
Opening this ticket to track slowness investigation. It routinely takes 1.5s to load an admin page on my dev machine. Given the server isn't even remote, I think we need to shoot for sub 1s on average.

Without having dug too deeply into the problem, I have some possible suspects:

* We're downloading too many small assets at once on page load.
* We're compiling too many Handlebar templates at page load.

Possible solutions we could implement (depending on root cause analysis):

* Bundle js/css assets
* Precompile templates
* Move app routing to the client

adleong:
On my home network the call to `https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,300,600` takes about 500ms and blocks further loading of the page while we wait for a response.  If we could not block on this we could probably shave those 500ms off.

rmars:
(the  requirejs optimizer http://requirejs.org/docs/optimization.html will probably also be helpful)

rmars:
Options:
- [ ] precompiling templates 
    - reduces current load times by ~1/3 
    - this is easy to do
    - doesn't require any changes to dev or release
- [ ] using r.js to optimize requirejs (serve all our js in a bundle)
    - reduces current load times by ~1/2
    - easy enough to do on the js side BUT
    - requires us to either set up a build tool or very manually switch between dev and the bundle we serve UNLESS we add some sort of flag when we run linkerd that will serve unbundled js
- adding a build tool e.g. grunt, gulp etc would add more complication to our dev process however then we could have nice things like es6 transpiling
-  for fonts, if we do move the font loading after the page loads, we get an unpleasant flicker as the whole page rerenders when the font loads

wmorgan:
It's a nice font but blocking for 500ms seems like a lot. We could serve it locally...

olix0r:
google claims that browsers should cache the fonts https://developers.google.com/fonts/faq#will_web_fonts_slow_down_my_page

Do we see this in practice?

re: build tool; As it is now, we only really depend on java/mvn, and then you run _./sbt linkerd/assembly_, and all of the dependency management just happens. I'm very squeamish about adding other build-time dependencies that can't be managed by sbt.  I just voliated this with protoc, in fact, and it's caused problems.  We can work around this by fetching/caching artifacts for the build (like we do with sbt) -- perhaps this could work for grung/gulp/etc, but I worry we biting off a bunch of dependencies.

[...]

PR #1028, opened 2017-02-02
Precompile handlebars templates

rmars:
for slight increase in page load times (~0.5 second, which brings us close to a 1sec page load time)

Part of #973


rmars:
Yep! Plugins can provide their own compiled or not compiled templates since all we do is include them with require.

You can still compile templates yourself, by using the old way, as I haven't removed the text plugin:
```
define(['Handlebars', 'text!template/logging_row.template'], function(Handlebars, notCompiledTemplate) {
  var template = Handlebars.compile(notCompiledTemplate);
}
```

Or by running `handlebars -a js/template/ > js/template/compiled_templates.js` to compile those templates in whatever plugin directory they are in and then including those.

admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
```diff
@@ -1,14 +1,15 @@
...
   'src/combined_client_graph',
-  'text!template/router_client_container.template'
-], function($, Handlebars,
+  'template/compiled_templates'
+], function($,
   RouterClient,
   CombinedClientGraph,
-  routerClientContainerTemplate) {
+  templates
+  ) {
```
esbie: weird formatting?

admin/src/main/resources/io/buoyant/admin/js/src/router_client.js
```diff
@@ -135,7 +134,7 @@ define([
     }
 
     return function (metricsCollector, routers, client, $container, routerName, colors, shouldExpandInitially) {
-      var metricPartial = Handlebars.compile(metricPartialTemplate);
+      var metricPartial = templates["metric.partial"];
```
esbie: huh, wonder if we should move away from .partial notation

admin/src/main/resources/io/buoyant/admin/README.md
```diff
@@ -32,3 +32,13 @@ when a file changes:
...
 karma start --autoWatch=true --singleRun=false
 ```
+
+## Handlebars templates
+
+For a slight speed boost we precompile handlebars templates. To add a new
+template, simply add the template in templates/ and then run
+
+```
+# npm install handlebars if needed
```
esbie: this instruction should just be `npm install` since handlebars is listed in devdeps, and it doesn't hurt to run `npm install` even if you're uptodate

admin/src/main/resources/io/buoyant/admin/js/src/router_client.js
rmars: yeah idk. I don't mind it that much, I do think it's informative?

esbie: if you like it, let's keep it

esbie:
thinking about this more... does this mean we can remove `lib/handlebars-v4.0.5` from the repo entirely?

rmars:
hmmmm not if we want to also support manually adding and compiling templates (e.g. https://github.com/linkerd/linkerd/pull/1025). if we want to force everyone to precompile templates, then yeah we can!

PR #1030, opened 2017-02-02
Use POST request to transfer Dtab in Namerd Admin delegator

Ashald:
We have a relatively large Dtab (>4k) and when it's being passed via query parameter it results into HTTP 414.

Switching from GET to POST method and passing Dtab in the body eliminates the issue.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
@@ -54,9 +54,12 @@ define([
         e.preventDefault();
         var path = $('#path-input').val();
         window.location.hash = encodeURIComponent(path);
-        var request = $.get(
-          "delegator.json?" + $.param({ path: path, dtab: dtabViewer.dtabStr(), namespace: namespace }),
-          renderAll.bind(this));
+        var request = $.ajax({
+            url: 'delegator.json',
+            method: 'POST',
+            contentType: 'application/json',
+            data: JSON.stringify({ namespace: namespace, dtab: dtabViewer.dtabStr(), path: path })
+        }).done(renderAll.bind(this));
         request.fail(function( jqXHR ) {
           $(".error-modal").html(templates.errorModal(jqXHR.responseText));
           $('.error-modal').modal();
--- a/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
@@ -237,6 +237,10 @@ object DelegateApiHandler {
 
     def writeStr[T](t: T): String = mapper.writeValueAsString(t)
     def writeBuf[T](t: T): Buf = Buf.ByteArray.Owned(mapper.writeValueAsBytes(t))
+    def readBuf[T: Manifest](buf: Buf): Try[T] = {
+      val Buf.ByteBuffer.Owned(bb) = Buf.ByteBuffer.coerce(buf)
+      Try { mapper.readValue[T](bb.array) }
+    }
   }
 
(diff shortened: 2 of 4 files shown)


Ashald:
Current implementation is more of PoC and is based on "least intrusion" strategy. :) I'm not sure that passing Dtab in request body while passing other parameters via query string is a good idea... Maybe we should move all parameters to post? And add a case class for de-serializaing that stuff with a bunch of `match`es to get around `Option`s while returning HTTP 400 on missing/invalid input.

olix0r:
I think I'd prefer that we:
- Leave GET as it is
- Add a POST, that takes a json object as the body

Also note that in POST requests, query parameters are always form-encoded into the body (and not in the HTTP resource).

Ashald:
OK, thanks for feedback.

Is it OK to return `HTTP 400` on JSON validation errors?

Also, the only usage of `delegator.json` with `GET` method Is aw was in Namerd Admin so I though it's OK to move it to POST. I'll just add a case for `POST` then.

Ashald:
@olix0r may I ask you to take another look at this PR?

admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
```diff
@@ -273,18 +283,25 @@ class DelegateApiHandler(
...
-          interpreters(ns) match {
-            case delegator: Delegator =>
-              getDelegateRsp(req.getParam("dtab"), req.getParam("path"), delegator)
-            case _ =>
-              err(Status.NotImplemented, s"Name Interpreter for $ns cannot show delegations")
-          }
-        case None =>
-          getDelegateRsp(req.getParam("dtab"), req.getParam("path"), ConfiguredNamersInterpreter(namers))
+    case Method.Post =>
+      Codec.readBuf[DelegationRequest](req.content) match {
```
olix0r: I think we need to look at req.contentType here...

Ashald: We want it to be set to some particular value?

Ashald:
@olix0r I think adding a test shouldn't be an issue. Should it be a test for `DelegateApiHandler` within `admin` module?

Ashald:
@olix0r please take another look. I found some tests for the `DelegateApiHandler` in `linkerd-admin` module so I just extended them rather then adding one more (4th :) ) test class for the `DelegateApiHandler`

admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
```diff
@@ -54,9 +54,11 @@ define([
         e.preventDefault();
         var path = $('#path-input').val();
         window.location.hash = encodeURIComponent(path);
-        var request = $.get(
-          "delegator.json?" + $.param({ path: path, dtab: dtabViewer.dtabStr(), namespace: namespace }),
-          renderAll.bind(this));
+        var request = $.post(
```
rmars: I checked this out and tested it in the UI - since we're now checking for content type json on the server, as is, I get a `POST http://localhost:9990/delegator.json 415 (Unsupported Media Type)` error when I try to test a delegation... so we'll need to switch this to use `$.ajax`  so we can specify a content type:

```
        var request = $.ajax({
            url: 'delegator.json',
            method: 'POST',
            contentType: 'application/json',
            data: JSON.stringify({ namespace: namespace, dtab: dtabViewer.dtabStr(), path: path })
        }).done(renderAll.bind(this));
```

Ashald: @rmars shame on me :(

Ashald:
@rmars pushed a fix, please take another look

Ashald:
Yaaay! Thanks!

PR #1031, opened 2017-02-02
[h2] Acknowledge window updates aggressively

### LINKED ISSUE #1013: flow control may starve gRPC streams

olix0r:
Problem

By default, netty does not send a WINDOW_UPDATE until the available window is at ≥50% capacity. This can lead to starvation (especially for gRPC).

Solution

Change the default window increment ratio from 0.5 to 0.95 so that WINDOW_UPDATE frames are sent much more aggressively.

This branch has been tested against the ver/grpc-interop-server branch (#1024, to be merged soon after this), confirming that this fixes #1013.

In a follow-up, these parameters will be made configurable.

### DIFF
--- a/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
+++ b/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
@@ -168,7 +168,7 @@ class H2FrameCodec(
 object H2FrameCodec {
 
   /** Aggressively acknowledge window updates */
-  val DefaultWindowUpdateRatio = 0.95f
+  val DefaultWindowUpdateRatio = 1.0f
 
   def client(
     settings: Http2Settings = new Http2Settings,
--- a/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
+++ b/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
@@ -15,15 +15,15 @@ import io.netty.handler.codec.http.HttpServerUpgradeHandler.UpgradeEvent;
  * Copyright 2016 The Netty Project
  */
 class H2FrameCodec(
-  connection: Http2Connection,
-  initialSettings: Http2Settings
+  http2Handler: Http2ConnectionHandler
 ) extends ChannelDuplexHandler {
 
   import H2FrameCodec._
 
   private[this] var channelCtx, http2HandlerCtx: ChannelHandlerContext = null
 
-  val connectionListener = new Http2ConnectionAdapter {
+  private[this] val connectionListener = new Http2ConnectionAdapter {
+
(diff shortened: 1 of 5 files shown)


finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -168,21 +167,50 @@ class H2FrameCodec(
 
 object H2FrameCodec {
 
-  private lazy val frameLogger = new Http2FrameLogger(io.netty.handler.logging.LogLevel.DEBUG, getClass)
+  /** Aggressively acknowledge window updates */
+  val DefaultWindowUpdateRatio = 0.95f
```
stevej: I have the same question here about why it's not 1.0f

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -0,0 +1,280 @@
...
+
+      case msg => ctx.write(msg, promise); ()
+    }
+
+}
+
+object H2FrameCodec {
+
+  /** Aggressively acknowledge window updates */
+  val DefaultWindowUpdateRatio = 1.0f
```
adleong: To make sure I understand: this means that any frames that are released will immediately trigger a window update?  Whereas a ratio of 0.95f would mean that releasing frames will not trigger a window update until 5% of the window size has been released?

olix0r: correct

olix0r:
It turns out that ratios have to be on `(0.0, 1.0)` (exclusive).

olix0r:
⭐️ myself

PR #1032, opened 2017-02-02
[h2] Configure HTTP/2 settings from the stack

olix0r:
Problem

HTTP/2 SETTINGS messages allow endpoints to advertise capabilities and buffer sizes. We have no means to configure these parameters.

Solution

Introduce com.twitter.finagle.buoyant.h2.param.Settings, containing the HeaderTableSize, InitialWindowSize, MaxConcurrentStream, MaxFrameSize, and MaxHeaderListSize params. Netty4H2Listener and Netty4H2Transproter use these parameters to configure servers and
clients, respectively.

Clients additionally disable the ENABLE_PUSH setting, since server-push is not supported.


olix0r:
Combining this with #1034

PR #1034, opened 2017-02-02
[h2] Expose HTTP/2-specific configuration

olix0r:
Problem

The HTTP/2 spec allows for a number of per-stream settings to advertise capabilities and constraints. None of these parameters are configurable in linkerd today.

Solution

Update H2Config to read specialized client and server types that allow several h2 settings to be configured.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/param.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/param.scala
@@ -1,11 +1,122 @@
 package com.twitter.finagle.buoyant.h2
 
 import com.twitter.finagle.Stack
+import com.twitter.util.StorageUnit
 
 package object param {
 
   case class ClientPriorKnowledge(assumed: Boolean)
   implicit object ClientPriorKnowledge extends Stack.Param[ClientPriorKnowledge] {
     val default = ClientPriorKnowledge(true)
+  }
+
+  object FlowControl {
+
+    /**
+     * Controls whether connection windows are automatically updated.
+     *
+     * When enabled, connection-level flow control is effectively
+     * disabled (i.e. constrained by SETTINGS_MAX_CONCURRENT_STREAMS *
+     * SETTINGS_INITIAL_WINDOW_SIZE).
+     */
+    case class AutoRefillConnectionWindow(enabled: Boolean)
+    implicit object AutoRefillConnectionWindow extends Stack.Param[AutoRefillConnectionWindow] {
+      val default = AutoRefillConnectionWindow(false)
+    }
+
+    /**
+     * Controls how frequently WINDOW_UPDATE messages are sent to the remote.
+     */
+    case class WindowUpdateRatio(ratio: Float) {
+      require(0.0 < ratio && ratio < 1.0, "ratio must be on (0.0, 1.0), exclusive")
+    }
+    implicit object WinodwUpdateRatio extends Stack.Param[WindowUpdateRatio] {
+      val default = WindowUpdateRatio(0.99f)
+    }
+  }
+
+  object Settings {
+
+    /**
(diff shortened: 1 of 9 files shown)


linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
```diff
@@ -101,11 +102,35 @@ class H2Config extends RouterConfig {
   }
 }
 
-class H2ServerConfig extends ServerConfig {
+trait InitialSettingsConfig {
+  var headerTableSize: Option[Int] = None
+  var initialWindowSize: Option[Int] = None
```
olix0r: I suppose these should be named like `initialWindowBytes`?

olix0r:
needs docs...

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/param.scala
```diff
@@ -1,11 +1,50 @@
...
   }
+
+  case class AutoRefillConnectionWindow(enabled: Boolean)
+  implicit object AutoRefillConnectionWindow extends Stack.Param[AutoRefillConnectionWindow] {
+    val default = AutoRefillConnectionWindow(false)
+  }
+
+  case class WindowUpdateRatio(ratio: Float)
+  implicit object WinodwUpdateRatio extends Stack.Param[WindowUpdateRatio] {
+    val default = WindowUpdateRatio(0.95f)
```
stevej: why not default to 1?

linkerd/examples/h2.yaml
```diff
@@ -12,8 +12,12 @@ routers:
...
     /h2/localhost:4142 => /$/inet/127.1/8888;
     /h2 => /srv;
-  servers:
-  - port: 4142
   identifier:
     kind: io.l5d.headerToken
     header: ":authority"
+  servers:
+  - port: 4142
+    maxConcurrentStreams: 3
```
stevej: why so low?

olix0r: just an example, no preference (though this fails now, anyway)

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/param.scala
```diff
@@ -1,11 +1,50 @@
...
+    }
+
+    case class InitialWindowSize(size: Option[StorageUnit])
+    implicit object InitialWindowSize extends Stack.Param[InitialWindowSize] {
+      val default = InitialWindowSize(None)
+    }
+
+    case class MaxConcurrentStreams(streams: Option[Long])
+    implicit object MaxConcurrentStreams extends Stack.Param[MaxConcurrentStreams] {
+      val default = MaxConcurrentStreams(None)
```
stevej: What does None mean here? No limit? (that seems ok to me)

olix0r:
```
 SETTINGS_MAX_CONCURRENT_STREAMS (0x3):  Indicates the maximum number
      of concurrent streams that the sender will allow.  This limit is
      directional: it applies to the number of streams that the sender
      permits the receiver to create.  Initially, there is no limit to
      this value.  It is recommended that this value be no smaller than
      100, so as to not unnecessarily limit parallelism.

      A value of 0 for SETTINGS_MAX_CONCURRENT_STREAMS SHOULD NOT be
      treated as special by endpoints.  A zero value does prevent the
      creation of new streams; however, this can also happen for any
      limit that is exhausted with active streams.  Servers SHOULD only
      set a zero value for short durations; if a server does not wish to
      accept requests, closing the connection is more appropriate.
```

adleong:
I think this needs to be re-diffed against master.

Issue #995, opened 2017-01-26
Remove tracer plugin interface


adleong:
The tracer plugin interface has been replaced by the telemeter plugin interface and should be removed.  Depends on https://github.com/linkerd/linkerd/issues/994.

PR #1041, opened 2017-02-03
Add zipkin telemeter plugin

### LINKED ISSUE #994: Add zipkin telemeter
Add a zipkin telemeter plugin.  This will act as a replacement for the zipkin tracer plugin and enable the deprecation or deletion of the tracer config section.

klingerf:
This telemeter is intended to replace the zipkin tracer that is currently provided, allowing us to deprecate the tracers config section. Fixes #994.

I was planning on removing the tracers config in a follow-up branch (#995), but I could do it as part of this branch if that's preferable.

### DIFF
--- a/linkerd/examples/zipkin.yaml
+++ b/linkerd/examples/zipkin.yaml
@@ -11,6 +11,7 @@ routers:
   - port: 4140
 
 telemetry:
+- kind: io.l5d.commonMetrics
 - kind: io.l5d.zipkin
   host: localhost
   port: 9410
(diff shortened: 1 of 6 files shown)


linkerd/examples/zipkin.yaml
```diff
@@ -10,6 +10,8 @@ routers:
   servers:
   - port: 4140
 
-tracers:
+telemetry:
 - kind: io.l5d.zipkin
```
adleong: should probably add the io.l5d.commonMetrics telemeter here as well so that the admin dashboard works

adleong:
I have no preference for removing tracers here vs in a followup.

PR #1042, opened 2017-02-03
Re-add commonMetrics as default telemeter

adleong:
io.l5d.commonMetrics was mistakenly removed as the default telemeter.  This adds it back.

### DIFF
--- a/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
@@ -20,6 +20,8 @@ import sun.misc.{Signal, SignalHandler}
  */
 object Main extends App {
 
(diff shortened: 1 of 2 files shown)


adleong:
@klingerf I started to go down this road in my original MetricsTree refactor but I felt like it was pretty weird and rolled it back.  But certainly it's not great that it's so easy to accidentally break the admin dashboard by not configuring any telemeters that provide admin/metrics.json.

klingerf:
@adleong Yeah, that's my main concern as well. It feels weird that by adding a `telemetry` section to a config that didn't previously have one, you could inadvertently disable the stats required for the admin dashboard to function.

adleong:
I do want to address this, but I'm not sure the best way.  I'll create a separate issue where we can discuss.

PR #1043, opened 2017-02-03
Make Consul Name case-insensitive

### LINKED ISSUE #1039: consul namer should be case insensitive
the consul namer caches service names. It should cache them lowercased, and do all cache lookups lowercase (because these names are generally case-insensitive in Http and consul).

Ashald:
Fixes #1039 ;)


Ashald:
@adleong happy to help! ;)

PR #1047, opened 2017-02-04
Convert usage telemeter to use MetricsTree

### LINKED ISSUE #1023: Use MetricsTree in usage telemeter
Depends on https://github.com/linkerd/linkerd/issues/934

The usage telemeter should pull metrics from the MetricsTree provided as a stack param instead of from the global commonMetrics registry.

adleong:
Fixes #1023

### DIFF
--- a/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
+++ b/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
@@ -12,6 +12,19 @@ trait MetricsTree {
 
   def metric: Metric
 
+  def counter: Option[Metric.Counter] = metric match {
+    case c: Metric.Counter => Some(c)
+    case _ => None
+  }
+  def gauge: Option[Metric.Gauge] = metric match {
+    case g: Metric.Gauge => Some(g)
+    case _ => None
+  }
+  def stat: Option[Metric.Stat] = metric match {
+    case s: Metric.Stat => Some(s)
+    case _ => None
+  }
+
   def mkCounter(): Metric.Counter
   def mkStat(): Metric.Stat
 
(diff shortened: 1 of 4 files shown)


telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
```diff
@@ -12,6 +12,19 @@ trait MetricsTree {
...
+  }
+  def gauge: Option[Metric.Gauge] = metric match {
+    case g: Metric.Gauge => Some(g)
+    case _ => None
+  }
+  def stat: Option[Metric.Stat] = metric match {
+    case s: Metric.Stat => Some(s)
+    case _ => None
+  }
+
```
olix0r: I don't _love_ adding these to the public interface for just the one use...

olix0r: specifically, i don't like encouraging an allocation just for the sake of convenience. I'd rather callers make that decision explicitly.

adleong: the allocation of the PF?  How would you feel about this if the PFs were val'd up?

adleong: or the allocation of the Option?

olix0r: well, the PF should def be val'd... i meant the option --  usually this is just an intermediary allocation like

`t.metric match {
  case c: Metric.Counter =>
    // do the thing
  case _ =>
    // don't do it
}
```

By providing `counter`, we encourage callers to do a needless option-allocation.  it's not a huge deal, i just tend to prefer public APIs that are minimal and don't encourage undesirable patterns

PR #1050, opened 2017-02-07
Do navbar highlighting on server side

### LINKED ISSUE #1017: Re-fix admin site tab highlighting
The admin site should highlight the current tab in the top nav.  

This broke in https://github.com/linkerd/linkerd/pull/948, was fixed in https://github.com/linkerd/linkerd/pull/970, and broke again in https://github.com/linkerd/linkerd/pull/1001.

rmars:
Also fixes page reloading that occurs when we change window.location - we don't need this since we get "selected router" from the server-data div.

Fixes #1017

### DIFF
--- a/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminFilter.scala
+++ b/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminFilter.scala
@@ -14,8 +14,9 @@ class AdminFilter(adminHandler: AdminHandler, css: Seq[String] = Nil) extends Si
     service: Service[Request, Response]
   ): Future[Response] = {
     service(request).map { rsp =>
+      val itemToHighlight = request.path.replace("/", "")
       if (rsp.contentType.contains(MediaType.Html))
-        rsp.contentString = adminHandler.html(rsp.contentString, csses = css)
+        rsp.contentString = adminHandler.html(rsp.contentString, csses = css, navHighlight = itemToHighlight)
       rsp
     }
   }
--- a/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
+++ b/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
@@ -8,11 +8,12 @@ import io.buoyant.linkerd.Build
 
 class AdminHandler(navItems: Seq[NavItem]) extends HtmlView {
 
-  private[this] val navHtml = navItems.map { item =>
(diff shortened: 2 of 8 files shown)


linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminFilter.scala
```diff
@@ -14,8 +14,9 @@ class AdminFilter(adminHandler: AdminHandler, css: Seq[String] = Nil) extends Si
     service: Service[Request, Response]
   ): Future[Response] = {
     service(request).map { rsp =>
+      val itemToHighlight = request.path.replace("/", "")
```
adleong: does this work if the admin page is served at a non-root url?

rmars: great question. it does! I tested it out with admin.yaml and it works. I asked @siggy and he says the admin scala is unaware of the root url.

if you use admin.yaml that serves admin at /foo/ and inspect request.path, it's unaware that it's at /foo/... so request.path here is still just "/requests", "/namerd" etc...

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/AdminHandler.scala
```diff
@@ -59,11 +60,18 @@ class AdminHandler(navItems: Seq[NavItem]) extends HtmlView {
...
   ): String = {
     val cssesHtml = csses.map { css =>
       s"""<link type="text/css" href="files/css/$css" rel="stylesheet"/>"""
     }.mkString("\n")
+
+    val navbarWithHighlight = navbar match {
+      case "" => navBar(navHighlight)
+      case _ => navbar
+    }
+
```
esbie: Hmm, this is a confusing change to how `html` works w/o navbar or navHighlight specified. `html` used to display the navbar if none was specified. In this version `html` displays no navbar if navbar & navHighlight are unspecified.

I suggest we change the method signature to make this logic easier to follow, something like:

```
customNavbar: Option[String] = None,
navHighlight: Option[String] = None
): String = {
...
val navbarWithHighlight = (customNavbar, navHighlight) match {
  case (Some(custom), _) => custom
  case (None, None) => navBar()
  case (None, Some(highlight)) => navBar(highlight)
}
```

or just get rid of the navbar arg entirely, as it's not used anywhere that I can see

rmars: okay! I'll get rid of it!

PR #1051, opened 2017-02-07
Allow export of traces via Zipkin HTTP

viglesiasce:
This adds the ability to use the Zipkin HTTP protocol for traces. I'm not a Scala expert but wanted to get this in front of you folks to so that its clear that this is easy to add. No hurt feelings if you choose to implement it some other way.

In my case, I needed this in order to make Stackdriver Trace's [stackdriver-zipkin](https://github.com/GoogleCloudPlatform/stackdriver-zipkin) proxy work with l5d.

screen shot 2017-02-06 at 6 58 50 pm

### DIFF
--- a/linkerd/tracer/zipkin/src/main/scala/io/buoyant/linkerd/tracer/ZipkinTracerInitializer.scala
+++ b/linkerd/tracer/zipkin/src/main/scala/io/buoyant/linkerd/tracer/ZipkinTracerInitializer.scala
@@ -1,5 +1,6 @@
 package io.buoyant.linkerd.tracer
 
+import zipkin.finagle.http.HttpZipkinTracer
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.twitter.conversions.time._
 import com.twitter.finagle._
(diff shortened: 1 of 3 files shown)


klingerf:
Hey @viglesiasce, this is great, thanks for contributing! I've also been looking into getting http trace transport working, since our conversation last week.

I talked it over with @olix0r, and we don't want to set a direct dependency on zipkin-finagle in this repo, since that project also depends on finagle, which would introduce a diamond dependency: we could end up depending on two incompatible versions of finagle, which would limit linkerd's ability to upgrade.

The solution we came up with is to move the new zipkin transports into an external telemeter plugin repo, which would produce new telemeters for writing trace data to zipkin via http and kafka. I actually have the new repo *almost* ready to go. Will try to get it up for review by the end of today. In the meantime, I can publish a docker image for you that includes linkerd with the zipkin http transport plugin, and you can try it out to see if it works in your setup. Does that work for you?

klingerf:
I also wanted to mention that we're in the process of streamlining our configurations in advance of the breaking 0.9.0 linkerd release. So all of the tracing functionality is being moved to the `telemeters` section of the config. This shouldn't be a huge issue, but you would need to change your config to use that section to configure the zipkin http tracer, instead of the `tracers` section. It will end up looking something like this:

```
telemetry:
- kind: io.l5d.commonMetrics
- kind: io.zipkin.http
  host: 127.0.0.1:9411
  sampleRate: 1.0
```

Will publish a docker image that works with that config shortly.

klingerf:
Ok, if you wouldn't mind giving the `buoyantio/linkerd:zipkin-http` docker image a shot, that would be great. We'll eventually publish that image elsewhere, but I pushed it to that tag for testing purposes.

Note, this image is built off of linkerd master, which includes some breaking changes in advance of the 0.9.0 release. Here are some notes about upgrading:

* The `tracers` section of the config has been removed. Instead, you should add this top-level section to your config (replacing the `host` value with the location of your stackdriver-zipkin proxy):

```
telemetry:
- kind: io.l5d.commonMetrics
- kind: io.zipkin.http
  host: 127.0.0.1:9411
  initialSampleRate: 1.0
```

* The `baseDtab` field has been renamed `dtab`.

There might be a few other miscellaneous changes too, but just ping us if you run into issues and we can help you get it sorted out.

klingerf:
The zipkin plugin repo is available for review here: linkerd/linkerd-zipkin#1

viglesiasce:
Thanks @klingerf! I just started paternity leave so will be about 3-4 weeks before I come around to this.

klingerf:
Awesome, congrats @viglesiasce! :champagne: There's no rush at all on this. I've been playing around with the other http tracer implementation and it's working for me.

viglesiasce:
@klingerf worked a treat!!!! Thanks for making this happen. Will close out this issue.

klingerf:
Awesome, thanks @viglesiasce! I'll be moving that repo to 0.9.0 stable soon, too.

Issue #1040, opened 2017-02-03
Remove io.l5d.commonMetrics telemeter


adleong:
Remove the io.l5d.commonMetrics telemeter and switch the default telemeter to io.l5d.adminMetricsExport

Depends on:
- [ ] https://github.com/linkerd/linkerd/issues/1023
- [ ] https://github.com/linkerd/linkerd/issues/938
- [ ] https://github.com/linkerd/linkerd/issues/966

PR #1053, opened 2017-02-07
Calculate metrics deltas in metrics_collector.js

### LINKED ISSUE #966: Calculate counter deltas client side for the admin dashboard

esbie:
Problem:
We'd like to drop support of delta values in `admin/metrics.json`.

Solution:
Calculate deltas client-side instead.
Fixes #966

This is a very surgical change in the interest of #1040 
Going forward I'd like to add a metrics endpoint that better reflects the MetricsTree structure. This would allow the frontend to simplify `routers.js` and any components currently using `query.js`

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
@@ -14,7 +14,7 @@ define(['jQuery'], function($) {
       handler: function called with incoming data of the form:
         {
           general: {}, // data obtained from /metrics.json
-          specific: {} // data obtained from /metrics?m=...
+          specific: {} // deltas derived from current and previous /metrics.json calls
         }
       metrics: returns a list of metrics the listener wants.
         Called with a list of metric names to choose from.
@@ -29,9 +29,9 @@ define(['jQuery'], function($) {
 
     function generateDeltaPayload(generalData, defaultMetrics, prevMetrics) {
       var metrics = _(listeners)
-        .map(function(listener){ return listener.metrics(defaultMetrics); })
-        .flatten()
+        .flatMap(function(listener){ return listener.metrics(defaultMetrics); })
         .uniq()
+        .compact()
         .value();
 
       return _.flatMap(metrics, function(metricName) {
--- a/admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
@@ -8,18 +8,7 @@ define(['jQuery'], function($) {
 
   var MetricsCollector = (function() {
     var generalUpdateUri = "admin/metrics.json";
-    var metricsUpdateUri = "admin/metrics";
     var listeners = [];
-
-    function requestBody(listeners, defaultMetrics) {
-      var params = _(listeners)
-        .map(function(listener){ return listener.metrics(defaultMetrics); })
-        .flatten()
-        .uniq()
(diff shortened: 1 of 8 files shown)


admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
```diff
@@ -38,46 +27,56 @@ define(['jQuery'], function($) {
       _.remove(listeners, function(l) { return l.handler === handler; });
     }
 
-    return function(initialMetrics) {
-      var defaultMetrics = _.keys(initialMetrics);
+    function generateDeltaPayload(generalData, defaultMetrics, prevMetrics) {
+      var metrics = _(listeners)
+        .map(function(listener){ return listener.metrics(defaultMetrics); })
+        .flatten()
+        .uniq()
```
rmars: a `.compact()` here might be nice to get rid of the `""` and `undefined`s

admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
```diff
@@ -38,46 +27,56 @@ define(['jQuery'], function($) {
       _.remove(listeners, function(l) { return l.handler === handler; });
```
rmars: I would also maybe update the description of `specific: {}` in the comment

adleong:
Created https://github.com/linkerd/linkerd/issues/1055 to track.

esbie:
Oh interesting! The `?m=` endpoint has a `delta` field for all stat types, but the field behaves differently based on stat type. 
The client doesn't receive any stat type information, so it'd be difficult to imitate this behavior (and also, the behavior is kind of crazy IMO so I don't want to recreate it). I think instead I'll update the RouterSummary and RouterClient components to be gauge-aware for the gauges they use. Then as part of #1055 we can send stat type in the api payload.

admin/src/main/resources/io/buoyant/admin/js/src/router_client.js
```diff
@@ -68,7 +68,7 @@ define([
     function getMetricDefinitions(routerName, clientName) {
       return _.map([
           {suffix: "requests", label: "Requests"},
-          {suffix: "connections", label: "Connections"},
+          {suffix: "connections", label: "Connections", isGauge: true},
           {suffix: "success", label: "Successes"},
           {suffix: "failures", label: "Failures"},
           {suffix: "loadbalancer/size", label: "Load balancer pool size"},
```
adleong: I believe these load balancer metrics are gauges also

esbie: good call

esbie:
Hm ok, show me your setup when you get a chance; connections look fine in my tests.

screen shot 2017-02-09 at 10 07 53 am

adleong:
Oh, it seems to be working now!  👍

esbie:
yay thank you for checking!

PR #1054, opened 2017-02-08
Add startTime to UsageMessage

esbie:
Problem:
UsageData's `startTime` field was not being populated


linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -79,6 +84,13 @@ private[telemeter] object UsageDataTelemeter {
   def mkContainerManager: Option[String] =
     sys.env.get("MESOS_TASK_ID").map(_ => "mesos")
 
+  def mkStartTime: Option[String] = {
+    val tz = TimeZone.getTimeZone("UTC")
+    val df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'")
+    df.setTimeZone(tz)
+    Some(df.format(new Date()))
```
adleong: This will get the current time when the client is created which isn't exactly the same as the linkerd start time (though maybe it's close enough?).  What about using the jvm uptime metric?  

Also: would we rather send `startTime` or `uptime` in the payload?  I guess one can always derive the other if the message timestamp is known.

esbie: Good idea. Thinking about it more, I'd rather send uptime. I'll update the proto accordingly

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -79,13 +81,22 @@ private[telemeter] object UsageDataTelemeter {
   def mkContainerManager: Option[String] =
     sys.env.get("MESOS_TASK_ID").map(_ => "mesos")
 
+  def mkStartTime(metrics: Map[String, Number]): Option[String] = {
+    val tz = TimeZone.getTimeZone("UTC")
+    val df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'")
+    df.setTimeZone(tz)
+    val uptime = metrics.get("jvm/uptime").map(_.intValue()).getOrElse(0)
```
adleong: I think this conflicts with https://github.com/linkerd/linkerd/pull/1047.
Sorry!!! 😢

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -78,18 +80,27 @@ private[telemeter] object UsageDataTelemeter {
   def mkContainerManager: Option[String] =
     sys.env.get("MESOS_TASK_ID").map(_ => "mesos")
 
+  def mkStartTime(metrics: MetricsTree): Option[String] = {
+    val tz = TimeZone.getTimeZone("UTC")
+    val df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'")
+    df.setTimeZone(tz)
+    val uptime = gaugeValue(metrics.resolve(Seq("jvm", "uptime")).metric).map(_.toInt).getOrElse(0)
```
adleong: Let's go with `toLong` to be safe.

linkerd/telemeter/usage/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -78,18 +80,27 @@ private[telemeter] object UsageDataTelemeter {
...
+    Some(df.format(new Date(System.currentTimeMillis() - uptime)))
+  }
+
   def mkNamers(config: Linker.LinkerConfig): Seq[String] =
     config.namers.getOrElse(Seq()).map(_.kind)
 
-  def gagueValue(metric: Metric): Option[Double] = metric match {
+  def gaugeValue(metric: Metric): Option[Double] = metric match {
```
adleong: whoooops, thank you.

PR #1057, opened 2017-02-08
Add the `clearContext` server configuration parameter

### LINKED ISSUE #1011: Configuration option to clear context on servers
When a server is exposed on the open net, it may be desirable to strip all context from incoming request (i.e. to scrub dtabs, etc).  Expose a config option on servers for this.

olix0r:
Problem

linkerd routers accept contextual information via request metadata. This
context--which includes Dtab overrides--should not be exposed on untrusted
networks, since attackers could use these headers towards malicious ends.

Solution

Add the `clearContext` server option, which causes servers to forcefully clear
the context before passing requests to the router. Additionally, protocol
initializers may extend this behavior to remove request metadata.

Fixes #1011


linkerd/core/src/main/scala/io/buoyant/linkerd/ClearContext.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.twitter.finagle._
+import com.twitter.finagle.context.Contexts
+import com.twitter.finagle.tracing.Trace
+
+/** Clears local and broadcast contexts from the request */
+object ClearContext {
+  val role = Stack.Role("ClearContext")
+  val description = "May clear local and broadcast contexts from inbound requests"
+
+  case class Param(enabled: Boolean)
```
adleong: Calling this Param is a bit confusing with Stack.Param.  What about `ClearContextEnabled` or something?

linkerd/core/src/main/scala/io/buoyant/linkerd/ClearContext.scala
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+  def filter[Req, Rsp]: Filter[Req, Rsp, Req, Rsp] =
+    new SimpleFilter[Req, Rsp] {
+      def apply(req: Req, svc: Service[Req, Rsp]) =
+        Contexts.local.letClear() {
+          Contexts.broadcast.letClear() {
+            Dtab.unwind {
+              Dtab.local = Dtab.empty
+              Trace.letClear {
```
adleong: Is this the right place to be clearing the trace context?  Has server receive already been recorded at this point?

linkerd/core/src/main/scala/io/buoyant/linkerd/ProtocolInitializer.scala
```diff
@@ -86,9 +87,16 @@ abstract class ProtocolInitializer extends ConfigInitializer { initializer =>
         override def close(d: Time) = Future.Unit
       }
 
-      val servable = servers.map { server =>
-        val stackServer = defaultServer.withParams(defaultServer.params ++ server.params)
-        ServerInitializer(protocol, server.addr, stackServer, adapted, server.announce)
+      val servable = servers.map { s =>
+        val stk = s.params[ClearContext.Param] match {
+          case ClearContext.Param(true) => clearServerContext(defaultServer.stack)
```
adleong: We only include the module if the param is enabled, and then check the param again in the module itself.  Isn't this redundant?

linkerd/core/src/main/scala/io/buoyant/linkerd/ProtocolInitializer.scala
```diff
@@ -108,13 +116,17 @@ abstract class ProtocolInitializer extends ConfigInitializer { initializer =>
    */
   protected type ServerReq
   protected type ServerRsp
+  protected type ServerStack = Stack[ServiceFactory[ServerReq, ServerRsp]]
 
-  /** Adapts a server to a router */
+  /** Adapts a server  to a router */
```
adleong: extra space

linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
```diff
@@ -74,6 +74,20 @@ object LinkerdHeaders {
...
 
+    val clearServerModule: Stackable[ServiceFactory[Request, Response]] =
+      new Stack.Module0[ServiceFactory[Request, Response]] {
+        val role = serverModule.role
+        val description = "Clears linkerd context from http request headers"
+
+        val deadline = new Deadline.ClearServerFilter
+        val dtab = new Dtab.ClearServerFilter
+        val sample = new Trace.ClearServerFilter
+        val trace = new Trace.ClearServerFilter
```
adleong: Do we want to do a more blanket clear of `l5d-ctx-*` or even `l5d-*` headers?

linkerd/core/src/main/scala/io/buoyant/linkerd/ClearContext.scala
olix0r: This idiom is borrowed from finagle (which does this all over the place). I'm fine with ClearContext.Enabled, though.

adleong: I guess if there's precedent then I'm okay with it.  It just made me do a double take when reading.

linkerd/core/src/main/scala/io/buoyant/linkerd/ProtocolInitializer.scala
olix0r: It is a little redundant, but I don't think it's harmful.  I think it's right for the ClearConext module to exhibit the behavior it does of checking whether it's enabled. And I think it's right that we only apply the protocol-specific transformation when ClearContext is enabled.  What may be confusing is using the same Param in both places... but I don't really think it's worth having an additional Param type for the protocol-specific setting.

linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
olix0r: Yeah, I as thinking about that... As much as I was trying to avoid iterating through headers, I think it's probably preferable to do it (additionally, after the targeted context clears have been performed)

linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
```diff
@@ -74,6 +74,20 @@ object LinkerdHeaders {
...
       }
 
+    val clearServerModule: Stackable[ServiceFactory[Request, Response]] =
+      new Stack.Module0[ServiceFactory[Request, Response]] {
+        val role = serverModule.role
+        val description = "Clears linkerd context from http request headers"
+
+        val deadline = new Deadline.ClearServerFilter
+        val dtab = new Dtab.ClearServerFilter
+        val sample = new Trace.ClearServerFilter
```
olix0r: should be Sample.ClearServerFilter

linkerd/core/src/main/scala/io/buoyant/linkerd/ClearContext.scala
olix0r: This installed at the bottom of the server stack (i.e. after all trace initialization is preserved).  From the tracing point of view, it will appear as if this request originates from linkerd.

olix0r: I suppose, if we want to maintain tracing from the server, we could simply not install this module for Http.  Then, we'd clear headers and let linkerd do what it wants with context... This module still seems like a good default behavior for other protocols

adleong: We don't want to accept trace headers but we probably do want to, for example, record the remote client address.

olix0r:
https://github.com/linkerd/linkerd/commit/98210aeb9dcf924f85155a365db93e2042dcf773

PR #1060, opened 2017-02-09
Add namerd client info to the namerd tab

rmars:
screen shot 2017-02-08 at 3 46 19 pm

Part of #937

I'm not super sure about the statsReceiver stuff that I've added in - help appreciated!

Also moving to handlebars.runtime which is smaller and meant for precompiled templates.


namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
```diff
@@ -3,19 +3,21 @@ package io.buoyant.namerd.iface
...
 import com.twitter.util.TimeConversions._
-import io.buoyant.namer.{Metadata, DelegateTree, Delegator}
+import io.buoyant.namer.{DelegateTree, Delegator, Metadata}
 import io.buoyant.namerd.iface.{thriftscala => thrift}
 import java.net.{InetAddress, InetSocketAddress}
 
 class ThriftNamerClient(
   client: thrift.Namer.FutureIface,
   namespace: String,
+  statsPrefix: String = "",
```
adleong: I think what you want is to pass a statsReceiver into the constructor here.  In NamerdInterpreterInitializer's `newInterpreter` you can extract the stats receiver from the params and pass it in.  At that point you can also scope it to indicate that it's for the namerd interpreter.

rmars: that's better, thanks!

PR #1061, opened 2017-02-09
Prometheus telemeter and remove commonMetrics

### LINKED ISSUE #938: Add Prometheus telemeter
Add a telemeter which exposes the linkerd metrics registry in prometheus format.  This will serve as a replacement for the prometheus endpoint that is currently part of the core linkerd admin server.  Making this a telemeter will be more modular and will also allow for more idiomatic prometheus stats because the tree structure of the MetricsTree will be available.

Depends on https://github.com/BuoyantIO/linkerd/issues/934
Relates to https://github.com/BuoyantIO/linkerd/issues/927

### LINKED ISSUE #1040: Remove io.l5d.commonMetrics telemeter

### LINKED ISSUE #1052: Default telemeter behavior is confusing
If no telemeters are specified in the linkerd config, the commonMetrics telemeter is used, which exports stats to /admin/metrics.json which powers the admin dashboard.  If a telemeters section is added that does not explicitly include commonMetrics (or another telemeter that provides /admin/metrics.json like adminMetricsExport) then the admin dashboard breaks.  This behavior is confusing.

adleong:
Fixes #938
Fixes #1040 
Fixes #1052 

* Create io.l5d.prometheus telemeter that exposes /admin/metrics/prometheus, is backed by MetricsTree, and does some stat labeling
* Delete the io.l5d.commonMetrics telemeter
* The io.l5d.adminMetricsExport telemeter is no longer a configurable plugin and is just always used.

Note that the io.l5d.adminMetricsExport telemeter does histogram snapshotting on the MetricsTree.  This allows other telemeters like the Prometheus telemeter to read the snapshotted histogram summaries from the MetricsTree without needing to do its own snapshotting.  The snapshot interval is controlled by the `-io.buoyant.telemetry.admin.histogramSnapshotInterval` flag.

Here is an example output of the prometheus telemeter: https://gist.github.com/adleong/633ac4e6eb88f59621b3d5e1d1e37102


adleong:
TODO: 
* fix typo in CHANGES.md
* add some scaladoc

siggy:
Some testing on k8s generates stats like:
```
requests{rt="outgoing", dst_id="%/io.l5d.k8s.daemonset/default/incoming/l5d/#/io.l5d.k8s/default/http/hello", type="counter"} 7308
requests{rt="incoming", dst_id="%/io.l5d.k8s.localnode/10.233.96.4/#/io.l5d.k8s/default/http/world-v1", dst_path="http/1.1/GET/world", type="counter"} 70101
```

Is it possible to break up the `dst_id` more? The `hello` and `world-v1` bits are most relevant. I think generally it's helpful to err on the side of more, shorter labels, as it's easy to query across multiple labels on the consumption side.

adleong:
@siggy not without adding namer specific logic.  Only a namer knows what the segments in its path represent.  That information is not encoded in the MetricsTree.

klingerf:
@adleong Makes sense, and probably something we could do with relabeling configs. But how would you feel about at least splitting the transformers out into a separate label?

adleong:
@klingerf it's possible, but it would have to be done via regex magic since the transformer prefix doesn't get any special treatment in the id.  I'm inclined to leave that to a post-processor.

adleong:
I've updated the stat naming slightly to leave placeholders like "rt", "dst_id", "dst_path", and "srv".  Updated example gist: https://gist.github.com/adleong/633ac4e6eb88f59621b3d5e1d1e37102

Let me know what you think.

adleong:
@siggy As mentioned earlier, pulling segments out of the `dst_id` will require namer specific knowledge.  I don't think we can build that in here, unfortunately.

I can definitely take out the `type` labels and change the histogram stats to include the stat (max, min, avg, etc) into the stat name instead of the label.  We won't match their format exactly since we export percentiles directly rather than cumulative bucket counts.

siggy:
I think if we're not matching prometheus' histogram format, the way you have it right now works best, at least all the components of the histogram are grouped under a single stat name.

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -0,0 +1,113 @@
...
+    // Re-write elements out of the prefix into labels
+    val (prefix1, labels1) = prefix0 match {
+      case Seq("rt", router) if !labelExists(labels0, "rt") =>
+        (Seq("rt"), labels0 :+ ("rt" -> router))
+      case Seq("rt", "dst", "path", path) if !labelExists(labels0, "dst_path") =>
+        (Seq("rt", "dst_path"), labels0 :+ ("dst_path" -> path))
+      case Seq("rt", "dst", "id", id) if !labelExists(labels0, "dst_id") =>
+        (Seq("rt", "dst_id"), labels0 :+ ("dst_id" -> id))
+      case Seq("rt", "dst_id", "path", path) if !labelExists(labels0, "dst_path") =>
+        (Seq("rt", "dst_id"), labels0 :+ ("dst_path" -> path))
```
klingerf: I'd like to add an additional scope to the metric name in this case, where the path is scoped by id.  With the current behavior, you're producing two stats that have the same metric name, with just one additional label to differentiate. For instance:

```
rt:dst_id:requests{rt="external", dst_id="$/inet/web/7000", dst_path="svc/192.168.99.100:8080", type="counter"} 455
rt:dst_id:requests{rt="external", dst_id="$/inet/web/7000", type="counter"} 455
```

I'm worried that with this setup it would be too easy to double count the stats from "$/inet/web/7000". You would need to setup a query that explicitly drops to deduplicates dst_path.  Instead, I would prefer for these two stats to end up as:

```
rt:dst_id:dst_path:requests{rt="external", dst_id="$/inet/web/7000", dst_path="svc/192.168.99.100:8080", type="counter"} 455
rt:dst_id:requests{rt="external", dst_id="$/inet/web/7000", type="counter"} 455
```

And in order to do that, I think you can just change line 68 to:

```
        (Seq("rt", "dst_id", "dst_path"), labels0 :+ ("dst_path" -> path))
```

Put another way, I think it would be a good rule of thumb to expect that every custom label produced by this code also shows up as a segment in the metric name.

adleong: done!  great suggestion, thanks!

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -0,0 +1,124 @@
...
+          sb.append("\n")
+        }
+        for (
+          (percentile, value) <- Seq(
+            "0.5" -> summary.p50, "0.9" -> summary.p90, "0.99" -> summary.p99,
+            "0.999" -> summary.p9990, "0.9999" -> summary.p9999
+          )
+        ) {
+          sb.append(key)
+          sb.append(formatLabels(labels1 :+ ("percentile" -> percentile)))
```
siggy: `quantile`, per https://prometheus.io/docs/concepts/metric_types/#summary

adleong: oops

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -0,0 +1,124 @@
...
+        sb.append(formatLabels(labels1))
+        sb.append(" ")
+        sb.append(g.get)
+        sb.append("\n")
+      case s: Metric.Stat =>
+        val summary = s.summary
+        for (
+          (stat, value) <- Seq(
+            "count" -> summary.count, "min" -> summary.min,
+            "max" -> summary.max, "sum" -> summary.sum, "avg" -> summary.avg
```
siggy: looking at the `go_gc_duration_seconds` summary, maybe we should set min as `quantile="0"` and max as `quantile="1"`?

```
# HELP go_gc_duration_seconds A summary of the GC invocation durations.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 6.409200000000001e-05
go_gc_duration_seconds{quantile="0.25"} 0.00010941100000000001
go_gc_duration_seconds{quantile="0.5"} 0.000138934
go_gc_duration_seconds{quantile="0.75"} 0.000173474
go_gc_duration_seconds{quantile="1"} 0.00030891400000000003
go_gc_duration_seconds_sum 0.29322348800000003
go_gc_duration_seconds_count 1854
```

telemetry/admin-metrics-export/src/main/scala/io/buoyant/telemetry/admin/AdminMetricsExportTelemeter.scala
```diff
@@ -122,3 +123,5 @@ class AdminMetricsExportTelemeter(
   }
 
 }
+
+object histogramSnapshotInterval extends GlobalFlag("60", "Interval, in seconds, to snapshot histrograms")
```
olix0r: You may just want to use a GlobalFlag[Duration] -- this allows specification of i.e. "60.seconds" "1.minute" etc -- and there's a default Flag handler for Duration so it all Just Works.

PR #1063, opened 2017-02-09
Show a warning when all clients are collapsed

rmars:
_Problem_
I think it could be confusing for users who open up the dashboard when a lot of clients are present and there's no requests in the top combined client requests graph.

_Solution_
Add a warning when all clients are collapsed

screen shot 2017-02-09 at 11 24 43 am

The think yellow bootstrap default warning is maybe a little much - I can tone it down.


esbie:
Hm yeah I agree with you, seeing an empty graph when you first get to the admin page is pretty confusing. 
I'd prefer to show all requests in the combined graph if all clients are closed. Then if the user ever bothers to expand the client we can go back to ignoring collapsed clients.

rmars:
Updated to show all clients if all clients are collapsed: 

screen shot 2017-02-09 at 2 28 42 pm

adleong:
Is there a performance concern with showing all the clients on the combined request graph when there are hundreds of clients?

rmars:
Hmmmm unclear. I tested a plain smoothie chart and it handled 400 plotted lines easily. I suspect our data processing to do other things in the dashboard cause us problems rather than the chart.

screen shot 2017-02-09 at 4 14 29 pm

adleong:
Ok, just thinking about the case where there are hundreds of clients and wondering if drawing them all in the combined graph will cause the page to grind to halt.  I'll defer to your judgement.

rmars:
Yeah it may be a good idea to just test ourselves and see what happens!

esbie:
Previous slowness from large numbers of clients was more likely due to # of DOM nodes rendered on the page & # of metrics parsed than to # of smoothie lines generated. We can always cap at the chart at 500 if it becomes an issue.

PR #1065, opened 2017-02-10
Introduce the io.linkerd.mesh gRPC interface

olix0r:
Problem

namerd's thriftmux interface, while multiplexed, suffers from lack of streaming primitives in thriftmux (which leads to operational complexities wrt load balancing). The HTTP interface, on the other hand, may stream messages but is not multiplexed and is JSON-encoded.

Solution

A new gRPC interface, io.linkerd.mesh, is introduced to eventualy supersede the existing namerd read interfaces. Furthermore, this interface will be expanded to cover a variety of additional linkerd management concerns.

A good deal of testing and validation remains to be done on this implementation. It is only suitable for evaluation purposes at this time.


mesh/core/src/main/protobuf/dtab.proto
```diff
@@ -0,0 +1,99 @@
...
+
+    message Prefix {
+      repeated Elem elems = 1;
+
+      message Elem {
+        oneof value {
+          bytes label = 1;
+          Wildcard wildcard = 2;
+        }
+        message Wildcard {}
```
adleong: what's the difference between this and a single valued enum?

mesh/core/src/main/protobuf/resolver.proto
```diff
@@ -0,0 +1,85 @@
...
+import "dtab.proto";
+
+/**
+ * Supports resolution of a concrete name to a replica set.
+ *
+ * Status: Experimental; this API is not yet finalized.
+ */
+service Resolver {
+
+  rpc GetReplicas (ReplicasReq) returns (Replicas) {}
```
adleong: "replicas" is a new piece of terminology we don't use elsewhere.  Likewise, in the existing namerd API, "resolve" means to go from a logical name directly to an address set.  The existing nomenclature isn't sacred but we should be deliberate about terminology.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -0,0 +1,305 @@
...
+    val rsps = open()
+
+    // As we receive streamed messages, we are careful not to release
+    // them until the state has been updated to a new value. This is
+    // intended to (1) integrate tightly with flow control and more
+    // importantly (2) later integrate with netty's reference counted
+    // bueffers.
+    @volatile var closed = false
+    def loop(releasePrior: () => Future[Unit]): Future[Unit] =
+      if (closed) Future.Unit
```
adleong: If the Var becomes unobserved, the prior message will never be released, even if the Var becomes observed again.  I think this is fine because when the Var is observed again a new stream will be created.  Will the old stream be torn down and collected if it has unreleased frames on it?

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -0,0 +1,305 @@
...
+    Closable.make { _ =>
+      closed = true
+      f.raise(Failure("closed", Failure.Interrupted))
+      rsps.reset(GrpcStatus.Ok())
+      Future.Unit
+    }
+  }
+
+  /**
+   * Constructs an ACtivity that, when observed, calls `open()` to obtain a
```
adleong: typo "ACtivity"

mesh/core/src/main/protobuf/resolver.proto
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * A concrete cluster id.
+   */
+  Path id = 1;
+}
+
+/**
+ * Describes a TCP endpoint.
+ */
+message Endpoint {
```
adleong: Endpoint or Replica?

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -0,0 +1,305 @@
...
+      Some(fromDtab(pdtab))
+    case _ => None
+  }
+
+  private[this] val _collectFromEndpoint: PartialFunction[mesh.Endpoint, Address] = {
+    case mesh.Endpoint(Some(_), Some(ipBuf), Some(port), pmeta) =>
+      val ipBytes = Buf.ByteArray.Owned.extract(ipBuf)
+      val ip = InetAddress.getByAddress(ipBytes)
+      val meta = Seq.empty[(String, Any)] ++
+        pmeta.flatMap(_.authority).map(Metadata.authority -> _) ++
```
adleong: I'm still not thrilled about explicitly listing the metadata fields in the proto and am still in favor of an untyped String->String map.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
olix0r: good catch. I think this should be `if (closed) releasePrior()`

mesh/core/src/main/protobuf/resolver.proto
olix0r: Yeah, I'm trying to use clearer language (and definitely move away from opaque finagle terms like Name, Addr, Address).  _Replicas_ explicitly indicates that the returned set of endpoints can be used interchangeably and that this API does not support other schemes (i.e. by surfacing shard IDs, etc).

mesh/core/src/main/protobuf/resolver.proto
olix0r: "Replica" describes the relationship.  Thing thing is an Endpoint.  Endpoints may be grouped in shards, etc.  (Really _Replicas_ could be called _ReplicatedEndpointSet_ or something if we want to be more verbose).

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
olix0r: I don't think the resolver API should be used to surface arbitrary configuration parameters. 

Instead, the path forward is to build additional service interfaces to satisfy these other configuration domains. And maybe it will be appropriate for some these interfaces to support arbitrary metadata, I just don't think it's appropriate to hang this all off of (only) concrete addresses.

I would even entertain the idea that metadata should be removed, though I think that, for instance, nodeName on the endpoint is a good use of metadata in that it provides additional diagnostic information and won't realistically be used to alter runtime behavior.

mesh/core/src/main/protobuf/dtab.proto
olix0r: i switched to message types everywhere.  enums were a premature optimization. i think it's better to use messages, so they may be extended in the future to support additional fields.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
olix0r: For now, I've removed Replicas.Meta entirely.  We still have Endpoints.Meta

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -0,0 +1,103 @@
+package io.buoyant.interpreter
+
+// import com.twitter.conversions.time._
```
adleong: rm if not needed

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -0,0 +1,103 @@
...
+}
+
+object MeshInterpreterInitializer extends MeshInterpreterInitializer
+
+object MeshInterpreterConfig {
+  private val log = Logger.get(getClass.getName)
+
+  val DefaultRoot = Path.Utf8("default")
+
+  // val defaultRetry = Retry(5, 10.minutes.inSeconds)
```
adleong: rm if not used

adleong: or add a TODO that retries will be added in a followup

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -0,0 +1,103 @@
...
+   */
+  @JsonIgnore
+  def newInterpreter(params: Stack.Params): NameInterpreter = {
+    val name = dst match {
+      case None => throw new IllegalArgumentException("`dst` is a required field") with NoStackTrace
+      case Some(dst) => Name.Path(dst)
+    }
+    val label = MeshInterpreterInitializer.configId
+
+    //val Retry(baseRetry, maxRetry) = retry.getOrElse(defaultRetry)
```
adleong: ditto

mesh/core/src/main/protobuf/resolver.proto
adleong: Here is how this updated terminology would affect our current nomenclature.

naming-v3

olix0r:
I'm removing this from 0.9.0. I'd rather spend some more time working through this branch than merge it in a state where we recommend against its use.

olix0r:
Manually tested with retries.  I think we can pull this in if we want.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -0,0 +1,324 @@
...
+      case mesh.DelegateTreeRsp(Some(ptree)) => Some(fromBoundDelegateTree(ptree))
+      case _ => None
+    }
+  }
+
+  /**
+   * Constructs a Var that, when observed, calls `open()` to obtain a
+   * gRPC stream, publishing updates into the Var's state.
+   *
+   * When the stream returns an error, the Var may be updated (as
```
adleong: a fatal error

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -0,0 +1,324 @@
...
+          backoffs match {
+            case wait #:: moreBackoffs =>
+              Future.sleep(wait).before(loop(open(), moreBackoffs, releasePriorNoError))
+
+            case _ => // empty: fail
+              releasePriorNoError().before(Future.exception(e))
+          }
+
+        case Throw(e) => // fatal
+          Future.exception(e)
```
adleong: should we set the state here?

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -0,0 +1,324 @@
...
+   * gRPC stream, publishing updates into the Var's state.
+   *
+   * When the stream returns an error, the Var may be updated (as
+   * according to `toT`) and no further updates will will be
+   * published.
+   */
+  private[this] def streamVar[S, T](
+    init: T,
+    open: () => Stream[S],
+    toT: Try[S] => Option[T],
```
adleong: We should either set the state on exceptions or change the signature of this to `S => Option[T]`

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -0,0 +1,109 @@
...
+    val name = dst match {
+      case None => throw new IllegalArgumentException("`dst` is a required field") with NoStackTrace
+      case Some(dst) => Name.Path(dst)
+    }
+    val label = MeshInterpreterInitializer.configId
+
+    val Retry(baseRetry, maxRetry) = retry.getOrElse(defaultRetry)
+    val backoffs = Backoff.exponentialJittered(baseRetry.seconds, maxRetry.seconds)
+
+    val client = H2.client
```
adleong: I think we want retries on the client itself in addition to retries within the stream processing.  That way if the initial request fails, we can retry it.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  val defaultRetry = Retry(1, 10.minutes.inSeconds)
+}
+
+case class Retry(
+  baseSeconds: Int,
+  maxSeconds: Int
+)
+
+case class MeshInterpreterConfig(
```
klingerf: Similar to the `NamerdInterpreterConfig`, I think the mesh config should implement the `/namerd` nav item in the admin dashboard. Right now, if you configure your routers with the `io.l5d.mesh` interpreter, we no longer display the "namerd" tab in the admin dashboard navigation.

namerd/iface/mesh/src/main/scala/io/buoyant/namerd/iface/MeshIfaceInitializer.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.twitter.finagle.{Path, Namer, Service, Stack}
+import com.twitter.finagle.buoyant.{H2, h2}
+import com.twitter.finagle.naming.NameInterpreter
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util.Duration
+import com.twitter.util.TimeConversions._
+import io.buoyant.grpc.runtime.ServerDispatcher
+import java.net.InetSocketAddress
+
+class MeshIfaceConfig extends InterfaceConfig {
```
klingerf: I couldn't figure out a way to configure the mesh interface to serve over TLS. Maybe it's not supported yet? Or am I missing the server TLS option? Interpreter client TLS seems to work.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  val DefaultRoot = Path.Utf8("default")
+
+  val defaultRetry = Retry(1, 10.minutes.inSeconds)
+}
+
+case class Retry(
+  baseSeconds: Int,
+  maxSeconds: Int
+)
```
klingerf: I'd prefer to unify this with [JitteredBackoffConfig](https://github.com/linkerd/linkerd/blob/master/linkerd/core/src/main/scala/io/buoyant/linkerd/Router.scala#L265).

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -0,0 +1,109 @@
...
+      val msg = "io.l5d.mesh: tls is configured with validation but `commonName` is not set"
+      throw new IllegalArgumentException(msg) with NoStackTrace
+  }
+}
+
+case class MeshClientTlsConfig(
+  disableValidation: Option[Boolean],
+  commonName: Option[String],
+  trustCerts: Seq[String] = Nil
+)
```
klingerf: It's a bit confusing that these are different configuration parameters than those found in [TlsClientConfig](https://github.com/linkerd/linkerd/blob/master/interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala#L41). Would be nice to standardize, but that would be a breaking change to the existing interpreter config.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
olix0r: copied from the thrift named interpreter config.  We should unify all of these.

namerd/iface/mesh/src/main/scala/io/buoyant/namerd/iface/MeshIfaceInitializer.scala
olix0r: Yep, that should definitely be added.

Ashald:
Meh, I didn't notice this earlier, but there not update to the docs. :(

namerd/iface/mesh/src/main/scala/io/buoyant/namerd/iface/MeshIfaceInitializer.scala
klingerf: #1191.

klingerf:
@Ashald Good call -- I opened #1190 to track adding the documentation.

PR #1068, opened 2017-02-13
Upgrade to jquery 3.1.1

rmars:
This started based on the map errors we were getting in Safari when we open the console.

We could remove the .map line in our jquery (though leaving it is nbd, per http://stackoverflow.com/a/18365316/6723657, we probably don't want users who open a console to be greeted by a bunch of js errors, even harmless ones)

Then I realized we could also just upgrade jquery, as later versions don't include the problematic `//@ sourceMappingURL=jquery-1.10.2.min.map` line. Upgrading didn't look like it would affect us any, and it would provide some minor useful things.

I also ran it with the jquery migrate tool (3.0.0), didn't see any warnings. [upgrade guide here.](https://jquery.com/upgrade-guide/3.0/) which I thought was a little weird. I manually went through the list and didn't see anything that would break and the dashboard works.

(I think I am mistaken about requirejs needing jquery, checking on that now)

This can wait until after 0.9.0


rmars:
Totally! I did do that just to see and I do get deprecated errors. Still so weird that we don't use any deprecated things tho!

screen shot 2017-02-15 at 2 00 11 pm

Issue #1072, opened 2017-02-15
Linkerd URL-decodes query parameter values when used as an HTTP proxy


Ashald:
Start a server
```
$ python -m SimpleHTTPServer $UID
```

Do rquest directly & via linkerd:
```
$ curl -Is "http://localhost:$UID/hello?x=R%26D"
$ http_proxy=localhost:4140 curl "http://localhost:$UID/hello?x=R%26D"
```

Check server logs:
```
127.0.0.1 - - [15/Feb/2017 15:31:00] "GET /hello?x=R%26D HTTP/1.1" 404 -
127.0.0.1 - - [15/Feb/2017 15:31:46] "GET /hello?x=R&D HTTP/1.1" 404 -
```

When request goes through linkerd the `&` is url-decoded and thus breaks the request.

adleong:
Fixed by #1073

PR #1073, opened 2017-02-15
Fix query string handling when Linkerd used as an HTTP proxy 

Ashald:
Addressed #1072

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/ProxyRewriteFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/ProxyRewriteFilter.scala
@@ -33,14 +33,20 @@ object ProxyRewriteFilter {
    * the URI is absolute.
    */
   private[this] def guessAbsolute(uri: String): URI =
-    if (uri.contains("://")) new URI(uri)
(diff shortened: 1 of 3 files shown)


router/http/src/main/scala/io/buoyant/router/http/ProxyRewriteFilter.scala
```diff
@@ -33,14 +33,20 @@ object ProxyRewriteFilter {
    * the URI is absolute.
    */
   private[this] def guessAbsolute(uri: String): URI =
-    if (uri.contains("://")) new URI(uri)
+    if (uri.contains("://"))
+      new URI(uri) // it's safe since single-parameter constructor doesn't mess with URL-encoding
```
Ashald: http://stackoverflow.com/a/4858211

Ashald:
Whaaat? I just fixed typo in commit message and tests failed T_T

Ashald:
@adleong hooray! Thanks for the review! So can this fix get into the release next week?

Ashald:
@olix0r thanks! Maybe not the most elegant one but I tried to keep it simple and concise :)

PR #1081, opened 2017-02-16
Prometheus telemeter now uses snapshotted values for histograms

### LINKED ISSUE #1076: Prometheus histograms not using fixed bucket window
Since #1061 shipped, I noticed that the prometheus histogram stats are now updating immediately, rather than being bucketed to a 1m window, and only updating once per minute. The behavior for these stats now differs between the `/admin/metrics/prometheus` and the `/admin/metrics.json` endpoints.

To reproduce, run a linkerd router with the prometheus telemeter enabled, and send it steady traffic. 

Then run:

```
watch 'curl -s :9990/admin/metrics/prometheus | grep "rt:srv:request_latency_ms"'
```

The stats change with each update. Compare that to:

```
watch 'curl -s :9990/admin/metrics.json?pretty=1 | grep "rt.*.*srv.*request_latency_ms"'
```

The stats only update once per minute.

adleong:
Fixes #1076 

# Problem

The Prometheus telemeter serves stats summaries from live histograms.  These histograms are reset once per minute.  This means that the summaries being served will always have a partial minute's worth of data.

# Solution

Move snapshotting into Metric.Stat.  This allows the Prometheus telemeter to serve the most recently snapshotted summary which will include a full minute's worth of data.

### DIFF
--- a/telemetry/admin-metrics-export/src/main/scala/io/buoyant/telemetry/admin/AdminMetricsExportTelemeter.scala
+++ b/telemetry/admin-metrics-export/src/main/scala/io/buoyant/telemetry/admin/AdminMetricsExportTelemeter.scala
@@ -51,15 +51,13 @@ class AdminMetricsExportTelemeter(
 
   private[this] def run0() = {
     val task = timer.schedule(snapshotInterval) {
-      summarySnapshots = snapshotHistograms(metrics)
+      snapshotHistograms(metrics)
     }
 
     new Closable with CloseAwaitably {
       override def close(deadline: Time): Future[Unit] = closeAwaitably(task.close(deadline))
     }
   }
-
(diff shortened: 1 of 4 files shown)


telemetry/admin-metrics-export/src/main/scala/io/buoyant/telemetry/admin/AdminMetricsExportTelemeter.scala
```diff
@@ -51,16 +51,14 @@ class AdminMetricsExportTelemeter(
 
   private[this] def run0() = {
     val task = timer.schedule(snapshotInterval) {
-      summarySnapshots = snapshotHistograms(metrics)
+      snapshotHistograms(metrics)
     }
```
olix0r: i think ideally it may be preferable to move this timer out into wherever the MetricsTree is configured, but definitely not a blocker

PR #1082, opened 2017-02-17
Switch PolymorphicConfig to abstract class

klingerf:
## Problem

Most of our plugin config traits extend the `PolymorphicConfig` trait, and traits have Java interoperability issues.

## Solution

Switch `PolymorphicConfig` and all objects that extend it to instead be abstract classes.

### DIFF
--- a/config/src/main/scala/io/buoyant/config/PolymorphicConfig.scala
+++ b/config/src/main/scala/io/buoyant/config/PolymorphicConfig.scala
@@ -3,8 +3,8 @@ package io.buoyant.config
 import com.fasterxml.jackson.annotation.{JsonProperty, JsonTypeInfo}
 
 /**
- * A trait that defines the property "kind" as both a var on the object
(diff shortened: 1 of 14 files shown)


klingerf:
@adleong Good question. I originally tried that and it compiles, but resulted in a bunch of runtime errors like this one:

```
$ ./sbt linkerd-examples/zipkin:run
...
com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of io.buoyant.telemetry.TelemeterConfig: abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
 at [Source: java.io.StringReader@5e779c2; line: 14, column: 3] (through reference chain: io.buoyant.linkerd.Linker$LinkerConfig["telemetry"]->com.fasterxml.jackson.module.scala.deser.BuilderWrapper[0])
```

adleong:
The intersection of scala magic and jackson magic is truly a vortex of mystery and wonder.

PR #1083, opened 2017-02-17
Pass stats receiver to namerd iface servers

### LINKED ISSUE #1062: Namerd should use configured stats receiver
In a number of places in namerd, a configured stats receiver is passed in but not used.  Most finagle clients and servers are created without the configured stats receiver, using the default stats receiver instead.  We should always use the configured stats receiver so that all metrics are available in the MetricsTree backed metrics endpoint.

adleong:
Fixes #1062

### DIFF
--- a/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlServiceConfig.scala
+++ b/namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlServiceConfig.scala
@@ -13,7 +13,7 @@ class HttpControlServiceConfig extends InterpreterInterfaceConfig {
     namers: Map[Path, Namer],
     store: DtabStore,
     stats: StatsReceiver
-  ): Servable = HttpControlServable(addr, store, delegate, namers)
+  ): Servable = HttpControlServable(addr, store, delegate, namers, stats)
 
   @JsonIgnore
   def defaultAddr = HttpControlServiceConfig.defaultAddr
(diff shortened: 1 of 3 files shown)


namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlServiceConfig.scala
```diff
@@ -28,11 +28,14 @@ case class HttpControlServable(
...
   delegate: Ns => NameInterpreter,
-  namers: Map[Path, Namer]
+  namers: Map[Path, Namer],
+  stats: StatsReceiver
 ) extends Servable {
   def kind = HttpControlServiceConfig.kind
   def serve(): ListeningServer = Http.server
+    .withLabel("httpController")
     .withStreaming(true)
+    .withStatsReceiver(stats)
```
olix0r: How does this end up scoped?

namerd/iface/control-http/src/main/scala/io/buoyant/namerd/iface/HttpControlServiceConfig.scala
```diff
@@ -28,11 +28,14 @@ case class HttpControlServable(
   addr: InetSocketAddress,
   store: DtabStore,
   delegate: Ns => NameInterpreter,
-  namers: Map[Path, Namer]
+  namers: Map[Path, Namer],
+  stats: StatsReceiver
 ) extends Servable {
   def kind = HttpControlServiceConfig.kind
   def serve(): ListeningServer = Http.server
+    .withLabel("httpController")
```
olix0r: what do you think about using the `kind` here?  (or is the kind not namespaced?) either way, maybe just reference the configId directly?

PR #1085, opened 2017-02-17
Usage reporting is now opt-out

adleong:
The usage reporter is no longer a telemeter plugin but now has its own top-level section in the linkerd config:

```
usage:
  orgId: foo
  enabled: true
```

If this block is absent, the default value of `enabled` is true which means this is on by default.  

Also in this PR I turned off retries and logging and stats for the usage http client.


linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
```diff
@@ -102,7 +103,8 @@ object Linker {
     namers: Option[Seq[NamerConfig]],
     routers: Seq[RouterConfig],
     telemetry: Option[Seq[TelemeterConfig]],
-    admin: Option[AdminConfig]
+    admin: Option[AdminConfig],
+    usage: Option[UsageDataTelemeterConfig]
```
olix0r: Can this exist in the telemetry section?

linkerd/core/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -117,13 +125,16 @@ private[telemeter] object UsageDataTelemeter {
     counters.toSeq
   }
 
+  type HttpRouter = { def identifier: Option[Seq[{ def kind: String }]] }
```
olix0r: i know that these can be costly at runtime. maybe worth a small note that this type is only used during configuration and not in the serving path?

linkerd/core/src/main/scala/io/buoyant/linkerd/Linker.scala
adleong: The telemetry section is a list of telemeter configs.  Having a list where there's a default item always included but that default item can be replaced if you add an item to the list that is the same kind... is pretty weird and complex.  I think having usage as a separate thing is cleaner.

linkerd/core/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
adleong: what is costly?  the reflexive calls?  these are actually used "in the serving path" but that path is only exercised once per hour (or when requested from the admin endpoint)

olix0r: yeah; reflection is slow. i think this is ok.

PR #1092, opened 2017-02-21
Indicate the primary path in the delegator ui

### LINKED ISSUE #1074: Namerd UI: distinguish between "active" path and possible alternatives
This is a follow-up on #604 . For the sake of demonstration the one can use the dtab from #604 or a slightly simplified one provided below.:

```
/bar=>/$/inet/456;
/foobar=>/bar | !;
/foobar=>/foo | !;
```

Expected behavior is such that resolution for `/foobar` must not succeed (due to explicit failure). This is how it works when you use the api:
```
$ curl -s "http://localhost:4180/api/1/resolve/bug?path=/foobar" | jq "."
{
  "type": "neg"
}
```
While the name resolution does fail Namerd UI shows that there are possible successful paths. This is great for debugging purposes (when you look how to fix your Dtab so it will give you desired result) but it's confusing when you trying to understand _how_ the Dtab works.

screen shot 2017-02-15 at 20 49 06

Briefly discussing this with @adleong in Slack we agreed that both of use-cases mentioned above make sense so we need to distinguish between an "active path" and "possible paths".
There are at least 2 options:
1) Styling: somehow highlight the "active path" / gray out "alternative paths"
2) Add a toggle to switch between different modes where only "active path" is displayed or all possible paths

adleong:
Fixes #1074 

# Problem

When the delegation tree has alternate branches, it can be unclear which branch will be used by linkerd.  

# Solution

screen shot 2017-02-21 at 3 00 21 pm

Visually indicate the "active" or "primary" path through the delegation tree.  I've attempted to do this by adding a green bar to the left hand side, but I'm not really happy with how it looks and I don't think it's very clear.  I need some help coming up with a better way of visually indicating this.

I tried giving solid or dashed borders to the "primary" path cells, but it looked pretty horrible.  Maybe with some more skilled use of CSS it might look better.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/css/delegator.css
+++ b/admin/src/main/resources/io/buoyant/admin/css/delegator.css
@@ -11,6 +11,10 @@
   box-shadow: none;
 }
 
+.panel-default>.panel-heading-primary, .panel>.list-group .list-group>.list-group-item-primary {
+  border-left: 10px solid green;
+}
+
 .panel {
   margin-bottom: 0px;
 }
--- a/admin/src/main/resources/io/buoyant/admin/css/delegator.css
+++ b/admin/src/main/resources/io/buoyant/admin/css/delegator.css
@@ -11,8 +11,9 @@
   box-shadow: none;
 }
 
-.panel-default>.panel-heading-primary, .panel>.list-group .list-group>.list-group-item-primary {
-  border-left: 10px solid green;
+.list-group-item {
+  background-color: #f5f5f5;
+  color: #333;
 }
 
 .panel {
--- a/admin/src/main/resources/io/buoyant/admin/css/delegator.css
+++ b/admin/src/main/resources/io/buoyant/admin/css/delegator.css
@@ -11,9 +11,19 @@
   box-shadow: none;
 }
 
-.list-group-item {
+.list-group-item.list-group-item-plain {
   background-color: #f5f5f5;
   color: #333;
+}
+
+.panel-default>.panel-heading-danger {
+  color: #a94442;
+  background-color: #f2dede;
+}
+
+.panel-default>.panel-heading-success {
+  color: #3c763d;
+  background-color: #dff0d8;
 }
 
 .panel {
--- a/admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
@@ -17,24 +17,40 @@ define([
       $('.result').html(renderNode(resp, true));
     }
 
-    function isNeg(obj){
+    function treeType(obj){
       switch(obj.type) {
-        case "delegate": return isNeg(obj.delegate);
-        case "transformation": return isNeg(obj.tree);
-        case "alt": return obj.alt.some(function(e,_i){
-            return isNeg(e);
+        case "delegate": return treeType(obj.delegate);
+        case "transformation": return treeType(obj.tree);
+        case "alt":
+          obj.alt.forEach(function(e,_i){
+            var subtreeType = treeType(e);
+            if (subtreeType != "neg") {
+              return subtreeType;
+            }
           });
-        case "union": return obj.alt.every(function(e,_i){
-            return isNeg(e);
-          });
-        case "neg": return true;
-        case "fail": return false;
-        case "leaf": return false;
-        case "exception": return false;
+          return "neg";
+        case "union":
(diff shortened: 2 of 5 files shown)


admin/src/main/resources/io/buoyant/admin/css/delegator.css
```diff
@@ -11,6 +11,10 @@
   box-shadow: none;
 }
 
+.panel-default>.panel-heading-primary, .panel>.list-group .list-group>.list-group-item-primary {
+  border-left: 10px solid green;
```
adleong: insert better styling here

esbie: Geez this is tough since we're hitting the limits of bootstrap. My thought is to ditch the css changes and instead set the active class on the panel heading and list group items.  We can stop using bootstrap for the delegator in a different CL. 

screen shot 2017-02-22 at 1 21 06 pm

adleong:
I've tried a new approach.  Now instead of coloring all the leaves (red for failure, green for success) we color just the primary path (the entire thing red for failure or green for success).  All nodes not in the primary path remain uncolored, regardless of success or failure.

screen shot 2017-02-24 at 2 27 21 pm

screen shot 2017-02-24 at 2 27 46 pm

Thoughts?

PS: I have no idea why that last node is hanging out all on it's lonesome way down there.

Ashald:
Sorry for jumping in but I like it! :)

admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
```diff
@@ -14,29 +14,80 @@ define([
...
 
-    function renderNode(obj, weight){
+    function treeType(obj){
       switch(obj.type) {
-        case "delegate": obj.isDelegate = true; obj.child = renderNode(obj.delegate); break;
+        case "delegate": return treeType(obj.delegate);
+        case "transformation": return treeType(obj.tree);
+        case "alt":
+          for (var i = 0; i < obj.alt.length; i++) {
+            var subtreeType = treeType(obj.alt[i]);
```
rmars: if we define subtreeType in here, of it's the case "union" etc then subtreeType isn't defined.....

```
Uncaught TypeError: subtreeType is not a function
    at http://localhost:9990/files/js/src/delegator.js:33:52
    at Array.some (native)
    at treeType (http://localhost:9990/files/js/src/delegator.js:33:25)
    at renderNode (http://localhost:9990/files/js/src/delegator.js:53:19)
    at HTMLButtonElement.renderAll (http://localhost:9990/files/js/src/delegator.js:17:25)
```

admin/src/main/resources/io/buoyant/admin/js/template/delegatenode.handlebars
```diff
@@ -1,6 +1,13 @@
 {{#if child}}
   <div class='panel panel-default'>
-    <div class='panel-heading'{{#with dentry}}
+    <div class='panel-heading
+  {{#if isPrimary}}
```
rmars: I think this could use an indent one level?

admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
```diff
@@ -14,29 +14,80 @@ define([
...
+        case "neg": return "neg";
+        case "fail": return "fail";
+        case "leaf": return treeType(obj.bound);
+        case "exception": return "fail";
+      }
+      if (obj.addr.type == "neg") {
+        return "neg";
+      } else {
+        return "success";
+      }
```
esbie: First of all, love this function, v useful. Adding more enum strings makes me think we should define type constants and use those instead, so the js runtime will complain if you make a typo.

There's pretty wide coverage for `const`: http://caniuse.com/#feat=const so I advocate adding it to our repo's lexicon (I should also list what browser versions we're realistically targeting).

I'd suggest something like
```
const neg = "neg"
const fail = "fail"
const success = "success"
...
```

admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
```diff
@@ -14,29 +14,80 @@ define([
...
+    }
+
+    function renderNode(obj, primary, weight){
+      var ttype = treeType(obj);
+      switch(obj.type) {
+        case "delegate": obj.isDelegate = true; obj.child = renderNode(obj.delegate, primary); break;
```
esbie: tioli this was already here, but for the record I prefer `break` to be on its own line

PR #1095, opened 2017-02-22
Hide the routers list dropdown on all pages except / and /delegator

### LINKED ISSUE #1091: Admin's router dropdown is ignored on :9990/namerd
The `/namerd` tab shows information about which router uses which namespace, so it was confusing to me that choosing a router from the router dropdown did not affect the page.

I'd suggest that we show the dropdown on `/` and `/delegator` and hide it on all other pages.

rmars:
Fixes #1091

I'm hiding after render rather than on the server side because in / and delegator, the flicker when you select a new router would be worse than the flicker on the other pages where the dropdown appears briefly and then disappears.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/main-linkerd.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/main-linkerd.js
@@ -38,10 +38,9 @@ require([
   namerd,
   loggingConfig
 ) {
-  var $routerDropdown = $(".navbar-right .dropdown");
   // poor man's routing
   if (window.location.pathname.endsWith("/")) {
-    adminPage.initialize(false).done(function(routerConfig) {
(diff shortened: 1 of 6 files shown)


klingerf:
Looks good -- just an optional suggestion. There's already some code in admin.js for manipulating the router dropdown menu. Maybe it makes sense to move the decision as to whether to show or hide the menu into that file as well? You could pass that information as an argument to the initialize method (which already takes an argument for showing or hiding the "all" dropdown option). Something like:

```
function initialize(showRouterDropdown, showAllOption) {
```

esbie:
If flicker is the issue, I'd rather put this logic on the server. Each admin handler knows whether or not it needs a dropdown.

rmars:
Oops, reverting that behaviour was not intentional. I'll check it out.

PR #1096, opened 2017-02-22
0.9.0

adleong:
* Admin dashboard:
  * Added reties stat, retry budget bar, and client pool bar.
  * Added colored border to clients to make them easier to distinguish.
  * Sorts clients and servers alphabetically.
  * Displays routers in the order that they are defined.
  * Namerd Admin now works with Dtabs of arbitrary size.
* Naming and Routing:
  * Rename `baseDtab` router property to `dtab`.
  * Change the default `dstPrefix` from the protocol name to `/svc`.
  * Change the default HTTP identifier to the `io.l5d.header.token` identifier.
  * Add the ability to route basted on the `dest` request header when using the
    TTwitter Thrift protocol.
* Metrics and Tracing:
  * Remove `io.l5d.commonMetrics` telemeter.
  * Add `io.l5d.prometheus` telemeter.
  * Remove the `tracers` router config in favor of the `io.l5d.zipkin` telemeter.
  * Add opt-out usage data collection.
* Namers:
  * Update Marathon namer to evaluate an app's running state.
  * Add `preferServiceAddress` option to `io.l5d.consul` namer
  * Make `io.l5d.consul` case-insensitive
* Add the `clearContext` server configuration option.
* Fix query parameter decoding when rewriting proxied requests

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,8 @@
 ## 0.9.0 2017-02-22
 
 * Admin dashboard:
-  * Added reties stat, retry budget bar, and client pool bar.
-  * Added colored border to clients to make them easier to distinguish.
+  * Add retries stat, retry budget bar, and client pool bar.
+  * Add colored border to clients to make them easier to distinguish.
   * Sorts clients and servers alphabetically.
   * Displays routers in the order that they are defined.
   * Namerd Admin now works with Dtabs of arbitrary size.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,29 +1,29 @@
-## Upcoming Release
+## 0.9.0 2017-02-22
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,29 +1,29 @@
-## Upcoming Release
+## 0.9.0 2017-02-22
 
-* Add `io.l5d.usage` telemeter for opt-in usage data collection
-* Admin dashboard
+* Admin dashboard:
+  * Added reties stat, retry budget bar, and client pool bar.
```
rmars: s/reties/retries

CHANGES.md
```diff
@@ -1,29 +1,29 @@
...
+  * Added colored border to clients to make them easier to distinguish.
```
rmars: total nit: the tense here is inconsistent with that below. s/Added/Add?

PR #1101, opened 2017-02-23
Addr.Neg addresses should be tinted red in delegator UI

adleong:
# Problem

If a transformer filters an address set down to empty, the result is an Addr.Neg.  In the delegator UI such addresses were being highlighted as green "Bound Address" when they should be highlighted as red "No further branch matches".  This was causing confusion for users who thought that the delegator UI was showing a successful delegation.

# Solution

Highlight Addr.Neg as red.


Ashald:
Reminds me about #1074 :)

adleong:
This is mutually exclusive with https://github.com/linkerd/linkerd/pull/1092

PR #1103, opened 2017-02-24
Fix TTwitter server filter leaking across connections

adleong:
# Problem 

The TTwitter server filter was incorrectly being shared across multiple connections.  This filter stores the state of whether the connection has been ttwitter upgraded.  By sharing this filter across connections we were treating un-upgraded connections as if they were upgraded.  This lead to exceptions like:
`org.apache.thrift.transport.TTransportException: Cannot read. Remote side has closed. Tried to read 459235171 bytes, but only got 29 bytes.`

# Solution

Create a new TTwitter server filter for each connection.

### DIFF
--- a/linkerd/protocol/thrift/src/main/scala/com/twitter/finagle/buoyant/linkerd/ThriftServerPrep.scala
+++ b/linkerd/protocol/thrift/src/main/scala/com/twitter/finagle/buoyant/linkerd/ThriftServerPrep.scala
@@ -21,13 +21,15 @@ object ThriftServerPrep {
     ): ServiceFactory[Array[Byte], Array[Byte]] = {
       val Label(label) = params[Label]
       val Thrift.param.ProtocolFactory(pf) = params[Thrift.param.ProtocolFactory]
-      val payloadSize = new PayloadSizeFilter[Array[Byte], Array[Byte]](
-        params[param.Stats].statsReceiver, _.length, _.length
-      )
       val uncaughtExceptionsFilter = new UncaughtAppExceptionFilter(pf)
-      val ttwitter = new TTwitterServerFilter(label, pf)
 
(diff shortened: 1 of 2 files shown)


linkerd/protocol/thrift/src/main/scala/com/twitter/finagle/buoyant/linkerd/ThriftServerPrep.scala
```diff
@@ -21,13 +21,15 @@ object ThriftServerPrep {
...
       val uncaughtExceptionsFilter = new UncaughtAppExceptionFilter(pf)
-      val ttwitter = new TTwitterServerFilter(label, pf)
 
-      payloadSize.andThen(ttwitter).andThen(uncaughtExceptionsFilter).andThen(next)
+      next.map { service =>
+        val payloadSize = new PayloadSizeFilter[Array[Byte], Array[Byte]](
+          params[param.Stats].statsReceiver, _.length, _.length
+        )
+        val ttwitter = new TTwitterServerFilter(label, pf)
+        payloadSize.andThen(ttwitter).andThen(uncaughtExceptionsFilter).andThen(service)
```
olix0r: can the payloadSize and uncaughtExceptionsFilter live outside of this?  Fewer allocations/resources per connection (request?)...

adleong: `uncaughtExceptionsFilter` is outside of the map.  `payloadSize` being inside is cargo-cult from finagle https://github.com/twitter/finagle/blob/695694ea710bda551ed3da7702a2bd5776ce4e20/finagle-thrift/src/main/scala/com/twitter/finagle/thrift/ThriftServerPreparer.scala

I'm not sure if there's a good reason but I feel like mirroring finagle is safer.

Issue #1106, opened 2017-02-27
Failure in running sample


djKooks:
Hello.
While following up from getting start, I got failure like below.
```
...
[info] Packaging /Users/kwangin/workspace/github_source/linkerd/grpc/gen/target/scala-2.11/protoc-gen-io.buoyant.grpc ...
[info] Compiling 1 Scala source to /Users/kwangin/workspace/github_source/linkerd/telemetry/zipkin/target/scala-2.11/classes...
[info] Compiling 1 Scala source to /Users/kwangin/workspace/github_source/linkerd/telemetry/tracelog/target/scala-2.11/classes...
[info] Compiling 3 Scala sources to /Users/kwangin/workspace/github_source/linkerd/interpreter/per-host/target/scala-2.11/classes...
[info] Compiling 2 Scala sources to /Users/kwangin/workspace/github_source/linkerd/interpreter/fs/target/scala-2.11/classes...
[info] Compiling 2 Scala sources to /Users/kwangin/workspace/github_source/linkerd/namer/k8s/target/scala-2.11/classes...
[info] Compiling 2 Scala sources to /Users/kwangin/workspace/github_source/linkerd/interpreter/k8s/target/scala-2.11/classes...
[info] Compiling 1 Scala source to /Users/kwangin/workspace/github_source/linkerd/telemetry/admin-metrics-export/target/scala-2.11/classes...
[info] Compiling 4 Scala sources to /Users/kwangin/workspace/github_source/linkerd/telemetry/recent-requests/target/scala-2.11/classes...
[info] Compiling 2 Scala sources to /Users/kwangin/workspace/github_source/linkerd/telemetry/prometheus/target/scala-2.11/classes...
[info] Compiling 8 Scala sources to /Users/kwangin/workspace/github_source/linkerd/namerd/core/target/scala-2.11/classes...
[info] Done packaging.
[info] Compiling 1 protobuf files to /Users/kwangin/workspace/github_source/linkerd/linkerd/core/target/scala-2.11/src_managed/main/compiled_protobuf
[info] Compiling schema /Users/kwangin/workspace/github_source/linkerd/linkerd/core/src/main/protobuf/usage.proto
[info] Compiling 11 Scala sources to /Users/kwangin/workspace/github_source/linkerd/namerd/iface/control-http/target/scala-2.11/classes...
[...]

PR #1109, opened 2017-02-27
Update install guide, about protobuf installation.

djKooks:
I updated quick start guide as @olix0r mentioned. This is for avoid failure in running as #1106 .
Please look on. Thanks.


olix0r:
thanks @djKooks!

PR #1110, opened 2017-02-27
Fix: rewrite Location & Refresh response headers when Linkerd rewrites request Host header

Ashald:
This is essentially similar to Nginx `proxy_redirect` option. When Linkerd rewrites `Host` header the endpoint doesn't know about the original `Host` value so it generates `Location`/`Refresh` headers with the value of host headers it gets  (those must be absolute URLs).

We ran into this issue after trying to sue Linkerd as an "edge" web server.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
@@ -1,17 +1,35 @@
 package io.buoyant.linkerd.protocol.http
 
 import com.twitter.finagle.client.AddrMetadataExtraction.AddrMetadata
-import com.twitter.finagle.http.{Request, Response}
+import com.twitter.finagle.http.{Fields, Request, Response}
 import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack}
 import io.buoyant.namer.Metadata
 
 object RewriteHostHeader {
 
+  val headers = Set(
+    Fields.Location,
+    "Refresh" // Not present in Fields... Probably because it's not widely used but still...
+  )
+
   case class Filter(host: String) extends SimpleFilter[Request, Response] {
     def apply(req: Request, svc: Service[Request, Response]) = {
-      // TODO: switch between Host/:authority for HTTP1.1/2.0 when Finagle will support it
+      // Location and Refresh response header MUST contain absolute URLs
+      // As we rewrite the Host header before passing it to a service there might be incorrect
+      // values in those headers if service relies solely on Host header to generate them
+      // Therefore we need to rewrite those response headers as nginx and Apache do
+      val originalHost = req.host
       req.host = host
-      svc(req)
+      svc(req).map { rsp =>
+        for {
+          orig <- originalHost
+          header <- headers
+          value <- rsp.headerMap.get(header)
+        } {
+          rsp.headerMap.set(header, value.replace(host, orig))
+        }
+        rsp
+      }
     }
   }
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 ## In the next release
 
 * Add tree and q params to /admin/metrics.json
+* Fix: rewrite Location & Refresh HTTP response headers when Linkerd
+  rewrites request Host header 
 
 ## 0.9.0 2017-02-22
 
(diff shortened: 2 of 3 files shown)


Ashald:
I'm also wondering whether it makes sense to set the `X-Forwarded-Host` header on the request:
```
a de facto standard for identifying the original host requested by the client in the Host HTTP request header, since the host name and/or port of the reverse proxy (load balancer) may differ from the origin server handling the request.
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
```diff
@@ -1,17 +1,33 @@
...
+    Fields.Location,
+    "Refresh" // Not present in Fields... Probably because it's not widely sued but still...
+  )
+
   case class Filter(host: String) extends SimpleFilter[Request, Response] {
     def apply(req: Request, svc: Service[Request, Response]) = {
-      // TODO: switch between Host/:authority for HTTP1.1/2.0 when Finagle will support it
+      val originalHost = req.host
       req.host = host
       svc(req)
```
adleong: I think this could use a commend explaining why we are replacing the host in the response headers.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
```diff
@@ -1,17 +1,33 @@
...
+        .map { rsp =>
```
adleong: style nit: map on the same line

```
svc(req).map { rsp =>
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
```diff
@@ -1,17 +1,33 @@
...
+          originalHost match {
+            case Some(originalHost) =>
+              rsp.headerMap
```
adleong: this is elegant but I think it might be less allocate-y and easier to read as:

```
for {
  header <- headers
  value <- rsp.headerMap.get(header)
} {
  rsp.headerMap.set(header, value.replace(host, originalHost))
}
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
```diff
@@ -1,17 +1,33 @@
...
+                .filterKeys(headers.contains)
+                .mapValues(_.replace(host, originalHost))
+                .foreach { case (k, v) => rsp.headerMap.set(k, v) }
+              rsp
+            case _ => rsp
```
adleong: style nit: this can only be `None` so I prefer to be explicit about that:
```
case None =>
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
Ashald: Yeah, sure!

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
Ashald: OK, thanks.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
Ashald: ZOMG, I forgot there is `for` in Scala. Thanks!

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RewriteHostHeader.scala
Ashald: Roger that. :)

Ashald:
> I think this makes sense. Just a couple nits.
You referred to the PR in general or to my comment saying that we might as well set `X-Forwarded-Host`?

BTW, tests have passed - that's just good luck after N-th re-run? :)

adleong:
The PR in general.

Ashald:
OK, thanks!

PR #1115, opened 2017-03-02
Convert frontend to use metrics.json tree structure

rmars:
I've deleted the query module, though maybe we could/should add some functions to consolidate all the places we do `['rt', router, 'dst', 'id', client]` etc.

This definitely performs better - with ~20 clients I can see jerking on the combined client graph whereas this one is smoother.


admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
```diff
@@ -27,53 +27,44 @@ define(['jQuery'], function($) {
...
     }
 
     return function(initialMetrics) {
       var prevMetrics = initialMetrics;
 
       function update(resp) {
-        var defaultMetrics = _.keys(resp);
-        var specific = generateDeltaPayload(resp, defaultMetrics, prevMetrics);
+        var metricsToGet = _.flatMap(listeners, function(l) { return l.metrics(resp); });
+        var metricsWithDeltas = getTreeDeltaPayload(metricsToGet, resp, prevMetrics);
```
esbie: Given the new API, I think we should take out the `.metrics()` function that listeners define and instead calculate deltas for every counter in the payload. I don't think it'd be a performance hit and it'd certainly be easier to maintain

admin/src/main/resources/io/buoyant/admin/js/src/routers.js
```diff
@@ -36,10 +35,6 @@ define([
    *   }
    */
   var Routers = (function() {
-
-    var clientRE = Query.clientQuery().withMetric("connects").build(),
-        serverRE = Query.serverQuery().withMetric("requests").build();
```
esbie: I'd like to get rid of `routers.js` because I don't think it adds a lot of value on top of the fully hydrated metrics tree. I guess the adding clients logic is nice, but could just as easily be implemented in `metrics_collector.js`

admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
rmars: yeahhh good point!

admin/src/main/resources/io/buoyant/admin/js/src/routers.js
rmars: agreed!

PR #1116, opened 2017-03-02
K8s Ingress Identifiers

esbie:
What is it:
Identifier that allows linkerd to act as an [ingress controller](https://kubernetes.io/docs/user-guide/ingress/#ingress-controllers) in a k8s setup. Watches ingress resources and uses them to identify incoming requests.

Adds:
* v1beta1 api (only has ingresses atm)
* IngressCache (very similar to ServiceCache and NsCache)
* http and h2 ingress identifiers

Todos:
* docs
* tls
* stripPrefix option


k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+ * IngressCache watches for ingress changes
+ * and checks incoming requests against cached ingress rules.
+ *
+ * @param namespace: The k8s namespace to filter on. If None, it watches all namespaces.
+ */
+
+class IngressCache(namespace: Option[String])
```
adleong: I think this can be greatly simplified if we make namespace mandatory (or optional with a default value) instead of allowing the watching of all namespaces.  If we do this I think we no longer need to subclass `Ns` or keep a 2-level cache.  The result should be very very similar to K8sDtabStore.  

Rather than exposing an `Activity[Map[NamespacedName, VarUp[IngressSpec]]]` we could simply expose an `Activity[IngressSpec]`.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IngressIdentifier.scala
```diff
@@ -0,0 +1,73 @@
...
+  override def apply(req: Request): Future[RequestIdentification[Request]] = {
+    val hostHeader = req.headerMap.get("Host")
+    val matchingPaths = ingressWithCache.get(None, None).getMatchingPath(hostHeader, req.path)
+    matchingPaths.flatMap { paths =>
+      paths.headOption match {
+        case None => Future.value(unidentified)
+        case Some(a) =>
+          val path = pfx ++ Path.Utf8(a.namespace, a.port, a.svc)
+          val dst = Dst.Path(path, baseDtab(), Dtab.local)
+          req.uri = "/"
```
adleong: I would ask if clearing the URI is part of the ingress controller spec, except that I'm guessing the ingress controller spec is not a thing that exists.  Does the golang ingress controller clear the uri like this?

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        val fallback = spec.backend.map(b => IngressPath(None, None, namespace, b.serviceName, b.servicePort))
+        Some(IngressSpec(ingress.metadata.flatMap(meta => meta.name), fallback, paths))
+      case None => None
+    }
+  }
+
+  def isPrefix(pfx: Path, str: Path): Boolean = pfx.showElems == str.take(pfx.size).showElems
+  def matchesNs(ns: String): Boolean = namespace.map(_ == ns).getOrElse(true)
```
adleong: Does this check that the namespace of the ingress resource is the same as the namespace that we use to fetch the ingress resource?  Isn't that always true?

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,188 @@
...
+          IngressPath(rule.host, path.path.map(Path.read(_)), namespace, path.backend.serviceName, path.backend.servicePort)
+        }
+
+        val fallback = spec.backend.map(b => IngressPath(None, None, namespace, b.serviceName, b.servicePort))
+        Some(IngressSpec(ingress.metadata.flatMap(meta => meta.name), fallback, paths))
+      case None => None
+    }
+  }
+
+  def isPrefix(pfx: Path, str: Path): Boolean = pfx.showElems == str.take(pfx.size).showElems
```
adleong: In the absence of a real ingress control spec, I guess we should just treat the golang implementation as the spec.  Does that do prefix matching?

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,188 @@
...
+  def matchesNs(ns: String): Boolean = namespace.map(_ == ns).getOrElse(true)
+
+  def getMatchingPath(hostHeader: Option[String], requestPath: String): Future[Option[IngressPath]] = {
+    val pathToMatch = Path.read(requestPath)
```
adleong: I prefer that we don't treat URI paths as finagle Paths except for where we have to.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IngressIdentifier.scala
esbie: Oops, this was for testing only! Nice catch. Traefik has a StripPrefix option that I think we should emulate later.

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
esbie: Thanks for looking at this. This was my attempt to treat prefixes the same way that the path identifier does. There are two controllers implemented by the kubernetes team (GCE & nginx) and they themselves disagree on the behaviour.

See https://github.com/kubernetes/contrib/issues/885

esbie:
thanks @adleong! I really went back and forth on whether to make namespace optional or required. On some level linkerd needs to be able to watch all namespaces, as this is what other implementations do by default. This is probably not useful, but here's an extremely long ticket where folks are confused that ingress resources are namespaced, but ingress controllers aren't: https://github.com/kubernetes/kubernetes/issues/17088#issuecomment-221442374

There could definitely be an argument for required namespaces, where instead of a single `IngressCache`, the identifier could use a cache per namespace. That'd add a bit of complexity in order to add/remove namespaces. It'd also keep a watch open per namespace when k8s has a perfectly reasonable all-namespaces api. And finally, the namespace is not known when receiving a request. So having a map of caches keyed off of namespaces isn't particularly useful... we have to look through every single ingress rule regardless.

Plz help me come up with something even better that still satisfies an all-namespaces use case :D

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
esbie: >_>

Yeah I think you're right

    <_<

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
esbie: Oh! Here I was thinking how nice I am to convert all the uris to fpaths. Seemed like a cleaner way to compare prefixes, but I'm also cool w/ string manip

olix0r: It's a good instinct, but finagle paths have a restricted character set. And when an unsupported character is in a path element, the entire path element is turned into escaped hex codes! So sometimes Real World paths get really bad really quick.

k8s/src/main/scala/io/buoyant/k8s/Ns.scala
```diff
@@ -15,7 +17,7 @@ abstract class Ns[O <: KubeObject: Manifest, W <: Watch[O]: Manifest, L <: KubeL
   private[this] val caches = Var[Map[String, Cache]](Map.empty[String, Cache])
   // XXX once a namespace is watched, it is watched forever.
   private[this] var _watches = Map.empty[String, Activity[Closable]]
-
+  val noNamespace = "all-namespaces"
```
adleong: I think this class specifically deals with a namescaped API so I don't think it makes sense to make namespace optional here.

esbie:
I confess I don't understand the suggestion. This PR already [defines an api](https://github.com/linkerd/linkerd/pull/1116/files#diff-8150e1d1ed0f27e61b5ab5ffa630e594R50) that extends `Version`. Then that api is used [here](https://github.com/linkerd/linkerd/pull/1116/files#diff-7e056dcc7844eb0f58a507bf1b29cff2R39) to make a non namespaced call.

adleong:
Ah yes, sorry, you're right.  My suggestion is to not create a `Ns` object that uses that API but just use that API directly.  

`Ns` is really purpose built to handle this 2 layer caching where changes to the namespaces are cached separately from changes to the actual resource.  This is necessary for namers where we want the NameTree activity to update independently from the Addr var.  But in this case, we just want a list of ingress resources so I don't think `Ns` is applicable.  

Put another way, we don't really need an `Activity[Map[String, VarUp[Ingress]]]` what we actually want is just an `Activity[Seq[Ingress]]`

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  private[k8s] def isPrefix(pfx: String, uri: String): Boolean = {
+    val pfxSegments = pfx.split("/")
+    val uriSegments = uri.split("/")
+    pfxSegments.corresponds(uriSegments.take(pfxSegments.size)) { _ == _ }
+  }
+
+  private[k8s] def getMatchingPath(hostHeader: Option[String], requestPath: String, ns: Option[String], ingresses: Seq[IngressSpec]): Option[IngressPath] = {
+    ingresses.flatMap { ingressResource =>
+      val matchingPath = ingressResource.rules.find { rule =>
```
adleong: TIOLI: consider moving the path matching logic into a method on the IngressPath class:
```
def matches(hostHeader: Option[String], requestPath: String): Boolean = ???
```

and then you can define a `getMatchingPath` method on the `IngressSpec` class and then finally this method can simply become something like

```
ingresses
  .toIterator // to make this lazy so we can stop after we find a match
  .flatMap(_.getMatchingPath(...)) // see if there's a match
  .take(1) // take the first match
  .toSeq.headOption // convert back to option
```

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  val api = namespace match {
+    case Some(ns) => v1beta1.Api(apiClient).withNamespace(ns).ingresses
+    case None => v1beta1.Api(apiClient).ingresses
+  }
+
+  private[this] object Closed extends Throwable
+  private[this] val state = Var.async[IngressState](Activity.Pending) { state =>
+    val closeRef = new AtomicReference[Closable](Closable.nop)
+    val pending = api.get(None, None, None, retryIndefinitely = true).respond {
```
adleong: I think those first parameters default to None so we can omit them.

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+      case Throw(e) => state.update(Activity.Failed(e))
```
adleong: more idiomatic is to use the syntactic sugar for update which is `state() = Activity.Failed(e)`

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+      case Return(ingressList) =>
+        val initState: Seq[IngressSpec] = ingressList.items.flatMap(item => mkItem(item))
```
adleong: style nit:

`.flatMap(mkItem)`

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+        state.update(Activity.Ok(initState))
+        val (stream, close) = api.watch(None, None, None)
```
adleong: I think we can get the resource version out of the initial ingressList and pass it in here.

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+        closeRef.set(close)
+        val _ = stream.foldLeft(initState) { (ingresses, watchEvent) =>
+          val newState: Seq[IngressSpec] = watchEvent match {
+            case v1beta1.IngressAdded(a) => mkItem(a).map(item => ingresses :+ item).getOrElse(ingresses)
```
adleong: TIOLI: `=> ingresses ++ mkItem(a)`

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+            case v1beta1.IngressModified(m) => mkItem(m).map(item => ingresses.filterNot(i => isNameEqual(i, item)) :+ item).getOrElse(ingresses)
```
adleong: TIOLI:
```
val item = mkItem(m)
val withItemRemoved = ingresses.filterNot(i => item.exists(isNameEqual(i, _)))
withItemRemoved ++ item
```

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+    Closable.make { t =>
+      pending.raise(Closed)
+      Closable.ref(closeRef).close(t)
+    }
+  }
+
+  private[this] val ingresses: Activity[Seq[IngressSpec]] = Activity(state)
+  private[this] def isNameEqual(x: IngressSpec, y: IngressSpec): Boolean = x.name == y.name && x.namespace == y.namespace
+  private[this] def mkItem(ingress: v1beta1.Ingress): Option[IngressSpec] = {
```
adleong: should this be called `mkIngress`?

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+    val namespace = ingress.metadata.flatMap(meta => meta.namespace).getOrElse("default")
+    ingress.spec match {
```
adleong: `ingress.spec.map { spec =>`

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+      case Some(spec) =>
+        val paths = for (
```
adleong: if you use `{` instead of `(` here, you don't need semicolons below

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+        val fallback = spec.backend.map(b => IngressPath(None, None, namespace, b.serviceName, b.servicePort))
+        Some(IngressSpec(ingress.metadata.flatMap(_.name), ingress.metadata.flatMap(_.namespace), fallback, paths))
+      case None => None
+    }
+  }
+
+  def matchPath(hostHeader: Option[String], requestPath: String): Future[Option[IngressPath]] =
+    ingresses.map { cache: Seq[IngressSpec] =>
+      getMatchingPath(hostHeader, requestPath, namespace, cache)
+    }.values.toFuture.flatMap(Future.const)
```
adleong: each time this is called, an observation is opened on the Activity until it produces a non-pending value, which then populates the Future, and closes the observation.  The result of this is that for each call to `matchPath` the Activity will be started and then stopped.  

Since we want identification to be fast, I think we should just hold the Activity open forever.  One way to do this would be to change the `ingresses` val to a lazy val and also hold open an observation:

```
private[this] lazy val ingresses: Activity[Seq[IngressSpec]] = {
  val act = Activity(state)
  val _ = act.states.respond(_ => ()) // register a listener forever to keep the Activity open
  act
}
```

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  private[k8s] def getMatchingPath(hostHeader: Option[String], requestPath: String, ns: Option[String], ingresses: Seq[IngressSpec]): Option[IngressPath] = {
+    ingresses.flatMap { ingressResource =>
+      val matchingPath = ingressResource.rules.find { rule =>
+        (rule.host, rule.uri, rule.namespace) match {
+          case (Some(host), Some(path), ns) =>
+            isPrefix(path, requestPath) && hostHeader.contains(host)
+          case (Some(host), None, ns) => hostHeader.contains(host)
+          case (None, Some(path), ns) => isPrefix(path, requestPath)
```
adleong: FWIW, [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/extensions/v1beta1/types.go#L690) says that the path is actually a regex.  We don't necessary have to support that, but it's worth being aware of.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/IngressIdentifier.scala
```diff
@@ -0,0 +1,68 @@
...
+  apiClient: Service[http.Request, http.Response]
+) extends Identifier[Request] {
+
+  private[this] val unidentified: RequestIdentification[Request] =
+    new UnidentifiedRequest(s"no ingress rule matches")
+
+  private[this] val ingressCache = new IngressCache(namespace, apiClient)
+
+  override def apply(req: Request): Future[RequestIdentification[Request]] = {
+    val hostHeader = req.headers.get("Host").lastOption
```
adleong: I think this should use `com.twitter.finagle.buoyant.h2.Headers.Authority` as the header namer

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/IngressIdentifier.scala
```diff
@@ -0,0 +1,68 @@
...
+    val matchingPath = ingressCache.matchPath(hostHeader, req.path)
+    matchingPath.flatMap { paths =>
```
adleong: flatMap and match can be collapsed:

```
matchingPath.flatMap {
  case None => 
  case Some(path) =>
}
```

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/IngressIdentifier.scala
```diff
@@ -0,0 +1,68 @@
...
+  port: Option[Port],
+  namespace: Option[String]
+) extends H2IdentifierConfig with ClientConfig {
+  override def portNum: Option[Int] = port.map(_.port)
+
+  @JsonIgnore
+  override def newIdentifier(params: Stack.Params) = {
+    val DstPrefix(pfx) = params[DstPrefix]
+    val BaseDtab(baseDtab) = params[BaseDtab]
+    val client = mkClient(Params.empty).configured(Label("ingress-identifier"))
```
adleong: probably good practice to pass `params` into `mkClient`.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IngressIdentifier.scala
```diff
@@ -0,0 +1,69 @@
...
+  apiClient: Service[http.Request, http.Response]
+) extends Identifier[Request] {
+
+  private[this] val unidentified: RequestIdentification[Request] =
+    new UnidentifiedRequest(s"no ingress rule matches")
+
+  private[this] val ingressCache = new IngressCache(namespace, apiClient)
+
+  override def apply(req: Request): Future[RequestIdentification[Request]] = {
+    val hostHeader = req.headerMap.get("Host")
```
adleong: `req.host`

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
esbie: Oh cool, I hadn't seen this comment. I'll switch to regex

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
esbie: Ah ok, I found `state()` more confusing since I'm not too familiar with `Updatable`. I'll switch them out.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/IngressIdentifier.scala
esbie: good catch!

esbie:
Thanks for the style review! Looks a lot better now

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,144 @@
...
+) {
+  def uriMatches(uri: String, path: String): Boolean = path.isEmpty || uri.matches(path)
+
+  def matches(hostHeader: Option[String], requestPath: String) = {
+    (host, uri) match {
+      case (Some(host), Some(path)) =>
+        uriMatches(requestPath, path) && hostHeader.contains(host)
+      case (Some(host), None) => hostHeader.contains(host)
+      case (None, Some(path)) => uriMatches(requestPath, path)
+      case (None, None) => false
```
adleong: I think if no host or path is specified, the rule should match everything.

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -0,0 +1,144 @@
...
+}
+
+case class IngressPath(
+  host: Option[String] = None,
+  uri: Option[String] = None,
+  namespace: String,
+  svc: String,
+  port: String
+) {
+  def uriMatches(uri: String, path: String): Boolean = path.isEmpty || uri.matches(path)
```
adleong: for performance reasons we may want to compile this regex and store it as a val on the class.

PR #1117, opened 2017-03-02
Add support for weights to the io.l5d.fs namer

olix0r:
Problem

Finagle supports the ability to specify a "weight" on an address, which impacts
load balancing decisions. Most of our namers don't expose any weighting
functionality, though.

Solution

As a first step towards broader support for endpoint weighting, we extend the
io.l5d.fs namer to accept an optional weight.

Now addresses are specified in the following format:

    HOST PORT ['*' WEIGHT]

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -55,7 +55,7 @@ apps    users   web
 $ cat config/web
 192.0.2.220 8080
 192.0.2.105 8080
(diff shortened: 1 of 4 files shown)


olix0r:
@ashald pull requests accepted 😉

Ashald:
@olix0r deal :D

PR #1122, opened 2017-03-02
Make long dtabs look a little better

### LINKED ISSUE #1098: Dtab in Namerd section of Linkerd UI v0.9 "wraps" long lines
Although long lines are not good the "view width" seems to be to narrow
screen shot 2017-02-23 at 13 09 43

rmars:
Fixes #1098

I also aligned the dtabs vertically. If you don't like this lmk.

Before:
screen shot 2017-03-02 at 3 15 20 pm

After:

screen shot 2017-03-02 at 3 42 52 pm

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
@@ -55,7 +55,6 @@ define(['jQuery'], function($) {
   DtabViewer.prototype._resize = function($prefix, $dst) {
     var halfPageWidth = ($(window).width() / 2) * 0.85;
     var maxWidth = 0;
-    var minWidth = halfPageWidth;
 
     $dst.width(halfPageWidth); // expand to find max content width
     $dst.find(".dst-content").map(function(_i, ea) {
--- a/admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
@@ -61,7 +61,6 @@ define(['jQuery'], function($) {
     $dst.find(".dst-content").map(function(_i, ea) {
       var w = $(ea).width();
       maxWidth = !w ? this.seenSoFar : w > maxWidth ? w : maxWidth;
-      console.log({maxWidth});
     }.bind(this));
 
     if (maxWidth < halfPageWidth) {
--- a/admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
@@ -6,9 +6,8 @@ define(['jQuery'], function($) {
     this.template = dentryTemplate;
 
     this.render();
-
-    this._resize($(".dentry-prefix"), $(".dentry-dst"))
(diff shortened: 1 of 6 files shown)


adleong:
Stretching a tiny dtab to fit the full width of the page looks a little goofy imo.

screen shot 2017-03-02 at 4 20 33 pm

rmars:
Ok.  This is how it looks if I remove the columns. For dtabs with more than one entry, I feel like columns are easier to read. wdyt? We could remove the columns for things with only one entry, though I'm sure if you have multiple short entries, it'd still look weird.

Long dentries (no columns):
screen shot 2017-03-02 at 5 58 58 pm

Long dentries (columns):
screen shot 2017-03-02 at 6 06 57 pm

Many short dentries (no columns):
screen shot 2017-03-02 at 6 03 43 pm


Many short dentries (columns, still looks strange):
screen shot 2017-03-02 at 6 04 09 pm

esbie:
ok honestly I think the best approach is a combination of the two. we find the longest line and then set that same length for the rest of the pfx&dsts. wdyt

rmars:
Okay, took a stab at the combined approach!

admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
```diff
@@ -49,9 +52,30 @@ define(['jQuery'], function($) {
...
   };
 
+  DtabViewer.prototype._resize = function($prefix, $dst) {
+    var halfPageWidth = ($(window).width() / 2) * 0.85;
+    var maxWidth = 0;
+
+    $dst.width(halfPageWidth); // expand to find max content width
+    $dst.find(".dst-content").map(function(_i, ea) {
+      var w = $(ea).width();
+      maxWidth = !w ? this.seenSoFar : w > maxWidth ? w : maxWidth;
```
esbie: tioli add parens around the inner ternary for readability

admin/src/main/resources/io/buoyant/admin/js/src/namerd.js
```diff
@@ -13,6 +13,8 @@ define([
   Delegator,
   templates
 ) {
+  var limitWidth = $(window).width() / 2 * 0.8;
```
esbie: multiplier is 0.8 here but 0.85 in `_resize`, intentional?

admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
```diff
@@ -20,6 +22,7 @@ define(['jQuery'], function($) {
         return { prefix: tuple[0], dst: tuple[1] };
       });
       this.render();
+      $(window).trigger("resize");
```
esbie: I'm a little confused here since triggering `resize` invokes `render`, but `render` is already invoked in the previous line...

admin/src/main/resources/io/buoyant/admin/js/src/namerd.js
rmars: yep, intentional - though maybe less confusing if I use `limitWidth = $namespacesContainer.width() / 2 - 60;`

admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
rmars: I think this is unnecessary and I forgot to remove it...

PR #1126, opened 2017-03-07
Upgrade to finagle 6.44.0

### LINKED ISSUE #1107: Upgrade to finagle 6.42.0

adleong:
#Fixes #1107


finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -210,7 +210,6 @@ object H2FrameCodec {
       val fr = new Http2InboundFrameLogger(new DefaultHttp2FrameReader, frameLogger)
       new DefaultHttp2ConnectionDecoder(conn, encoder, fr)
     }
-    decoder.localSettings(settings)
```
adleong: @olix0r: this API has been removed.  what was the intention behind setting the decoder's local settings?  does removing this cause any problems?

olix0r: Well, I *think* it prevents any of the configured http2 settings from being sent during connection establishment.  We're going against the grain in netty (as you'll see at the top of this file}, so we probably need to do some poking to figure out if this feature has been removed from netty entiirely.  I don't think it's a hard blocker, but we may need to remove a bunch of the settings from the config.

olix0r: Actually, I may be wrong about that. Would be best to setup a simple test -- i usually run `nghttpd --no-tls` and have linkerd connect to it.  nghttpd will dump all of the settings received, which will be helpful.  The `nghttp` client can be used to initiate an h2 request through linkerd, and will dump any settings received.

adleong:
This is blocked until https://github.com/twitter/finagle/commit/08c2171dd2189c0bc3eddd5b7d8db67b03274fe5 makes it into a finagle release.

olix0r:
It looks like [Finagle 6.43](https://finagle.github.io/blog/2017/03/14/release-notes-6-43/) is out.  Are we waiting for 6.44?

adleong:
Yeah, the above commit just barely missed 6.43

PR #1135, opened 2017-03-10
Set zipkin service name to host header on http requests

### LINKED ISSUE #1132: Improve service name annotations in Zipkin UI
The Zipkin UI throws errors on service names that include transformers (e.g. `%/io.l5d.k8s.daemonset/default/incoming/l5d/#/io.l5d.k8s/default/http/hello`). Here's a sample error:

screen shot 2017-03-08 at 3 02 17 pm

The dependency graph also looks pretty crazy, and the service names include node IPs, which makes for a pretty convoluted graph:

screen shot 2017-03-07 at 3 51 22 pm

To fix this, I think we should switch away from using `dst.id` as the annotation endpoint service name on client spans, and instead use something more meaningful.

klingerf:
## Problem

Finagle's default ClientTracingFilter sets the `ServiceName` annotation on traced client spans to be the value of `param.Label` on the client stack. In Linkerd's case, `param.Label` is the destination ID, but our destination IDs don't display well in the Zipkin UI. For instance, a traced request to the hello world services running in kubernetes daemonsets:

screen shot 2017-03-09 at 4 07 53 pm

## Solution

For HTTP requests, override the client span's `serviceName` to be the HTTP Host header, if available. For comparison, the same hello world request:

screen shot 2017-03-09 at 4 14 27 pm

We might want to consider doing this for other protocols as well, but I didn't do that as part of this branch. Fixes #1132.

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/TracingFilter.scala
@@ -41,7 +41,9 @@ class TracingFilter extends SimpleFilter[Request, Response] {
       Trace.recordBinary("http.uri", req.uri)
       Trace.recordBinary("http.req.method", req.method.toString)
       req.host match {
-        case Some(h) => Trace.recordBinary("http.req.host", h)
+        case Some(h) =>
+          Trace.recordServiceName(h)
+          Trace.recordBinary("http.req.host", h)
         case None =>
       }
       Trace.recordBinary("http.req.version", req.version.toString)
--- a/router/http/src/test/scala/io/buoyant/router/http/TracingFilterTest.scala
+++ b/router/http/src/test/scala/io/buoyant/router/http/TracingFilterTest.scala
@@ -34,6 +34,7 @@ class TracingFilterTest extends FunSuite with Awaits {
 
       val reqEvents = tracer.iterator.toSeq
       assert(reqEvents.exists(_.annotation == Annotation.Rpc("HEAD /foo")))
+      assert(reqEvents.exists(_.annotation == Annotation.ServiceName("monkeys")))
       assert(reqEvents.exists(_.annotation == Annotation.BinaryAnnotation("http.uri", "/foo?bar=bah")))
       assert(reqEvents.exists(_.annotation == Annotation.BinaryAnnotation("http.req.method", "HEAD")))
       assert(reqEvents.exists(_.annotation == Annotation.BinaryAnnotation("http.req.host", "monkeys")))


adleong:
How would this look in the case of path based routing?  Would all the service names just be `localhost:4140`?  What about using the logical name instead?  e.g. `/svc/hello`

klingerf:
@adleong Good point about the path identifier. I had considered using the logical name for the client span's `ServiceName` annotation, but we're already surfacing that as part of the parent server span's `Rpc` annotation. So it felt a bit redundant to include it again.

I tend to think about the division between server annotations and client annotations as being protocol-agnostic and protocol-specific. The server annotations are universally the same, regardless of protocol (e.g. `dst.id`, `dst.path`, `router.label`, etc.). And the client annotations vary widely by protocol (e.g. `http.req.method`, `http.uri`, etc.). So my preference would be to use a protocol-specific `ServiceName` annotation as well, and I think host header makes the most sense, even in the context of the path identifier.

I grabbed some sample traces that show the current behavior for the path identifier, and the proposed new behavior. Here's what a trace of a demo app running locally using the path identifier and the current tracing code from master:

screen shot 2017-03-14 at 12 01 40 pm

You can see the dst ids in the Services column (in this demo setup they map pretty closely to the logical service names, but that's not the case in most environments, and especially not when transformers are present). Here's the same trace with the proposed change from this branch:

screen shot 2017-03-14 at 1 18 35 pm

I can't say this approach is necessarily better when using the path identifier, but it feels more consistent to me, at least. Anyway, happy to discuss more options too. Just wanted to expand on my initial reasoning for this change.

adleong:
Thanks for the context.  I think it makes sense that there are protocol-specific client annotations since those annotations describe properties of the (protocol-specific).  Service name, on the other hand, represents the name of the service that the request is destined for; and that doesn't seem like a protocol-specific thing to me.  

I think my personal preference would be to keep the current behavior of naming the service name on the client span as the concrete id.  It's inscrutable, but I think it's most correct and I prefer inscrutable to misleading.  My worry with hostname is that the service name looks like a resolvable address when it isn't one.  For example, in the screenshot above, spans with the service name "localhost:8080" are not actually destined for localhost:8080 at all.  I think it can be the job of additional tooling on top of the zipkin data to distill more human-readable information out of it and let the zipkin data itself just be low-level and accurate.

That's just my opinion, I'm happy to be outvoted on this.

klingerf:
Ok, thanks for considering it. That works for me -- will close this PR.

If I get a chance I'll submit a separate PR to Zipkin so that their UI properly escapes our service names. Otherwise we'll see the "Required String parameter 'serviceName' is not present" error that's shown in #1132.

It's also a bit unfortunate that dst ids using the localnode transformer include the node IP. That means that we end up with a separate serviceName for each node in our cluster, even if the destinations are targeting the same service. I can't think of another way to fix that though, but will leave #1132 open to track it.

PR #1137, opened 2017-03-10
Allow k8s namer to accept port numbers and make it case insensitive

### LINKED ISSUE #1123: k8s namer should be case-agnostic
k8s services (all resources?) start must start with lowercase letters.  grpc services tend to have uppercase names.  If we make the k8s namer case-agnostic, this will all work smoothly with a simple:
```
/grpc/com/foo => /#/io.l5d.k8s/default/grpc ;
/svc => /$/io.buoyant.http.domainToPathPfx/grpc ;
```
to route a grpc service like com.foo.Bar to the "bar"service.

cc @christopheclc

### LINKED ISSUE #1124: K8s namer should accept both named ports and port numbers
The k8s namer accepts names of the form `/#/io.l5d.k8s/default/http/foo` which uses the port named `http`.  It should also be able to accept a port number in the same position: `/#/io.l5d.k8s/default/8888/foo`.

adleong:
Fixes #1124 
Fixes #1123

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -26,23 +26,28 @@ class EndpointsNamer(
    * and attempts to bind an Addr by resolving named endpoint from the
    * kubernetes master.
    */
-  def lookup(path: Path): Activity[NameTree[Name]] = (path.take(variablePrefixLength), labelName) match {
-    case (id@Path.Utf8(nsName, _, _), None) =>
-      val residual = path.drop(variablePrefixLength)
-      log.debug("k8s lookup: %s %s", id.show, path.show)
-      lookupServices(endpointNs.get(nsName.toLowerCase, None), id, residual)
+  def lookup(path: Path): Activity[NameTree[Name]] = {
+    val lowercasePath = path.take(variablePrefixLength) match {
+      case Path.Utf8(segments@_*) => Path.Utf8(segments.map(_.toLowerCase): _*)
+    }
+    (lowercasePath, labelName) match {
+      case (id@Path.Utf8(nsName, _, _), None) =>
+        val residual = path.drop(variablePrefixLength)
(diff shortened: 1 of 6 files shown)


k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -52,24 +52,29 @@ class EndpointsNamer(
...
+              case None =>
+                val state: Var[Activity.State[NameTree[Name.Bound]]] = service.port(portName.toLowerCase).map {
+                  case Some(addr) =>
+                    log.debug("k8s ns %s service %s port %s found + %s", nsName.toLowerCase, serviceName.toLowerCase, portName.toLowerCase, residual.show)
+                    Activity.Ok(NameTree.Leaf(Name.Bound(addr, idPrefix ++ id, residual)))
+                  case None =>
+                    log.debug("k8s ns %s service %s port %s missing", nsName.toLowerCase, serviceName.toLowerCase, portName.toLowerCase)
+                    Activity.Ok(NameTree.Neg)
+                }
+                Activity(state)
```
olix0r: my only concern here is that we seem to be calling toLowerCase _a lot_, which makes me worried that it's easy to miss this in one place..  Is there any way we can do this somewhere once than throughout the code?  For instance, can we just alter the Path once at the start?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -102,112 +111,72 @@ class EndpointsNamer(
...
+  private[this] def getPorts(subsets: Seq[v1.EndpointSubset]): Map[String, Int] = {
+    val portSet = mutable.Map.empty[String, Int]
+    for {
+      subset <- subsets
+      ports <- subset.ports
+      port <- ports
+      name <- port.name
+    } {
+      val proto = port.protocol.map(_.toUpperCase).getOrElse("TCP")
+      if (proto == "TCP") {
```
olix0r: tioli: i like consts

PR #1138, opened 2017-03-11
Add protoc script which downloads and runs protoc

### LINKED ISSUE #1033: fetch protoc binary on-demand
The linkerd repo now has a build-time on `protoc`, but we don't fetch this dependency as part of the build.

We can add a ./protoc script that works sort of like ./sbt:
- detect OS (I.e. via `uname`
- check if we already have a protoc binary cached locally
- if not, fetch appropriate binary from release and cache locally
- verify SHA of fetched binary

Then, our grpc sbt config can be changed to use `./protoc` rather than look for protoc on the PATH.

adleong:
Fixes #1033 

The `./protoc` script fetches and cashes the correct version of `protoc` as `.protoc` and runs it.  Note that the binaries differ depending on os and arch so we don't checksum the binary.

### DIFF
--- a/protoc
+++ b/protoc
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -e
+
+if [ `uname -s` = "Darwin" ]; then
+  os=osx
+else
+  os=linux
+fi
+arch=`uname -m`
+
+protocbin=.protoc
+protocversion=3.0.0
+protocurl=https://github.com/google/protobuf/releases/download/v$protocversion/protoc-$protocversion-$os-$arch.zip
+
+if [ ! -f $protocbin ]; then
+  echo "downloading $protocbin" >&2
+  tmp=`mktemp -d`
+  curl -L --silent --fail -o $tmp/$protocbin.zip $protocurl
+  unzip -q $tmp/$protocbin.zip -d $tmp
+  mv $tmp/bin/protoc $protocbin
+  rm -rf $tmp
+fi
+
+./$protocbin $@
--- a/protoc
+++ b/protoc
@@ -2,24 +2,24 @@
 
 set -e
 
-if [ `uname -s` = "Darwin" ]; then
+if [ "$(uname -s)" = "Darwin" ]; then
   os=osx
 else
   os=linux
 fi
-arch=`uname -m`
+arch=$(uname -m)
 
 protocbin=.protoc
 protocversion=3.0.0
-protocurl=https://github.com/google/protobuf/releases/download/v$protocversion/protoc-$protocversion-$os-$arch.zip
+protocurl="https://github.com/google/protobuf/releases/download/v${protocversion}/protoc-${protocversion}-${os}-${arch}.zip"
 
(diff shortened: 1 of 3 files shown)


protoc
```diff
@@ -0,0 +1,25 @@
...
+fi
+arch=`uname -m`
+
+protocbin=.protoc
+protocversion=3.0.0
+protocurl=https://github.com/google/protobuf/releases/download/v$protocversion/protoc-$protocversion-$os-$arch.zip
+
+if [ ! -f $protocbin ]; then
+  echo "downloading $protocbin" >&2
+  tmp=`mktemp -d`
```
klingerf: In my shell this requires a template argument -- e.g. `mktemp -d -t protoc.XXXXXX`

```
$ mktemp -d
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 
$ echo $?
1
```

protoc
```diff
@@ -0,0 +1,25 @@
...
+
+protocbin=.protoc
+protocversion=3.0.0
+protocurl=https://github.com/google/protobuf/releases/download/v$protocversion/protoc-$protocversion-$os-$arch.zip
+
+if [ ! -f $protocbin ]; then
+  echo "downloading $protocbin" >&2
+  tmp=`mktemp -d -t protoc.XXX`
+  curl -L --silent --fail -o $tmp/$protocbin.zip $protocurl
+  unzip -q $tmp/$protocbin.zip -d $tmp
```
olix0r: it would be Nice to check the sha of the binary (like we do for sbtlaunch) to check i.e. if the urlc hangges.

protoc
```diff
@@ -0,0 +1,25 @@
...
+if [ `uname -s` = "Darwin" ]; then
+  os=osx
+else
+  os=linux
+fi
+arch=`uname -m`
+
+protocbin=.protoc
+protocversion=3.0.0
+protocurl=https://github.com/google/protobuf/releases/download/v$protocversion/protoc-$protocversion-$os-$arch.zip
```
olix0r: prefer ${protocversion} to make it clearer.  Also we should probably quote this Just In Case.

protoc
```diff
@@ -0,0 +1,25 @@
...
+
+if [ ! -f $protocbin ]; then
```
olix0r: probably should quote protocbin

protoc
```diff
@@ -0,0 +1,25 @@
...
+  echo "downloading $protocbin" >&2
+  tmp=`mktemp -d -t protoc.XXX`
+  curl -L --silent --fail -o $tmp/$protocbin.zip $protocurl
+  unzip -q $tmp/$protocbin.zip -d $tmp
+  mv $tmp/bin/protoc $protocbin
+  rm -rf $tmp
+fi
+
+./$protocbin $@
```
olix0r: probably should quote $@

protoc
adleong: would be a different sha depending on os and arch so I didn't bother.

PR #1147, opened 2017-03-17
fix small typo in README

jtaylor32:


wmorgan:
Thanks @jtaylor32!

jtaylor32:
No problem @wmorgan. I think the releases link is broken as well. I think it needs to reference (/releases/latest) rather than just release. 

https://help.github.com/articles/linking-to-releases/

I can fix that as well if you'd like.

jtaylor32:
No problem @adleong!

PR #1148, opened 2017-03-17
Add support for Marathon HTTP basic authentication

hhtpcd:
Running standalone Marathon (without DCOS), it is possible to [enable HTTP basic authentication](https://mesosphere.github.io/marathon/docs/ssl-basic-access-authentication.html#enabling-basic-access-authentication). Currently Linkerd only supports getting access tokens from DCOS. This is my first time on a Scala codebase but I've got HTTP authentication working to Marathon, but it's probably not very neat or tidy. 

Keen to get Linkerd working with HTTP auth as hoping to be able to use it in a Mesos/Marathon env without DCOS, so very happy to work on the implementation.

### DIFF
--- a/namer/marathon/src/main/scala/io/buoyant/namer/marathon/MarathonInitializer.scala
+++ b/namer/marathon/src/main/scala/io/buoyant/namer/marathon/MarathonInitializer.scala
@@ -59,7 +59,8 @@ case class MarathonSecret(
  *   - https://github.com/mesosphere/universe/search?utf8=%E2%9C%93&q=DCOS_SERVICE_ACCOUNT_CREDENTIAL
  */
 object MarathonSecret {
-  val EnvKey = "DCOS_SERVICE_ACCOUNT_CREDENTIAL"
+  val DCOSEnvKey = "DCOS_SERVICE_ACCOUNT_CREDENTIAL"
+  val basicEnvKey = "MARATHON_HTTP_AUTH_CREDENTIAL"
 
   case class Invalid(secret: MarathonSecret) extends NoStackTrace
 
@@ -71,7 +72,7 @@ object MarathonSecret {
   }
 
   def load(): Option[MarathonSecret] =
-    sys.env.get(EnvKey) match {
+    sys.env.get(DCOSEnvKey) match {
       case None => None
       case Some(json) =>
         Api.readJson[MarathonSecret](Buf.Utf8(json)) match {
@@ -125,11 +126,15 @@ case class MarathonConfig(
       .filtered(SetHost(host0))
       .newService(dst0)
 
-    val service = MarathonSecret.load() match {
-      case None => client
-      case Some(secret) =>
+    val service = (MarathonSecret.load(), sys.env.get(MarathonSecret.basicEnvKey)) match {
+      case (Some(secret), _) =>
         val auth = MarathonSecret.mkAuthRequest(secret)
         new Authenticator.Authenticated(client, auth)
+      case (None, Some(http_auth_token)) =>
+        val filter = new BasicAuthenticatorFilter(http_auth_token)
+        filter.andThen(client)
+      case (None, None) => client
+
     }
 
     val uriPrefix0 = uriPrefix.getOrElse("")
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -376,7 +376,7 @@ appId | yes | The app id of a marathon application. This id can be multiple path
 
 ### Marathon Authentication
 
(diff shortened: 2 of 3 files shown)


marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -175,7 +175,7 @@ object Api {
     }
 }
 
-private class AppIdApi(client: Api.Client, apiPrefix: String, useHealthCheck: Boolean)
+private class AppIdApi(client: Api.Client, apiPrefix: String, useHealthCheck: Boolean, httpAuth: String)
```
adleong: I think this should be an Option[String] throughout and only add the header if the Option is a Some.

hhtpcd:
Thanks both for the feedback, glad to hear this could be worthwhile. Agree about the env var however I don't know enough about how DCOS authentication is set up to know if it can be configured with HTTP auth at the same time, I'll look into that and try to justify exclusivity, or lack of.

Will rework adding the auth header as well, into the existing initializer.

siggy:
Thanks for digging into this! I think it's totally acceptable to make these settings mutually exclusive. My assumption is that if someone is running DC/OS, they can use either setting, and if they're just on Marathon/Mesos, they can use this new setting.

nm-harry:
I've moved the basic auth into the existing Authenticator flow and set up an environment variable, on the assumption that you can use one or the other. I haven't included any checks for providing both. 

Let me know if this is the sort of implementation you were thinking of. As before, I expect this is not ideal Scala so keen to get feedback. If this is the right direction, I can amend tests etc after.

hhtpcd:
I agree with the suggestions and pulling the basic auth out into a separate object as it looks like it will keep the auth code tidier. My understanding is that DCOS is more of a focus for Mesosphere than standalone Mesos/Marathon atm so the Basic auth option may go away or become a lot less relevant/supported. Keeping it separate will help further down the line if it needs to be removed.

I expect the match block after `MarathonSecret.load()` is not ideal, but conceptually feels much nicer than forcing everything into `AuthRequest`. Thanks for the continued feedback, it's much appreciated.

namer/marathon/src/main/scala/io/buoyant/namer/marathon/BasicAuthenticator.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  private[marathon] case class UnauthorizedResponse(rsp: http.Response) extends Throwable
+  private[this] val closedException = Failure("closed", Failure.Interrupted)
+  private[this] val closedExceptionF = Future.exception(closedException)
+
+  class Authenticated(client: Api.Client, http_auth_token: String) extends Api.Client {
+    private[this] sealed trait State
+    private[this] object Closed extends State
+    private[this] object Init extends State
+    private[this] val state = new AtomicReference[State](Init)
```
siggy: i think we can do away with most of the state management, as most of that was motivated by the dc/os auth module hold the state of the authentication flow.

```scala
class BasicAuthenticatorFilter(http_auth_token: String) extends Filter[http.Request, http.Response] {
  def apply(req: http.Request, svc: Service[http.Request, http.Response]): Future[http.Response] = {
    req.headerMap.set("Authorization", s"Basic $token")
    svc(req)
  }
}
```

hhtpcd:
Following your comments, I've removed all the added logic from the `Authenticator` object for basic auth and implemented the `BasicAuthenticatorFilter` as suggested, which I now realise is very similar to the Consul API namer.

namer/marathon/src/main/scala/io/buoyant/namer/marathon/MarathonInitializer.scala
```diff
@@ -125,11 +126,16 @@ case class MarathonConfig(
       .filtered(SetHost(host0))
       .newService(dst0)
 
-    val service = MarathonSecret.load() match {
-      case None => client
-      case Some(secret) =>
-        val auth = MarathonSecret.mkAuthRequest(secret)
+    val service = (sys.env.get(MarathonSecret.DCOSEnvKey), sys.env.get(MarathonSecret.basicEnvKey)) match {
```
adleong: You can update these two lines to:

```
val service = (MarathonSecret.load(), sys.env.get(MarathonSecret.basicEnvKey)) match {
  case (Some(secret), _) =>
```

This way `secret` is a String instead of an `Option[String]` and you don't need to call `.get` below.

hhtpcd:
I've updated the docs with the env var details, but it looks like the tests failed with an inotify error.

PR #1155, opened 2017-03-23
Allow per-client configuration

adleong:
This is a config breaking change (see TLS configuration, below)!

Goal: to be able to apply client configuration more narrowly on a per-client basis.

The `client` section of a router config now admits a `kind` property (`io.l5d.default` is assumed if no kind is specified).

## io.l5d.default

All client configuration is applied globally to all created clients.

```
routers:
- protocol: http
  client:
    # kind: io.l5d.default is implied
    loadBalancer:
      kind: ewma
    hostConnectionPool:
      maxSize: 100
```

## io.l5d.static

With the `io.l5d.static` config, a list of client config objects must be specified under the `configs` property.  Each of these configs must specify a prefix.  The client config options specified in that config will apply only to clients which match the prefix.

```
routers:
- protocol: http
  client:
    kind: io.l5d.static
    configs:
    - prefix: /#/io.l5d.fs
      loadBalancer:
        kind: ewma
      hostConnectionPool:
        maxSize: 100
    - prefix: /#/io.l5d.fs/dog
      loadBalancer:
        kind: p2c
```

## TLS configuration

Since we can now apply configurations on a per-client basis, there is no longer a need for the pluggable `boundPath` tls config.  Instead we can use simpler per-client tls configs.  We adopt the same config structure as `MeshClientTlsConfig`.  This is a config breaking change.

```
routers:
- protocol: http
  client:
    kind: io.l5d.static
    configs:
    - prefix: /#/io.l5d.fs/dog
      tls:
        commonName: dog.linkerd.io
        trustCerts:
        - /certs/cacert.pem
    - prefix: /#/io.l5d.fs/cat
      tls:
        commonName: cat.linkerd.io
        trustCerts:
        - /certs/cacert.pem
```

You can also use variable capture from the prefix in commonNames:

```
routers:
- protocol: http
  client:
    kind: io.l5d.static
    configs:
    - prefix: "/#/io.l5d.fs/{service}"
      tls:
        commonName: "{service}.linkerd.io"
        trustCerts:
        - /certs/cacert.pem
```

## Implementation Notes

Since we now have the ability to pass different Stack.Params to each client stack, there's no longer a need for a TlsClientPrep module.  Instead, we can simply set the TransportSecurity and Trust params directly on each client stack as appropriate.  Therefore we have removed this module.


esbie:
Re config format: love the tls changes. Very surprised we haven't named anything `io.l5d.default` until now! This is the only place I know of where the `kind:` field is optional. Since this is going to be a breaking change anyway, I propose we rename `io.l5d.default` to `io.l5d.global` and make the `kind:` field required.

olix0r:
@esbie plugin names are not in a global namespace, so it's possible that we have things called io.l5d.default in other interfaces (i'm not sure).

@siggy there will be more kinds, possibly with configs sections

esbie:
@olix0r right, it's possible, just hasn't happened up until this point (I already grepped for it). Either way, the name `io.l5d.global` gives better context

adleong:
@siggy: Anything is possible 🌈 but as @olix0r mentions there will be a number of different kinds that this block can be (global, per-client static, file watching, api watching, etc)

@esbie io.l5d.global is a good suggestion, thanks.  I think we can still make io.l5d.global the default kind so that it doesn't need to be specified.  That way, we keep the TLS blocks as the only breaking change.  If you don't have a TLS block, your config won't need to be updated.

adleong:
@olix0r from the client config docs:
>If a client matches more than one config's prefix, all parameters from the
>matching configs will be applied, with parameters from later configs taking
>precedence.

I will add an example as well.

PR #1159, opened 2017-03-24
Warn users that dtab playground saves are temporary

### LINKED ISSUE #1158: admin interface dtabs "edit/save" does not change dtab rules
It is not clear that editing the dtab rules is for the purposes of the playground only.

rmars:
* Warn users that dtab playground saves are temporary
* Resize the dtab text area when edit is clicked - long dtabs overflow this

screen shot 2017-03-24 at 1 13 03 pm

Fixes #1158


wmorgan:
I know I'm late to this, but I think changing the "Save*" to "Apply" might help. Save implies storage, at least to my ancient brain.

(I also think the path should be underneath the dtab area, but that's another discussion)

PR #1162, opened 2017-03-25
Support case-insensitive naming for Marathon services

### LINKED ISSUE #1161: Marathon namer: make case insensitive
The namer should be case insensitive, e.g. `Bar` and `bar` will both resolve to the same service. 

This is especially useful for `gRPC`, where services often are named `com.foo.Bar` whilst the service in Marathon is named `bar`.

macalinao:
Fixes #1161

### DIFF
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
@@ -42,8 +42,10 @@ class AppIdNamer(
       val possibleIds = (1 to path.size).map(path.take(_))
       appsActivity.map { apps =>
         Trace.recordBinary("marathon.path", path.show)
-        val found = possibleIds.collectFirst {
-          case app if apps(app) =>
+        val found = possibleIds.toStream.map { app =>
+          apps.find(_.show.toLowerCase == app.show.toLowerCase)
+        }.collectFirst {
+          case Some(app) =>
             Trace.recordBinary("marathon.appId", app.show)
             val residual = path.drop(app.size)
             val id = prefix ++ app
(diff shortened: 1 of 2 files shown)


marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -42,8 +42,10 @@ class AppIdNamer(
       val possibleIds = (1 to path.size).map(path.take(_))
       appsActivity.map { apps =>
         Trace.recordBinary("marathon.path", path.show)
-        val found = possibleIds.collectFirst {
-          case app if apps(app) =>
+        val found = possibleIds.toStream.map { app =>
```
macalinao: Let me know if there's a more idiomatic way to do this.

macalinao:
^Hmm looks like CircleCI failed, but not because my test didn't work. I didn't run the test locally-- does anyone have any tips on how to do this?

olix0r:
that's an odd error. i hit rebuild, which i expect to fix it

macalinao:
Looks like the end-to-end tests are still OOMing... 🤔 Could my change be the reason?

olix0r:
@macalinao it's definitely not your branch. seems like something's flakey in the ci running scripts. i've bounced it again, but in the worst case we'll test locally before merging.  (I'd expect early this coming week.)

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
adleong: Another option would be to call `.toLowerCase` before `Path.read` in `Api.toAppIds` to make sure all ids in `appsActivity` are lowercase.  Then we can convert `path` to lowercase by doing 

```
val Path.Utf8(segs@_*) = path
val lowercasePath = Path.Utf8(segs.map(_.toLowerCase): _*)
```

this way we can avoid relying on `path.show` which is probably unreliable for equality

adleong:
hey @macalinao, what are your thoughts on my suggestion above?  other than that one comment this is looking great and I'd like to get it merged soon.

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
macalinao: I'm not sure why `path.show` is unreliable for equality.

macalinao:
@adleong Sorry for getting back to you so late. I see you've already merged things.

PR #1165, opened 2017-03-27
Move timeoutMs from router section to client section in linkerd config.

adleong:
So that per-client timeouts can be specified.


adleong:
I'm realizing now that this setting used to set a timeout both on the client stack AND on the server stack.  This meant that if a timeout was set, each individual try of that request was subject to that timeout but also the entire request including all retries was also subject to the same timeout.  We have a few options on how to proceed:

1. Move the timeout into the client config and remove the server stack timeout entirely.  This is what this change has done so far.
2. Move the timeout into the client config but also allow a timeout to be configured per-server in the server config.
3. Keep the timeout in the router config where it will apply to both the client and server stacks, but allow it to be overrideable in the client config.
4. Keep the timeout in the router config where it will apply to both the client and server stacks, but allow it to be overrideable in the client config AND overrideable in the server config.

3 and 4 would be non-breaking changes since they allow a timeout in the router block.  4 is most flexible, but also potentially the most confusing.  3 may be a good balance of flexibility and clarity.

siggy:
i like 2. it provides the same level of flexibility as 3 and 4, correct?

overrides within a config get super confusing, it's never obvious which things affect which other things.

as we're aiming for 1.0, i'm all for breaking changes right now.

klingerf:
👍 to @siggy's comment. I think the existing behavior of setting the client and server timeout to the same value is incorrect, since the client requests could be retried. Moving the timeout configurations into the client and server sections of the config makes the most intuitive sense to me.

adleong:
Ok, as suggested here I have implemented option 2.  

`timeoutMs` in the client section now configures a timeout for individual requests (tries) which can be configured on a per-client basis.  

`timeoutMs` in the server section now configures a timeout for the whole request including all retries (note that retries may happen on different clients and each try is subject to the timeout of that client that it is tried on).  

`timeoutMs` can not be specified directly on the router section.

adleong:
Client timeouts are per-try (pre-retry) and server timeouts are total across all retries (post-retry).  In your example:

Suppose the request is dispatched to client A (which has a timeout of 1s) after 1s has elapsed and no response has been received, the client's 1s timeout is hit and the request is retried.  Retries happen in the path stack so the retry is dispatched to client B this time (which has a timeout of 3s).  After another 1s has elapsed, the client still has 2s left towards its timeout but the server has hit its 2s timeout and the request is failed.  

I don't think we can do post-retry timeouts at the client level because retries happen in the path stack.

This is described briefly in the config docs.  It's a complex enough feature that it could probably use a section on linkerd.io/documentation describing it.

olix0r:
We certainly could have a timeout filter above the retry filter in the path stack to achieve post-retry client timeouts.

All-in-all I *think* this is okay, but for 1.0, we should write a doc on timeouts so that our thinking is clear.  We may want/need to fill in some gaps/rename some things for clarity before cutting f'real 1.0.

adleong:
Merging this.  As discussed above, further work remains to be done.

PR #1166, opened 2017-03-28
Port numbers in k8s names should apply the service's port mapping

### LINKED ISSUE #1143: io.l5d.k8s namer: port should map to service port instead of pod port
When using the io.l5d.k8s namer, routing to `/#/io.l5d.k8s/default/http` (the name of the port in the service) routes to the pods correctly, using their endpoint address as the destination.  However, routing to `/#/io.l5d.k8s/default/80` routes to port 80 of the internal pod IPs.

Example service:
```kubectl describe service sas
Name:            sas
Namespace:        default
Labels:            <none>
Selector:        app=sas
Type:            ClusterIP
IP:            10.100.31.239
Port:            http    80/TCP
Endpoints:        10.102.14.30:3000,10.102.5.39:3000
Session Affinity:    None
```

I would expect that `/#/io.l5d.k8s/default/80` routes to 10.102.14.30:3000 and 10.102.5.39:3000, which is the behavior of the `/#/io.l5d.k8s/default/http` route, however it is routing to 10.102.14.30:80 and 10.102.5.39:80.

We are following the convention of exposing services on port 80, but have not been strict about requiring the port name to match http.  This is the behavior that https://github.com/kubernetes/contrib/tree/master/service-loadbalancer uses.

adleong:
Fixes #1143 

# Problem

When a port number is specified in a k8s name such as `/#/io.l5d.k8s/ns/80/srv` the resulting address list will simply use the given port on each endpoint, ignoring any port mapping specified in the service definition.

# Solution

The port and targetPort from the service definition is taken into account when evaluating k8s names with port numbers.

# Result

* If a port number is specified that has a target port, the name will resolve using the target port.
* If a port number is specified that has no target port, the target port is assumed to be the same as the service port.
* If a port number is specified that does not exist in the service definition, Neg is returned.

This should result in more compatible behavior when linkerd is used as an ingress controller and port numbers are specified in the ingress resource.

## Implementation Note

Note that the targetPort is not available from the EndpointPort resource in the endpoints API.  Therefore, we must establish a watch on the services API in addition to the watch on the endpoints API.


adleong:
@olix0r I could use a review on this now that 1.0.0 is out the door

PR #1167, opened 2017-03-28
Automatically upgrade all HTTP/1.0 messages to HTTP/1.1

### LINKED ISSUE #1133: linkerd routers should respond with request protocol, not the downstream response protocol
(am not entirely sure this writeup captures the issue, so please feel free to adjust as necessary)

Am seeing an issue running a linkerd http router in front of an HTTP/1.0 webserver. If I make an HTTP/1.1 request to linkerd, and it forwards it to a service that returns an HTTP/1.0 response, then the response from linkerd is also HTTP/1.0. In talking with @olix0r, I think it makes more sense for linkerd to still respond with HTTP/1.1, even if it received an HTTP/1.0 response.

Here's what the curl output looks like for this type of request:

```
$ http_proxy=$LINKERD_URL curl -v http://hello
* Rebuilt URL to: http://hello/
*   Trying 192.168.99.101...
* TCP_NODELAY set
* Connected to 192.168.99.101 (192.168.99.101) port 30665 (#0)
> GET http://hello/ HTTP/1.1
> Host: hello
> User-Agent: curl/7.50.3
> Accept: */*
> Proxy-Connection: Keep-Alive
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: text/html; charset=utf-8
< Content-Length: 39
< Server: Werkzeug/0.11.15 Python/2.7.12
< Date: Thu, 09 Mar 2017 00:16:21 GMT
< Via: 1.0 linkerd, 1.0 linkerd
* HTTP/1.0 connection set to keep alive!
< Connection: keep-alive
< 
* Curl_http_done: called premature == 0
* Connection #0 to host 192.168.99.101 left intact
Hello (172.17.0.9) world (172.17.0.10)!
```

This is using the hello world app from the linkerd-examples repo.

adleong:
Fixes #1133 

Linkerd should only ever send HTTP/1.1 messages, regardless of the version it receives.  We do this upgrade in the `ViaHeaderAppenderFilter` so that we can set the `Via` header with the protocol version that was received.


router/http/src/e2e/scala/io/buoyant/router/HttpEndToEndTest.scala
```diff
@@ -231,80 +231,4 @@ class HttpEndToEndTest extends FunSuite with Awaits {
...
-      assert(err == None)
-      assert(rsp2.status == Status.Ok)
-      assert(downstreamCounter("connects") == Some(4))
-
-    } finally {
-      await(client.close())
-      await(router.close())
-      await(downstream.server.close())
-    }
-  }
```
olix0r: this test should still be relevant, no?

When a client connects to linkerd with HTTP/1.0 and no Connection header, we can't reuse that connection... right?

adleong: I... don't know.  I guess I assumed that the client would close the connection after receiving the response.  linkerd returns a 1.1 response and there's no reason why it couldn't accept another request on the same connection if the client were to send one.  But I don't know what the spec dictates in this case...

adleong:
Yup, @klingerf and I have both tested this manually.

PR #1171, opened 2017-03-29
The daemonset transformer should return Addr.Neg for empty sets

### LINKED ISSUE #1169: PortTransformer should return neg resolution if empty
When following the daemonset example on CNI/Calico, resolution fails but the admin ui shows it has a `Bound Match` of `/`. Ideally we return neg with some sort of indication of what the issue is.

screen shot 2017-03-29 at 11 06 15 am

related to https://linkerd.io/getting-started/k8s-daemonset/#cni-setups
related to https://github.com/BuoyantIO/linkerd-examples/issues/119

adleong:
Fixes #1169

If the daemonset transformation results in an empty address set, the delegator UI will now display the tree in red instead of green.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
@@ -67,6 +67,9 @@ define([
           obj.tree.transformation = obj.name;
           obj.tree.isTransformation = true;
           obj.bound.transformed = renderNode(obj.tree, primary);
+          if (ttype != success) {
+            obj.bound.addr.type = "neg";
+          }
           obj.child = renderNode(obj.bound, primary);
           break;
         case "alt":
--- a/interpreter/subnet/src/main/scala/io/buoyant/transformer/SubnetGatewayTransformer.scala
+++ b/interpreter/subnet/src/main/scala/io/buoyant/transformer/SubnetGatewayTransformer.scala
@@ -63,7 +63,10 @@ class GatewayTransformer(
           // select the gateway addresses that share a subnet with addr
           gatewayAddrs.filter(gatewayPredicate(addr, _))
         }
-        Addr.Bound(selected, meta)
+        if (selected.isEmpty)
+          Addr.Neg
+        else
+          Addr.Bound(selected, meta)
       case List(addr, _) => addr
     }
     bound.id match {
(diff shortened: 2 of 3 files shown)


admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
```diff
@@ -67,6 +67,9 @@ define([
           obj.tree.transformation = obj.name;
           obj.tree.isTransformation = true;
           obj.bound.transformed = renderNode(obj.tree, primary);
+          if (ttype != success) {
+            obj.bound.addr.type = "neg";
+          }
```
esbie: for my own edification, why is this needed? I figured a neg was a neg transformer or no

interpreter/subnet/src/main/scala/io/buoyant/transformer/SubnetGatewayTransformer.scala
```diff
@@ -63,7 +63,10 @@ class GatewayTransformer(
           // select the gateway addresses that share a subnet with addr
           gatewayAddrs.filter(gatewayPredicate(addr, _))
         }
-        Addr.Bound(selected, meta)
+        if (selected.isEmpty)
+          Addr.Neg
```
esbie: Yay! Could we do something like info log "no addrs found were on this subnet" if selected is empty but `addrs` is not?

admin/src/main/resources/io/buoyant/admin/js/src/delegator.js
adleong: Good question.  A transformation object has 2 children: `obj.tree` which is the result of the transformation and `obj.bound` which is the bound address pre-transformation.  We show both so that we can see the effect of the transformation.  

In the case where the transformation results in a failure we want to color the whole path red, even if the pre-transformed value is valid.  That's what this does.  Without it, the pre-transformed value would be green while the rest of the path would be red.

interpreter/subnet/src/main/scala/io/buoyant/transformer/SubnetGatewayTransformer.scala
adleong: I hesitate because returning Neg here doesn't necessarily mean something is wrong.  For example, you could imagine building a fallback dtab that attempts to route to a daemonset but falls back to something else if no matching daemonset pods exist.  In that case we'd log on every request even though the requests get successfully routed to the fallback.

esbie: Right, it could be extremely useful or extremely annoying. I'm fine leaving as is

PR #1173, opened 2017-03-31
Expire and prune idle metrics

adleong:
If a counter or stat goes no updates for a configurable TTL (default: 24 hours) then we remove the metric from the MetricsTree.  This is done as part of the periodic snapshotting loop of the AdminMetricsExportTelemeter.  

The ttl can be configured with the `-io.buoyant.telemetry.admin.metricsExpiryTtl` flag.

Gauges are never expired by time, but will be pruned if they are manually removed.  Gauges created using `StatsReceiver`'s `addGauge` method are weakly referenced and will be removed and pruned from the MetricsTree after the gauge falls out of scope.  This should pair very nicely with the client expiration that is currently in Finagle that we will pick up on the next Finagle release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## In the next release...
+
+* Expire and prune metrics which are idle for 24 hours.
+
 ## 1.0.0 2017-04-24
 
 * Configuration:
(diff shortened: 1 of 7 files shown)


adleong:
@olix0r I could use a review on this

adleong:
Nevermind, I've changed my mind on how I want to do this.  This approach has a few drawbacks:

Expiring gauges is tricky because they have to be manually deregistered (which not everyone does) or rely on a WeakReferences to deregister when they are no longer referenced.  Unfortunately, there are gauges that seem to stick around forever even though they don't appear to be referenced anywhere.

Additionally, expiring idle counters may just be a bad idea in general.  Counters that update very rarely may still contain valuable information.  For example, binding cache one-shots: you may want to know if a one-shot has ever happened.  This information would be lost after the TTL period.

Instead of expiring idle metrics based on a TTL I think we should more explicitly remove entire subtrees from the metrics tree when the relevant ServiceFactory is closed.

PR #1184, opened 2017-04-06
introduce io.l5d.influxdb telemeter

### LINKED ISSUE #474: Make linkerd stats compatible with InfluxDB LINE
(this is a request from the customer survey)

https://docs.influxdata.com/influxdb/v0.13/write_protocols/line/#main-nav

We can start with the prometheus stats rewriter.

siggy:
fixes #474

### DIFF
--- a/telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
+++ b/telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
@@ -1,0 +1,118 @@
+package io.buoyant.telemetry.influxdb
+
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{MediaType, Request, Response}
+import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.finagle.tracing.NullTracer
+import com.twitter.util.{Awaitable, Closable, Future}
+import io.buoyant.admin.Admin
+import io.buoyant.telemetry.{Metric, MetricsTree, Telemeter}
+
+/**
+ * This telemeter exposes metrics data in the InfluxDb LINE format, served on
+ * an admin endpoint.  It does not provide a StatsReceiver or Tracer.  It reads
+ * histogram summaries directly off of the MetricsTree and assumes that stats
+ * are being snapshotted at some appropriate interval.
+ */
+class InfluxDbTelemeter(metrics: MetricsTree) extends Telemeter with Admin.WithHandlers {
+
+  private[influxdb] val handler = Service.mk { request: Request =>
+    val response = Response()
+    response.version = request.version
+    response.mediaType = MediaType.Txt
+    val sb = new StringBuilder()
+    val host = request.host.getOrElse("none")
+    writeMetrics(metrics, sb, Nil, Seq("host" -> host))
+    response.contentString = sb.toString
+    Future.value(response)
+  }
+
+  val adminHandlers: Seq[Admin.Handler] = Seq(
+    Admin.Handler("/admin/metrics/influxdb", handler)
+  )
+
+  val stats = NullStatsReceiver
+  def tracer = NullTracer
+  def run(): Closable with Awaitable[Unit] = Telemeter.nopRun
+
+  private[this] val disallowedChars = "[^a-zA-Z0-9:]".r
+  private[this] def escapeKey(key: String) = disallowedChars.replaceAllIn(key, "_")
+
+  private[this] val first: ((String, String)) => String = _._1
+  private[this] def formatLabels(labels: Seq[(String, String)]): String =
+    labels.sortBy(first).map {
+      case (k, v) =>
+        s"""$k=$v"""
+    }.mkString(",")
+
+  private[this] def labelExists(labels: Seq[(String, String)], name: String) =
+    labels.toIterator.map(first).contains(name)
+
+  private[this] def writeMetrics(
+    tree: MetricsTree,
+    sb: StringBuilder,
+    prefix0: Seq[String],
+    tags0: Seq[(String, String)]
+  ): Unit = {
+    // TODO: share with prometheus telemeter?
+    // Re-write elements out of the prefix into labels
+    val (prefix1, tags1) = prefix0 match {
+      case Seq("rt", router) if !labelExists(tags0, "rt") =>
+        (Seq("rt"), tags0 :+ ("rt" -> router))
+      case Seq("rt", "dst", "path", path) if !labelExists(tags0, "dst_path") =>
+        (Seq("rt", "dst_path"), tags0 :+ ("dst_path" -> path))
+      case Seq("rt", "dst", "id", id) if !labelExists(tags0, "dst_id") =>
+        (Seq("rt", "dst_id"), tags0 :+ ("dst_id" -> id))
+      case Seq("rt", "dst_id", "path", path) if !labelExists(tags0, "dst_path") =>
+        (Seq("rt", "dst_id", "dst_path"), tags0 :+ ("dst_path" -> path))
+      case Seq("rt", "srv", srv) if !labelExists(tags0, "srv") =>
+        (Seq("rt", "srv"), tags0 :+ ("srv" -> srv))
+      case _ => (prefix0, tags0)
+    }
+
+    // gather all metrics directly under a common parent
+    val fields =
+      tree.children.toSeq.flatMap {
+        case (name, child) =>
+          // write metrics deeper in the tree (side effect)
+          writeMetrics(child, sb, prefix1 :+ name, tags1)
+
+          child.metric match {
+            case c: Metric.Counter => Seq(name -> c.get.toString)
+            case g: Metric.Gauge => Seq(name -> g.get.toString)
+            case s: Metric.Stat =>
+              val summary = s.snapshottedSummary
+              if (summary != null) {
+                Seq(
+                  name + "_count" -> summary.count.toString,
+                  name + "_sum" -> summary.sum.toString,
+                  name + "_avg" -> summary.avg.toString,
+                  name + "_min" -> summary.min.toString,
+                  name + "_max" -> summary.max.toString,
+                  name + "_p50" -> summary.p50.toString,
+                  name + "_p90" -> summary.p90.toString,
+                  name + "_p95" -> summary.p95.toString,
+                  name + "_p99" -> summary.p99.toString,
+                  name + "_p999" -> summary.p9990.toString,
+                  name + "_p9999" -> summary.p9999.toString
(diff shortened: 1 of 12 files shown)


telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,123 @@
...
+      labels.sortBy(_._1).map {
+        case (k, v) =>
+          s"""$k=$v"""
+      }.mkString(",")
+    } else {
+      ""
+    }
+
+  private[this] def labelExists(labels: Seq[(String, String)], name: String) =
+    labels.exists(_._1 == name)
```
adleong: This will allocate a function for each call to `labelExists`.  Not sure if there's a super great way to avoid this, best I could come up with is:

```
val first = ((String, String)) => String = _._1
private[this] def labelExists(labels: Seq[(String, String)], name: String) = labels.toIterator.map(first).contains(name)
```

Unfortunately that's not super readable, but I THINK it saves on allocation which might matter depending on how often this endpoint is going to be scraped.

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  val stats = NullStatsReceiver
+  def tracer = NullTracer
+  def run(): Closable with Awaitable[Unit] = Telemeter.nopRun
+
+  private[this] val disallowedChars = "[^a-zA-Z0-9:]".r
+  private[this] def escapeKey(key: String) = disallowedChars.replaceAllIn(key, "_")
+
+  private[this] def formatLabels(labels: Seq[(String, String)]): String =
+    if (labels.nonEmpty) {
```
adleong: `Seq().mkString(",")` produces an empty string so I don't think you have to special case when the seq is empty.

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,123 @@
...
+      labels.sortBy(_._1).map {
```
adleong: can val this function up (see below).

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,123 @@
...
+      case Seq("rt", "dst_id", "path", path) if !labelExists(tags0, "dst_path") =>
+        (Seq("rt", "dst_id", "dst_path"), tags0 :+ ("dst_path" -> path))
+      case Seq("rt", "srv", srv) if !labelExists(tags0, "srv") =>
+        (Seq("rt", "srv"), tags0 :+ ("srv" -> srv))
+      case _ => (prefix0, tags0)
+    }
+
+    // gather all metrics directly under a common parent
+    val fields = {
+      tree.children.toSeq map {
```
adleong: How about flatMap here and return just Seq[(String, String)] below instead of Seq[Option[(String, String)]]?  Then you can get rid of BOTH calls to flatten!

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,123 @@
...
+      if (tags1.nonEmpty) {
+        sb.append(",")
+        sb.append(formatLabels(tags1))
+      }
+      sb.append(" ")
+      sb.append(formatLabels(fields))
+      sb.append("\n")
+    }
+
+    return
```
adleong: Don't need this

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,123 @@
...
+        (Seq("rt", "dst_id"), tags0 :+ ("dst_id" -> id))
+      case Seq("rt", "dst_id", "path", path) if !labelExists(tags0, "dst_path") =>
+        (Seq("rt", "dst_id", "dst_path"), tags0 :+ ("dst_path" -> path))
+      case Seq("rt", "srv", srv) if !labelExists(tags0, "srv") =>
+        (Seq("rt", "srv"), tags0 :+ ("srv" -> srv))
+      case _ => (prefix0, tags0)
+    }
+
+    // gather all metrics directly under a common parent
+    val fields = {
```
adleong: I don't think you need this brace for a single expression

esbie:
Not familiar with influxdb, but I looked over the telemeter and it lgtm.

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
siggy: i get:
```
[error] /Users/sig/code/linkerd/telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala:116: discarded non-Unit value
[error]       sb.append("\n")
[error]                ^
```

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
siggy: I got it working with:
```
val first = (tuple: (String, String)) => { tuple._1 }
```
... it was complaining about the tuple not being named. if you know a better way i'm open to it.

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
adleong: Oh.  The problem is that the method return type is `Unit` but `append` returns a stringbuilder.  You can do `val _ = sb.append("\n")` to explicitly discard the returned stringbuilder.

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
adleong: This works in the repl for me `val first: ((String, String)) => String = _._1`

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,118 @@
...
+      case Seq("rt", "dst_id", "path", path) if !labelExists(tags0, "dst_path") =>
+        (Seq("rt", "dst_id", "dst_path"), tags0 :+ ("dst_path" -> path))
+      case Seq("rt", "srv", srv) if !labelExists(tags0, "srv") =>
+        (Seq("rt", "srv"), tags0 :+ ("srv" -> srv))
+      case _ => (prefix0, tags0)
+    }
+
+    // gather all metrics directly under a common parent
+    val fields =
+      tree.children.toSeq flatMap {
```
adleong: `.flatMap {`

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,118 @@
...
+ * are being snapshotted at some appropriate interval.
+ */
+class InfluxDbTelemeter(metrics: MetricsTree) extends Telemeter with Admin.WithHandlers {
+
+  private[influxdb] val handler = Service.mk { request: Request =>
+    val response = Response()
+    response.version = request.version
+    response.mediaType = MediaType.Txt
+    val sb = new StringBuilder()
+    val host = request.host.getOrElse("none")
```
adleong: does this write the host header of the request to the metrics endpoint as a tag?  influx is weird.

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,118 @@
...
+    writeMetrics(metrics, sb, Nil, Seq(("host", host)))
```
adleong: TIOLI: I like arrow notation for pairs to avoid the double parens.  Total TIOLI though.
`Seq("host" -> host)`

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    // gather all metrics directly under a common parent
+    val fields =
+      tree.children.toSeq flatMap {
+        case (name, child) =>
+          // write metrics deeper in the tree (side effect)
+          writeMetrics(child, sb, prefix1 :+ name, tags1)
+
+          child.metric match {
+            case c: Metric.Counter => Seq((name, c.get.toString))
```
adleong: Could use arrow notation here to avoid the double parens if you wanted.

telemetry/influxdb/src/main/scala/io/buoyant/telemetry/influxdb/InfluxDbTelemeter.scala
siggy: That's correct. read on for the gritty details:

Although we're providing an endpoint in LINE format, InfluxDB expects data to be pushed, so we need Telegraf. Telegraf has a number of input plugins, such as Prometheus, which, given a list of endpoints, collects data and automatically tags them with the source. Surprisingly, there is no input plugin specifically for LINE endpoints, so we must set the `host` tag ourselves. There is an open issue in Telegraf for this:
https://github.com/influxdata/telegraf/issues/813

Because there is no LINE input plugin in Telegraf, we rely on the `inputs.exec` input plugin, which executes arbitrary commands, like this:
https://github.com/linkerd/linkerd-examples/pull/124/files#diff-898d1cb8b1ac4b01cdfca408c9069081R2

leozc:
Feature request : possible to support data dog as well?

PR #1185, opened 2017-04-06
Add service section to linkerd config for per-path configuration.

adleong:
We add a `service` section to the linkerd router config to configure per-logical name params.  ie per-path stack configuration.  This is analogous to the `client` section for per-client configuration and admits the same kinds: `io.l5d.global` (default) and `io.l5d.static`.

```
routers:
- protocol: http
  service:
    kind: io.l5d.static
    configs:
    - prefix: /svc
      retries:
        budget:
          percentCanRetry: 0.3
    - prefix: /svc/foo
      totalTimeoutMs: 300
```

The following config keys changes are included:

* `retries` moved from `router` to `service`
* `failFast` moved from `router` to `client`
* `timeoutMs` (client) renamed to `requestAttemptTimeoutMs`
* `timeoutMs` (server) removed
* `totalTimeoutMs` added to `service` and a TotalTimeout module added to the path stack above the classified retries module.

This means that `requestAttemptTimeoutMs` is the pre-retry timeout and is configured on the client and `totalTimeoutMs` is the post-retry timeout and is configured on the service.  

## Naming

I very much considered naming the `service` section `routes` instead but eventually decided against it.  When thinking about a traditional HTTP router, you typically define "routes" which are mappings from URIs to destinations.  More abstractly: a route is a mapping from an id to a destination.  In the case of linkerd, this corresponds to dtabs which map logical names (ids) to concrete names (destinations).  My fear is that if we had a section in the config called `routes:` that it would seem like routing information (dtabs) should go there.  Instead, we want a section for configuration/policy of those logical names (ids).  

Service is a severely overloaded term in our space which is why we hesitate to use it.  I believe that it's use is unavoidable so we should not shy away from it, but instead be definitive and consistent with its meaning.  Our default `dstPrefix` is `/svc` which already indicates that logical names are service names.  It seems natural and fitting that the `services` section configures `/svc` style names.


siggy:
I think this general direction makes sense, glad we are doing this pre-1.0.

I'm still struggling a little bit with `io.l5d.static` and `io.l5d.global` coupled with the use of a `configs` section or not, respectively. This shouldn't necessarily motivate a change, but perhaps just more explicit explanation/documentation on usage. The fact that it mirrors `client` is helpful, though.

My understanding is that these two configs blocks are valid:

```yaml
  service:
    kind: io.l5d.static
    configs:
    - prefix: /svc
      retries:
```

```yaml
  service:
    kind: io.l5d.global
    retries:
```

I think it feels a bit weird that the `kind` field affects what is possible in the rest of the object. For example, would the following config blocks be invalid, and if so, is there a way to enforce it more structurally in the yaml:

```yaml
  service:
    kind: io.l5d.global
    configs:
    - prefix: /svc
      retries:
```

```yaml
  service:
    kind: io.l5d.static
    retries:
```

adleong:
@siggy thanks for the feedback!  Your understanding is correct about the valid/invalid configs.  I'm not sure what you mean about structural enforcement; invalid configs will fail to parse.  There's also president all over our configs for changing what's allowed in the object based on the value of the `kind` field.  Consider:

```
namers:
- kind: io.l5d.fs
  rootDir: /disco
- kind: io.l5d.k8s
  host: localhost
  port: 8001
```

siggy:
@adleong all sounds good. per offline convo, my concern stems from sibling objects affecting each other. this is not necessarily actionable, just something to consider as we document and articulate this approach. being explicit that `kind` is a special thing is helpful.

esbie:
This looks great! I like the mirroring of the client section. I have some thoughts, but none are actionable. 

- I wish the words "client" and "server" were at the same level in the config hierarchy.
- I think there's a potential source of confusion that `prefix` means logical name under `service.configs` but then means concrete name under `client.configs`.
- Retries are still really hard for me to grok. Not really a fault of the config though.

Nice jorb 👍

adleong:
@esbie lmk if I've misunderstood your comment but `client`, `service`, and `servers` are all siblings.

esbie:
Oh! yeah my mistake. I thought servers were tucked under configs.

adleong:
Here's a proposal on what to do about retry budgets and response classification:

# Retry Budgets
The objective of retry budgets is to allow linkerd to retry as much as possible without overloading itself or downstream systems.  Currently linkerd has a single retry budget per router and both application failures (retries) and transport failures (requeues) count against this budget.  One problem with this is that a single failing client or service can eat up the entire retry budget and allow no other retries to take place.

We would like to continue to allow linkerd to retry as much as safely possible while also protecting the budget against failing clients and keeping the configuration sane.  Thankfully, retry budgets Just Work with the default settings and changing the budget should be a rarely used advanced feature.

I propose that each service have its own retry budget that is not shared with other services and is not shared with any client.  Likewise, each client would have its own retry budget (for requeues) that is not shared with other clients or with any service.  Each retry filter and requeue filter would be responsible for its own depots and withdrawals.  This means that a failing client or service can only deplete its own budget.  A `retries` config may be specified as a `service` config that configures the retry budget and backoff for that service.  A `requeueBudget` config may be specified as a `client` config to configure the requeue budget for that client.

# Response Classification
Response classification feeds into stats reporting (successes vs failures) and application retries.  The response classifier that is configured on the router today is used in 4 places: client level stats, service level stats, application failure retries, and server level stats.  However, having one response classifier for all services is problematic because the definition of failure or retryability/idempotency is different for different services.  Therefore, we would like to move the response classifier to the `service` config.  However, we also need the results of that classification to be reported in the client and server level stats.

I propose that response classification be configured as a `service` config.  The response classification module in the path (service) stack should do 3 things:

1. Set the response classifier into a `LocalKey` (ie Context).  The client stats module will then use the response classifier from the LocalKey to report success/failure.  In other words, it will use the response classifier request's service.
2. Use the response classifier for service level stats and application failure retries.
3. Set a `l5d-ctx-success` header on the response indicating the success or failure (or partial success?) of the request.  The server stats module will read this header and use it to report stats.

In this way, the response classifier is only set at the service level, but is used to report consistent stats at the client, service, and server levels.

olix0r:
@adleong I like this proposal, assuming the classification stuff actually works.  minor quibble that i'd call it `l5d-rsp-class` or something like that -- i don't think this is really _context_ in the same way other things are on requests.  For instance, when a single response transits multiple linkerds, the clients and services of the upward linkers shouldn't honor this header blindly.

linkerd/docs/routers.md
```diff
@@ -186,9 +178,10 @@ Key | Default Value | Description
 hostConnectionPool | An empty object | see [hostConnectionPool](#host-connection-pool).
 tls | no tls | The router will make requests using TLS if this parameter is provided.  It must be a [client TLS](#client-tls) object.
 loadBalancer | [p2c](#power-of-two-choices-least-loaded) | A [load balancer](#load-balancer) object.
-retries | see [retries](#retries) | A [retry policy](#retries) for all clients created by this router.
+failFast | `false` | If `true`, connection failures are punished more aggressively. Should not be used with small destination pools.
+requeueBudget | see [retries](#retries) | A [requeue budget](#retries) for connection-level retries.
 failureAccrual | 5 consecutive failures | a [failure accrual policy](#failure-accrual) for all clients created by this router.
-timeoutMs | no timeout | Per-request timeout in milliseconds.
+requestAttemptTimeoutMs | no timeout | The timeout, in milliseconds, for each attempt (original or retry) of the request made by this client.
```
olix0r: how does this interact with requeues?

linkerd/examples/retries.yaml
```diff
@@ -6,11 +6,11 @@ namers:
...
 - protocol: http
   dtab: |
-    /svc => /#/io.l5d.fs
-  responseClassifier:
-    # Retry 5XXs for GETs, HEADs, OPTIONs, PUTs, and DELETEs.
+    /svc/* => /#/io.l5d.fs/dog & /#/io.l5d.fs/cat
+  service:
+    responseClassifier:
+      # Retry 5XXs for GETs, HEADs, OPTIONs, PUTs, and DELETEs.
     kind: io.l5d.retryableIdempotent5XX
```
olix0r: indentation looks off here, but might just be github

linkerd/examples/retries.yaml
```diff
@@ -20,5 +20,23 @@ routers:
...
+    # cat client.
+    totalTimeoutMs: 1000
+  client:
+    kind: io.l5d.static
+    configs:
+    - prefix: /#/io.l5d.fs/dog
+      # The timeout for a single request attempt on the dog client.  If this
+      # timeout is reached, the request attempt will be canceled and the request
+      # will be retried if there is still time remaining in the total timeout.
+      # Note that the retry may happen on the dog or the cat client.
```
olix0r: note re other client isn't necessary here since it's mentioned above

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/ResponseClassificationEndToEndTest.scala
```diff
@@ -0,0 +1,200 @@
...
+    assert(stats.counters(Seq("http", "srv", "127.0.0.1/0", "success")) == 1)
+    assert(stats.counters.get(Seq("http", "srv", "127.0.0.1/0", "failures")) == None)
+    assert(stats.counters(Seq("http", "dst", "path", "svc/foo", "success")) == 1)
+    assert(stats.counters.get(Seq("http", "dst", "path", "svc/foo", "failures")) == None)
+    assert(stats.counters(Seq("http", "dst", "id", s"$$/inet/127.1/${downstream.port}", "path", "svc/foo", "success")) == 1)
+    assert(stats.counters(Seq("http", "dst", "id", s"$$/inet/127.1/${downstream.port}", "path", "svc/foo", "failures")) == 1)
+  }
+
+  test("per service classification") {
+    var i = 0
```
olix0r: technically should be `@volatile` since accessed across threads

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/ResponseClassificationEndToEndTest.scala
```diff
@@ -0,0 +1,200 @@
...
+    req.host = "foo"
+    await(client(req))
+
+    assert(stats.counters(Seq("http", "dst", "id", s"$$/inet/127.1/${downstream.port}", "path", "svc/foo", "failures")) == 1)
+    assert(stats.counters(Seq("http", "dst", "path", "svc/foo", "failures")) == 1)
+    assert(stats.counters(Seq("http", "srv", "127.0.0.1/0", "failures")) == 1)
+  }
+
+  test("retryable failure") {
+    var i = 0
```
olix0r: technically should be @volatile since accessed across threads

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/ResponseClassificationEndToEndTest.scala
```diff
@@ -0,0 +1,200 @@
...
+    assert(stats.counters(Seq("http", "dst", "path", "svc/foo", "success")) == 1)
+    assert(stats.counters.get(Seq("http", "dst", "path", "svc/foo", "failures")) == None)
+    assert(stats.counters(Seq("http", "dst", "id", s"$$/inet/127.1/${downstream.port}", "path", "svc/foo", "success")) == 1)
+    assert(stats.counters(Seq("http", "dst", "id", s"$$/inet/127.1/${downstream.port}", "path", "svc/foo", "failures")) == 1)
+  }
+
+  test("per service classification") {
+    var i = 0
+    val downstream = Downstream("ds", Service.mk { req =>
+      if (i%2 == 0) {
```
olix0r: nit: let it breathe `i % 2 == 0`

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -114,6 +114,36 @@ trait HttpClientConfig extends ClientConfig {
...
+  visible = true,
+  defaultImpl = classOf[HttpDefaultSvc]
+)
+@JsonSubTypes(Array(
+  new JsonSubTypes.Type(value = classOf[HttpDefaultSvc], name = "io.l5d.global"),
+  new JsonSubTypes.Type(value = classOf[HttpStaticSvc], name = "io.l5d.static")
+))
+abstract class HttpSvc extends Svc
+
+class HttpDefaultSvc extends HttpSvc with DefaultSvc with HttpSvcConfig
```
olix0r: i think this makes it clear that DefaultSvc should be `trait DefaultSvc { _: Svc => ... }`

linkerd/protocol/mux/src/main/scala/io/buoyant/linkerd/protocol/MuxInitializer.scala
```diff
@@ -27,6 +27,7 @@ class MuxConfig extends RouterConfig {
 
   var servers: Seq[ServerConfig] = Nil
   var client: Option[Client] = None
+  var service: Option[Svc] = None
```
olix0r: nit: ordering

router/core/src/main/scala/com/twitter/finagle/buoyant/TotalTimeout.scala
```diff
@@ -0,0 +1,42 @@
+package com.twitter.finagle.buoyant
+
+import com.twitter.finagle.service.TimeoutFilter
+import com.twitter.finagle._
+import com.twitter.util.{Duration, Timer}
+
+/**
+ * This stack module is intended to enforce a total timeout on a request
+ * including all retries.  To do this it must be installed above any retry
+ * filters in the stack.
```
olix0r: "any" or "all"? i think we mean all...

router/core/src/main/scala/io/buoyant/router/LocalClassifierStatsFilter.scala
```diff
@@ -0,0 +1,43 @@
+package io.buoyant.router
+
+import com.twitter.finagle.{Filter, Path, Stack, Stackable, ServiceFactory, param}
+import com.twitter.finagle.buoyant.Dst
+import com.twitter.finagle.service.StatsFilter
+import com.twitter.finagle.stats.{ExceptionStatsHandler, StatsReceiver}
+import io.buoyant.router.context.ResponseClassifierCtx
+
+object LocalClassifierStatsFilter {
```
olix0r: what does _local_ refer to here?  that it's using the context?

router/core/src/main/scala/io/buoyant/router/Router.scala
```diff
@@ -302,6 +281,23 @@ object StackRouter {
       val default: PerClientParams = PerClientParams(Seq.empty)
     }
 
+    case class PathParams(prefix: PathMatcher, mk: Map[String, String] => Stack.Params)
```
olix0r: what's this Map[String, String] all about?  perhaps a typedef would make this clearer?

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -0,0 +1,51 @@
...
+
+class ClassifierFilter(classifier: ResponseClassifier) extends SimpleFilter[Request, Response] {
+
+  private[this] val successClass = classifier.andThen(_.fractionalSuccess).lift
+
+  def apply(req: Request, svc: Service[Request, Response]): Future[Response] = {
+    svc(req).map { rep =>
+      successClass(ReqRep(req, Return(rep))) match {
+        case Some(success) =>
+          rep.headers.set("l5d-success-class", success.toString)
```
olix0r: please reuse the const

router/http/src/main/scala/io/buoyant/router/http/ClassifierFilter.scala
```diff
@@ -0,0 +1,47 @@
...
+
+class ClassifierFilter(classifier: ResponseClassifier) extends SimpleFilter[Request, Response] {
+
+  private[this] val successClass = classifier.andThen(_.fractionalSuccess).lift
+
+  def apply(req: Request, svc: Service[Request, Response]): Future[Response] = {
+    svc(req).map { rep =>
+      successClass(ReqRep(req, Return(rep))) match {
+        case Some(success) =>
+          rep.headerMap("l5d-success-class") = success.toString
```
olix0r: reuse const

linkerd/docs/routers.md
```diff
@@ -207,3 +200,82 @@ minSize | `0` | The minimum number of connections to maintain to each host.
...
+
+<aside class="notice">
+These parameters are available to the service regardless of protocol. Services may also have protocol-specific parameters.
+</aside>
+
+Key                 | Default Value            | Description
+------------------- | -------------------------| -----------
+retries             | see [retries](#retries)  | A [retry policy](#retries) for applicaiton-level retries.
+totalTimeoutMs      | no timeout               | The timeout for an entire request, including all retries, in milliseconds.
+responseClassifier  | `io.l5d.nonRetryable5XX` | A (sometimes protocol-specific) [response classifier](#http-response-classifiers) that determines which responses should be considered failures and, of those, which should be considered [retryable](#retries).
```
olix0r: I think we need to explain how this setting influences server and client modules.

linkerd/core/src/main/scala/io/buoyant/linkerd/Svc.scala
```diff
@@ -0,0 +1,66 @@
...
+  def pathParams = PerPathParams(Seq(PathParams(matchAll, mk)))
+}
+
+class DefaultSvcImpl extends Svc with DefaultSvc
+
+/**
+ * StaticSvc consists of a list of PrefixConfigs.  This is a trait so that it
+ * can be mixed in to protocol specific versions.
+ */
+trait StaticSvc {
```
adleong: it can't extend Svc because it's a trait but yeah, self type is a good idea.

linkerd/docs/routers.md
adleong: It doesn't.  Timeouts are not requeuable but they may be retryable.

linkerd/docs/routers.md
adleong: Yeah?  I don't think we talk about stacks or modules anywhere else in this documentation.

olix0r: hm, ok. fair.

linkerd/docs/routers.md
olix0r: meaning that this timeout spans requeues... ok not worth documenting just confirming my understanding

adleong: Oh I see what you mean.  Retries are above timeouts in the client stack.  So timeouts don't get requeued but if a requeue happens, the requeue gets a new timeout.

router/core/src/main/scala/io/buoyant/router/LocalClassifierStatsFilter.scala
adleong: yeah that it gets the classifier from the local context

router/core/src/main/scala/io/buoyant/router/Router.scala
adleong: These are the values for the captured path segments from the prefix matcher.  E.g. if the prefix is `/svc/{service}` and the path is `/svc/foo` then this map is populated with `"service" -> "foo"`.  None of the service configs use this yet, but I included it to be consistent with client configs.

linkerd/docs/retries.md
```diff
@@ -15,21 +15,36 @@ routers:
...
 budget | See [retry budget](#retry-budget-parameters) | Object that determins _how many_ failed requests are eligible to be retried.
 backoff | See [retry backoff](#retry-backoff-parameters) | Object that determines which backoff algorithm should be used.
 
+## Retries
+
+Retries are for application-level failures (such as 5XX responses in the case of HTTP) as determined by the response
+classifier.  If the response classifier determines that a request is a retryable failure, and there is retry budget
+available, then the request will be retried.  Retries are configured by the `retries` parameter on the `service` object.
+On the `retries` object you may specify the retry budget and retry backoff schedule.  Each service has its own retry
+budget that is not shared with other services or clients.
```
klingerf: For here and the rest of the docs changes, can you wrap these lines at 80 charactoers?

linkerd/docs/retries.md
```diff
@@ -15,21 +15,36 @@ routers:
...
 
 Key | Default Value | Description
 --- | ------------- | -----------
 budget | See [retry budget](#retry-budget-parameters) | Object that determins _how many_ failed requests are eligible to be retried.
 backoff | See [retry backoff](#retry-backoff-parameters) | Object that determines which backoff algorithm should be used.
 
+## Retries
+
+Retries are for application-level failures (such as 5XX responses in the case of HTTP) as determined by the response
+classifier.  If the response classifier determines that a request is a retryable failure, and there is retry budget
```
klingerf: Typo here: there is _a_ retry budget available...

linkerd/docs/routers.md
```diff
@@ -186,9 +178,10 @@ Key | Default Value | Description
 hostConnectionPool | An empty object | see [hostConnectionPool](#host-connection-pool).
 tls | no tls | The router will make requests using TLS if this parameter is provided.  It must be a [client TLS](#client-tls) object.
 loadBalancer | [p2c](#power-of-two-choices-least-loaded) | A [load balancer](#load-balancer) object.
-retries | see [retries](#retries) | A [retry policy](#retries) for all clients created by this router.
+failFast | `false` | If `true`, connection failures are punished more aggressively. Should not be used with small destination pools.
+requeueBudget | see [retries](#retries) | A [requeue budget](#retries) for connection-level retries.
```
klingerf: It looks like `requestBudget` is of type `RetryBudgetConfig` rather than type `RetriesConfig`, so I think this should be:

```
budget | See [retry budget](#retry-budget-parameters) | A [requeue budget](#retry-budget-parameters) for connection-level retries.
```

linkerd/docs/routers.md
```diff
@@ -207,3 +200,82 @@ minSize | `0` | The minimum number of connections to maintain to each host.
 maxSize | Int.MaxValue | The maximum number of connections to maintain to each host.
 idleTimeMs | forever | The amount of idle time for which a connection is cached in milliseconds.
 maxWaiters | Int.MaxValue | The maximum number of connection requests that are queued when the connection concurrency exceeds maxSize.
+
+<a name="service-configuration"></a>
```
klingerf: This anchor isn't necessary since it matches the h2 header exactly. Same goes for the rest of the headers in this file. Can you remove all of them?

linkerd/docs/routers.md
```diff
@@ -207,3 +200,82 @@ minSize | `0` | The minimum number of connections to maintain to each host.
...
+
+```
+
+This service configuration allows you to specify [service parameters](#service-parameters)
+which will be applied to all services that match a specified prefix.  The service
+configuration must contain a property called `configs` which contains a list
+of config objects.  Each config object must specify a prefix and the
+[service parameters](#service-parameters) to apply to services that match that prefix.
+If a service matches more than one config's prefix, all parameters from the
+matching configs will be applied, with parameters from later configs taking
```
klingerf: Maybe clearer to say "with parameters defined later in the configuration file taking precedence of those defined earlier" (if that's actually the case.

linkerd/docs/routers.md
```diff
@@ -186,9 +247,10 @@ Key | Default Value | Description
 hostConnectionPool | An empty object | see [hostConnectionPool](#host-connection-pool).
 tls | no tls | The router will make requests using TLS if this parameter is provided.  It must be a [client TLS](#client-tls) object.
 loadBalancer | [p2c](#power-of-two-choices-least-loaded) | A [load balancer](#load-balancer) object.
-retries | see [retries](#retries) | A [retry policy](#retries) for all clients created by this router.
+failFast | `false` | If `true`, connection failures are punished more aggressively. Should not be used with small destination pools.
+requeueBudget | see [retry budget](#retry-budget-parameters) | A [requeue budget](#retries) for connection-level retries.
```
klingerf: Thanks for updating! I'd also link the second `requeue budget` link to `#retry-budget-parameters`.

linkerd/docs/routers.md
```diff
@@ -93,19 +88,91 @@ tls | no tls | The server will serve over TLS if this parameter is provided. see
...
+          percentCanRetry: 0.5
+          ttlSecs: 15
+        backoff:
+          kind: jittered
+          minMs: 10
+          maxMs: 10000
+    - prefix: /svc/foo
+        totalTimeoutMs: 500ms
+    - prefix: /svc/bar
+        totalTimeoutMs: 200ms
```
olix0r: indent looks weird here:
```yaml
  - prefix: /svc/foo
      totalTimeoutMs: 200ms
```

router/core/src/main/scala/io/buoyant/router/context/responseClassifier.scala
```diff
@@ -0,0 +1,5 @@
+package io.buoyant.router.context
+
+import com.twitter.finagle.param
+
+object ResponseClassifierCtx extends LocalKey[param.ResponseClassifier]("ResponseClassifier")
```
olix0r: probably worth a quick comment about what this is intended for

PR #1195, opened 2017-04-12
Upgrade repo to CircleCI 2.0

siggy:
Problem: CircleCI has been flakey.

Solution: Upgrade to CircleCI 2.0. This change allows for a more
controlled Docker environment to run our tests in. Also fixed an issue
in package.json where npm was failing when attempting to pull a package
over ssh.

Validation: Tests passing in 2.0. CI time decreased from ~19min to ~6.5min.
Passing CI run: https://circleci.com/gh/linkerd/linkerd/4425


.circleci/config.yml
```diff
@@ -0,0 +1,104 @@
...
+          CI_TERRIBLENESS: 30.seconds
+        parallelism: 2
+    working_directory: ~/linkerd
+    steps:
+      - checkout
+
+      - setup_remote_docker
+
+      - restore_cache:
+          key: linkerd-{{ checksum "project/LinkerdBuild.scala" }}-{{ checksum "admin/src/main/resources/io/buoyant/admin/package.json" }}
```
olix0r: this whole file could use some contents.  why are checksums based on these two files?  LinkerdBuild rarely changes...

siggy: will add some comments.

re: LinkerdBuild.scala, i tried to pick a file which, when changed, would best motivate invalidating `~/.ivy2` in the circleci cache. if there's a better file, i'm open to it.

related: any idea why we weren't caching `~/.ivy2` in circleci 1.0? it's a huge time saver.

.circleci/config.yml
```diff
@@ -0,0 +1,137 @@
...
+
+      # This allows us to run docker commands Since we're already in a docker
+      # container, this step sets up a remote docker environment for us.
+      - setup_remote_docker
+
+      # Restore build dependencies and downloaded files from the save_cache
+      # step. We use LinkerdBuild.scala as a proxy for Scala dependency changes,
+      # and package.json as a proxy for JavaScript dependency changes.
+      - restore_cache:
+          key: linkerd-{{ checksum "project/LinkerdBuild.scala" }}-{{ checksum "admin/src/main/resources/io/buoyant/admin/package.json" }}
```
olix0r: unfortunately, LinkerdBuild isn't really a good proxy for dependency changes. All of the files in files in projects basically need to be accounted for....

siggy: ok makes sense. i think there's a way to key off of entire directories will check.

.circleci/config.yml
```diff
@@ -0,0 +1,146 @@
...
+      # container, this step sets up a remote docker environment for us.
+      - setup_remote_docker
+
+      # cat the entire contents of the project/ directory, to be used in cache
+      # keys.
+      - run:
+          name: Generate cache seeds
+          command: |
+            set -x
+            find ~/linkerd/project/ -type f -exec cat {} \; > project-seed.txt
```
olix0r: might as well append admin/src/main/resources/io/buoyant/admin/package.json into this as well.  also i might be inclined to put this in /tmp

.circleci/config.yml
```diff
@@ -0,0 +1,146 @@
...
+
+      # Restore build dependencies and downloaded files from the save_cache
+      # step. We use project-seed.txt, generated from the entire contents of the
+      # project/ directory, as a proxy for Scala dependency changes. We use
+      # package.json as a proxy for JavaScript dependency changes.
+      - restore_cache:
+          key: linkerd-{{ checksum "project-seed.txt" }}-{{ checksum "admin/src/main/resources/io/buoyant/admin/package.json" }}
```
olix0r: and then this can be simplified

.circleci/config.yml
```diff
@@ -0,0 +1,146 @@
...
+      # Download and install protoc. Skip download if already in cache.
+      - run:
+          name: Install protoc
+          command: |
+            set -x
+            if [ ! -f protoc-3.1.0-linux-x86_64.zip ]; then
+              curl -sLO https://github.com/google/protobuf/releases/download/v3.1.0/protoc-3.1.0-linux-x86_64.zip
+            fi
+            unzip -p protoc-3.1.0-linux-x86_64.zip bin/protoc > /bin/protoc
+            chmod 755 /bin/protoc
```
olix0r: i don't think this is relevant any longer since we now use `./protoc` (which is what's actually called from within sbt).  And, instead, we have to cache `.protoc`

PR #1202, opened 2017-04-13
Sync up http and h2 identifier ids (#1176)

### LINKED ISSUE #1176: Reconcile HTTP/1.1 and HTTP/2 config kind ids
In working on our documentation, I've noticed some slight discrepancies in the kind ids for our http/1.1 and http/2 implementations, which I think we should fix as a breaking change.

In the linkerd/http project, we have:

File | ID
---- | --
HeaderIdentifierConfig.scala | `io.l5d.header`
HeaderTokenIdentifierConfig.scala | `io.l5d.header.token`
IngressIdentifier.scala | `io.l5d.http.ingress`
MethodAndHostIdentifierConfig.scala | `io.l5d.methodAndHost`
PathIdentifierConfig.scala | `io.l5d.path`
ResponseClassifiers.scala | `io.l5d.nonRetryable5XX`
ResponseClassifiers.scala | `io.l5d.retryableIdempotent5XX`
ResponseClassifiers.scala | `io.l5d.retryableRead5XX`

Whereas in the linkerd/h2 project, we have:

File | ID
---- | ---
HeaderPathIdentifier.scala | `io.l5d.headerPath`
HeaderTokenIdentifier.scala | `io.l5d.headerToken`
IngressIdentifier.scala | `io.l5d.h2.ingress`
ResponseClassifiers.scala | `io.l5d.h2.nonRetryable5XX`
ResponseClassifiers.scala | `io.l5d.h2.retryableIdempotent5XX`
ResponseClassifiers.scala | `io.l5d.h2.retryableRead5XX`

I'd like to suggest that we remove the protocol from these ids, and use the same id in both protocols for plugins that do roughly the same thing in each protocol. So this would mean:

* Rename `io.l5d.http.ingress` to `io.l5d.ingress`
* Rename `io.l5d.headerPath` to `io.l5d.path`
* Rename `io.l5d.headerToken` to `io.l5d.header`
* Rename `io.l5d.h2.ingress` to `io.l5d.ingress`
* Rename `io.l5d.h2.nonRetryable5XX` to `io.l5d.nonRetryable5XX`
* Rename `io.l5d.h2.retryableIdempotent5XX` to `io.l5d.retryableIdempotent5XX`
* Rename `io.l5d.h2.retryableRead5XX` to `io.l5d.retryableRead5XX`

Questions: Is this even possible? Do all plugins for a given initializer have to have a globally-unique identifier? Or can the IDs overlap as long as they're not part of the same protocol?

I'm also open to other suggestions too if we think there are better ways to go about.

klingerf:
## Problem

The existing http and h2 identifier IDs are slightly inconsistent (e.g.
`io.l5d.header.token` and `io.l5d.headerToken`). Additionally, some of
the ids are scoped by protocol, and others are not.

## Solution

Rename identifier IDs to make them more consistent. Update the
documentation accordingly, and fix some issues with broken anchors.

Fixes #1176.

### DIFF
--- a/config/src/main/scala/io/buoyant/config/Parser.scala
+++ b/config/src/main/scala/io/buoyant/config/Parser.scala
@@ -82,9 +82,9 @@ object Parser {
     mapper.setSerializationInclusion(Include.NON_ABSENT)
     mapper.setVisibility(PropertyAccessor.ALL, Visibility.PUBLIC_ONLY)
 
-    // Subtypes must not conflict
+    // Subtypes with the same config deserializer parent must not conflict
     for (kinds <- configInitializers) {
-      ensureUniqueKinds(kinds)
+      kinds.groupBy(_.configClass.getSuperclass).values.foreach(ensureUniqueKinds)
       for (k <- kinds) k.registerSubtypes(mapper)
     }
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,11 @@
 * Add timeoutMs option to server section in linkerd config.
 * Automatically upgrade all HTTP/1.0 messages to HTTP/1.1.
 * Add `io.l5d.influxdb` LINE telemeter
+* Refactor http and h2 identifiers for consistency:
+  * The `io.l5d.headerToken` id has been renamed to `io.l5d.header.token`.
+  * The `io.l5d.headerPath` id has been renamed to `io.l5d.header.path`.
+  * The `io.l5d.h2.ingress` id has been renamed to `io.l5d.ingress`.
+  * The `io.l5d.http.ingress` id has been renamed to `io.l5d.ingress`.
 
 ## 0.9.1 2017-03-15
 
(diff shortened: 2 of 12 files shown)


olix0r:
What are the http1 identifiers named?

klingerf:
@olix0r There's more context for this change in #1176. And after a bit more investigation I might be making a few more changes to this branch -- see my last comment there.

config/src/main/scala/io/buoyant/config/Parser.scala
```diff
@@ -84,7 +84,7 @@ object Parser {
 
     // Subtypes must not conflict
     for (kinds <- configInitializers) {
-      ensureUniqueKinds(kinds)
+      kinds.groupBy(_.configClass.getSuperclass).values.foreach(ensureUniqueKinds)
```
klingerf: How do we feel about this change? Happy to add a test but wanted to run it by folks first. The previous code checked for uniqueness among all IDs for a given initializer class. The new code checks for uniqueness among all IDs for a given initializer class _and_ the parent config class.

adleong: I think conflicting kinds between response classifiers for different protocols is going to be a problem at deserialization time because they are all subtypes of ResponseClassifierConfig.  So I think we want to keep this check.

klingerf: Ok, I added a test for this behavior in b7124fd.

PR #1209, opened 2017-04-14
Fix h2 memory leak in Netty4DispatcherBase

### LINKED ISSUE #1154: HTTP/2 stream memory leak
With this config:
```
- kind: io.l5d.recentRequests
  sampleRate: 1.0
  capacity: 10
```

A linkerd instance @ 6 req/sec with 1G heap can have ~700mb allocated to a recentRequests related class:
image

zackangelo:
Fixes #1154 

Fixes a leak caused by a bug in the H2 state machine logic in `Netty4StreamTransport`. 

The transport state was not transitioning to `LocalClosed` as intended when an h2 frame with the `endStream` flag was observed from `send`. This was due to the `Future` returned by `writeStream` not completing when the end of the stream was reached. 

When the remote end sent an `endStream`, the transport was permanently left in a half open state instead of closing, causing the corresponding state in `Netty4DispatcherBase` to never exit `StreamOpen`.  Since the `streams` table there grows without bound and `StreamOpen` instances have the underlying stream instance as a member, this caused a significant memory leak. 

Verified heap stability with ~70K requests, usage was stable at 42MB after (was >1GB previously).

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
@@ -531,8 +531,11 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
   private[this] def writeStream(stream: Stream): Future[Unit] = {
     def loop(): Future[Unit] =
       stream.read().rescue(wrapLocalEx)
-        .flatMap(writeFrame)
-        .before(loop())
+        .flatMap { f =>
+          writeFrame(f).flatMap { _ =>
+            if (!f.isEnd) loop() else Future.Unit
+          }
+        }
 
     localResetOnCancel(loop())
   }


wmorgan:
Nice!

adleong:
Amazing work!  I just tested it and verified that memory usage is way way down.  I think that we still may be leaking `StreamClosed` objects but these are tiny compared to leaking the actual streams (and we can fix that issue in a followup).

zackangelo:
@adleong thanks! yeah, that table still grows without bound, but will get gc'd when the channel closes. @olix0r pointed out that the `StreamClosed` object is a singleton, so we're only accumulating stream IDs at that point.

olix0r:
It's technically bounded, just with a very high bound... something like `2^31-1 / 2` ;)

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -531,8 +531,11 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
   private[this] def writeStream(stream: Stream): Future[Unit] = {
     def loop(): Future[Unit] =
       stream.read().rescue(wrapLocalEx)
-        .flatMap(writeFrame)
-        .before(loop())
+        .flatMap { f =>
+          writeFrame(f).flatMap { _ =>
```
olix0r: i'm still stunned that `writeFrame(f).before {` doesn't work...

wmorgan:
We owe you a hoodie and a beer :)

PR #1212, opened 2017-04-14
Jitter usage data telemeter (#1194)

### LINKED ISSUE #1194: jitter usage telemeter
currently we report at exact hour increments, which means it's possible/likely to get report storms when a cluster is started simultaneously.  if we jitter this timer, it may be a bit kinder on the service that receives these events.

esbie:
Fixes #1194 

Slightly randomizes reporting times to avoid traffic spikes.

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
@@ -30,6 +30,7 @@ import java.util.{Date, TimeZone}
 import java.util.concurrent.atomic.AtomicBoolean
 import scala.collection.JavaConverters._
 import scala.language.reflectiveCalls
+import scala.util.Random
 
 private[telemeter] object UsageDataTelemeter {
   val DefaultPeriod = 1.hour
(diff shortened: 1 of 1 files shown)


linkerd/core/src/main/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeter.scala
```diff
@@ -206,20 +207,25 @@ case class UsageDataTelemeter(
...
   private[this] def run0() = {
-    val _ = client(metrics)
+    val initialCall = DefaultTimer.twitter.schedule(jitter(1.minute)) {
+      val _ = client(metrics)
+    }
 
-    val task = DefaultTimer.twitter.schedule(DefaultPeriod) {
+    val recurringCall = DefaultTimer.twitter.schedule(DefaultPeriod.plus(jitter(1.minute))) {
       val _ = client(metrics)
     }
```
olix0r: Doesn't this initiate _two_ request loops?  i.e. `initialCall` is a task that will continue to submit updates every ~minute?

Should the first one just be `Future.sleep(jitter(1.minute).before(client(metrics))` ?

esbie: ah, duh, yes thank you!

PR #1220, opened 2017-04-17
Use service and client as dst stat names

adleong:
The names dst_id and dst_path which occur in linkerd's metrics, trace annotions
and headers are arcane, unintuitive, and not consistently used.

Rename dst_id to client and dst_path to service in metrics keys, trace
annotation keys, and header names.

### DIFF
--- a/linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
+++ b/linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
@@ -379,11 +379,11 @@ object LinkerdHeaders {
    * next hop.
    */
   object Dst {
-    val Path = Prefix + "dst-logical"
-    val Bound = Prefix + "dst-concrete"
+    val Path = Prefix + "dst-service"
+    val Bound = Prefix + "dst-client"
     val Residual = Prefix + "dst-residual"
 
-    /** Encodes `l5d-dst-path` on outgoing requests. */
+    /** Encodes `l5d-dst-service` on outgoing requests. */
     class PathFilter(path: Path) extends SimpleFilter[Request, Response] {
       private[this] val pathShow = path.show
       def apply(req: Request, service: Service[Request, Response]) = {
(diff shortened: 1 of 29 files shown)


linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
```diff
@@ -379,8 +379,8 @@ object LinkerdHeaders {
    * next hop.
    */
   object Dst {
-    val Path = Prefix + "dst-logical"
-    val Bound = Prefix + "dst-concrete"
+    val Path = Prefix + "dst-service"
+    val Bound = Prefix + "dst-client"
     val Residual = Prefix + "dst-residual"
 
     /** Encodes `l5d-dst-path` on outgoing requests. */
```
klingerf: I think this comment should read `l5d-dst-service`.

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -383,8 +383,8 @@ object Headers {
    * next hop.
    */
   object Dst {
-    val Path = Prefix + "dst-logical"
-    val Bound = Prefix + "dst-concrete"
+    val Path = Prefix + "dst-service"
+    val Bound = Prefix + "dst-client"
     val Residual = Prefix + "dst-residual"
 
     /** Encodes `l5d-dst-path` on outgoing requests. */
```
klingerf: `l5d-dst-service`

adleong:
Thanks for the thorough review, @klingerf!  I think I've updated all the sites now, but let me know if you see anything else that I've missed.  The linkerd dashboard seems to work now, although there is an unrelated bug about certain failures being reported as success: https://github.com/linkerd/linkerd/pull/1228

@rmars requesting a review from you since this touches a bunch of the admin js

adleong:
I will update srv to server in a followup, if that works for you

PR #1222, opened 2017-04-17
Fix handling of failure accrual default

adleong:
If a client config section does not include a failureAccrual property, the
default failure accrual policy will be applied, overriding any failure accrual
policy set in previous matching client configs.

We refactor slightly to use .maybeWith so that the failureAccrual param will
only be added if the property is defined.  A default value is supplied as a
router config that can be overwritten.

A test case in ClientTest validates this.

### DIFF
--- a/linkerd/core/src/test/scala/io/buoyant/linkerd/ClientTest.scala
+++ b/linkerd/core/src/test/scala/io/buoyant/linkerd/ClientTest.scala
@@ -3,14 +3,18 @@ package io.buoyant.linkerd
 import com.twitter.conversions.time._
 import com.twitter.finagle.buoyant.TlsClientPrep.Trust
 import com.twitter.finagle.loadbalancer.{Balancers, DefaultBalancerFactory, LoadBalancerFactory}
+import com.twitter.finagle.service.exp.FailureAccrualPolicy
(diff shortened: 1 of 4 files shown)


linkerd/core/src/test/scala/io/buoyant/linkerd/ClientTest.scala
```diff
@@ -88,4 +92,37 @@ class ClientTest extends FunSuite {
...
+                          |  failureAccrual:
+                          |    kind: io.l5d.foo
+                          |- prefix: "/#/io.l5d.fs/foo"
+                          |  # this should inherit the failure accrual policy
+                          |  # from above
+                          |""".stripMargin)
+
+    val fooParams = client.clientParams.paramsFor(Path.read("/#/io.l5d.fs/foo"))
+    val policy = fooParams[FailureAccrualFactory.Param]
+      .asInstanceOf[{ def failureAccrualPolicy: () => FailureAccrualPolicy }]
```
adleong: to cut through finagle's insanely restrictive visibility

olix0r: woa

adleong:
The new test case catches this.  Notice that failureAccrual is not specified in the second config block.

Issue #1240, opened 2017-04-24
Port namerd's thrift interface to netty4


klingerf:
As part of #1239, we realized that namerd's `io.l5d.thriftNameInterpreter` interface still uses netty3, which requires openssl for serving the interface with tls. We should move that interface to netty4, and then switch namerd's base docker image to match linkerd's base docker image (library/openjdk:8-jre-alpine), which does not include openssl.

adleong:
Unfortunately, moving to netty4 while still supporting pkcs#1 keys means that we need to use the LegacyKeyServerEngineFactory which depends on openssl.  We can't switch docker images until we drop support for pkcs#1.

siggy:
Related: #1366.

PR #1244, opened 2017-04-25
Revert "Use the openjdk:8-jre-alpine base docker image (#1214)"

klingerf:
## Problem

With the buoyantio/linkerd:1.0.0 docker image, it's not possible to configure a netty3 router to serve with tls. It fails with the following error:

```
W 0425 21:16:20.761 UTC THREAD29 org.jboss.netty.channel.socket.nio.AbstractNioSelector: Failed to initialize an accepted socket.
java.io.IOException: Cannot run program "openssl": error=2, No such file or directory
```

## Solution

Revert to the java base image that we previously used pre-1.0.0. Relates to #1240. As part of this change, I'm also going to republish the buoyantio/linkerd:1.0.0 image to docker hub.

### DIFF
--- a/project/Base.scala
+++ b/project/Base.scala
@@ -137,7 +137,7 @@ class Base extends Build {
     assemblyJarName in assembly := s"${name.value}-${version.value}-${configuration.value}-exec",
     docker <<= docker dependsOn (assembly in configuration),
     dockerEnvPrefix := "",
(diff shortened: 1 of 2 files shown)


adleong:
Would dropping support for netty3 also be a solution?  Are we ready to do that?

olix0r:
That's a potential solution... but it will require changing TLS key formats...  So not without going to 1.1.0

PR #1256, opened 2017-04-27
Discard responses when retrying

### LINKED ISSUE #1178: Leaking connection on 500 retrys
Posting this as a followup to a Slack conversation with Alex. This is against 0.9.0 of linkerd

We are noticing that linkerd's connection pool for certain endpoints grows rapidly during error conditions, then doesn't shrink. When the same app gets a burst of normal traffic, the pool shrinks back to 0 fairly quickly.

The timeline is:
* Request comes in. We send that request to linkerd.
* Linkerd returns an error to us after 20 seconds(our request timeout)
* Connection pool has ~83 idle connections that do not disappear
* metrics.json showed 83 500 status errors and 83 ResponseClassificationSyntheticException
* a `netstat -atpn | grep <L5d.pid> | grep CLOSE_WAIT` shows 83 sockets in CLOSE_WAIT
* It happened on Friday, and the connections are still there the following Monday

The service recovered, and is still working now. The connection pool number is correct, linkerd is holding onto CLOSE_WAIT sockets. This will eventually become an issue requiring a restart.

This happened 2 times, so the numbers in the attached metrics.json are a bit different. This also occurred on 3 separate linkerd instances at about the same time. The cause was the way the downstream app responded, but it affected 3 independent linkerds. Alex identified this in slack:
> I think you got stuck in a retry loop.  `"rt/cgp/dst/path/svc/live/appName/retries/total": 274,`  something about that one request triggered a 5xx response from the server and it kept being retried and failing until either the retry budget was exhausted or the request timed out
> it looks like there were 99 requests, 97 of which were successful, and 2 of which were failures and those failures triggered 274 retries

We have retryableRead500 tuned on, which is likely the reason for the large number of connections. This is a server in our staging environment, so it gets very little traffic. In the case of the error, a single request came in that caused 83 abandoned connections.

[metrics.json](https://github.com/linkerd/linkerd/files/893581/metrics.txt)
Compare number of connections in metrics with CLOSE_WAIT
```
root@ip-10-122-17-136:~# netstat -atpn | grep 3887 |grep CLOSE_WAIT |wc
    166    1162   16102
```
Showing all CLOSE_WAIT are to the server that errored out.
```
root@ip-10-122-17-136:~# netstat -atpn | grep 3887 |grep CLOSE_WAIT | grep 10.122.3.77:31392|wc
    166    1162   16102
```

And this server is still serving traffic on that ip:port. We just had some sort of disruption for this brief window.

I have the metrics being captured by datadog, so we have how the metrics changed on a 30 second window.
This shows the connection count
connections
This shows the count of 500 errors
500statuscounter
And this is the bytes_sent stat for the application
bytespersec
 

# Debugging attempts
We attempted to see if using the connections would result in them being cleaned up. I flooded a server that had 159 connections and 159 sockets in CLOSE_WAIT. The pool grew to 360+, then returned back to 159

# Dead sockets
This is a sample of netstat showing the dead sockets that match the leaked connection counts
```
tcp        1      0 10.122.8.59:56734       10.122.3.77:31392       CLOSE_WAIT  4228/java       
tcp        1      0 10.122.8.59:35906       10.122.3.77:31392       CLOSE_WAIT  4228/java 
```

# Linkerd log
```2017/03/31 18:17:11 I 0331 18:17:11.572 UTC THREAD25 TraceId:4f2d95dc50b33383: FailureAccrualFactory marking connection to "#/consul/us-east-1-vpc-XXXX/live/appName" as dead
. Remote Address: Inet(/10.122.3.77:31392,Map())
2017/03/31 18:17:16 I 0331 18:17:16.862 UTC THREAD24 TraceId:4f2d95dc50b33383: FailureAccrualFactory marking connection to "#/consul/us-east-1-vpc-XXXX/live/appName" as dead
. Remote Address: Inet(/10.122.3.77:31392,Map())
2017/03/31 18:17:26 I 0331 18:17:26.970 UTC THREAD23 TraceId:4f2d95dc50b33383: FailureAccrualFactory marking connection to "#/consul/us-east-1-vpc-XXXX/live/appName" as dead
. Remote Address: Inet(/10.122.3.77:31392,Map())
2017/03/31 18:17:30 E 0331 18:17:30.874 UTC THREAD10: service failure
2017/03/31 18:17:30 com.twitter.finagle.IndividualRequestTimeoutException: exceeded 20.seconds to 0.0.0.0/4140 while waiting for a response for an individual request, excluding retries. Remo
te Info: Not Available
2017/03/31 18:17:30
2017/03/31 18:17:31 E 0331 18:17:31.060 UTC THREAD24 TraceId:4f2d95dc50b33383: service failure
2017/03/31 18:17:31 Failure(20.seconds, flags=0x03) with RemoteInfo -> Upstream Address: /127.0.0.1:54910, Upstream Client Id: Not Available, Downstream Address: /10.122.3.77:31392, Downstream Client Id: #/consul/us-east-1-vpc-XXXX/live/appName, Trace Id: 4f2d95dc50b33383.4f2d95dc50b33383<:4f2d95dc50b33383
2017/03/31 18:17:31 Caused by: com.twitter.util.TimeoutException: 20.seconds
2017/03/31 18:17:31     at com.twitter.util.Future$$anonfun$within$1.apply(Future.scala:992)
```

# Config file
```
admin:
  port: 9990
routers:
- protocol: http
  httpAccessLog: /opt/proxy/logs/access.log
  label: cgp
  timeoutMs: 2000
  identifier:
  - kind: io.l5d.header
    header: X-Route-Id
  interpreter:
    kind: io.l5d.namerd
    dst: /$/inet/namerd/4100
    namespace: header
  responseClassifier:
    kind: io.l5d.retryableRead5XX
  client:
    loadBalancer:
      kind: ewma
    failureAccrual:
      kind: io.l5d.successRate
      successRate: 0.9
      requests: 20
      backoff:
        kind: constant
        ms: 10000
  servers:
  - port: 4140
    ip: 0.0.0.0
```

adleong:
Fixes #1178

When an HTTP response with a chunked body is classified as a retryable failure
and is retried, the chunked body is never read and the request is discarded.
The connection is never released because the response has unread chunks.  This
results in a connection leak.

Introduce a ResponseDiscarder stack param which is invoked on responses when
the RetryFilter decides to retry and thus discard the response.  For HTTP we
supply a ResponseDiscarder that explicitly discards chunked bodies.

Tested manually against a node.js server that returns chuncked 500s and added
an e2e test.  Shout out to @taer for providing the node.js server and @fantayeneh
for providing the test case.

### DIFF
--- a/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/RetriesEndToEndTest.scala
+++ b/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/RetriesEndToEndTest.scala
@@ -5,6 +5,7 @@ import com.twitter.finagle.buoyant.linkerd.Headers
 import com.twitter.finagle.http.{Method, Response, Request}
 import com.twitter.finagle.stats.{InMemoryStatsReceiver, NullStatsReceiver}
 import com.twitter.finagle._
+import com.twitter.io.Buf
 import com.twitter.util.{Future, Time, Var}
 import com.twitter.finagle.tracing.NullTracer
 import io.buoyant.test.FunSuite
@@ -619,5 +620,55 @@ class RetriesEndToEndTest extends FunSuite {
       await(server.close())
       await(router.close())
     }
+  }
+
+  test("chunked error responses should not leak connections on retries") {
+    val stats = new InMemoryStatsReceiver
+    val tracer = NullTracer
+
+    val downstream = Downstream("dog", Service.mk { req =>
+      val rsp = Response()
+      rsp.statusCode = 500
+      rsp.setChunked(true)
+      rsp.close()
+      Future.value(rsp)
+    })
+
+    val label = s"$$/inet/127.1/${downstream.port}"
+    val dtab = Dtab.read(s"/svc/dog => /$label;")
+    val yaml =
+      s"""|routers:
+          |- protocol: http
+          |  dtab: ${dtab.show}
+          |  service:
+          |    responseClassifier:
+          |      kind: io.l5d.http.retryableRead5XX
+          |  servers:
+          |  - port: 0
+          |""".stripMargin
+    val linker = Linker.load(yaml)
+      .configured(param.Stats(stats))
+      .configured(param.Tracer(tracer))
+    val router = linker.routers.head.initialize()
+    val server = router.servers.head.serve()
+    val client = upstream(server)
+
+    try {
+
+      val req = Request()
+      req.host = "dog"
+      val errrsp = await(client(req))
+      assert(errrsp.statusCode == 500)
+      assert(stats.counters.get(Seq("http", "server", "127.0.0.1/0", "requests")) == Some(1))
+      assert(stats.counters.get(Seq("http", "client", label, "requests")) == Some(101))
+      assert(stats.gauges.get(Seq("http", "client", label, "connections")).map(_.apply.toInt) == Some(1))
+
+    } finally {
+      await(client.close())
+      await(downstream.server.close())
+      await(server.close())
+      await(router.close())
+    }
+
   }
 }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## In the next release...
+
+* Fix connection leak when retrying on responses with chunked bodies.
+
 ## 1.0.0 2017-04-24
 
 * Configuration:
--- a/router/core/src/main/scala/com/twitter/finagle/buoyant/RetryFilter.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/buoyant/RetryFilter.scala
@@ -17,7 +17,8 @@ class RetryFilter[Req, Rep](
   retryPolicy: RetryPolicy[(Req, Try[Rep])],
   timer: Timer,
   statsReceiver: StatsReceiver,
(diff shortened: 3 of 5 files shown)


linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/RetriesEndToEndTest.scala
```diff
@@ -620,4 +621,54 @@ class RetriesEndToEndTest extends FunSuite {
...
+      .configured(param.Stats(stats))
+      .configured(param.Tracer(tracer))
+    val router = linker.routers.head.initialize()
+    val server = router.servers.head.serve()
+    val client = upstream(server)
+
+    try {
+
+      val req = Request()
+      req.host = "dog"
```
leozc: cat ..

adleong: what?

leozc:
Thanks when would this CR be released?

olix0r:
@leozc i expect that we'll do a release by early next week, depending on progress of some other pending issues.

leozc:
Thanks, I will do coordination of release cycle on my side.

Cheers
Leo

PR #1257, opened 2017-04-28
Upgrade to scala 2.12

adleong:


project/Base.scala
```diff
@@ -47,7 +47,8 @@ class Base extends Build {
     organization := "io.buoyant",
     version := Git.version,
     homepage := Some(url("https://linkerd.io")),
-    scalaVersion in GlobalScope := "2.11.7",
+    scalaVersion in GlobalScope := "2.12.1",
```
zackangelo: Just a heads up, 2.12.2 is out now.

adleong: Thanks!  Even though it might not be strictly necessary, I think we're going to mirror the scala version that finagle uses for now.

olix0r:
I believe finagle has 2.12 builds now, but we should check that all of our
dependencies are covered

finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
```diff
@@ -78,6 +78,7 @@ class Netty4StreamTransportTest extends FunSuite {
 
       eventually { assert(transport.onReset.isDefined) }
       assert(await(transport.onReset.liftToTry) == Throw(StreamError.Remote(rst)))
+      ()
     }
```
olix0r: So I understand... what does assert return now?

namerd/iface/interpreter-thrift/src/test/scala/io/buoyant/namerd/iface/ThriftNamerInterfaceTest.scala
```diff
@@ -16,7 +16,7 @@ import org.scalatest.FunSuite
 class ThriftNamerInterfaceTest extends FunSuite with Awaits {
   import ThriftNamerInterface._
 
-  def retryIn() = 1.second
+  val retryIn = () => 1.second
```
olix0r: so I understand, why is this necessary?

adleong: `Eta-expansion of zero-argument method values is deprecated.`

It works if we replace usages of `retryIn` with `retryIn _` but I figured just val'ing up the function to begin with would be easier to understand.

finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
adleong: It returns [Assertion](http://doc.scalatest.org/3.0.1/index.html#org.scalatest.compatible.Assertion) which is just a marker trait.

Issue #1163, opened 2017-03-27
Provide an option to turn off the error messages 


naveensrinivasan:
The linkerd does not provide an option to turn off error messages like  `No hosts are available for /svc/code.xyz.io, Dtab.base=[/srv=>/#/io.l5d.k8s/default/http;/host=>/srv;/svc=>/host;/svc/zipkin=>/$/fail;/svc/zipkin-collector=>/$/fail], Dtab.local=[]. Remote Info: Not Available` this from being retruned to the client. 
This could be useful in development mode but in production we wouldn't want the clients to know it is being served by linkerd.

linkerd should provide an option to turn on and off these messages based on ENV variable.

olix0r:
It think it would be best to tie this behavior to the `clearContext` setting, so it can be controlled per-server (and not on the entire process).  I think the semantics are: if the l5d-err header is set, drop the message body.

KarolisL:
Hello,

Even with `clearContext: true`, we get verbose errors from linkerd:
```
$ curl -v localhost:8081/ ; echo
* About to connect() to localhost port 8081 (#0)
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8081 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.29.0
> Host: localhost:8081
> Accept: */*
>
< HTTP/1.1 400 Bad Request
< l5d-err: Unknown+destination%3A+Request%28%22GET+%2F%22%2C+from+%2F10.0.0.15%3A33672%29+%2F+not+enough+segments+in+path
< Content-Type: text/plain
< Content-Length: 93
<
* Connection #0 to host localhost left intact
Unknown destination: Request("GET /", from /10.0.0.15:33672) / not enough segments in path
```

Our config includes `clearContext` option. Is this the expected behaviour? According to the commited tests, `linkerd` should return empty body on such errors.

Our config:

```

    - protocol: http
      label: mobile-api-router
      identifier:
        kind: io.l5d.path
        segments: 2
      dtab: |-
        /payments_srv => /$/inet/momo-payments/8080;
       <...snip...>
        /svc/v1/operations => /payments_srv;
       <...snip...>
      servers:
      - ip: 0.0.0.0
        port: 8081
        clearContext: true
      service:
        responseClassifier:
          kind: io.l5d.http.nonRetryable5XX
      client:
        kind: io.l5d.global
        failFast: false
        failureAccrual:
          kind: none
```

[...]

PR #1258, opened 2017-04-28
Remove linkerd headers and body when clearContext is set (#1163)

fantayeneh:


adleong:
Thanks, @fantayeneh!!!

PR #1264, opened 2017-05-02
Add path stack registry

adleong:
There is little visibility into what path stacks have been created and how they
have been configured.

Add a path stack registry that exposes information about the modules in each
path stack and the params that configure them.  This information is exposed
in the global registry which can be viewed at /admin/registry.json on the
admin site.  Also improve the readability of the ResponseClassifier and
Trust.Config params.

screen shot 2017-05-02 at 10 54 51 am

### DIFF
--- a/router/core/src/main/scala/io/buoyant/router/ClassifiedRetries.scala
+++ b/router/core/src/main/scala/io/buoyant/router/ClassifiedRetries.scala
@@ -35,7 +35,7 @@ object ClassifiedRetries {
     private[this] val underlying = a orElse b
     def isDefinedAt(reqRep: ReqRep): Boolean = underlying.isDefinedAt(reqRep)
     def apply(reqRep: ReqRep): ResponseClass = underlying(reqRep)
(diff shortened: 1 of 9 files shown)


router/core/src/main/scala/io/buoyant/router/ClassifiedRetries.scala
```diff
@@ -29,6 +29,16 @@ object ClassifiedRetries {
   }
 
   /**
+   * A replacement for ResponseClassifier's orElse that gives a nice toString.
+   */
+  def orElse(a: ResponseClassifier, b: ResponseClassifier): ResponseClassifier = new ResponseClassifier {
+    private[this] val underlying = a orElse b
+    def isDefinedAt(reqRep: ReqRep): Boolean = underlying.isDefinedAt(reqRep)
+    def apply(reqRep: ReqRep): ResponseClass = underlying(reqRep)
+    override def toString: String = s"$a; $b"
```
olix0r: maybe `s/;/ orElse/?`

PR #1266, opened 2017-05-02
Client stats should use service response classifier

adleong:
If a response classifier is configured on a service, that response classifier
is not used to report client level success and failure metrics.  This can lead
to inconsistent metric counts.

Modify the PerDstStatsFilter to use the response classifier from the service
(stored in a local context).  Also add io.l5d.{http, h2}.allSuccessful response
classifiers to aid with testing and for general usefulness.


linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ResponseClassifiers.scala
```diff
@@ -36,3 +37,14 @@ class NonRetryable5XXInitializer extends ResponseClassifierInitializer {
...
+class AllSuccessfulConfig extends ResponseClassifierConfig {
+  def mk: ResponseClassifier = ClassifiedRetries.Default
+}
+
+class AllSuccessfulInitializer extends ResponseClassifierInitializer {
+  val configClass = classOf[AllSuccessfulConfig]
+  override val configId = "io.l5d.h2.allSuccessful"
+}
+
+object AllSuccessfulInitializer extends AllSuccessfulInitializer
```
esbie: nit: eof newline

PR #1269, opened 2017-05-03
Support ServiceInstance objects with a custom payload

dhay:
This change allows the Curator namer to support ServiceInstance objects that have a custom payload.  Assumes that the class referenced by the ServiceInstance record is in the classpath (e.g. in the /plugins directory)


wmorgan:
Thanks, @dhay! We'll take a look!

PR #1273, opened 2017-05-04
Namespaced Kubernetes Namer

### LINKED ISSUE #1199: New k8s namer that only resolves in 1 namespace
The current [Kubernetes namer](https://linkerd.io/config/head/linkerd/index.html#kubernetes-service-discovery) that we offer takes 4 parameters: namespace, port name, service name, and label value. This allows linkerd to route to any service in all available Kubernetes namespaces. In larger production setups, it may be better to restrict Linkerd to only route to services in the namespace where it is currently running. We should introduce a new namer to accomplish that, and figure out the best way to retrieve the current namespace where Linkerd is running from the Kubernetes downward API.

esbie:
Problem:
Users have requested a version of the k8s namer that is restricted
to only route to services in the namespace where linkerd is currently
running.

Solution:
This version of the k8s namer reads the namespace name from the
downward api and uses that namespace instead of specifying the
namespace in the dtab. Fixes #1199

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -10,14 +10,15 @@ import io.buoyant.namer.{EnumeratingNamer, Metadata}
 import java.net.{InetAddress, InetSocketAddress}
 import scala.collection.mutable
 
-class EndpointsNamer(
+class MultiNsNamer(
   idPrefix: Path,
   labelName: Option[String],
   mkApi: String => NsApi,
   backoff: Stream[Duration] = Backoff.exponentialJittered(10.milliseconds, 10.seconds)
-)(implicit timer: Timer = DefaultTimer.twitter) extends EnumeratingNamer {
+)(implicit timer: Timer = DefaultTimer.twitter) extends EndpointsNamer(idPrefix, mkApi, backoff)(timer) {
 
-  import EndpointsNamer._
+  val PrefixLen = 3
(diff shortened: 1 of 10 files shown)


k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -49,12 +50,67 @@ class EndpointsNamer(
...
+  mkApi: String => NsApi,
+  backoff: Stream[Duration] = Backoff.exponentialJittered(10.milliseconds, 10.seconds)
+)(implicit timer: Timer = DefaultTimer.twitter) extends EndpointsNamer(idPrefix, mkApi, backoff)(timer) {
+
+  val PrefixLen = 2
+  private[this] val variablePrefixLength = PrefixLen + labelName.size
+
+  /**
+   * Accepts names in the form:
+   *   /<port-name>/<svc-name>/residual/path
```
adleong: Maybe add a note here and in the scaladoc for the multins lookup method about the optional label segment.

linkerd/docs/namer.md
```diff
@@ -318,6 +318,123 @@ port-name | yes | The port name.
...
+      labels:
+        app: l5d
+    spec:
+      volumes:
+      - name: l5d-config
+        configMap:
+          name: "l5d-config"
+      containers:
+      - name: l5d
+        image: buoyantio/linkerd:1.0.0-SNAPSHOT
```
adleong: what do you think about using the `latest` tag so that we don't have to update this for each version bump?

esbie: that's a great idea, thanks!

PR #1274, opened 2017-05-05
Expire metrics for idle services and clients

adleong:
If a service or client is idle for long enough (10 minutes by default), the
ServiceFactory will be closed and removed from cache but the metrics for that
service or client will live on forever.

We add the ability to prune MetricsTrees and delete all metrics for a service
or client when that ServiceFactory is closed.  We also expose a config option
for setting the ttl for idle ServiceFactories.

Tested manually.  Unable to create an end-to-end test because IdlingFactory
uses Stopwatch to determine idle time which does not respect TimeControl.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 * Upgrade to scala 2.12.
 * Fix connection leak when retrying on responses with chunked bodies.
+* Expire idle services and clients.
 
 ## 1.0.0 2017-04-24
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,13 +1,35 @@
-## In the next release...
+## In the latest release...
+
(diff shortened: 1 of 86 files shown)


adleong:
Good question.  If a service or client goes idle for the idleTtl and then becomes active again, the metrics will be deleted and recreated, effectively resetting the counters.  This could definitely impact the usage data.  We could lengthen the default idle ttl to longer than the 1 hour reporting interval.

olix0r:
I have a feeling the usage data reporter doesn't need any stats that would be reaped, but we should confirm. Even if it does, i think I'd prefer to try to get the usage reporter to compensate for this in some other way (like polling stats more aggressively than the idle timeout and saving them to be reported later).

PR #1280, opened 2017-05-08
Ensure h2 streams are fully received before closing (#1245)

### LINKED ISSUE #1245: H2 connection starts to deadline every request on a connection after random interval
We are mirroring some gRPC production traffic through linkerd and after a random interval (anywhere from 30 seconds to 2 hours) every request over the connection starts to deadline. After this point all of the request charts in the admin UI go to zero. If the connection is recreated through either restarting linkerd or restarting the client service, traffic flow is temporarily restored. 

There are no abnormal log messages with `TRACE` verbosity turned on when linkerd gets in this state. 

- We did a tcpdump to verify that the traffic was reaching linkerd (it is). We observe that the client service is sending h2 request frames (`DATA`, `HEADERS`) over the stream, and then after the deadline interval sends a `RST_STREAM`, which is intended behavior. Nothing during this interval is sent from linkerd other than TCP ACK packets. 
- We also ran a test without linkerd to see if it works (it does).
- We tried disabling failure accrual on the client linkerd and the server linkerd, it didn't make a difference.

Metrics snapshot below.

klingerf:
## Problem

A race condition existed that would cause a stream to be closed by the remote sender before the local receiver had fully received all frames in an h2 stream. This would result in the final frame of an h2 response never being sent back to the caller, causing the caller's connection to hang indefinitely.

## Solution

Make the receiver fully responsible for closing streams, to ensure that they are fully received before closing.

## Validation

I've deployed this branch to a test environment and it has successfully served gRPC traffic for the past 48 hours, whereas without this fix the connection hangs within about 20 minutes.

As part of this change I've included some small refactors for readability in the Netty4StreamTransport. Will also try to add a unit test in a follow-up commit. Fixes #1245.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
@@ -189,23 +189,23 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
     // appropriate.
     remoteMsgP.setInterruptHandler {
       case err: Reset =>
-        log.debug("[%s] remote message interrupted %s", prefix, err)
+        log.debug(err, "[%s] remote message interrupted", prefix)
         localReset(err)
 
       case Failure(Some(err: Reset)) =>
-        log.debug("[%s] remote message interrupted %s", prefix, err)
+        log.debug(err, "[%s] remote message interrupted", prefix)
         localReset(err)
 
       case f@Failure(_) if f.isFlagged(Failure.Interrupted) =>
-        log.debug("[%s] remote message interrupted %s", prefix, f)
+        log.debug(f, "[%s] remote message interrupted", prefix)
         localReset(Reset.Cancel)
 
       case f@Failure(_) if f.isFlagged(Failure.Rejected) =>
-        log.debug("[%s] remote message interrupted %s", prefix, f)
+        log.debug(f, "[%s] remote message interrupted", prefix)
         localReset(Reset.Refused)
 
       case e =>
-        log.debug("[%s] remote message interrupted %s", prefix, e)
+        log.debug(e, "[%s] remote message interrupted", prefix)
         localReset(Reset.InternalError)
     }
 
@@ -505,7 +505,7 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
     val streamFF = headersF.map(_ => writeStream(msg.stream))
 
     val writeF = streamFF.flatten
-    onReset.onFailure(writeF.raise)
+    onReset.onFailure(writeF.raise(_))
     writeF.respond {
       case Return(_) =>
         closeLocal()
@@ -534,7 +534,7 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
     streamFF
   }
 
-  private[this] def writeHeaders(hdrs: Headers, eos: Boolean): Future[Unit] =
+  private[this] val writeHeaders: (Headers, Boolean) => Future[Unit] = { (hdrs, eos) =>
     stateRef.get match {
       case Closed(rst) => Future.exception(StreamError.Remote(rst))
       case LocalClosed(_) => Future.exception(new IllegalStateException("writing on closed stream"))
@@ -543,9 +543,10 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
           Future.exception(StreamError.Local(Reset.ProtocolError))
         } else localResetOnCancel(transport.write(streamId, hdrs, eos))
     }
+  }
 
   /** Write a request stream to the underlying transport */
-  private[this] def writeStream(stream: Stream): Future[Unit] = {
+  private[this] val writeStream: Stream => Future[Unit] = { stream =>
     def loop(): Future[Unit] =
       stream.read().rescue(wrapLocalEx)
         .flatMap { f =>
@@ -568,7 +569,7 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
     p
   }
 
-  private[this] def writeFrame(frame: Frame): Future[Unit] =
+  private[this] val writeFrame: Frame => Future[Unit] = { frame =>
     stateRef.get match {
       case Closed(rst) => Future.exception(StreamError.Remote(rst))
       case LocalClosed(_) => Future.exception(new IllegalStateException("writing on closed stream"))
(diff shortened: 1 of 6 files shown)


mattolson:
🙏

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -187,28 +184,28 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
       case f@Failure(_) if f.isFlagged(Failure.Rejected) =>
-        log.debug(f, "[%s] remote message interrupted", prefix)
+        log.debug("[%s] remote message interrupted %s", prefix, f)
         localReset(Reset.Refused)
 
       case e =>
-        log.debug(e, "[%s] remote message interrupted", prefix)
+        log.debug("[%s] remote message interrupted %s", prefix, e)
         localReset(Reset.InternalError)
     }
```
olix0r: the prior format is used so that we print stack traces when they are enabled (this can be controlled at start time via a system property).

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -551,7 +568,7 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
     p
   }
 
-  private[this] val writeFrame: Frame => Future[Unit] = { frame =>
+  private[this] def writeFrame(frame: Frame): Future[Unit] =
```
olix0r: why is this necessary?  We prefer to use val'd functions when possible, since they save on unnecessary anonymous function allocations.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -488,7 +505,7 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
     val streamFF = headersF.map(_ => writeStream(msg.stream))
 
     val writeF = streamFF.flatten
-    onReset.onFailure(writeF.raise(_))
+    onReset.onFailure(writeF.raise)
```
olix0r: the prior form is used to be explicit that this allocates.  We use val'd functions and the paren-less form to indicate that there is no allocation performed, and we use parens to illustrate that an anonymous function is allocated (i.e. when we call a function rather than pass a val'd function).

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -435,29 +441,40 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
+            if (resetFromLocal(remote, Reset.InternalError)) false
+            else recv(data)
+
+          case Open(remote@RemoteStreaming()) =>
+            if (recvFrame(toFrame(data), remote)) true
+            else recv(data)
+
+          case LocalClosed(remote@RemoteStreaming()) =>
+            if (recvFrame(toFrame(data), remote)) true
+            else recv(data)
```
olix0r: Is this change related to the bugfix in some way or is this just a refactor?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -274,7 +271,6 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
 
       case state@RemoteClosed() =>
         if (stateRef.compareAndSet(state, Closed(Reset.NoError))) {
-          state.close()
```
olix0r: Is it worth adding a comment here explaining how state.close is expected to be called?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
klingerf: Ah, this is good to know. Which system property enables stack traces? Does that also world for Failure types (cases 3 and 4)? I can revert this change.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
klingerf: This is just a refactor. I kept getting confused that we process `Http2HeadersFrame.isEndStream` and `Http2HeadersFrame` in separate cases, but we process all `Http2DataFrame` in the same case. Breaking data frames out into two separate cases feels more consistent with how we're handling header frames.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
klingerf: Ok, good to know. Will revert.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
klingerf: I changed this for consistency with the `writeHeaders` and `writeStream` defs defined above. I can revert this, but I might also convert those other defs to vals.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -442,8 +442,10 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
 
           case Open(remote@RemoteStreaming()) =>
             if (stateRef.compareAndSet(state, remote.toRemoteClosed)) {
-              if (recvFrame(toFrame(data), remote)) true
-              else throw new IllegalStateException("stream queue closed prematurely")
+              if (recvFrame(toFrame(data), remote)) {
+                remote.close()
+                true
+              } else throw new IllegalStateException("stream queue closed prematurely")
```
zackangelo: Should there be `remote.close()` here as well if the stream queue is closed?

klingerf: If `recvFrame` returns false, that means that the call to `remote.offer` failed, due to the stream already being closed. So I think in this case calling `remote.close` again would be redundant.

PR #1284, opened 2017-05-09
Add error messages to namerd's http iface BadRequest responses

esbie:
It can be hard to debug dtab syntax. This change shows the illegal argument exception message to the caller. eg:

```bash
$ curl -v -X PUT -d"/foo ==> /bar" -H "Content-Type: application/dtab" $NAMERD_URL/api/1/dtabs/baz
...
< HTTP/1.1 400 Bad Request
< Content-Length: 47

'>' expected but '=' found at '/foo =[=]> /bar'
```


esbie:
I had a dream someone submitted this same CL

PR #1285, opened 2017-05-09
Fix deprecated eta-expansion of zero-argument method values

elecnix:
As per [SI-7187](https://issues.scala-lang.org/browse/SI-7187), eta-expansion is deprecated in Scala 2.12.

This was shown as an error in Eclipse Scala IDE 4.6.0-rc2.


wmorgan:
Thanks @elecnix! We'll take a look.

PR #1286, opened 2017-05-09
Escape metrics labels in the Prometheus Telemeter

### LINKED ISSUE #1236: Escape labels in prometheus telemeter
If for some reason our entities get labeled as their unicode hex values, prometheus chokes:

for example something like, `\x5b\x31\x32\x33\x2e\x31\x32\x33\x2e\x31\x32\x33\x2e\x31\x32\x33\x5d` (`[123.123.123.123]`) causes an error like `text format parsing error in line 95: invalid escape sequence '\x'` in prometheus. In PrometheusTelemeter, we escape the metric name but we should escape the label as well. 

From https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details:
```
metric_name and label_name have the usual Prometheus expression language restrictions. label_value can be any sequence of UTF-8 characters, but the backslash, the double-quote, and the line-feed characters have to be escaped as \\, \", and \n, respectively. 
```

We should maybe echo what the official clients do e.g. https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Collector.java#L163
https://github.com/prometheus/common/blob/master/expfmt/text_parse.go#L699

we can escape `\` and `"` by the following rather hard to follow regexps:
```
  private[this] val unescapedQuotes = """\\([\s\S])|(")""".r
  private[this] val escapeBackslashes = """(?<!\\)\\(?=(?:\\\\)*(?!\\)(?!\"))""".r

  private[this] def escapeLabelVal(str: String) = {
    val allQuotesEscaped = unescapedQuotes.replaceAllIn(str, """\\$1$2""")
    escapeBackslashes.replaceAllIn(allQuotesEscaped, """\\\\""")
  }
```
via [here](https://gist.github.com/getify/3667624) and [here](http://stackoverflow.com/questions/14977410/how-to-escape-all-unescaped-slashes-in-python) - though it would be easier to just replace them like we have been doing for the metric key name

rmars:
Problem:
if someone names their labels with unicode hex characters, or single quotes,
prometheus chokes because we don't escape them in the prometheus telemeter

Solution:
Escape these characters in the telemeter

Validation:
Wrote a test, ran prometheus and checked for errors

Fixes #1236

### DIFF
--- a/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
+++ b/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
@@ -34,14 +34,20 @@ class PrometheusTelemeter(metrics: MetricsTree) extends Telemeter with Admin.Wit
   def tracer = NullTracer
   def run(): Closable with Awaitable[Unit] = Telemeter.nopRun
 
-  private[this] val disallowedChars = "[^a-zA-Z0-9:]".r
-  private[this] def escapeKey(key: String) = disallowedChars.replaceAllIn(key, "_")
+  private[this] val metricNameDisallowedChars = "[^a-zA-Z0-9:]".r
+  private[this] def escapeKey(key: String) = metricNameDisallowedChars.replaceAllIn(key, "_")
+
+  private[this] val labelKeyDisallowedChars = """[^a-zA-Z0-9_]""".r
+  private[this] def escapeLabelKey(key: String) = labelKeyDisallowedChars.replaceAllIn(key, "_")
+
+  private[this] val labelValDisallowedChars = "(\\\\+|\"|\n)".r
+  private[this] def escapeLabelVal(key: String) = labelValDisallowedChars.replaceAllIn(key, "\\\\\\\\")
 
   private[this] def formatLabels(labels: Seq[(String, String)]): String =
     if (labels.nonEmpty) {
       labels.map {
         case (k, v) =>
-          s"""$k="$v""""
+          s"""${escapeLabelKey(k)}="${escapeLabelVal(v)}""""
       }.mkString("{", ", ", "}")
     } else {
       ""
--- a/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
+++ b/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
@@ -37,11 +37,12 @@ class PrometheusTelemeter(metrics: MetricsTree) extends Telemeter with Admin.Wit
   private[this] val metricNameDisallowedChars = "[^a-zA-Z0-9:]".r
   private[this] def escapeKey(key: String) = metricNameDisallowedChars.replaceAllIn(key, "_")
 
-  private[this] val labelKeyDisallowedChars = """[^a-zA-Z0-9_]""".r
+  private[this] val labelKeyDisallowedChars = "[^a-zA-Z0-9_]".r
   private[this] def escapeLabelKey(key: String) = labelKeyDisallowedChars.replaceAllIn(key, "_")
 
-  private[this] val labelValDisallowedChars = "(\\\\+|\"|\n)".r
-  private[this] def escapeLabelVal(key: String) = labelValDisallowedChars.replaceAllIn(key, "\\\\\\\\")
+  // https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details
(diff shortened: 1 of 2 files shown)


telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -34,14 +34,20 @@ class PrometheusTelemeter(metrics: MetricsTree) extends Telemeter with Admin.Wit
...
 
-  private[this] val disallowedChars = "[^a-zA-Z0-9:]".r
-  private[this] def escapeKey(key: String) = disallowedChars.replaceAllIn(key, "_")
+  private[this] val metricNameDisallowedChars = "[^a-zA-Z0-9:]".r
+  private[this] def escapeKey(key: String) = metricNameDisallowedChars.replaceAllIn(key, "_")
+
+  private[this] val labelKeyDisallowedChars = """[^a-zA-Z0-9_]""".r
+  private[this] def escapeLabelKey(key: String) = labelKeyDisallowedChars.replaceAllIn(key, "_")
+
+  private[this] val labelValDisallowedChars = "(\\\\+|\"|\n)".r
```
adleong: Would triple quotes make this easier to read?  I'm not sure what's going on here.

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -34,14 +34,20 @@ class PrometheusTelemeter(metrics: MetricsTree) extends Telemeter with Admin.Wit
...
+  private[this] def escapeLabelVal(key: String) = labelValDisallowedChars.replaceAllIn(key, "\\\\\\\\")
```
adleong: I can't even.

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
rmars: oh yeah, they would! let me fix

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -34,14 +34,21 @@ class PrometheusTelemeter(metrics: MetricsTree) extends Telemeter with Admin.Wit
...
-  private[this] def escapeKey(key: String) = disallowedChars.replaceAllIn(key, "_")
+  private[this] val metricNameDisallowedChars = "[^a-zA-Z0-9:]".r
+  private[this] def escapeKey(key: String) = metricNameDisallowedChars.replaceAllIn(key, "_")
+
+  private[this] val labelKeyDisallowedChars = "[^a-zA-Z0-9_]".r
+  private[this] def escapeLabelKey(key: String) = labelKeyDisallowedChars.replaceAllIn(key, "_")
+
+  // https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-details
+  private[this] val labelValDisallowedChars = """(\\|\"|\n)""".r
+  private[this] def escapeLabelVal(key: String) = labelValDisallowedChars.replaceAllIn(key, """\\\\""")
```
adleong: Why 4 backslashes instead of 2?  Is there another layer of escaping that I'm missing?

rmars: yeah - we want to replace any `\` with `\\`, and each `\` here has to be escaped, even in the triple quotes. the backslash character in the replace string is still interpreted as an escape character...! 
https://www.scala-lang.org/api/current/scala/util/matching/Regex.html

```
scala> ("abc".r).replaceAllIn("abcdef", """\\""")
res106: String = \def

scala> ("abc".r).replaceAllIn("abcdef", """\\\\""")
res107: String = \\def

scala> ("abc".r).replaceAllIn("abcdef", """\""")
java.lang.IllegalArgumentException: character to be escaped is missing
  at java.util.regex.Matcher.appendReplacement(Matcher.java:809)
  at java.util.regex.Matcher.replaceAll(Matcher.java:955)
  at scala.util.matching.Regex.replaceAllIn(Regex.scala:450)
  ... 33 elided
```

adleong: omg it's backslashes all the way down

PR #1292, opened 2017-05-10
Stop tracking all closed streams in h2 dispatchers (#1215)

### LINKED ISSUE #1215: HTTP/2 dispatchers should not track all Closed streams
As discussed in #1154, com.twitter.finagle.buoyant.h2.Netty4DispatcherBase has ConcurrentHashMap of all stream states.  When a stream is closed, the _Closed_ singleton is inserted in the value of the stream.  This means that, through the lifetime of a connection, we end up with an entry in the map for each stream (request, response).

This is done so that, when we receive a frame after the stream has closed (i.e. due to a reset), we can handle it appropriately.

Instead, we should perhaps track the highest stream id below which all stream ids are closed so that the stream state cache only needs active or recently-closed streams.  Or maybe we don't need to keep Closed states at all really?

klingerf:
## Problem

Netty4DispatcherBase maintains a ConcurrentHashMap with all known stream states, including closed streams. This map grows unbounded as the number of closed streams increase over the lifetime of a connection, causing a memory leak.

## Solution

Remove closed streams from the map, and store the highest closed stream id. Streams with an id <= to the max and not found in the map are assumed to be closed.

## Validation

Tested locally and verified that the memory leak does not happen.

Fixes #1215.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
@@ -8,7 +8,7 @@ import com.twitter.util._
 import io.netty.handler.codec.http2.{Http2Frame, Http2GoAwayFrame, Http2StreamFrame}
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
-import java.util.function.IntUnaryOperator
+import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 
 trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
@@ -38,7 +38,11 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
   protected[this] def isClosed = closed.get
 
   private[this] val closedId: AtomicInteger = new AtomicInteger(0)
-  protected[this] def maxClosedId = closedId.get
+  @tailrec private[this] def addClosedId(id: Int): Unit = {
+    val i = closedId.get
+    val max = if (id > i) id else i
+    if (!closedId.compareAndSet(i, max)) addClosedId(id)
+  }
 
   protected[this] def demuxing: Future[Unit]
 
@@ -53,7 +57,8 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
     log.debug("[%s S:%d] initialized stream", prefix, id)
     val _ = stream.onReset.respond {
       case Return(_) =>
-        closedId.updateAndGet(new IntUnaryOperator { def applyAsInt(i: Int) = if (id > i) id else i })
+        // Free and clear.
+        addClosedId(id)
         streams.remove(id)
         log.debug("[%s S:%d] stream closed", prefix, id)
 
@@ -116,7 +121,7 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
 
           case id =>
             streams.get(id) match {
-              case null if id <= maxClosedId =>
+              case null if id <= closedId.get =>
                 // The stream has been closed and should know better than
                 // to send us messages.
                 writer.reset(id, Reset.Closed)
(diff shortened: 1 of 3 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -50,10 +53,9 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
     log.debug("[%s S:%d] initialized stream", prefix, id)
     val _ = stream.onReset.respond {
       case Return(_) =>
-        // Free and clear.
-        if (streams.replace(id, open, StreamClosed)) {
-          log.debug("[%s S:%d] stream closed", prefix, id)
-        }
+        closedId.updateAndGet(new IntUnaryOperator { def applyAsInt(i: Int) = if (id > i) id else i })
```
klingerf: There's gotta be a better way to go about this, but I couldn't figure it out.

zackangelo: Unless you want to pull in scala-java8-compat, there probably isn't. Java 8 lets you specify these operators as lambdas. 

https://github.com/scala/scala-java8-compat/blob/master/src/test/scala/scala/compat/java8/FunctionConvertersTest.scala#L388

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -132,8 +139,8 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
                 if (closed.get) Future.Unit
                 else transport.read().transform(loop)
 
-              case StreamClosed | StreamRemoteReset =>
-                // The stream has been closed and should know better than
+              case StreamRemoteReset =>
+                // The stream has been reset and should know better than
```
olix0r: Doesn't this make the match non-exhaustive?  Should we remove the StreamRemoteClosed state?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
adleong: might be able to avoid allocating a new IntUnaryOperator here by doing something like defining 

```
def addClosedId(id: Int): Unit = {
  val i = closedId.get
  val max = if (id > i) id else i
  if (!closedId.compareAndSet(i, max)) addClosedId(id()
}
```

which, I think, is pretty similar to what updateAndGet does anyway, based on the scaladoc.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
klingerf: Yep, am removing `StreamClosed` on line 31 in this file.

olix0r: whoops, missed that. thanks

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -37,6 +37,13 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
   private[this] val closed: AtomicBoolean = new AtomicBoolean(false)
   protected[this] def isClosed = closed.get
 
+  private[this] val closedId: AtomicInteger = new AtomicInteger(0)
+  @tailrec private[this] def addClosedId(id: Int): Unit = {
+    val i = closedId.get
+    val max = if (id > i) id else i
+    if (!closedId.compareAndSet(i, max)) addClosedId(id)
+  }
```
olix0r: On second, thought, is this safe?

```
t0: closedId.get = 1
t1: closedId.get = 1
t1: max = 4
t0: max = 2
t0: cas(2, 1): true
t1: cas(4, 1): false
t2: closedId.get = 2
...
```

Maybe we don't actually care in this case, but it seems possible to miscalculate the max id

adleong: wouldn't this cause `t1` to try again with `id = 4` (and keep trying until successful)?

olix0r: yes, that's right. carry on.

mattolson:
Awesome

PR #1294, opened 2017-05-11
Accept private keys in PKCS#1 format

adleong:
As of finagle 6.43, private keys are expected to explicitly be PKCS#8
PEM-encoded keys.  This means that PKCS#1 keys that were previously accepted
no longer work.

We use the deprecated LegacyKeyServerEngineFactory in order to continue
accepting PKCS#1 keys.  This has the added benefit that PKCS#1 keys can now
be used for both the Netty3 and Netty4 HTTP engines.

Manually tested client and server TLS with both netty3 and netty4 all using
PKCS#1 keys.

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/Server.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/Server.scala
@@ -1,13 +1,15 @@
 package io.buoyant.linkerd
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.github.ghik.silencer.silent
 import com.twitter.concurrent.AsyncSemaphore
 import com.twitter.conversions.time._
 import com.twitter.finagle.filter.RequestSemaphoreFilter
-import com.twitter.finagle.ssl.{ApplicationProtocols, CipherSuites, KeyCredentials, TrustCredentials}
-import com.twitter.finagle.ssl.server.SslServerConfiguration
+import com.twitter.finagle.ssl._
+import com.twitter.finagle.ssl.server.{LegacyKeyServerEngineFactory, SslServerConfiguration, SslServerEngineFactory}
 import com.twitter.finagle.transport.Transport
 import com.twitter.finagle.service.TimeoutFilter
+import com.twitter.finagle.ssl.client.SslClientEngineFactory
 import com.twitter.finagle.{ListeningServer, Path, Stack}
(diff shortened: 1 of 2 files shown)


olix0r:
Does this remove support for pkcs#8, or simply augment the existing support?

adleong:
I just verified that this works with pkcs8 keys with both netty3 and netty4.  Looking at the code, it looks like it shells out to openssl to convert the key to pkcs12.  Does this have implications for which base docker image we can ship this with?

linkerd/core/src/main/scala/io/buoyant/linkerd/Server.scala
```diff
@@ -119,12 +121,16 @@ class ServerConfig { config =>
       case Some(ps) => ApplicationProtocols.Supported(ps)
       case None => ApplicationProtocols.Unspecified
     }
-    Transport.ServerSsl(Some(SslServerConfiguration(
+    // The deprecated LegacyKeyServerEngineFactory allows us to accept PKCS#1 formatted keys.
+    // We should remove this and replace it with Netty4ServerEngineFactory once we no longer allow
+    // PKCS#1 keys.
+    @silent val sslServerEngine = SslServerEngineFactory.Param(LegacyKeyServerEngineFactory)
```
olix0r: what does `@silent` do?

olix0r:
@adleong the current netty3 behavior requires openssl to shell-out. We were hoping to move to an openssl-less image after the netty4 upgrade...  I think the ideal behavior would be to only invoke openssl when the key is in the legacy format.

linkerd/core/src/main/scala/io/buoyant/linkerd/Server.scala
olix0r: Is there any way for us to print some deprecation warning when legacy keys are used?

adleong: Not easily, I don't think.  We'd need to open up the key ourselves and figure out the format.

olix0r: Seems like something we want to do for migration, but not a blocker.

adleong: I think we should probably just make a clean break in a future major release where we drop support for netty3 and require all keys to be pkcs#8

PR #1297, opened 2017-05-11
Fix shutdown documentation

drewrobb:
This is a pretty trivial fix:

```
$ curl localhost:9990/admin/shutdown
ignoring [/admin/shutdown] from 127.0.0.1, because it is a GET, not a POST
$ curl -X POST localhost:9990/admin/shutdown
[/admin/shutdown] from 127.0.0.1 quitting
```

happy to follow the rest of the guidelines, but wasn't sure it was necessary for something this trivial


adleong:
Good catch, thanks!  Would you mind signing our CLA so that we can accept this fix? https://buoyant.io/cla/

drewrobb:
done

PR #1298, opened 2017-05-11
Update h2 headers API with method for fetching single header value

klingerf:
## Problem

In a local CPU profile, `Netty4Message.Headers.get` showed up as a minor hotspot (0.1%). The method returns all values for a header, but in many of the places where we call it, we immediately follow the call with a `headOption` or `lastOption` call.

## Solution

Rename the existing method to `getAll`, and add a new `get` method that returns a single option value.

## Validation

With this change, neither `get` nor `getAll` show up as hotspots, but this could also just be a result of splitting 1 method into 2. That said, I think the new API is cleaner and would like to keep it.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
@@ -61,8 +61,8 @@ object Headers {
     def toSeq = synchronized(current)
     override def toString = s"""Headers(${toSeq.mkString(", ")})"""
     def contains(key: String) = toSeq.exists { case (k, _) => key == k }
-    def get(key: String) = toSeq.collectFirst { case (k, v) if key == k => v }
-    def getAll(key: String) = toSeq.collect { case (k, v) if key == k => v }
+    def get(key: String) = toSeq.find(_._1 == key).map(_._2)
+    def getAll(key: String) = toSeq.filter(_._1 == key).map(_._2)
     def add(key: String, value: String) = synchronized {
       current = current :+ (key -> value)
     }
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
@@ -62,7 +62,7 @@ object Headers {
     override def toString = s"""Headers(${toSeq.mkString(", ")})"""
     def contains(key: String) = toSeq.exists { case (k, _) => key == k }
     def get(key: String) = toSeq.find(_._1 == key).map(_._2)
-    def getAll(key: String) = toSeq.filter(_._1 == key).map(_._2)
+    def getAll(key: String) = toSeq.collect { case (k, v) if key == k => v }
     def add(key: String, value: String) = synchronized {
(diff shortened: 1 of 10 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
```diff
@@ -60,7 +61,8 @@ object Headers {
     def toSeq = synchronized(current)
     override def toString = s"""Headers(${toSeq.mkString(", ")})"""
     def contains(key: String) = toSeq.exists { case (k, _) => key == k }
-    def get(key: String) = toSeq.collect { case (k, v) if key == k => v }
+    def get(key: String) = toSeq.collectFirst { case (k, v) if key == k => v }
```
adleong: If this is a super hot method that we're trying to optimize, you could try getting rid of the partial function allocation and just iterate over the seq and build the result.

klingerf: Cool, yeah, I switched this to a `.find(...).map(...)`.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
```diff
@@ -60,7 +61,8 @@ object Headers {
     def toSeq = synchronized(current)
     override def toString = s"""Headers(${toSeq.mkString(", ")})"""
     def contains(key: String) = toSeq.exists { case (k, _) => key == k }
-    def get(key: String) = toSeq.collect { case (k, v) if key == k => v }
+    def get(key: String) = toSeq.find(_._1 == key).map(_._2)
+    def getAll(key: String) = toSeq.filter(_._1 == key).map(_._2)
```
olix0r: any reason you prefer filter/map over collect? i find collect to be clearer (and i think it's slightly less costly).

klingerf: I removed it per [alex's comment](https://github.com/linkerd/linkerd/pull/1298#discussion_r116559219), but easy enough to revert this. I could also put together a benchmark to see which is faster.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
adleong: Sorry, I should have been more clear.  I think we can probably do this without allocating any functions by doing it purely imperatively (eg no higher order functions).  But I would only bother doing that for the hottest of codepaths.

klingerf: Ah, ok -- thanks for the clarification. Will revert to the collect methods, but we can revisit later if this remains a hotspot.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
klingerf: Ok, with some console testing, looks like `collect` and `.filter.map` are roughly equal, but `.find.map` is faster than `collectFirst`. Will revert getAll to use collect. Sorry for all of the back-and-forth on this.

PR #1304, opened 2017-05-12
Don't use legacy key ssl server engine for h2

adleong:
TLS cannot be used with the H2 protocol because the LegacyKeyServerEngineFactory is not compatible with ApplicationProtocols.

Use the Netty4ServerEngineFactory instead of the LegacyKeyServerEngineFactory for H2.  This is fine to do because H2 already requires pkcs#8 formatted private keys.

Verified manually that TLS can be used with H2.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## In the next release...
+
+* Fix issue where TLS could not be used with H2.
+
 ## 1.0.1 2017-05-12
 
 * Upgrade to scala 2.12.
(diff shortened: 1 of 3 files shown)


olix0r:
Will we need to do this for all netty4 servers?

Have we tested the clientside?

adleong:
The issue is specifically with ApplicationProtocols which non-H2 servers don't set.  I don't think there's anything to test clientside because we don't support client private keys (yet).

Issue #1287, opened 2017-05-09
dispatcher client stats breaks admin dashboard


klingerf:
I'm looking at the admin dashboard for the [gob app](https://github.com/linkerd/linkerd-examples/tree/master/gob) running in Kubernetes. When I try to view the "http" router, I see:

![screen shot 2017-05-09 at 4 03 42 pm](https://cloud.githubusercontent.com/assets/9226/25876316/24e309be-34d1-11e7-9163-76be08ef3e79.png)

Here's the metrics payload that produced the error:

[metrics.txt](https://github.com/linkerd/linkerd/files/988248/metrics.txt)

I think the issue is:

```
{
  "rt": {
    "http": {
      "client": {
        "$/inet/127.1/8282": {
          ...
        },
        "dispatcher": {
          "serial": {
            "queue_size": {
              "gauge": 0
            }
          }
        }
      }
    }
  }
}
```

The "dispatcher" key in the "client" map doesn't match the [TRANSFORMER_RE](https://github.com/linkerd/linkerd/blob/master/admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js#L15) that we use to parse client ids in the js code. Maybe we should just skip all client ids that don't match the regex? Or maybe those dispatcher stats aren't scoped properly?

esbie:
I agree that the dispatcher map entry is the problem. Very curious, since our current fixture data has dispatcher as a child of each client (so like under `$/inet/127.1/8282` in this example). I wonder if something has changed recently leaving the dispatcher unscoped.

I'll fix the UI to not error on unexpected client ids and also investigate what's up with the stats scoping.

One thing I noticed is the `metrics.txt` linked has "outgoing" and "incoming" router stats, but there aren't any routers labeled that in the gob example app. @klingerf is that expected or...?

klingerf:
@esbie Good catch -- this is from our k8s test env. Will ping you the link.

esbie:
Update: verified this dispatcher metric appears when the first client is created. Occurs in both netty3 and netty4.

adleong:
I believe this is caused by https://github.com/twitter/finagle/pull/629

adleong:
Blocked on https://github.com/linkerd/linkerd/issues/1595

adleong:
Fixed!

PR #1306, opened 2017-05-13
Do not try to render bad client metrics

esbie:
Related to #1287

Gracefully handle the js edge case where client metric names are not
formatted as we expect.

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
@@ -40,6 +40,8 @@ define([
         return initializeClient(client);
       });
 
+      routerClients = _.compact(routerClients);
+
       if (routerClients.length == 0) {
         $clientEl.hide();
       }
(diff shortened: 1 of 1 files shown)


adleong:
I notice that "dispatcher" is in `metrics.js`.  Is it possible to add a test that catches this error?

esbie:
Yes I noticed that too! I'm a little embarrassed to admit we don't have tests for this component yet. It is certainly possible to do, would you be alright adding tests in a followup PR?

adleong:
Mostly I just was looking for a way of testing that the issue is resolved.  If you've tested manually and are confident then 🚀.

admin/src/main/resources/io/buoyant/admin/js/src/router_clients.js
```diff
@@ -40,6 +40,8 @@ define([
         return initializeClient(client);
       });
 
+      routerClients = _.compact(routerClients);
```
rmars: tioli: `_.chain()` this to the map in the previous line so these don't get separated

PR #1309, opened 2017-05-13
Dtab playground UI tweaks

### LINKED ISSUE #523: dtab playground UI issues

rmars:
Addresses most of the remaining items in #523:
* re-run the delegation after saving a new dtab in the textarea if you've previously run one
* add an explanation that the dtab rules from namerd are not editable
* make the text area the same size as the path textfield
* place the cancel/save/reset buttons at the bottom of the textarea
* add a reset button to wipe edits and restore the original base dtab

Fixes #523

screen shot 2017-05-12 at 6 00 27 pm


admin/src/main/resources/io/buoyant/admin/js/src/dtab_viewer.js
```diff
@@ -25,8 +29,22 @@ define(['jQuery'], function($) {
...
       this.render();
       this._toggleEdit();
+      this._rerunDelegation();
+    }.bind(this));
+
+    $('#reset-dtab-btn').click(function(_e) {
+      $('.confirm-modal').modal();
+    });
+
+    $('.confirm-modal .confirm').click(function(_e) {
```
esbie: if you're not using the passed in event object, just leave it out? `function() {`

rmars:
reset is now a link:
screen shot 2017-05-15 at 5 07 28 pm

PR #1315, opened 2017-05-15
Mutual Auth

adleong:
While TLS server auth allowed clients to verify the identity of the server
that they are talking to, there is no way for servers to verify the identity
of clients that are talking to them.

Add `requireClientAuth` property to the server TLS config to require that only
clients that provide valid certificates may issue requests to that server.
Add a `clientAuth` section to the client TLS config where client credentials
can be provided.

Manually tested and integration test added.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## In the next release
+
+* Add support for mutual auth.
+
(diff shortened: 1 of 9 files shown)


adleong:
Posted prematurely.  There are still some things I need to fix.

PR #1316, opened 2017-05-15
Add tests for RouterClients

### LINKED ISSUE #1307: Tests for router_clients.js
We test the metricscollector and the barchart, but there aren't any tests that exercise the handlebars template rendering. A test of router_clients.js or similar would help improve coverage.

rmars:
Fixes #1307

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/spec/RouterClientsSpec.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/spec/RouterClientsSpec.js
@@ -4,15 +4,21 @@ define([
   'jQuery',
   'lodash',
   'src/router_clients',
-  'src/metrics_collector',
-  'spec/fixtures/metrics',
   'template/compiled_templates'
-], function($, _, RouterClients, MetricsCollector, metricsJson, templates) {
+], function($, _, RouterClients, templates) {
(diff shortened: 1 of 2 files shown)


admin/src/main/resources/io/buoyant/admin/js/spec/RouterClientsSpec.js
```diff
@@ -0,0 +1,88 @@
...
+          "/%/io.l5d.port/4141/#/io.l5d.fs/foo",
+          "does-not-compute/",
+          "does_not_obey_rules"
+        ]
+      },
+      nothing_here: { clients: [] }
+    }
+
+    beforeEach(function () {
+      collector = MetricsCollector(metricsJson);
```
esbie: Given that the collector api is pretty small, I'm wondering if we should just use a stub here instead of loading up the metricsJson fixtures. Something like
```
{
        start: function() {},
        registerListener: function() {},
        deregisterListener: function() {},
        onAddedClients: function() {},
}
```
 Wdyt?

rmars: oh yeah, good idea!

PR #1319, opened 2017-05-16
Mutual Auth

adleong:
While TLS server auth allowed clients to verify the identity of the server
that they are talking to, there is no way for servers to verify the identity
of clients that are talking to them.

Add requireClientAuth property to the server TLS config to require that only
clients that provide valid certificates may issue requests to that server.
Add a clientAuth section to the client TLS config where client credentials
can be provided.  Also entirely remove TlsClientPrep and Netty4ClientTls
since it is now possible to specify everything that we need (including
application protocols) directly with the Finagle supplied SSL configuration
params.

Added an integration test for client auth with HTTP.  Also manually tested the following configurations:
* HTTP
* HTTP with TLS
* HTTP with TLS and Netty4
* H2
* H2 with TLS (requires PKCS#8 for server key)
* namerd thrift interface
* namerd thrift interface with TLS
* namerd mesh interface
* HTTP mutual auth (requires PKCS#8 for client key)
* HTTP mutual auth with Netty4 (requires PKCS#8 for client key)
* H2 mutual auth (requires PKCS#8 for client and server keys)


CHANGES.md
```diff
@@ -1,6 +1,7 @@
-## In the latest release...
+## In the next release...
 
 * Expire idle services and clients.
+* Add support for mutual auth.
```
olix0r: Maybe spell this out a little more...  At the least, mention TLS.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2ApplicationProtocol.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import com.twitter.finagle.Stack.Module
+import com.twitter.finagle.ssl.ApplicationProtocols
+import com.twitter.finagle.transport.Transport
+import com.twitter.finagle.{ServiceFactory, Stack}
+import io.netty.handler.ssl.ApplicationProtocolNames
+
+object H2ApplicationProtocol {
+
+  val role = Stack.Role("ApplicationProtocol")
```
olix0r: Ideally, this should be generalized in linkerd-core or somewhere this module can be applied to other protocols (http for instance)

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4H2Transporter.scala
```diff
@@ -35,33 +33,30 @@ object Netty4H2Transporter {
...
-
-        case TransportSecurity.Secure() =>
-          // Netty4Transporter has already installed `ssl` and
-          // `sslConnect` channel handlers. The Netty4ClientTls handler
-          // replaces these handlers with the `tls` and `tlsConnect`
-          // handlers, which are configured to advertise h2 support.
-          p => {
-            p.addLast(Netty4ClientTls.handler(params))
+      if (params[Transport.ClientSsl].e.isDefined) {
+        // secure
```
olix0r: What does this actually mean?  Perhaps a clearer comment like `// TLS is disabled.`

But I'm slightly mystified as to how `ClientSsl.e.isDefined` means that tls should be disabled... Are you sure that this shouldn't be `isEmpty`?

adleong: TLS is enabled for this branch

olix0r: oh duh, i got confused with the matching blocks below... got it now

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2ApplicationProtocol.scala
adleong: This is in finagle-h2 which doesn't depend on linkerd-core.  I'm open to refactoring this up in the future if it's used elsewhere.

PR #1320, opened 2017-05-17
Remove namespace from io.l5d.k8s.ns dstId (#1308)

### LINKED ISSUE #1308: Namespaced k8s namer's dst id erroneously includes namespace
The io.l5d.k8s.ns final dst id is formatted like `/#/io.l5d.k8s.ns/ns/http/{service}`, but I would expect it to be `/#/io.l5d.k8s.ns/http/{service}` (because this is how it appears in the dtab).

This surprising difference makes it harder for users to set up static client tls.

esbie:
Fixes #1308
Refactor Single and MultiNs Endpoint namers so that namespace isn't
appended to the dstId during naming.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -32,15 +32,20 @@ class MultiNsNamer(
       case Path.Utf8(segments@_*) => Path.Utf8(segments.map(_.toLowerCase): _*)
     }
     (lowercasePath, labelName) match {
-      case (id@Path.Utf8(nsName, _, _), None) =>
+      case (id@Path.Utf8(nsName, portName, serviceName), None) =>
         val residual = path.drop(variablePrefixLength)
         log.debug("k8s lookup: %s %s", id.show, path.show)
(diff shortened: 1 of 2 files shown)


k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -104,50 +113,49 @@ abstract class EndpointsNamer(
...
     cache: NsCache,
     serviceCache: ServiceCache,
     id: Path,
     residual: Path
-  ): Activity[NameTree[Name]] = id.take(idLen) match {
-    case id@Path.Utf8(nsName, portName, serviceName) =>
-      cache.services.flatMap { services =>
-        log.debug("k8s ns %s initial state: %s", nsName, services.keys.mkString(", "))
-        services.get(serviceName) match {
+  ): Activity[NameTree[Name]] = cache.services.flatMap { services =>
```
adleong: Did anything change in this method other than taking the ns, port, and service names from method parameters instead of extracting them from the id?  The diff is a bit tough to read because it looks like some cases got re-ordered.

esbie: Argh this diff is nigh impossible. That's right, nothing changed besides the method params stuff.

PR #1330, opened 2017-05-19
Improve scoping of metrics for namers and transformers

### LINKED ISSUE #642: k8s namers should use configured stats receiver
Currently the k8s namer (1) only records client stats (no namer-specific stats and (2) reports these stats to DefaultStatsReceiver.  We should use the configured (scoped) stats receiver built in LinkerConfig.mk.

### LINKED ISSUE #864: Tranformer client stats not scoped properly
We see stats like `clnt/daemonsetTransformer/available`.  These stats should be scoped under the namer they apply to and use the full module name like _io.l5d.k8s.daemonset_ -- maybe something like `[namer, io.l5d.k8s, trans, io.l5d.k8s.daemonset]` ?

### LINKED ISSUE #865: prevent autoloading of namers/tracers that export stats
We see stats like:
```
  "zipkin/zipkin/log_span/try_later" : 0,
  "zk2/inet/dns/cache/evicts" : 0.0,
```

Even when there is no configuration germane to these subsystems.

Finagle has a way to block-autoload of some plugins -- it would be Nice (slightly less confusing for users) if these stats were only exported when the relevant subsystem was actually configured.

adleong:
Transformer metrics are scoped at the root of the metrics tree instead of
inheriting the scope of their containing namer or interpreter.  Namer metrics
are scoped under a `namer/namer` redundant scope.  Unwanted metrics such as
`zk2` metrics are included in the metrics tree by way of DefaultStatsReceiver.

We rescope the metrics such that namer client metrics appear under
`/namer/<namer>/clnt` and transformer client metrics appear under
`/namer/<namer>/transformer/<transformer>/clnt` or
`/rt/<router>/transformer/<transformer>/clnt`.  We also set the
LoadedStatsReceiver to NullStatsReceiver so that the metrics tree does not
pick up any unwanted stats.

Fixes #642
Fixes #864
Fixes #865


interpreter/k8s/src/main/scala/io/buoyant/transformer/k8s/DaemonSetTransformerInitializer.scala
```diff
@@ -41,8 +41,8 @@ case class DaemonSetTransformerConfig(
   private[this] val netmask = InetAddress.getByName("255.255.255.0")
 
   @JsonIgnore
-  override def mk(): NameTreeTransformer = {
-    val client = mkClient(Params.empty).configured(Label("daemonsetTransformer"))
+  override def mk(params: Params): NameTreeTransformer = {
+    val client = mkClient(params).configured(param.Label("clnt"))
```
olix0r: not a huge fan of this abbreviation. Maybe just 'client'?

adleong: Yeah... I don't love it either.  "clnt" is the scope that finagle uses for its ClientStatsRecevier so there is some precedent for this abbreviation.  My main motivation was to distinguish stats for clients that we explicitly create (e.g. for talking to the k8s api etc) from stats for clients that are provisioned by linkerd routers.  ie "client" has a specific meaning in the context of linkerd and I don't want to conflate that with the one-off finagle clients created by namers and transformers.

Advice appreciated.

olix0r: My gut feeling is that the conflict on "client" isn't really that onerous -- on one hand we're scoped under rt/_/... on the other, we're scoped under namer/...

I don't feel that strongly, though

PR #1333, opened 2017-05-19
Remove expired clients from the dashboard

### LINKED ISSUE #1314: Remove expired clients from dashboard
When an idle client expires, it's metrics are removed from /admin/metrics.json.  The client should be removed from the linkerd admin dashboard as well.

rmars:
Fixes #1314 

Problem:
Since #1274, idle clients' metrics are reaped, but this expiry is
not reflected in the admin dashboard.

Solution:
* Remove expired clients' sections from the dashboard, and deregister
them from the MetricsCollector.
* In anticipation of having to do lots of lookups of handlers in the
MetricsCollector, I've changed the implementation to be map based
instead of searching through arrays to find handlers. (So more like
AddEventListener)
* Also add the concept of expired clients to the combined client graph
so that expired clients can be removed

Adding and removing clients in this manner creates some UI questions:

* clients can expire and then be revived. the client section stores the client and its originally assigned color. so if certain clients expire and certain others don't, the color combinations can look weird (e.g. only two purple graphs remain). I don't want to keep redrawing the client colors though, as this could get confusing.
* figuring out what to do between collapsing/expanding clients when they are minimized and/or expired:
_Current behavior_ (not caused by this branch): when a client is collapsed, it's removed from the top line graph of all the clients. when all clients are collapsed, instead of showing an empty graph, we show all clients. if any client is expanded, all non-expanded clients are hidden from the top graph. if there are currently less than 6 clients, any new client added will be auto-expanded. after 6, new clients are auto-collapsed.
_Consequence_: if you have less than 6 clients, and collapse them all, any new client up to 6 will be auto expanded, causing the clients you were previously seeing in the combined graph to disappear, since now some clients are expanded and some are not. it could confuse users as to why some clients appear in the top graph and some do not. this only gets worse when you add in that expired clients can be readded and suddenly appear (or not appear if minimized) in this chart

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/combined_client_graph.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/combined_client_graph.js
@@ -5,7 +5,6 @@ define([
 ], function(Utils) {
   var CombinedClientGraph = (function() {
     var ignoredClients = {};
-    var expiredClients = {};
 
     function clientToMetric(client) {
       return { name: client }; //TODO: move to clientName only after v2 migration
@@ -20,29 +19,17 @@ define([
       };
     }
 
-    function getExpiredClients(routerName) {
-      return _.reduce(expiredClients[routerName], function(mem, v, k) {
-        if(v) { mem.push(k); }
-        return mem;
-      }, []);
-    }
-
     function getClientsToDisplay(clients, routerName) {
       var nonIgnoredClients = _(clients).map(function(clientData, client) {
-        return !ignoredClients[routerName][client] ? client : null;
+        // expired clients have empty data
+        return !ignoredClients[routerName][client] && !_.isEmpty(clientData) ? client : null;
       }).compact().value();
 
-      // if all clients are collapsed, let the combined graph show all non-expired clients
-      if (_.isEmpty(nonIgnoredClients)) {
-        return _.difference(_.keys(clients), getExpiredClients(routerName));
-      } else {
-        return nonIgnoredClients;
-      }
+      return _.isEmpty(nonIgnoredClients) ? _.keys(clients) : nonIgnoredClients;
     }
 
     return function(metricsCollector, initialData, routerName, $root, colors) {
       ignoredClients[routerName] = {}; // clients that are minimized in the UI
-      expiredClients[routerName] = {}; // clients that are hidden from the UI
 
       var chart = new Utils.UpdateableChart(
         {
@@ -101,14 +88,6 @@ define([
 
         unIgnoreClient: function(client) {
           ignoredClients[routerName][client] = false;
-        },
-
-        expireClient: function(client) {
-          expiredClients[routerName][client] = true;
-        },
-
-        unexpireClient: function(client) {
-          expiredClients[routerName][client] = false;
         },
 
         updateColors: function(newColors) {
--- a/admin/src/main/resources/io/buoyant/admin/js/src/combined_client_graph.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/combined_client_graph.js
@@ -5,6 +5,7 @@ define([
 ], function(Utils) {
   var CombinedClientGraph = (function() {
     var ignoredClients = {};
+    var expiredClients = {};
 
     function clientToMetric(client) {
       return { name: client }; //TODO: move to clientName only after v2 migration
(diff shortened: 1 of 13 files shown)


admin/src/main/resources/io/buoyant/admin/js/src/combined_client_graph.js
```diff
@@ -89,6 +103,14 @@ define([
           ignoredClients[routerName][client] = false;
         },
 
+        expireClient: function(client) {
+          expiredClients[routerName][client] = true;
```
siggy: possibly naive question: rather than keeping track of expired clients, is it possible to just remove all references to them?

rmars: that is definitely possible - however, if we keep track of them we can put them back in their original color (both the client section and the combined graph) - which is possibly less confusing than picking a new color every time this client is unexpired. I guess we don't really know how likely it is that clients will come back, or go for good

rmars:
ooooooooh, this was caused by "dispatcher" being among the clients returned in metrics.json, and me not accounting for the case that things can be returned in metrics.json that don't have valid client names. added a check and a test for this case.

admin/src/main/resources/io/buoyant/admin/js/src/combined_client_graph.js
siggy: ah, that's an interesting use case. what's the TTL on clients? i suppose the longer it is, the more ok it is to allow the color to change. i guess it's really a question of trading code complexity here for user friendliness in the ui.

adleong: default is clients expire after being idle for 10 minutes.  but it's configurable.

esbie: Sounds like a pretty small edge case to me, I'd go for removing all references.

admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
```diff
@@ -7,19 +7,19 @@ define(['jQuery'], function($) {
...
-    function registerListener(handler) {
-      listeners.push({handler: handler});
+    function registerListener(listenerId, handler) {
+      listeners[listenerId] = handler;
     }
 
-    function deregisterListener(handler) {
-      _.remove(listeners, function(l) { return l.handler === handler; });
+    function deregisterListener(listenerId) {
+      delete(listeners[listenerId]);
```
esbie: I'm not opposed to moving to a map, but I don't understand what additional lookups you're anticipating making? This diff has the same number of lookups.

admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
```diff
@@ -76,8 +76,10 @@ define(['jQuery'], function($) {
 
         prevMetrics = resp;
 
-        _.each(listeners, function(listener) {
-          listener.handler(resp);
+        _.each(listeners, function(handler) {
+          if(handler) {
```
esbie: Why an `if` here? Might be more performant in `registerListener`.

admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
rmars: ah! I guess what I was saying was that since clients can now be expired and unexpired at any time, we'd call register/deregister more often than before, where we never deregistered anything unless the user minimized the client section

admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
rmars: good point, moved!

rmars:
Okay, I removed the handling of expired clients from router_clients, and each client now controls its own isExpired and updates it on calls to metricsHandler. One result of this is that, since metricsHandler isn't called on minimized clients, a collapsed client won't be hidden unless the user expands it, reactivating the metricsHandler, enabling us to tell whether the client is expired.

Good idea, re the states of RouterClients. Since I'll be revamping expand/collapse in that branch, my plan is to leave the above behaviour as is for now.

PR #1334, opened 2017-05-20
Cache dtab observations in consul dtab store

### LINKED ISSUE #1328: Namerd doesn't cache NameInterpreters
According to our observations, on each call to Namerd (both via `io.l5d.mesh` and via http control interface) to resolve a name, the
```
def observe(ns: Ns): Activity[Option[VersionedDtab]]
```
is also called. So that even though "name observation" are cached, the "namespace observations" are not cached and rather "duplicated" so there is distinct observation for each name that is being watched.

This not only leads to excess resource usage but also introduces an extra delay when a name is observed for the first time.

In our environment we use a custom dtabs storage plugin so we were able to workaround the issue with a read-through cache:
```
  private case class ActClose[T](activity: Activity[T], closable: Closable)

  private lazy val observationCache = CacheBuilder
    .from(cacheSpec.getOrElse("maximumSize=10,concurrencyLevel=1"))  // comes from config
    .removalListener(new RemovalListener[Ns, ActClose[Option[VersionedDtab]]] {
      override def onRemoval(notification: RemovalNotification[Ns, ActClose[Option[VersionedDtab]]]): Unit = {
        val _ = notification.getValue.closable.close()
      }
    })
    .build[Ns, ActClose[Option[VersionedDtab]]](
    new CacheLoader[Ns, ActClose[Option[VersionedDtab]]] {
      override def load(key: Ns): ActClose[Option[VersionedDtab]] = {
        val act = observeNamespace(key)  // starts actual observation

        // we want to keep the activity observed as long as it's in the cache
        val closable = act.run.changes.respond(_ => ())
        ActClose(act, closable)
      }
    }
  )

  def observe(ns: Ns): Activity[Option[VersionedDtab]] = observationCache.get(ns).activity
```

We found that `concurrencyLevel=1` is important to ensure that we won't create duplicate namespace observations no matter what with neglectable cost as synchronization happens only on writes and therefore occurs 1 time for the namespace observation lifetime.

adleong:
Each call to ConsulDtabStore.observe creates a new observation, even if there
is already an observation active for that namespace.  This means that namerd
will have N identical watches on a dtab namespace where N is the number of
names observed in that namespace.

Add a cache of activities so that there is only one activity per namespace.
These activities are reference counted and only kept open if observed so there
is little cost to caching them forever.

Manually verified that establishing multiple dtab watches on the same namespace
via the http or grpc interface only results in a single connection to consul.

Fixes #1328


adleong:
cc @Ashald

Ashald:
This looks great to me! I don't know how important is this, but I noticed that unless you set the concurrency level within cache to 1 there is a pretty good chance to end with several duplicated activities.

PR #1338, opened 2017-05-22
Better namerd admin ui with no namespaces (#447)

### LINKED ISSUE #447: Better namerd admin for 0 namespaces

siggy:
Problem
With no namespaces configured/found, namerd displays "Namespaces:"
without any context.

Solution
Display a helpful error message informing the user that no namespaces
are found, and point them to the namerd documentation.

Validation
Visual UI check.

fixes #447 

screen shot 2017-05-22 at 4 53 29 pm

### DIFF
--- a/namerd/main/src/main/scala/io/buoyant/namerd/DtabListHandler.scala
+++ b/namerd/main/src/main/scala/io/buoyant/namerd/DtabListHandler.scala
@@ -18,11 +18,28 @@ class DtabListHandler(
     }
 
   def render(list: Iterable[String]) = {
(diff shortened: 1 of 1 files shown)


adleong:
Awesome!  Maybe reword 1. slightly to say that no namespaces have been created in the backend store, since namespace aren't configured in the namerd config (except for the in-memory store).

PR #1341, opened 2017-05-23
Add TLS for mesh and HTTP interfaces (#1191)

### LINKED ISSUE #1191: Support TLS for io.l5d.mesh interpreter
The io.l5d.mesh interpreter was added in #1065. In order for that interpreter to work with TLS, we need to add server TLS support for the io.l5d.mesh interface in namerd.

siggy:
Problem
io.l5d.namerd, io.l5d.namerd.http, io.l5d.thriftNameInterpreter all
support tls, while io.l5d.mesh and io.l5d.httpController do not.

Solution
Add TLS support to io.l5d.mesh and io.l5d.httpController, enabling
TLS for all namerd interfaces.

Validation
Added tests for MeshIfaceInitializer, MeshInterpreterInitializer,
HttpControlServiceConfig, and ThriftInterpreterInterfaceConfig.
Modified linkerd/namerd-tls and namerd/tls example configs to use
tls for mesh and http.

fixes #1191

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsServerConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsServerConfig.scala
@@ -1,0 +1,45 @@
+package com.twitter.finagle.buoyant
+
+import com.twitter.finagle.Stack
+import com.twitter.finagle.ssl.{ClientAuth => FClientAuth, _}
+import com.twitter.finagle.ssl.server.{SslServerConfiguration, SslServerEngineFactory}
+import com.twitter.finagle.transport.Transport
+import java.io.File
+
+case class TlsServerConfig(
+  certPath: String,
+  keyPath: String,
+  caCertPath: Option[String] = None,
+  ciphers: Option[Seq[String]] = None,
+  requireClientAuth: Option[Boolean] = None
+) {
+  def params(
+    alpnProtocols: Option[Seq[String]],
+    sslServerEngine: SslServerEngineFactory
+  ): Stack.Params = {
+    val trust = caCertPath match {
+      case Some(caCertPath) => TrustCredentials.CertCollection(new File(caCertPath))
+      case None => TrustCredentials.Unspecified
+    }
+    val cipherSuites = ciphers match {
+      case Some(cs) => CipherSuites.Enabled(cs)
+      case None => CipherSuites.Unspecified
+    }
+    val appProtocols = alpnProtocols match {
+      case Some(ps) => ApplicationProtocols.Supported(ps)
+      case None => ApplicationProtocols.Unspecified
+    }
+    val clientAuth = requireClientAuth match {
+      case Some(true) => FClientAuth.Needed
+      case _ => FClientAuth.Off
+    }
+
+    Stack.Params.empty + Transport.ServerSsl(Some(SslServerConfiguration(
+      clientAuth = clientAuth,
+      keyCredentials = KeyCredentials.CertAndKey(new File(certPath), new File(keyPath)),
+      trustCredentials = trust,
+      cipherSuites = cipherSuites,
+      applicationProtocols = appProtocols
+    ))) + SslServerEngineFactory.Param(sslServerEngine)
+  }
+}
--- a/interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
+++ b/interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
@@ -3,7 +3,7 @@ package io.buoyant.interpreter
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.twitter.conversions.time._
 import com.twitter.finagle._
-import com.twitter.finagle.buoyant.{H2, TlsClientConfig}
+import com.twitter.finagle.buoyant.{TlsClientConfig, H2}
 import com.twitter.finagle.naming.NameInterpreter
 import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.ssl.ApplicationProtocols
@@ -64,9 +64,10 @@ case class MeshInterpreterConfig(
 
     val Retry(baseRetry, maxRetry) = retry.getOrElse(defaultRetry)
     val backoffs = Backoff.exponentialJittered(baseRetry.seconds, maxRetry.seconds)
+    val tlsParams = tls.map(_.params).getOrElse(Stack.Params.empty)
 
     val client = H2.client
-      .withParams(H2.client.params ++ params)
+      .withParams(H2.client.params ++ tlsParams ++ params)
       .newService(name, label)
 
     root.getOrElse(DefaultRoot) match {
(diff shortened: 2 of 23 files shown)


finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
```diff
@@ -54,3 +54,14 @@ case class TlsClientConfig(
 }
 
 case class ClientAuth(certPath: String, keyPath: String)
+
+case class ClientTlsConfig(commonName: String, caCert: Option[String]) {
```
adleong: I believe this config format is only used by the namerd thrift interpreter.  Why move it here?

interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -43,7 +43,7 @@ case class Retry(
 case class MeshInterpreterConfig(
   dst: Option[Path],
   root: Option[Path],
-  tls: Option[TlsClientConfig],
+  tls: Option[ClientTlsConfig],
```
adleong: Isn't this a breaking change?  I think we'd like to standardize on TlsClientConfig as the proper way to specify client tls configuration.  It's somewhat unfortunate that the namerd thrift interpreter uses a different format but we're stuck with it until a breaking release I think.

siggy: AFAICT this was just a placeholder, as `io.l5d.mesh` wasn't supported on the server?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
siggy: Similar to the other comment, my reading of the code was that we were only supporting TLS for the thrift interpreter, so in adding http and mesh, it made sense to make all 3 the same. Open to discussion though.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsServerConfig.scala
```diff
@@ -0,0 +1,47 @@
...
+case class TlsServerConfig(
+  certPath: String,
+  keyPath: String,
+  caCertPath: Option[String] = None,
+  ciphers: Option[Seq[String]] = None,
+  requireClientAuth: Option[Boolean] = None
+) {
+  def params(
+    alpnProtocols: Option[Seq[String]],
+    sslServerEngine: SslServerEngineFactory
```
adleong: It looks like sslServerEngine is just passed through.  Could it just be added as a param at the call site instead of being wired through this method?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsServerConfig.scala
```diff
@@ -0,0 +1,47 @@
...
+    }
+    val appProtocols = alpnProtocols match {
+      case Some(ps) => ApplicationProtocols.Supported(ps)
+      case None => ApplicationProtocols.Unspecified
+    }
+    val clientAuth = requireClientAuth match {
+      case Some(true) => FClientAuth.Needed
+      case _ => FClientAuth.Off
+    }
+    // The deprecated LegacyKeyServerEngineFactory allows us to accept PKCS#1 formatted keys.
```
adleong: This comment is specific to LegacyKeyServerEngineFactory and should be moved to the place where the LegacyKeyServerEngineFactory is specified.

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdHttpInterpreterInitializer.scala
```diff
@@ -27,7 +28,7 @@ case class NamerdHttpInterpreterConfig(
   dst: Option[Path],
   namespace: Option[String],
   retry: Option[Retry],
-  tls: Option[ClientTlsConfig]
+  tls: Option[TlsClientConfig]
```
adleong: Just to confirm my understanding: this is a breaking change of something that was already broken, right?  Namerd's http interface never supported TLS on the server side?

namerd/iface/mesh/src/main/scala/io/buoyant/namerd/iface/MeshIfaceInitializer.scala
```diff
@@ -30,7 +30,14 @@ class MeshIfaceConfig extends InterfaceConfig {
         val resolver = mesh.ResolverService(namers, stats)
         ServerDispatcher(codec, interpreter, delegator, resolver)
       }
-      H2.serve(addr, dispatcher)
+
+      // val params = (tls match {
```
adleong: rm

namerd/core/src/main/scala/io/buoyant/namerd/InterfaceConfig.scala
```diff
@@ -21,6 +26,10 @@ abstract class InterfaceConfig extends PolymorphicConfig {
   )
 
   @JsonIgnore
+  @silent
+  def tlsParams = tls.map(_.params(None, LegacyKeyServerEngineFactory)).getOrElse(Stack.Params.empty)
```
adleong: Does this work for the mesh interface?  I would have thought that `Some(Seq(ApplicationProtocolNames.HTTP_2))` would need to be specified for the `alpnProtocols` instead of `None`.

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdHttpInterpreterInitializer.scala
siggy: That's correct. Note the changes in `HttpControlServiceConfig.scala`

namerd/core/src/main/scala/io/buoyant/namerd/InterfaceConfig.scala
siggy: It does work, tested with:

```bash
./sbt "namerd-examples/tls:run -log.level=DEBUG -com.twitter.finagle.tracing.debugTrace=true"
```

```bash
./sbt "linkerd-examples/namerd-tls:run -log.level=DEBUG -com.twitter.finagle.tracing.debugTrace=true"
```

However, I've updated mesh interface to use `ApplicationProtocolNames.HTTP_2` and `Netty4ServerEngineFactory()`, and added a test to verfiy `Netty4ServerEngineFactory`.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsServerConfig.scala
siggy: You're correct we could avoid passing it through. The call sites are different enough where I thought this approach minimized code complexity, since we already are passing `aplnProtocols` through anyway.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsServerConfig.scala
```diff
@@ -0,0 +1,45 @@
...
+
+case class TlsServerConfig(
+  certPath: String,
+  keyPath: String,
+  caCertPath: Option[String] = None,
+  ciphers: Option[Seq[String]] = None,
+  requireClientAuth: Option[Boolean] = None
+) {
+  def params(
+    aplnProtocols: Option[Seq[String]],
```
adleong: whoops, I think this should have been ALPN all along: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation

PR #1351, opened 2017-05-29
Use backticks when matching proto enum values

nvartolomei:
Using lowercase/snakecase enum values generates variable pattern
match.

Scala 2.11 compiles the code with a warning, Scala 2.12 will throw an
error. Although proto3 style guide says to use CAPITALS_WITH_UNDERSCORES
this is not an enforced rule across proto ecosystem.

Using backticks in generated code matches variable name.

### DIFF
--- a/grpc/eg/src/main/protobuf/eg.proto
+++ b/grpc/eg/src/main/protobuf/eg.proto
@@ -15,7 +15,7 @@ enum Enumeration {
 
 message Message {
   enum Enumeration {
-    ZERO = 0;
+    zero = 0;
     ONE = 1;
     TWO = 2;
     THREEFOUR = 34;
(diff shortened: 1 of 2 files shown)


wmorgan:
Thanks, @nvartolomei! We'll take a look!

grpc/eg/src/main/protobuf/eg.proto
```diff
@@ -15,7 +15,7 @@ enum Enumeration {
 
 message Message {
   enum Enumeration {
-    ZERO = 0;
+    zero = 0;
```
olix0r: This seems like it might be a mistake?

nvartolomei: What do you mean? This is here to break previous tests, which are fixed by adding backticks.

CLA signed.

PR #1352, opened 2017-05-31
Add Path to Host identifier in PLUGINS.md

dario-simonetti:
The Path to Host identifier sets the Host header to the first segment in
the path. This is useful when hitting linkerd directly (as opposed to
using it as an HTTP proxy). If you want to use linkerd in this mode
while hitting external services (e.g. 3rd part APIs), setting the Host
this way is mandatory or the request will fail as the Host header will
be set to your linkerd host+port and the server will likely respond with
404 Not Found.

Config example:
```
...

routers:
- ...
  identifier:
  - kind: com.askattest.pathToHost
  - ...

...
```


wmorgan:
Very cool!

PR #1353, opened 2017-05-31
Simplify states of router clients expand/collapse

rmars:
Problem:
The router clients can be in a number of states with regards
to whether they are collapsed or expanded automatically, and whether
added clients should be expanded or collapsed

Solution:
Implement the following states (based on discussion in #1333):

      RouterClients can be in 1 of 3 states: COLLAPSED, EXPANDED, or CUSTOMIZED.

      COLLAPSED:
        Occurs if there are more than THRESHOLD clients on pageload.
        Any added clients are collapsed. Dropping below 6 clients transitions to EXPANDED.
        User opening any clients transitions to CUSTOMIZED.

      EXPANDED:
        Occurs if there are less than THRESHOLD clients on pageload.
        Any added clients are expanded.
        Adding more than 6 clients transitions to COLLAPSED.
        User closing any clients transitions to CUSTOMIZED.

      CUSTOMIZED:
        Any added clients are expanded.
        There is no transition away from this state.

Also add links to expand all and collapse all clients, to get us out
of the custom state, since now, it could be annoying that half your
clients are expanded.

This is a Follow up to #1333 
* Fixed the issue where collapsed clients wouldn't expire until they'd been re-expanded. The fix: don't deregister the client from the metricscollector on collapse. Since we check to see if the client has gotten any data before executing the rest of the handler, I'm hoping this won't be too bad peformance-wise.
* I added expand all/collapse all links because I feel like this functionality would be useful. It also unsets the custom behaviour. 
* The expand/collapse links aren't very pretty. Suggestions welcome, I could also remove this functionality if it causes too much confusion

screen shot 2017-05-31 at 12 48 15 pm


rmars:
(actually, please hold off on looking at this for now, I've discovered a bug in some states :/)

rmars:
Okay, this should now be behaving as expected.
Notes: if the expand all/collapse all links are clicked, when clients are unexpired, they'll obey this rather than 'custom'. However /newly added/ clients will still be collapsed if there are more than 6.

rmars:
I like middot!
screen shot 2017-06-02 at 2 18 42 pm

Here's slash for comparison:
screen shot 2017-06-02 at 2 19 16 pm

admin/src/main/resources/io/buoyant/admin/js/template/router_container.handlebars
```diff
@@ -14,7 +14,11 @@
     </div>
 
     <div class="clients router-clients">
-      <div class="router-subsection-title">Clients</div>
+      <div class="pull-left router-subsection-title">Clients</div>
+      <div class="pull-right client-toggle">
+        <a target="blank" class="expand-all">expand all</a> &middot;
+        <a target="blank" class="collapse-all">collapse all</a>
+      </div>
```
klingerf: TIOLI: it might be worth hiding these two links if there are no clients, and therefore nothing to expand / collapse. Otherwise it looks like this:

screen shot 2017-06-02 at 2 47 14 pm

rmars: Oooh good catch, thanks!

PR #1355, opened 2017-06-02
Add specific host transformer

dario-simonetti:
**Problem**
When deploying linkerd in a linker-to-linker + per-host configuration
(i.e. one linkerd container per machine with an incoming router and
outgoing router), the incoming router needs to be able to route the
incoming request to the destination container running on the same
Docker host (i.e. machine).

The destination app container is likely to be registered against
the namer using the IP address of the Docker host it is running on and
the port that is exposed in its Docker container. The problem is that
the incoming router is unaware of the IP address of the Docker host it's
running on so won't know which of the resolved names live on the same
Docker host.

At the moment there is no transformer that can solve this problem as
`io.l5d.localhost` will only look at network interfaces inside the
Docker container while `io.l5d.k8s.daemonset` and `io.l5d.k8s.localnode`
are k8s-specific.

**Solution**
The specific host transformer filters the list of addresses down to only
addresses that have the same IP address as the specified host. This
transformer can be used by an incoming router to only route traffic to
specific destinations. This way, when linkerd is running inside a Docker
container the traffic can be routed to another Docker container in the
same host.


wmorgan:
@dario-simonetti Thank you! We'll take a look!

PR #1358, opened 2017-06-06
Reconcile namerd config docs and code (#1357)

### LINKED ISSUE #1357: reconcile namerd config docs and code
the namerd docs state that `interfaces` and `namers` sections are not required:
https://linkerd.io/config/1.0.2/namerd/index.html

...but the code requires them:
https://github.com/linkerd/linkerd/blob/master/namerd/core/src/main/scala/io/buoyant/namerd/NamerdConfig.scala#L24

siggy:
Problem
Namerd configs require 'interfaces' and 'namers' sections. The docs say
they are optional.

Solution
Make the `namers` section optional, and fix the docs to indicate
'interfaces' is required.

Validation
Test updates, added a minimal.yaml namerd config.

Fixes #1357

### DIFF
--- a/namerd/core/src/main/scala/io/buoyant/namerd/NamerdConfig.scala
+++ b/namerd/core/src/main/scala/io/buoyant/namerd/NamerdConfig.scala
@@ -17,11 +17,10 @@ import scala.util.control.NoStackTrace
 private[namerd] case class NamerdConfig(
   admin: Option[AdminConfig],
   storage: DtabStoreConfig,
-  namers: Seq[NamerConfig],
+  namers: Option[Seq[NamerConfig]],
   interfaces: Seq[InterfaceConfig],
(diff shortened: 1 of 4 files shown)


namerd/core/src/main/scala/io/buoyant/namerd/NamerdConfig.scala
```diff
@@ -72,7 +71,7 @@ private[namerd] case class NamerdConfig(
   }
 
   private[this] def mkNamers(params: Stack.Params): Map[Path, Namer] =
-    namers.foldLeft(Map.empty[Path, Namer]) {
+    namers.getOrElse(Seq[NamerConfig]()).foldLeft(Map.empty[Path, Namer]) {
```
klingerf: TIOLI `namers.getOrElse(Nil)` is a bit more concise.

Issue #1250, opened 2017-04-26
Display services in linkerd admin dashboard


adleong:
The current admin dashboard shows a list of clients with traffic volume and success rate for each but no mention of services.

It would be cool to have a visualization of all services (using the /rt/*/service metrics) and all clients (using the /rt/*/client metrics) as well as the traffic flow between them (using the /rt/*/client/*/service metrics).

```
/svc/foo ----------100rps--------> /#/io.l5d.k8s/default/http/foo
/svc/bar ----------20rps---------> /#/io.l5d.k8s/default/http/bar
/svc/bas ----------5rps----------> /#/io.l5d.k8s/default/http/bas-v1
         \
           --------95rps---------> /#/io.l5d.k8s/default/http/bas-v2
```

PR #1362, opened 2017-06-07
Begin work on a Service oriented dashboard

rmars:
Add the concept of Services as a dashboard page.
This page is a bare-bones list of the services per router, their clients, and a basic indication of traffic flow between them.

The page won't be linked from anywhere until it's ready, but can be found at e.g. localhost:9990/services.

Not intended to be a well designed page - I just want to get the basic service/client stats wired thru.
Next up: adding service stats to the page.

screen shot 2017-06-07 at 2 01 15 pm

Part of #1250


admin/src/main/resources/io/buoyant/admin/js/src/dashboard.js
```diff
@@ -2,18 +2,20 @@
...
 ], function(
   $, bootstrap,
+  RouterServices,
   MetricsCollector,
   ProcInfo,
   RequestTotals,
   RouterController
 ) {
-  return function(routerConfig) {
+  return function(routerConfig, isServiceMap) {
```
siggy: this feels a bit overloaded. is there some way to factor out the common bits?

admin/src/main/resources/io/buoyant/admin/js/src/metrics_collector.js
```diff
@@ -56,8 +56,8 @@ define(['jQuery'], function($) {
     }
 
     function onAddedClients(handler) {
```
siggy: not sure i fully grock what's going on here. i see two calls to `onAddedClients`:
1. `metricsCollector.onAddedClients(addServices)`
1. `metricsCollector.onAddedClients(addClients);`

... and the functions passed in have different signatures:
1. `function addServices(addedClients, metricsJson)`
1. `function addClients(addedClients)`

is this expected?

rmars: ah! yeah expected. `addServices` and `addClients` are handlers that could be called if new clients are detected (one is for the Service-focussed dashboard and one is for the current (client) dashboard. they both listen of the 'addedClients' event, but since they belong to different dashboards they won't both be triggered).

I modified the handler in metricsCollector to include the most current metricsJson because I needed to check which services added clients belong to in order to fit them under the correct service in the Serivce dashboard. the clients dashboard doesn't care about this, so it ignores that second argument (metricsRsp) passed in on line 60 of this file. 

Maybe `onAddedClients(addServices)` is confusingly named - the addServices method is really "check if any of the added clients are under a service we haven't seen before, and if they are, add that service" - perhaps I should name it something other than addServices?

admin/src/main/resources/io/buoyant/admin/js/src/dashboard.js
rmars: Yeah, I think `isServiceMap` was also a bad arg name :) I originally called the service dashboard "serviceMap" but forgot to rename it in all places.

Would it be less confusing if this argument was called `dashboardType` instead?

Renaming `RouterController` to `ClientDashboard` and `RouterServices` to `ServiceDashboard` could also help. (They both need the same setup things, e.g. initializing ProcInfo and RequestTotals)

Then line 43 would be something like:
```
if(dashboardType === "service") {
  ServiceDashboard(metricsCollector, initialData, $(".service-dashboard-container"));
} else {
  ClientDashboard(metricsCollector, selectedRouter, initialData, $(".client-dashboard-container"), routerConfig);
}
```

This file could be called something better, like `dashboard_initializer` which is maybe more indicative of what it is doing.

admin/src/main/resources/io/buoyant/admin/js/src/router_services.js
```diff
@@ -0,0 +1,85 @@
...
+      }
+
+      _.each(routerData.services, function(svc) {
+        initializeService($routerContainer, metricsCollector, router, svc);
+      });
+    });
+  }
+
+  function initializeService($routerContainer, metricsCollector, router, svc) {
+    $routerContainer.find(".svc-subsection").removeClass("hidden");
```
esbie: I'd rather not do a DOM find for every added service if we don't need to (and I don't think we do for initializeServices). For addServices, you really only need to check size at the end, maybe something like `_.flatten(svcsByRouter).length > 0`

admin/src/main/resources/io/buoyant/admin/js/src/router_service.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
+
+define([
+  'jQuery',
+  'template/compiled_templates',
+  'bootstrap'
+], function($,
+  templates
+) {
```
esbie: this argument style is kinda wonky, either `$` should get its own line or all args on one line.

admin/src/main/resources/io/buoyant/admin/js/spec/RouterServicesSpec.js
```diff
@@ -0,0 +1,93 @@
...
+        adder: { services: ["svc"] },
+        multiplier: { services: ["svc"]}
+      }
+      var realCollector = MetricsCollector(metricsJson);
+
+      RouterServices(realCollector, initialData, $container);
+      realCollector.__update__(metricsJson);
+
+      var $multiplierRouter = $($container.find(".router-multiplier")[0]);
+      var multiplierSvcs = $multiplierRouter.find(".svc-container");
```
esbie: this... is a jquery object too right? might as well put a $ before it like the others. (applies to other tests too)

PR #1373, opened 2017-06-09
Use the default (INFO) tracer log level

thedebugger:
- tracer's TRACE log level doesn't log anything; not sure why

### DIFF
--- a/linkerd/docs/config.md
+++ b/linkerd/docs/config.md
@@ -40,7 +40,6 @@ telemetry:
   sampleRate: 0.01
 - kind: io.l5d.tracelog
   sampleRate: 0.2
-  level: TRACE
 ```
 
 Welcome to the Configuration Reference for linkerd!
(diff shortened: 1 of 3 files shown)


adleong:
The log level in the tracelog config indicates the level at which traces is logged at.  When set to TRACE, this means that traces are logged at the TRACE level.  If this is below your logging threshold, you won't see them.  The log threshold can be configured with the `-log.level=TRACE` flag or at runtime in the logging tab of the linkerd admin dashboard.

thedebugger:
@adleong ahh, that makes sense. I thought we are setting the logger level to the given level. Actually, we are trying to log the traces with the given level. I'll update the docs to make it more clear. Thanks!

adleong:
Thanks for the doc update!  Can you revert your changes to the example configs?

klingerf:
Fixed by #1514.

PR #1374, opened 2017-06-09
Add Istio namer

adleong:
Add a namer which retrieves addresses from Istio-Manager's Service Discovery
Service.  Each lookup is implemented as a polling loop against SDS.

Accepts names in the form:
`/<namespace>/<port-name>/<svc-name>/<labels>/residual/path`
and initiates a polling loop against the SDS API.  The labels segment must be a :: delimited
list of label:value pairs.  To avoid duplication, the labels must be in alphabetical order
by label name.  e.g.

```
/default/http/foo/az:us-west::env:prod::version:v1
```

Use :: alone to specify an empty list of labels

```
/default/http/foo/::
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 * Add TLS support for `io.l5d.httpController` and `io.l5d.mesh`.
 * Add support for HTTP/2 tracing.
 * Cache dtab observations in the io.l5d.consul store.
+* Add Istio namer that uses Istio-Manager's Service Discovery Service.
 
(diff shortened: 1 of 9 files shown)


adleong:
There's nothing explicitly coupled to k8s but the `namespace`, `port`, `service` concepts are k8s concepts.  In general, it seems like istio leaks a lot of k8s details through the abstraction.  I imagine that the istio apis are going to need to go through some refactors to make them work in other environments.

PR #1375, opened 2017-06-09
Generate protobuf maps as Map instead of Seq of tuples

esbie:
**Problem:**
I have json (generated via golang protobuf) that contains a map created from a `map<string,string>` field. I wasn't able to parse the map into a scala protobuf object, because linkerd protobuf maps are generated as repeated entries. Here's how the generator currently works:

for each map, a map entry is created eg:
```
case class FooEntry(
  `key`: Option[String] = None,
  `value`: Option[String] = None
)
```

then the map is defined as:
```
Seq[FooEntry]
```

**Solution:**
I've changed all map fields to be generated as:
`Map[String, String]`

This enables jackson parsing directly into the generated protobuf case classes.

**Validation:**
`map<string, string>` declaration added to EgTest

### DIFF
--- a/grpc/gen/src/main/scala/io/buoyant/grpc/gen/Generator.scala
+++ b/grpc/gen/src/main/scala/io/buoyant/grpc/gen/Generator.scala
@@ -1,10 +1,10 @@
 package io.buoyant.grpc.gen
 
 import com.google.protobuf.DescriptorProtos._
-import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Label._
 import com.google.protobuf.DescriptorProtos.FieldDescriptorProto.Type._
 import com.google.protobuf.compiler.PluginProtos.{CodeGeneratorRequest, CodeGeneratorResponse}
-import com.twitter.util.{Future, Return, Throw, Try}
+import com.twitter.util.{Return, Throw, Try}
+import io.buoyant.grpc.gen.ProtoFile.TypeRef
 import java.io.InputStream
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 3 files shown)


grpc/gen/src/main/scala/io/buoyant/grpc/gen/Generator.scala
```diff
@@ -184,19 +184,37 @@ object Generator {
...
-        val typ = genFieldType(f, translateType)
-        val boxed = if (f.isRepeated) s"Seq[${typ}]" else s"Option[${typ}]"
-        val default = if (f.isRepeated) "Nil" else "None"
-        FieldArg(name, typ, boxed, default, Left(f))
+        f.typeRef match {
+          case ProtoFile.TypeRef.Message(msg) if f.isMap =>
+            try {
+              // find the nested message that represents entries in the map
+              // we expect this message to have "key" and "value" fields
+              mapMessages.find(m => msg.endsWith(m.name)).map { mp =>
```
adleong: I'd replace 
```
try { find(...).map(...).get } catch { ... throw ... }
``` 
with 
```
mapMessages.collectFirst {
  case mp if msg.endsWith(mp.name) => ...
}.getOrElse(throw new IllegalArgumentException(...))
```

grpc/gen/src/main/scala/io/buoyant/grpc/gen/Generator.scala
```diff
@@ -184,19 +184,37 @@ object Generator {
...
+                val key = mp.fields.find(_.name.equals("key")).map(genFieldType(_, translateType)).getOrElse("String")
```
adleong: TIOLI: could use collectFirst instead of find -> map

PR #1378, opened 2017-06-10
Istio routing rule protobuf import

esbie:
Adding local copies of istio route-rule protobuf to generate scala bindings. Requires https://github.com/linkerd/linkerd/pull/1375

### DIFF
--- a/istio/src/main/protobuf/README.md
+++ b/istio/src/main/protobuf/README.md
@@ -1,0 +1,6 @@
+**google/protobuf**: Replicated from 
+[the google protobuf repo](https://github.com/google/protobuf/tree/2f4489a3e504e0a4aaffee69b551c6acc9e08374/src/google/protobuf).
+Ideally we wouldn't need these and could use the protobuf-java library directly. The java packages have been changed to
+ `com.google.local` to avoid protobuf-java conflicts.
+
+**proxy/v1**: Replicated from [the istio api repo](https://github.com/istio/api/tree/master/proxy)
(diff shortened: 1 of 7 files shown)


adleong:
Can we add a README.md or something indicating where these file are copied from?

istio/src/main/protobuf/http_fault.proto
```diff
@@ -0,0 +1,104 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+
+syntax = "proto3";
+
+import "duration.proto";
```
siggy: in the istio/api repo i see this line as:
```
import "google/protobuf/duration.proto";
```

is it possible for us to conform to a directory structure such that the files will not require modification?

related: for the istio telemeter i'm going to need proto files under `api/mixer/v1/config`, can we move `http_fault.proto` and friends down into a directory structure that mirrors the api repo they came from?

esbie: I bet we could do that, makes sense if we need to import a bunch more. However I also edited the google protobuf files to change their package `option java_package = "com.google.local";` otherwise they conflict with java's protoc lib. If you figure out a better solution, I'm all ears.

esbie:
Added README.md, reverted import statements, and moved to nested directory structure. Thank you!

PR #1382, opened 2017-06-12
Add service level stats to the Service page

rmars:
* Add combined client graph on service level
* Add service metrics section
* Add service success rate graph
* Add Service latencies table
* Add colour to client request stats

Still not pretty, gonna get design input and refine afterwards. This update does get all the service level stats displayed on the page, and the color mapping for the clients, should we need that.

screen shot 2017-06-12 at 11 27 43 am

Part of #1250


adleong:
@rmars closing this PR as we're not actively working on this.  Go ahead and keep the branch around though.

PR #1386, opened 2017-06-13
[WIP] Istio Identifier

esbie:
Adds an identifier that reads istio configuration from the istio pilot/manager apiserver. Currently only chooses routes based on hostname and precedence.

TODO:
- implement matching against more than just hostname
- query rds for port number
- identifier unit tests
- additional manual testing

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
@@ -1,0 +1,42 @@
+package io.buoyant.k8s
+
+import com.twitter.util._
+import io.buoyant.k8s.PilotClient.RouteRuleConfig
+import java.util.concurrent.atomic.AtomicReference
+import istio.proxy.v1.config.RouteRule
+import io.buoyant.namer.RichActivity
+
+class RouteManager(api: PilotClient, pollInterval: Duration) {
+
+  private[this] val states = Var.async[Activity.State[Map[String, RouteRule]]](Activity.Pending) { state =>
+    val close = new AtomicReference[Closable](Closable.nop)
+    close.set(Closable.make { _ =>
+      log.trace("istio config watch was closed")
+      Future.Unit
+    })
+    val pending = api.getRouteRules().respond {
+      case Throw(e) => state.update(Activity.Failed(e))
+      case Return(routeList) =>
+        val initState: Map[String, RouteRule] = mkRouteMap(routeList)
+        state.update(Activity.Ok(initState))
+
+        val _ = api.watchRouteRules(pollInterval).run.map {
+          case Activity.Ok(rsp) => state.update(Activity.Ok(mkRouteMap(rsp)))
+          case Activity.Pending => //do nothing
+          case Activity.Failed(e) => log.error(e, "failed to update istio routing rules")
+        }
+    }
+    Closable.ref(close)
+  }
+
+  private[this] def mkRouteMap(routeList: Seq[RouteRuleConfig]): Map[String, RouteRule] =
+    routeList.map(r => r.name.get -> r.spec.get).toMap[String, RouteRule]
+
+  private[this] lazy val routeRules: Activity[Map[String, RouteRule]] = {
+    val act = Activity(states)
+    val _ = act.states.respond(_ => ()) // register a listener forever to keep the Activity open
+    act
+  }
+
+  def getRules(): Future[Map[String, RouteRule]] = routeRules.toFuture
+}
--- a/k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
@@ -1,36 +1,29 @@
 package io.buoyant.k8s
 
 import com.twitter.util._
-import io.buoyant.k8s.PilotClient.RouteRuleConfig
+import io.buoyant.k8s.IstioPilotClient.RouteRuleConfig
 import java.util.concurrent.atomic.AtomicReference
 import istio.proxy.v1.config.RouteRule
 import io.buoyant.namer.RichActivity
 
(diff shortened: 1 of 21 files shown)


k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
```diff
@@ -0,0 +1,35 @@
+package io.buoyant.k8s
+
+import com.twitter.util._
+import io.buoyant.k8s.IstioPilotClient.RouteRuleConfig
+import java.util.concurrent.atomic.AtomicReference
+import istio.proxy.v1.config.RouteRule
+import io.buoyant.namer.RichActivity
+
+class RouteManager(api: IstioPilotClient, pollInterval: Duration) {
+  private[this] val states: Var[Activity.State[Map[String, RouteRule]]] = api.watchRouteRules(pollInterval).run.map {
```
adleong: I think this can just be
```
private[this] val routeRules: Activity[Map[String, RouteRule]] = api.watchRouteRules(pollInterval).map(mkRouteMap)
```
although you'd lose the logging...

k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
```diff
@@ -0,0 +1,35 @@
...
+    case Activity.Pending =>
+      log.trace("istio route manager is pending")
+      Activity.Pending
+    case Activity.Failed(e) =>
+      log.error(e, "failed to update istio routing rules")
+      Activity.Failed(e)
+  }
+
+  private[this] def mkRouteMap(routeList: Seq[RouteRuleConfig]): Map[String, RouteRule] =
+    routeList.flatMap {
```
adleong: I think this can just be
```
routeList.collect {
  case RouteRuleConfig(...) => name -> spec
}
```

k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
```diff
@@ -0,0 +1,35 @@
...
+      case RouteRuleConfig(typ, Some(name), Some(spec)) => Some(name -> spec)
+      case _ => None
+    }.toMap[String, RouteRule]
```
adleong: You might be able to leave off the type annotation here

k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
adleong: Or even make this a lazy val and do the observation here.

k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  private[this] lazy val routeRules: Activity[Map[String, RouteRule]] = {
+    val act = Activity(states)
+    val _ = act.states.respond(_ => ()) // register a listener forever to keep the Activity open
+    act
+  }
+
+  def getRules(): Future[Map[String, RouteRule]] = routeRules.toFuture
+}
```
adleong: We're going to need the RouteManager in the identifier AND in the interpreter.  What do you think about having an `apply` method on the companion to create RouteManagers that caches based on the URL for the client or something like that?  That way when the identifier and the interpreter both try to create a RouteManager, only one will be created and they can share it.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -0,0 +1,80 @@
...
+import io.buoyant.config.types.Port
+import io.buoyant.k8s.{ClientConfig, IstioPilotClient, RouteManager}
+import io.buoyant.linkerd.IdentifierInitializer
+import io.buoyant.linkerd.protocol.HttpIdentifierConfig
+import io.buoyant.router.RoutingFactory.{IdentifiedRequest, Identifier, RequestIdentification, UnidentifiedRequest}
+import com.twitter.conversions.time._
+import istio.proxy.v1.config.RouteRule
+
+class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeManager: RouteManager) extends Identifier[Request] {
+  private[this] val unidentified: RequestIdentification[Request] =
```
adleong: Should we fail the request if no routes match?  Or forward the request to it's stated destination (I _think_ this is what istio does) ?  I was thinking we could use `/svc/dest/<hostname>/<port>` in the case where no rules match.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -0,0 +1,80 @@
...
+    new UnidentifiedRequest(s"no matching istio rules found")
+
+  override def apply(req: Request): Future[RequestIdentification[Request]] = {
+    routeManager.getRules().map { rules =>
+      val filteredRules = rules.filter {
+        case (_, r) => r.`destination` == req.host
```
adleong: maybe add a TODO comment here about checking more route conditions

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
esbie: gc, forwarding would be a lot more user friendly

k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
esbie: sgtm we can put logging in the client or identifier etc

esbie:
okay I'm going to merge this into the `istio` branch as is and address comments in a followup, thank you!

PR #1389, opened 2017-06-14
Update to finagle 6.45.0, netty 4.1.10

klingerf:
As part of this change I'm creating a local copy of finagle's [AnyToHeapInboundHandler](https://github.com/twitter/finagle/blob/c86789cf0e064483ebf4509b52c9a216c31dd134/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/channel/AnyToHeapInboundHandler.scala), which was removed as part of this release. I'm also moving DstBindingFactory and DynBoundFactory to package com.twitter.finagle.naming.buoyant, to mirror the change in package for finagle's NameTreeFactory.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/UnpoolHandler.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/UnpoolHandler.scala
@@ -1,0 +1,56 @@
+package com.twitter.finagle.buoyant.h2
+package netty4
+
+import io.netty.buffer.{ByteBuf, ByteBufHolder, EmptyByteBuf, Unpooled}
+import io.netty.channel.ChannelHandler.Sharable
+import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandlerAdapter}
+
+/**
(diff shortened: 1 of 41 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/UnpoolHandler.scala
```diff
@@ -0,0 +1,56 @@
+package com.twitter.finagle.buoyant.h2
+package netty4
+
+import io.netty.buffer.{ByteBuf, ByteBufHolder, EmptyByteBuf, Unpooled}
+import io.netty.channel.ChannelHandler.Sharable
+import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandlerAdapter}
+
+/**
+ * This was originally copied from finagle's implementation of AnyToHeapInboundHandler:
```
adleong: This seems like the right thing to do for now.  Is there a ticket for figuring out buffer management in the long term?

klingerf: Created #1391 to track it.

PR #1390, opened 2017-06-14
RouteManager cache and updated path format

esbie:
Adds `RouteManager` cache and changes identifier path format to accommodate request forwarding. Paths will either be returned in `/svc/route/route-name` format for requests that match an istio route-rule or `/svc/dest/host/port` for requests that do not match.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
@@ -26,24 +26,18 @@ class RouteManager(api: IstioPilotClient, pollInterval: Duration) {
 }
 
 object RouteManager {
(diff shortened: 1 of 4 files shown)


k8s/src/main/scala/io/buoyant/k8s/RouteManager.scala
```diff
@@ -1,35 +1,49 @@
...
+    states
   }
 
   def getRules(): Future[Map[String, RouteRule]] = routeRules.toFuture
 }
+
+object RouteManager {
+  private[this] val managers = Var(Map.empty[(String, Int), RouteManager])
+
+  def getManagerFor(host: String, port: Int): RouteManager = {
```
adleong: TIOLI: could use com.twitter.util.Memoize instead of a Var.  Or even just a mutable Map.

PR #1392, opened 2017-06-14
Add Istio Interpreter

adleong:
Add the Istio Interpreter and the Istio Namer that it uses.

TODOS:
* wire it up to RouteManager
* actually map routes into dentries
* DRY up Istio-Pilot clients and reuse the underlying finagle client

### DIFF
--- a/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
+++ b/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
@@ -35,7 +35,7 @@ object IstioInterpreter {
 
     val routesDtab = routes.map { routeTable =>
       val dentries = routeTable.map { _ =>
-        val route = "my-cool-route"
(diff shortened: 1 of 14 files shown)


interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
```diff
@@ -0,0 +1,58 @@
...
+
+    // val routes = routerManager.routes
+    // TODO: Get routes from the route manager
+    val routes: Activity[IndexedSeq[Unit]] = Activity.value(IndexedSeq(()))
+
+    val routesDtab = routes.map { routeTable =>
+      val dentries = routeTable.map { _ =>
+        // TODO: map route into dentry
+        val route = "reviews-default"
+        val ns = "default"
```
esbie: heads up that the route-rules from the api server don't include a namespace field, so we need to derive that from the fully qualified domain name or something

PR #1396, opened 2017-06-15
Wire IstioInterpreter up to RouteManager

adleong:
IstioInterpreter now gets its list of routes from the RouteManager and uses those routes to build the route dentries.

### DIFF
--- a/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
+++ b/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
@@ -3,8 +3,9 @@ package io.buoyant.interpreter.k8s
 import com.twitter.finagle._
 import com.twitter.finagle.buoyant.PathMatcher
 import com.twitter.finagle.naming.NameInterpreter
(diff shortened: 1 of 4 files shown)


interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
```diff
@@ -28,25 +29,42 @@ object IstioInterpreter {
...
+            )
+          }
+          val prefix = Dentry.Prefix.read(s"/svc/route/$name")
+          Some(Dentry(prefix, NameTree.Union(branches: _*)))
+        case _ =>
+          None
+      }
+    }
+
+  def apply(routeManager: RouteManager, istioNamer: Namer): NameInterpreter = {
```
adleong: IstioNamers still can be configured directly if you wanted to use one standalone, but for the IstioInterpreter it has an IstioNamer that it uses internally that is not configured directly.  This is similar to how in a linkerd config that uses namerd, namers won't be configured directly in the linkerd config because the namers are "internal" to namerd.

We pass the istioNamer in here so we can pass it to the ConfiguredDtabNamer.

PR #1398, opened 2017-06-15
Introduce Istio telemeter and Mixer client

siggy:
Problem
We have no Istio Mixer integration or telemetry

Solution
Introduce an Istio Telemeter and Mixer client. This change adds support
for enabling an Istio Telemeter that connects to the Mixer gRPC API. The
Mixer client sends fake telemetry data every two seconds. Next steps
will be to wire up Mixer Client to a filter in the client stack.

Validation
./sbt linkerd-examples/istio:run
./sbt telemetry/test


telemetry/istio/src/main/scala/io/buoyant/telemetry/istio/MixerClient.scala
```diff
@@ -0,0 +1,114 @@
...
+
+private[telemetry] case class MixerClient(
+  service: Service[h2.Request, h2.Response]
+) {
+  import MixerClient._
+
+  def apply(): Future[ReportResponse] = {
+    val reportRequest = mkReportRequest()
+    val stream = h2.Stream()
+    val req = h2.Request("http", h2.Method.Post, "", "/", stream)
```
adleong: stream and req are unused I think.

telemetry/istio/src/main/scala/io/buoyant/telemetry/istio/IstioTelemeterInitializer.scala
```diff
@@ -0,0 +1,53 @@
...
+
+private[istio] case class IstioConfig(
+  hostname: Option[String],
+  port: Option[Int]
+) extends TelemeterConfig {
+  import IstioConfig._
+
+  @JsonIgnore override val experimentalRequired = true
+
+  @JsonIgnore private[this] val log = Logger.get("io.l5d.istio")
```
adleong: Might just use `Logger()` here since io.l5d.istio could refer to a number of different things

linkerd/docs/telemetry.md
```diff
@@ -54,6 +54,30 @@ Exposes admin endpoints:
...
+## Istio (experimental)
+
+> Example Istio config
+
+```yaml
+telemetry:
+- kind: io.l5d.istio
+  experimental: true
+  hostname: 127.0.0.1
+  port: 9091
```
esbie: might consider naming these `mixerHost` `mixerPort` to be explicit (the other istio components need different host/port combos so I could see this getting confusing)

linkerd/docs/telemetry.md
```diff
@@ -54,6 +54,30 @@ Exposes admin endpoints:
...
+kind: `io.l5d.istio`
+
+[Istio Mixer](https://github.com/istio/mixer) metrics exporting. This telemeter
+connects to a given Istio Mixer server via gRPC. ReportRequest objects are sent
+at request time.
+
+Key | Default Value | Description
+--- | ------------- | -----------
+experimental | _required_ | Because this telemeter is still considered experimental, you must set this to `true` to use it.
+hostname | `127.0.0.1` | Hostname of the Istio Mixer server.
```
esbie: Default should be `istio-mixer.default.svc.cluster.local` imo

telemetry/istio/src/main/scala/io/buoyant/telemetry/istio/MixerClient.scala
```diff
@@ -0,0 +1,114 @@
...
+import com.twitter.finagle.buoyant.h2
+import com.twitter.finagle.Service
+import com.twitter.util.Future
+import istio.mixer.v1.{Attributes, Mixer, ReportRequest, ReportResponse, StringMap}
+
+private[telemetry] object MixerClient {
+
+  // TODO: build up this dictionary as requests are sent out,
+  // also consider using defaultWords, once per ReportRequest, rather than per Attributes
+  private[telemetry] val words = Seq[String](
```
ZackButcher: There's no need to repeat the well-known attribute names in your `words` list: they already exist in the global list, so you can index into that list for their names. Then, you only need to include your own custom words in your per-request list. That should cut down on a lot of data over the wire.

The global list is defined here: https://github.com/istio/api/blob/master/mixer/v1/global_dictionary.yaml

You'll want to include the [global word count](https://github.com/istio/api/blob/master/mixer/v1/report.proto#L56) with each of your requests anyway (IIRC Mixer will reject the call if it can't agree on a dictionary), so you need either a copy of the global list or to hard-code the number of entries anyway. We do not anticipate changing the global list frequently, and any change made to it will be a breaking API change, so there will be plenty of warning. [We embed it directly into the Mixer.](https://github.com/istio/mixer/blob/master/pkg/api/globalWordList.go)

telemetry/istio/src/main/scala/io/buoyant/telemetry/istio/MixerClient.scala
```diff
@@ -0,0 +1,111 @@
...
+    "TARGET_LABELS_APP",
+    "version",
+    "TARGET_LABELS_VERSION"
+  )
+
+  private[telemetry] def mkReportRequest(): ReportRequest =
+    ReportRequest(
+      attributes = Seq[Attributes](
+        Attributes(
+          words = words,
```
ZackButcher: Since you're sharing the same word list for each of these `Attributes` messages, you can specify it once in the [`default_words`](https://github.com/istio/api/blob/master/mixer/v1/report.proto#L52) field of the `ReportRequest` and not have to repeat it per `Attributes` (even though there's only one `Attributes` right now).

telemetry/istio/src/main/scala/io/buoyant/telemetry/istio/MixerClient.scala
siggy: Thanks for the feedback @ZackButcher ! I didn't realize the global dictionary was embedded in Mixer, we'll remove this duplication in that case. We may still end up reading/importing `global_dictionary.yaml` since we'll need to reference those strings in `Attributes`, but good to know we don't need to send them over the wire.

telemetry/istio/src/main/scala/io/buoyant/telemetry/istio/MixerClient.scala
siggy: Makes sense. Our plan is to eventually bundle multiple Attributes, and will switch to `default_words` in that case. This code as it exists right now is mostly boiler to confirm we are speaking gRPC correctly. Thanks again!

PR #1399, opened 2017-06-15
Add port-names and tests to istio identifier

esbie:


linkerd/examples/istio.yaml
```diff
@@ -2,7 +2,9 @@ routers:
 - protocol: http
   identifier:
     kind: io.l5d.istio
-    host: localhost
+    apiserverHost: localhost
+    discoveryHost: localhost
+    discoveryPort: 8082
```
adleong: 8082?

esbie: ah whoops yeah I was using 8080 for `kubectl proxy`

PR #1414, opened 2017-06-17
Fix up the default routing in istio

adleong:
Simplify the way routing/naming works in istio to make routing when no rules match work properly.

Requests that match a rule are now named:
```
/svc/route/<route-name>/<port-name>
```

Requests that are for a valid cluster but don't match any rules:
```
/svc/dest/<cluster-name>/::/<port-name>
```

Requests that don't match a cluster name:
```
/svc/ext/<host>/<port-number>
```

This greatly simplifies the dtab that needs to be constructed.  The istio namer has also been changed to accept names in the form:
```
/prefix/<cluster-name>/<labels>/<port-name>
```


interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
```diff
@@ -1,54 +1,46 @@
...
 import istio.proxy.v1.config.RouteRule
 
 object IstioInterpreter {
-  private val noLabelsPfx = "/#/no-labels"
   private val istioPfx = "/#/io.l5d.k8s.istio"
   private val defaultRouteDtab = Dtab.read(s"""
-    |/svc/dest => /$$/inet ;
-    |/destPath/local/cluster/svc => $noLabelsPfx ;
-    |/svc/dest => /$$/io.buoyant.http.domainToPathPfx/destPath ;
+    |/svc/ext => /$$/inet ;
```
adleong: In the future we may want to send this to the egress linkerd instead of directly to the internet.

Issue #1402, opened 2017-06-16
Support for source and sourceTag MatchConditions


esbie:
IstioIdentifier should filter route-rules based on [MatchConditions](https://istio.io/docs/reference/config/traffic-rules/routing-rules.html#istio.proxy.v1.config.MatchCondition)

These match conditions are based off of request headers as well as the source's tags and FQDN (how do we get source information?)

PR #1418, opened 2017-06-20
Istio identifier header match conditions

esbie:
Adds match condition checks based on request header. Related to #1402 

The match condition's proto file defines matches as a `oneof`, which I couldn't get 
jackson to parse without adding a serializer/deserializer. This isn't a great long 
term solution since there are other `oneof` definitions that will likely exhibit the same issue.
Ulitmately the grpc-gen could generate these (de)serializers for all `oneof`s.


k8s/src/main/scala/io/buoyant/k8s/istio/PollingApiClient.scala
```diff
@@ -16,6 +16,9 @@ abstract class PollingApiClient(client: Service[Request, Response]) {
   protected val log = Logger()
   protected val mapper = Parser.jsonObjectMapper(Nil)
   mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+  registerModules(mapper)
+
+  def registerModules(mapper: ObjectMapper) {}
```
adleong: I think scala gets annoyed if you don't put in the Unit return type and the `=`.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -23,14 +24,36 @@ class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeCache: RouteCache, c
     }
   }
 
+  def compareStringMatch(headerValue: String, stringMatch: StringMatch): Boolean = {
```
adleong: TIOLI: naming this something like `headerMatches` would make it a bit more clear to me that it returns true for a match and false if it does not match.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -23,14 +24,36 @@ class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeCache: RouteCache, c
...
+    stringMatch.`matchType` match {
+      case Some(OneofMatchType.Exact(value)) => headerValue == value
+      case Some(OneofMatchType.Prefix(pfx)) => headerValue.startsWith(pfx)
+      case Some(OneofMatchType.Regex(r)) => headerValue.matches(r)
```
adleong: It's too bad we can't compile and store the regex on the StringMatch object.  This is not actionable, I'm just complaining.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -23,14 +24,36 @@ class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeCache: RouteCache, c
...
+      case None => throw new IllegalArgumentException("stringMatch missing matchType")
+    }
+  }
+
+  def compareMatchConditions(req: Request, matchCondition: MatchCondition): Boolean = {
```
adleong: TIOLI: what about `matchesAllConditions` or `matchesCondition`?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -23,14 +24,36 @@ class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeCache: RouteCache, c
...
+    val matchesHeaders = matchCondition.`httpHeaders`.map {
```
adleong: could do:
```
matchCondition.httpHeaders.forall {
  ...
}
```
instead of an explicit fold

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -23,14 +24,36 @@ class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeCache: RouteCache, c
...
+      case (headerName, stringMatch) =>
+        println(headerName, stringMatch)
```
adleong: rm println before shipping please

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -23,14 +24,36 @@ class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeCache: RouteCache, c
...
+        // return false if the headerName does not appear in the request's headerMap
+        req.headerMap.get(headerName).map(compareStringMatch(_, stringMatch)).getOrElse(false)
```
adleong: could also do
```
req.headerMap.get(headerName) match {
  case Some(headerValue) => headerMatches(headerValue, stringMatch)
  case None => false
}
```

(.get.map.getOrElse is also fine, just throwing this out there.  go with whatever seems clearer to you)

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -23,14 +24,36 @@ class IstioIdentifier(pfx: Path, baseDtab: () => Dtab, routeCache: RouteCache, c
...
     req.host match {
       case Some(host) =>
         Future.join(clusterCache.get(host), routeCache.getRules).map {
           case (Some(Cluster(dest, port)), rules: Map[String, RouteRule]) =>
             val filteredRules: Seq[(String, RouteRule)] = rules.filter {
-              //TODO: add more route conditions
-              case (_, r) => r.`destination` == Some(dest)
+              case (_, r) if r.`destination` == Some(dest) =>
+                // return true if no match conditions were defined on the route-rule
+                r.`match`.map(compareMatchConditions(req, _)).getOrElse(true)
```
adleong: could do match instead of .map.getOrElse here as well if you think it improves clarity.

esbie:
thanks for the style pointers!

PR #1423, opened 2017-06-20
Display server connections in the dashboard

### LINKED ISSUE #1421: Display server connections on admin dashboard
Super minor, but it would be helpful to see the number of server connections on the admin dashboard. We already display client connections, but not server connections. Something like:

screen shot 2017-06-20 at 10 57 08 am

I think if we add that, then the client stat grid and the server stat grid will be at parity, but we may also want to consider standardizing on "Successes" or "Success Rate", since those differ at present. Personally I like the "Successes" treatment that the server uses better.

rmars:
* Displays the number of server connections, like we do for clients
* Standardize the client and server displays
    - place the stats in the same order
    - display the successes in the client as number/percent, like we do for servers

Fixes #1421
screen shot 2017-06-20 at 1 14 39 pm


rmars:
Yeah! I think the border around the client sections makes it weird to also add the green/red vertical lines there (there's slightly less space). There are some simple things we could do without thinking too much about the design. wdyt?

I could make it uniform and take away the vertical lines from the server section:
screen shot 2017-06-23 at 11 32 36 am


Or, I could add vertical white lines in the server section, further differentiating it from the client:
screen shot 2017-06-23 at 11 31 52 am

Alternatively, we could have some sort of border around the server section:
screen shot 2017-06-23 at 11 35 07 am

klingerf:
😍  Ooh, I like option 2 -- vertical white lines.

esbie:
😍  I think I like the second one best! first one is good too

PR #1426, opened 2017-06-21
Use SingletonPool for H2

adleong:
Concurrent requests to an H2 client will create multiple connections.  This is
unnecessary because H2 is a multiplexing protocol.  One connection should be
sufficient.

Replace DefaultPool (which uses WatermarkPool) with SingletonPool.  The former
creates a connection for each concurrent request and the latter uses a single
connection for all requets.

Yet to be tested/validated.


adleong:
Closed in favor of https://github.com/linkerd/linkerd/pull/1444

Issue #1410, opened 2017-06-16
Support Mixer telemetry reporting


esbie:
Mixer provides a [report api](https://istio.io/docs/reference/api/mixer/mixer-service.html#report) that linkerd can use to report logs and metrics.

- [x] #1398 Introduce Istio telemeter and Mixer client
- [x] #1413 use Mixer's global dictionary
- [ ] switch to using Mixer global dictionary on the server
- [x] #1435 wire up request filter
- [ ] export `source.uid`, deprecate `source.labels` and `target.labels`

adleong:
Getting source.uid and target.uid is a bit tricky because linkerd is not running as a sidecar.  A couple options:

1.  For outgoing traffic, linkerd gets the remoteAddress of the request (ie source.ip).  Linkerd keeps a watch on the k8s API to keep a mapping of IP to pod name.  When this request reaches the incoming router of the destination linkerd, that linkerd looks at the remoteAddress of the response (ie target.ip) and uses its mapping to get the target.uid.  It encodes this as a header on the response it sends back to the source linkerd.  The source linkerd now has the source.uid and the target.uid.  
2. We run linkerd-tcp as a sidecar in addition to the linkerd which is run as a daemonset.  Since linkerd-tcp is running as a sidecar, it can get the pod name from the downward API.  The source linkerd-tcp can write the source.uid onto a request header and the target linkerd-tcp can write the target.uid onto a response header.  This gives the linkerds access to both pieces of information.

Option 1 requires each linked to keep an additional watch on the k8s api for the list of pods on its node.  To complicate things further, `remoteAddress` is not a field on response and we may need to make a finagle change to get access to this information. 

Option 2 is more complex since it requires both a sidecar linkerd-tcp and a daemonset linkerd (each request would make 5 hops from source app to destination app).  However, we may be moving in this direction anyway to have linkerd-tcp initiate and terminate TLS as a sidecar.  This option would require adding code in linkerd-tcp to read the environment variable and write the header.

esbie:
I mentioned this last week, but the istio identifier needs to know source id and source tags too (for [MatchConditions](https://github.com/linkerd/linkerd/issues/1402)). So whichever solution we decide on here will be used by the identifier as well.

PR #1435, opened 2017-06-22
http router loggers and mixer integration

siggy:
looking for feedback on:
- overall structure
- where in the stack to insert these loggers? (see `class HttpInitializer extends ProtocolInitializer.Simple`)
- best way to get `source.uid`, `target.uid`, noted in `mkReportRequest()`.  (perhaps map from `remoteAddress` via k8s API, per https://github.com/linkerd/linkerd/issues/1410#issuecomment-309193732 ?) OR, just send source and target labels, which still may need to come from a k8s watch?

part of #1410


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -209,6 +210,10 @@ case class HttpConfig(
   )
 
   @JsonIgnore
+  private[this] val combinedLoggers = loggers.map { configs =>
+    HttpLoggerConfig.param.Logger(configs.map(_.mk()) reduce (_ andThen _))
```
adleong: style nit: avoid infix notation

`configs.map(_.mk()).reduce(_ andThen _)`

k8s/src/main/scala/io/buoyant/k8s/istio/MixerClient.scala
```diff
@@ -25,14 +26,17 @@ private[telemetry] object MixerClient {
     mapper.readValue[Seq[String]](yaml).asInstanceOf[Seq[String]]
   }
 
-  private[telemetry] def indexOf(dict: Seq[String], str: String): Int =
+  private[this] def indexOf(dict: Seq[String], str: String): Int =
     -1 * (dict.indexOf(str) + 1)
 
-  private[telemetry] def mkReportRequest(): ReportRequest = {
+  // TODO: http1-only for now
+  private[istio] def mkReportRequest(req: Request, rsp: Option[Response]): ReportRequest = {
```
adleong: I think there's additional information we want to pass in here such as request duration, dst, and possibly other params from the endpoint stack.  We could add all of those things as parameters to this method or we could move the body of this method into the istio logger so it has direct access to that information.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpLoggerConfig.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.{Filter, Service, ServiceFactory, Stack, Stackable}
+import com.twitter.finagle.http.{Request, Response}
+import com.twitter.util.Future
+import io.buoyant.config.PolymorphicConfig
+
+abstract class HttpLoggerConfig extends PolymorphicConfig {
+  @JsonIgnore
+  def mk(): Filter[Request, Response, Request, Response]
```
adleong: probably want to add `params: Stack.Params` as a parameter to `mk` so that the filter has access to the stack params (like dst, and endpointAddr, for example).

k8s/src/main/scala/io/buoyant/k8s/istio/MixerClient.scala
adleong: if you make this take a `Try[Response]` instead of an `Option[Response]` then we'll have access to the exception in the case of a failure.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -35,6 +34,7 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .prepend(Headers.Dst.BoundFilter.module)
     val clientStack = Http.router.clientStack
       .prepend(http.AccessLogger.module)
+      .prepend(HttpLoggerConfig.module)
```
adleong: I think we want this to be somewhere in the endpoint stack so that is has access to the endpoint addr.  How about 
```
.insertAfter(ClientDestTracingFilter.Role, HttpLoggerConfig.module)
```

k8s/src/main/scala/io/buoyant/k8s/istio/MixerClient.scala
siggy: yeah! i had thought about moving the body into IstioLogger to make MixerClient not http1-specific, so rather than `logHttp`, we'd have:
```
MixerClient {
  def log(rsp: ResposeRequest)
}
```

my only concern was exposing the IstioLogger to the Mixer protobuf, but perhaps that's better than exposing MixerClient to all the http1-specifics ?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+  // Some((Path(%,io.l5d.k8s.daemonset,default,incoming,l5d,#,io.l5d.k8s.istio,reviews.default.svc.cluster.local,az:us-west::env:prod::version:v1,http),Path()))
+  val pathLength = 10
+  val pathServiceIndex = 7
+  val pathLabelsIndex = 8
+}
+
+class IstioLogger(mixerClient: MixerClient, params: Stack.Params) extends Filter[Request, Response, Request, Response] {
+  import IstioLogger._
+
+  private[this] val log = Logger.get()
```
adleong: I think we usually do `Logger()`

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  def apply(req: Request, svc: Service[Request, Response]) = {
+    val start = Time.now
+
+    svc(req).respond { ret =>
+      val duration = Time.now - start
```
adleong: code in finagle for this kind of thing usually uses `com.twitter.util.Stopwatch`.  Not sure what the difference is.... maybe stopwatch is more accurate?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+      val responseCode = {
```
adleong: maybe something like
```
ret.toOption.map(_.statusCode).getOrElse(Status.InternalServerError)
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+        ret match {
+          case Return(rsp) => rsp
+          case Throw(e) =>
+            // map exceptions to 500
+            log.warning(s"IstioLogger exception: $e")
```
adleong: this is an exception from the target service, right? not from istio?  probably shouldn't warn on that

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+            case path: Path if (path.elems.length == pathLength) => Some(path)
+            case _ => None
+          }
+        case None => None
+      }
+
+      val targetService = istioPath match {
+        case Some(path) =>
+          val elem = path.showElems(pathServiceIndex)
+          if (elem.indexOf(".svc.cluster.local") != -1) {
```
adleong: A bit worried about this being brittle.  I think this suffix might actually be configurable in kubernetes.  I'd be in favor blindly trusting what's in that path segment.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+          } else {
+            unknown
+          }
+        case None => unknown
+      }
+
+      val version = istioPath match {
+        case Some(path) =>
+          path.showElems(pathLabelsIndex).split("::").find {
+            e => e.indexOf("version:") == 0
```
adleong: could also do `e.startsWith`

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+          case Return(rsp) => rsp
+          case Throw(e) =>
+            // map exceptions to 500
+            log.warning(s"IstioLogger exception: $e")
+            Response(com.twitter.finagle.http.Status.InternalServerError)
+        }
+      }.statusCode
+
+      // check for valid istio path
+      val istioPath = DstBoundCtx.current match {
```
adleong: perfect opportunity for flatMap here:

`DstBoundCtx.current.flatMap { bound =>`

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+        case Some(bound) =>
+          bound.id match {
+            case path: Path if (path.elems.length == pathLength) => Some(path)
+            case _ => None
+          }
+        case None => None
+      }
+
+      val targetService = istioPath match {
```
adleong: Let's defer the default value of "unknown" until later.  Instead, we can make this an Option[String] by using `istioPath.map`.  This will pay off shortly...

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+          path.showElems(pathLabelsIndex).split("::").find {
+            e => e.indexOf("version:") == 0
+          } match {
+            case Some(label: String) => label.split(":").last
+            case _ => unknown
+          }
+        case None => unknown
+      }
+
+      val targetLabelApp = if (targetService != unknown) {
```
adleong: ...and the payoff!  `targetService.flatMap(_.split('.').headOption)`

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+          val elem = path.showElems(pathServiceIndex)
+          if (elem.indexOf(".svc.cluster.local") != -1) {
+            elem
+          } else {
+            unknown
+          }
+        case None => unknown
+      }
+
+      val version = istioPath match {
```
adleong: we can make this an Option too

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+      val targetLabelApp = if (targetService != unknown) {
+        targetService.split('.').head
+      } else {
+        unknown
+      }
+
+      val _ = mixerClient.report(
+        responseCode,
+        req.path,
+        targetService, // target in prom (~reviews.default.svc.cluster.local)
```
adleong: here's where we can `.getOrElse(unknown)` on all our Option[String] values

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,129 @@
...
+        unknown, // source in prom (~productpage)
+        targetLabelApp, // service in prom (~reviews)
```
adleong: I think this field is supposed to be the value of the "app" label of the target pod.  We populate it with the name of the target service.  This is good enough for now because (by convention) folks will usually name these the same.  But we should keep in mind (and probably add a comment) that we should use the correct value once we're able to get label values (or uids).

k8s/src/main/scala/io/buoyant/k8s/istio/MixerClient.scala
```diff
@@ -0,0 +1,134 @@
...
+    ).mkString
+
+    val mapper = new ObjectMapper(new YAMLFactory) with ScalaObjectMapper
+    mapper.registerModule(DefaultScalaModule)
+    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+
+    mapper.readValue[Seq[String]](yaml).asInstanceOf[Seq[String]]
+  }
+
+  private[this] def indexOf(dict: Seq[String], str: String): Int =
```
siggy: it was for istio > 0.1.6. i'll remove it for now.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioLogger.scala
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+}
+
+case class IstioLoggerConfig(
+  mixerHost: Option[String],
+  mixerPort: Option[Port]
+) extends HttpLoggerConfig {
+  @JsonIgnore
+  val DefaultMixerHost = "istio-mixer.default.svc.cluster.local"
```
adleong: A bunch of similar defaults were just moved into https://github.com/linkerd/linkerd/pull/1458/files#diff-cc9d18fb542f89cb0c372bf6de429a58

We should do the same for consistency.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/IstioLoggerTest.scala
```diff
@@ -0,0 +1,117 @@
...
+    val config = mapper.readValue[HttpLoggerConfig](yaml).asInstanceOf[IstioLoggerConfig]
+    val logger = config.mk(Stack.Params.empty)
+    assert(config.host == config.DefaultMixerHost)
+    assert(config.port == config.DefaultMixerPort)
+    // val req = Request()
+    // req.headerMap.set("l5d-name", "/one/two/three")
+    // assert(
+    //   await(identifier(req)).asInstanceOf[IdentifiedRequest[Request]].dst ==
+    //     Dst.Path(Path.read("/one/two/three"))
+    // )
```
olix0r: should these commented sections be removed?  or uncommented?

siggy:
i'lll work on creating a separate PR against master without all the istio changes.

siggy:
linkerd interface change in new PR against master: https://github.com/linkerd/linkerd/pull/1463

PR #1440, opened 2017-06-23
Support ingress to daemonset request path

### LINKED ISSUE #1433: Make k8s ingress controller class configurable
Ingress resources [meant for istio](https://istio.io/docs/tasks/ingress.html#configuring-ingress-http) are annotated with `kubernetes.io/ingress.class: istio`, and linkerd ignores them to avoid ingress controller collisions. If istio uses linkerd as its ingress controller, linkerd should be configured to read the istio annotated resources.

esbie:
In order to route from a Deployment of Linkerd Ingress Controllers to a Daemonset of Linkerds controlled via istio apis, some tweaks were needed.

- **Problem**: Host information is lost after identifying destination service via the `IngressIdentifier`, so the Daemonset instance would not know where to send traffic. **Solution**: combination `IstioIngressIdentifier` sends traffic directly to services.
- **Problem**: `destination` can be applied at the `destinationWeight` level in addition to the whole route-rule's `destination`. [See routing-rule reference](https://istio.io/docs/reference/config/traffic-rules/routing-rules.html). **Solution**: use destinationWeight's `destination` with a fallback to global `destination`.
- **Problem**: The route-rule `route` field is not strictly required. **Solution**: assign a default NameTree if it's missing
- **Problem**: Jackson does not respect default values if the key is not defined in json. **Solution**: convert `null` routes into an empty Seq. This fix isn't a good one, but I wasn't able to find jackson magic that does this for you...
- Fixes #1433: ingress class annotation is now configurable

**Validation:**
Additional unit tests, manual minikube testing of [enabling ingress traffic](https://istio.io/docs/tasks/ingress.html).

### DIFF
--- a/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
+++ b/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
@@ -34,13 +34,18 @@ object IstioInterpreter {
           .map { case (k, v) => s"$k:$v" }
           .mkString("::")
         val labelSegment = if (labels.isEmpty) "::" else labels
+        val clusterSegment = weightedDest.`destination`.getOrElse(cluster)
         NameTree.Weighted(
           weightedDest.weight.getOrElse(0).toDouble,
-          NameTree.Leaf(Path.read(s"/#/io.l5d.k8s.istio/$cluster/$labelSegment"))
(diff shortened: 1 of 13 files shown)


esbie:
Discussed bullet point 1 with @adleong. Decided to attempt a combination ingress-istio identifier & send  traffic directly to the service in lieu of forwarding traffic to the daemonset.

esbie:
OK this is ready for review, thank you!

PR #1444, opened 2017-06-26
H2 should multiplex streams over a single connection

adleong:
Finagle's DefaultPool (WatermarkPool specifically) will only hand out a service
to one caller at a time.  This means that each connection will only have one
(bidirectional) stream on it at any given time.

Replace DefaultPool with SingletonPool which hands out a single refcounted
service to all callers.  This takes advantage of H2's multiplexing.  To make
this work we fixed a couple issues:
* Added a delayed release module for H2 that delays release of a service until
  both the request and response streams have completed.
* Fixed a race condition in Netty4ClientDispatcher where the initial frame for
  a stream with a higher stream id could be written to the transport before
  the initial frame for a stream with a lower stream id.

I've manually tested this up to about 150rps with a concurrency of about 3.
More industrial strength load testing is required.

### DIFF
--- a/finagle/h2/src/main/java/io/netty/handler/codec/http2/BetterHttp2ConnectionEncoder.java
+++ b/finagle/h2/src/main/java/io/netty/handler/codec/http2/BetterHttp2ConnectionEncoder.java
@@ -1,0 +1,517 @@
+/*
+ * Copyright 2014 The Netty Project
+ *
+ * The Netty Project licenses this file to you under the Apache License, version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License. You may obtain a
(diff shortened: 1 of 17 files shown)


olix0r:
🎉 thanks for doing so much work to get to the bottom of this @adleong @klingerf!

For those following along:

If a linkerd h2 endpoint (either client or server) received a RST_STREAM message as it was trying to send a message on that stream, netty could throw an exception because it thought a write was being attempted on a non-existent stream. It would then try to _create_ the stream (even though it already been closed), but would fail when trying to allocate a client stream on the server (or vice versa).

The fix @adleong pulled in prevents the h2 encoder from trying to create a stream that has already been used.

PR #1446, opened 2017-06-26
Add HeadFilter to admin service (#1412)

### LINKED ISSUE #1412: Spurious log message for HEAD /admin/ping
I'm testing the liveliness of linkerd. I'm not interested in the word "pong" returned by the endpoint. I'm making a HEAD request to /admin/ping to get only the HTTP response code, since that's the only part that matters. 

Every time I make the request, linkerd logs the following useless log message:
`E 0617 00:08:35.145 UTC THREAD34: Received response to HEAD request (Request("HEAD /admin/ping", from /127.0.0.1:15000)) that contained a static body of length 4. Discarding body. If this is desired behavior, consider adding HeadFilter to your service`

hawkw:
Previously, the `admin/ping` route returns a message body for all request verbs, including HEAD.
This is not standards-compliant and produces a spurious finagle log message.

I've added a `HeadFilte`r to the admin service. This ensures that all HEAD requests have empty
message bodies.

`curl -X HEAD http://localhost:9990/admin/ping` no longer returns a message body.

Fixes #1412

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/Admin.scala
+++ b/admin/src/main/scala/io/buoyant/admin/Admin.scala
@@ -3,6 +3,7 @@ package io.buoyant.admin
 import com.twitter.app.{App => TApp}
 import com.twitter.finagle._
 import com.twitter.finagle.http.{HttpMuxer, Request, Response}
+import com.twitter.finagle.http.filter.HeadFilter
 import com.twitter.finagle.stats.NullStatsReceiver
 import com.twitter.finagle.tracing.NullTracer
 import com.twitter.logging.Logger
@@ -114,7 +115,7 @@ class Admin(val address: SocketAddress) {
         log.debug(s"admin: $url => ${service.getClass.getName}")
         muxer.withHandler(url, service)
     }
-    notFoundView.andThen(muxer)
+    HeadFilter.andThen(notFoundView.andThen(muxer))
   }
 
   def serve(app: TApp, extHandlers: Seq[Handler]): ListeningServer =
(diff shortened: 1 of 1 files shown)


admin/src/main/scala/io/buoyant/admin/Admin.scala
```diff
@@ -114,7 +115,7 @@ class Admin(val address: SocketAddress) {
         log.debug(s"admin: $url => ${service.getClass.getName}")
         muxer.withHandler(url, service)
     }
-    notFoundView.andThen(muxer)
+    HeadFilter.andThen(notFoundView.andThen(muxer))
```
adleong: I'm pretty sure filter composition is associative so this doesn't really matter but it's more common to do `A.andThen(B).andThen(C)` instead of `A.andThen(B.andThen(C))`

hawkw: Ah, of course – I'll make it prettier :)

admin/src/main/scala/io/buoyant/admin/Admin.scala
```diff
@@ -114,7 +115,7 @@ class Admin(val address: SocketAddress) {
         log.debug(s"admin: $url => ${service.getClass.getName}")
         muxer.withHandler(url, service)
     }
-    notFoundView.andThen(muxer)
+    HeadFilter andThen notFoundView andThen muxer
```
olix0r: we tend to prefer dots and parens over spaces... but this is probably fine in this contextx

hawkw: Also good to know, thanks; I'll keep that in mind. I wasn't sure about coding style, all I saw was "use Effective Scala"

PR #1448, opened 2017-06-27
Add jitter to Marathon namer poll interval (#171)

### LINKED ISSUE #171: Add jitter to marathon api polling

hawkw:
For very large clusters, if a bunch of linkerds all poll the Marathon namer at exactly the same time, the namer can be overloaded.

I've added a slight random jitter to poll attempts, based on the formula suggested by @adelong in comments on PR #772. The jitter amount is configurable in YAML. Note that I decided to just pass a function from `MarathonInitializer` to `AppIdNamer`, rather than pass a `Stream` as Alex [recommended](https://github.com/linkerd/linkerd/pull/772#issuecomment-256993836) – this required less modifications to existing code in `AppIdNamer`, and it
seemed simpler & thus more elegant. Annotating the jitter function with `@inline` should prevent it from creating a new call stack, so performance is probably about the same as using a `Stream`.

I've added a handful of tests for this: we test that the correct values are parsed from the config file,
that a respectable sample of the jittered TTLs are within the expected range, and that at least 50% of
the random values are unique.

Fixes #171

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/Admin.scala
+++ b/admin/src/main/scala/io/buoyant/admin/Admin.scala
@@ -3,6 +3,7 @@ package io.buoyant.admin
 import com.twitter.app.{App => TApp}
 import com.twitter.finagle._
 import com.twitter.finagle.http.{HttpMuxer, Request, Response}
+import com.twitter.finagle.http.filter.HeadFilter
 import com.twitter.finagle.stats.NullStatsReceiver
 import com.twitter.finagle.tracing.NullTracer
 import com.twitter.logging.Logger
@@ -114,7 +115,7 @@ class Admin(val address: SocketAddress) {
         log.debug(s"admin: $url => ${service.getClass.getName}")
         muxer.withHandler(url, service)
     }
-    notFoundView.andThen(muxer)
+    HeadFilter.andThen(notFoundView.andThen(muxer))
   }
 
   def serve(app: TApp, extHandlers: Seq[Handler]): ListeningServer =
--- a/admin/src/main/scala/io/buoyant/admin/Admin.scala
+++ b/admin/src/main/scala/io/buoyant/admin/Admin.scala
@@ -115,7 +115,7 @@ class Admin(val address: SocketAddress) {
         log.debug(s"admin: $url => ${service.getClass.getName}")
         muxer.withHandler(url, service)
     }
-    HeadFilter.andThen(notFoundView.andThen(muxer))
+    HeadFilter andThen notFoundView andThen muxer
   }
 
   def serve(app: TApp, extHandlers: Seq[Handler]): ListeningServer =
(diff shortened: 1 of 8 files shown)


admin/src/main/scala/io/buoyant/admin/Admin.scala
```diff
@@ -114,7 +115,7 @@ class Admin(val address: SocketAddress) {
         log.debug(s"admin: $url => ${service.getClass.getName}")
         muxer.withHandler(url, service)
     }
-    notFoundView.andThen(muxer)
+    HeadFilter andThen notFoundView andThen muxer
```
adleong: I think you need to merge master.

namer/marathon/src/test/scala/io/buoyant/namer/marathon/MarathonTest.scala
```diff
@@ -37,6 +40,88 @@ class MarathonTest extends FunSuite {
...
+    val minTtl = baseTtl - 50.millis
+    val maxTtl = baseTtl + 50.millis
+
+    for (jitter <- jitters.take(300)) {
+      assert(jitter >= minTtl)
+      assert(jitter <= maxTtl)
+    }
+
+    // assert that we're generating a reasonable amount of distinct TTLs
+    assert(jitters.take(300).toSet.size >= 50)
```
adleong: I'm mildly worried about this being flaky.  Back of the napkin math suggests the expected number of distinct elements is about 95 so this is probably ok, but to be ultra-conservative I'd be tempted to only check that there's more than 1 distinct value.  What do you think?

admin/src/main/scala/io/buoyant/admin/Admin.scala
hawkw: Yeah, I screwed up the merge, my bad.

namer/marathon/src/test/scala/io/buoyant/namer/marathon/MarathonTest.scala
hawkw: Yeah, I had this expecting 150 for a while (I hadn't had coffee yet, and instead of doing the math, I thought "half seems reasonable, right?" 😓) and it was failing with lengths around 95-98. 

After I worked out the percentage to actually expect, I was thinking about setting it to 95 but figured I'd lowball it because I didn't want to have a test that breaks occasionally due to probability. I can take it down to 2 but I suspect the odds of it dipping under 50 are pretty low...hmm.

hawkw: Changed this in 5d781156316f1ceb7d58e21a7fd8489c89e8aaaf. I took down the expected number of unique values to 2, and added a (perhaps too) long [comment](https://github.com/linkerd/linkerd/pull/1448/files#diff-99dfffd6beb2fa09f9d5c7dc5a7e9968R12) for historical reasons.

I think that seems fair?

adleong: Looks great!

hawkw: Okay, should I go ahead and merge this, or wait for @rmars?

PR #1451, opened 2017-06-28
Decide to use default failure accrual backoff lazily (#1439)

### LINKED ISSUE #1439: FailureAccrual backoff only uses the default values
I was playing with the failure accrual example and found that no matter what settings you put in the backoff configuration. 
I have tried using both constant and jittered backoff with a server that will send back 500 errors for all requests. Neither of the settings are used and the default values of jittered with a minimum of 500 ms and a maximum 30000ms.

hawkw:
## Problem:
As described in issue #1439, custom failure accrual back off policies specified in YAML are currently ignored – linkerd uses the default backoff policy regardless of what is specified.

## Solution:
I traced the issue to the `backoffOrDefault` value member in `FailureAccrualInitializer`. This value, which is set to the class's `backoff` member if it is defined, and to the default otherwise, was
being evaluated eagerly when instances of the initialiser are constructed. However, the `backoff` member, which is mutable, is *always* set to `None` initially, and is mutated to the config value
*after* the class is instantiated. Thus, `backoffOrDefault` is always set to the default.

I've fixed the issue simply by changing `backoffOrDefault` to a lazily-evaluated value.

## Validation:
In an attempt to localize this fault, I've added a number of unit tests.

I initially thought the fault was in how we parse configs for failure accrual policies, so I've added several tests for config parsing. I then proceeded to add tests which simulate failure accrual by repeatedly calling the failure method on a failure accrual policy and asserting the resulting Stream of values contains the correct probe durations.

Some of my unit tests were written with the ScalaCheck property testing
library, so our tests now depend on ScalaCheck.

Fixes #1439

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
@@ -16,7 +16,7 @@ abstract class FailureAccrualConfig extends PolymorphicConfig {
   var backoff: Option[BackoffConfig] = None
 
   @JsonIgnore
-  lazy val backoffOrDefault =
+  def backoffOrDefault: Stream[Duration] =
     backoff.map(_.mk).getOrElse(FailureAccrualConfig.defaultBackoff)
 }
 
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
@@ -16,7 +16,7 @@ abstract class FailureAccrualConfig extends PolymorphicConfig {
   var backoff: Option[BackoffConfig] = None
 
   @JsonIgnore
-  val backoffOrDefault =
+  lazy val backoffOrDefault =
     backoff.map(_.mk).getOrElse(FailureAccrualConfig.defaultBackoff)
 }
 
--- a/linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
+++ b/linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
@@ -3,15 +3,14 @@ package io.buoyant.linkerd.failureAccrual
 import io.buoyant.config.Parser
 import io.buoyant.linkerd.{ConstantBackoffConfig, FailureAccrualConfig, JitteredBackoffConfig}
 import io.buoyant.test.FunSuite
-import org.scalatest.{Matchers, OptionValues, OutcomeOf}
+import org.scalatest.OptionValues
 import com.twitter.conversions.time._
 import org.scalatest.prop.PropertyChecks
 import org.scalacheck.Gen
 
 class ConfigTest extends FunSuite
   with OptionValues
-  with PropertyChecks
-  with OutcomeOf {
+  with PropertyChecks {
 
   def parse(yaml: String): FailureAccrualConfig = {
     val mapper = Parser.objectMapper(yaml, Seq(Seq(
--- a/linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
+++ b/linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
@@ -7,15 +7,19 @@ import io.buoyant.linkerd.{ConstantBackoffConfig, FailureAccrualConfig, Jittered
 import io.buoyant.test.FunSuite
 import org.scalatest.{Matchers, OptionValues, OutcomeOf}
 import com.twitter.conversions.time._
-import org.scalatest.prop.TableDrivenPropertyChecks
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.prop.{PropertyChecks, TableDrivenPropertyChecks}
+import org.scalacheck.Gen
+import org.junit.runner.RunWith
 
 /**
  * Created by eliza on 6/27/17.
  */
+@RunWith(classOf[JUnitRunner])
 class ConfigTest extends FunSuite
   with Matchers
   with OptionValues
-  with TableDrivenPropertyChecks
+  with PropertyChecks
   with OutcomeOf {
 
   def parse(yaml: String): FailureAccrualConfig = {
@@ -75,6 +79,32 @@ class ConfigTest extends FunSuite
              """.stripMargin
       parse(yaml).backoff.value shouldEqual backoff
     }
+  }
+
+  private[this] val positiveInts = for { n <- Gen.choose(1, Integer.MAX_VALUE) } yield n
+
+  test("constant backoff configs produce streams of constant durations") {
+    forAll { (n: Int) => ConstantBackoffConfig(n).mk.take(100).forall(_ == n.millis) }
+  }
+
+  test("jittered backoff configs produce streams of durations between the minimum and maximum") {
+    forAll((positiveInts, "min"), (positiveInts, "max")) { (min: Int, max: Int) =>
+      whenever(min < max) {
+        JitteredBackoffConfig(Some(min), Some(max)).mk
+          .take(100)
+          .forall(d => d >= min.millis && d <= max.millis)
+      }
+  }
+
+  test("jittered backoff configs throw exceptions when passed invalid min/max") {
+    forAll { (min: Int, max: Int) =>
+      whenever(min >= max || min <= 0 || max <= 0 ) {
+        an[IllegalArgumentException] should be thrownBy JitteredBackoffConfig(Some(min), Some(max)).mk
+      }
+    }
+    an[IllegalArgumentException] should be thrownBy JitteredBackoffConfig(None, Some(1000)).mk
+    an[IllegalArgumentException] should be thrownBy JitteredBackoffConfig(Some(1000), None).mk
+    an[IllegalArgumentException] should be thrownBy JitteredBackoffConfig(None, None).mk
   }
 
 }
--- a/linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
+++ b/linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
@@ -9,7 +9,6 @@ import org.scalatest.prop.PropertyChecks
 import org.scalacheck.Gen
 
 class ConfigTest extends FunSuite
-  with Matchers
   with OptionValues
   with PropertyChecks
   with OutcomeOf {
@@ -49,14 +48,14 @@ class ConfigTest extends FunSuite
 
   test("configs parse to the correct kinds") {
     // TODO: this could look nicer as a table-based property check
-    parse("kind: io.l5d.consecutiveFailures") shouldBe a[ConsecutiveFailuresConfig]
-    parse("kind: io.l5d.successRate") shouldBe a[SuccessRateConfig]
-    parse("kind: io.l5d.successRateWindowed") shouldBe a[SuccessRateWindowedConfig]
-    parse("kind: none") shouldBe a[NoneConfig]
+    assert(parse("kind: io.l5d.consecutiveFailures").isInstanceOf[ConsecutiveFailuresConfig])
+    assert(parse("kind: io.l5d.successRate").isInstanceOf[SuccessRateConfig])
+    assert(parse("kind: io.l5d.successRateWindowed").isInstanceOf[SuccessRateWindowedConfig])
+    assert(parse("kind: none").isInstanceOf[NoneConfig])
   }
 
   test("unspecified backoffs should parse to None") {
-    forAll(kinds) { parse(_).backoff shouldBe None }
+    forAll(kinds) { cfg => assert(parse(cfg).backoff.isEmpty) }
   }
 
   test("configs with backoff configurations have the correct backoff") {
@@ -69,7 +68,7 @@ class ConfigTest extends FunSuite
              |backoff:
              |$config
              """.stripMargin
-      parse(yaml).backoff.value shouldEqual backoff
+      assert(parse(yaml).backoff.value == backoff)
     }
   }
 
@@ -94,7 +93,9 @@ class ConfigTest extends FunSuite
   test("jittered backoff configs produce streams containing at least two unique durations") {
     forAll((positiveInts, "min"), (positiveInts, "max")) { (min: Int, max: Int) =>
       whenever(min < max) {
-        JitteredBackoffConfig(Some(min), Some(max)).mk.take(300).toSet.size should be >= 2
+        val n_unique = JitteredBackoffConfig(Some(min), Some(max)).mk
+                          .take(300).toSet.size
+        assert(n_unique >= 2)
       }
     }
   }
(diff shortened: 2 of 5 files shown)


hawkw:
This was a fun one – it was a bit trickier than I expected, and then the fix was deceptively simple :)

linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
```diff
@@ -16,7 +16,7 @@ abstract class FailureAccrualConfig extends PolymorphicConfig {
   var backoff: Option[BackoffConfig] = None
 
   @JsonIgnore
-  val backoffOrDefault =
+  lazy val backoffOrDefault =
```
adleong: this is probably fine but it might be more technically correct for this to be a `def` so that it remains correct if the `backoff` var changes.

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+import io.buoyant.test.FunSuite
+import org.scalatest.{Matchers, OptionValues, OutcomeOf}
+import com.twitter.conversions.time._
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.prop.{PropertyChecks, TableDrivenPropertyChecks}
+import org.scalacheck.Gen
+import org.junit.runner.RunWith
+
+/**
+ * Created by eliza on 6/27/17.
```
adleong: rm this comment please

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+ */
+@RunWith(classOf[JUnitRunner])
```
adleong: For consistency with our other tests you should be able to omit the annotation and extend `io.buoyant.test.FunSuite`.

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+        """|  kind: jittered
+          |  minMs: 300
+          |  maxMs: 400""".stripMargin,
+        JitteredBackoffConfig(Some(300), Some(400))
+      )
+    )
+
+  test("configs parse to the correct kinds") {
+    // TODO: this could look nicer as a table-based property check
+    parse("kind: io.l5d.consecutiveFailures") shouldBe a[ConsecutiveFailuresConfig]
```
adleong: `shouldBe` is cool but I'd recommend `assert(...isInstanceOf[...])` to be consistent with other tests

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      val yaml =
+        s"""$kind
+             |backoff:
+             |$config
+             """.stripMargin
+      parse(yaml).backoff.value shouldEqual backoff
+    }
+  }
+
+  private[this] val positiveInts = for { n <- Gen.choose(1, Integer.MAX_VALUE) } yield n
```
adleong: Since I haven't used scalacheck much, can you briefly explain what the difference between `for { n <- Gen.choose(1, Integer.MAX_VALUE) } yield n` and simply `Gen.choose(1, Integer.MAX_VALUE)` is?

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+  test("configs parse to the correct kinds") {
+    // TODO: this could look nicer as a table-based property check
+    parse("kind: io.l5d.consecutiveFailures") shouldBe a[ConsecutiveFailuresConfig]
+    parse("kind: io.l5d.successRate") shouldBe a[SuccessRateConfig]
+    parse("kind: io.l5d.successRateWindowed") shouldBe a[SuccessRateWindowedConfig]
+    parse("kind: none") shouldBe a[NoneConfig]
+  }
+
+  test("unspecified backoffs should parse to None") {
+    forAll(kinds) { parse(_).backoff shouldBe None }
```
adleong: I think this is a property with an assertion in it.  I think we should either iterate over kinds and assert for each one or remove the assertion and just have this be a property.

adleong: Wait nevermind, is this `forAll` from TableDrivenPropertyChecks or from `org.scalacheck.Prop`?

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  test("configs with backoff configurations have the correct backoff") {
```
adleong: does this test actually fail when one of these cases fails?  or does it succeed and produce a sequence of failed Outcomes?

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+             |$config
+             """.stripMargin
+      parse(yaml).backoff.value shouldEqual backoff
+    }
+  }
+
+  private[this] val positiveInts = for { n <- Gen.choose(1, Integer.MAX_VALUE) } yield n
+
+  test("constant backoff configs produce streams of constant durations") {
+    forAll { (n: Int) => ConstantBackoffConfig(n).mk.take(100).forall(_ == n.millis) }
```
adleong: Doesn't this just define a property?  Don't we have to check it as well?  And assert that the check was successful?

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      whenever(min < max) {
+        JitteredBackoffConfig(Some(min), Some(max)).mk.take(300).toSet.size should be >= 2
+      }
+    }
+  }
+
+  test("jittered backoff configs throw exceptions when passed invalid min/max") {
+    forAll { (min: Int, max: Int) =>
+      whenever(min >= max || min <= 0 || max <= 0) {
+        an[IllegalArgumentException] should be thrownBy JitteredBackoffConfig(Some(min), Some(max)).mk
```
adleong: Use `assertThows` or `intercept` for consistency

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConsecutiveFailuresTest.scala
```diff
@@ -16,4 +30,26 @@ class ConsecutiveFailuresTest extends FunSuite {
...
+  test("constant backoff") {
+    val yaml = """|kind: io.l5d.consecutiveFailures
+                  |failures: 5
+                  |backoff:
+                  |  kind: constant
+                  |  ms: 10000""".stripMargin
+    val policy = parse(yaml).policy()
+    val probeDelays = Stream.continually(policy.markDeadOnFailure()).take(20)
+    probeDelays.take(4) should contain only (None)
+    probeDelays should contain inOrderOnly (None, Some(10000.millis))
```
adleong: how about checking that `probeDelays.drop(4)` only contains the Somes

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      new ConsecutiveFailuresInitializer,
+      new NoneInitializer,
+      new SuccessRateInitializer,
+      new SuccessRateWindowedInitializer
+    )))
+    mapper.readValue[FailureAccrualConfig](yaml)
+  }
+
+  val kinds =
+    Table(
```
adleong: For my education: what's the advantage of table based checks rather than iterating over a collection and asserting?

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: Oops! I should probably tell IDEA to stop making those...

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: This [should](https://github.com/linkerd/linkerd/pull/1451/files/af864d60f3b6dcb11f39776ba5ccda51a15683eb#diff-870b1825491434809bccc89f7d11fe56R7) already extend `io.buoyant.test.FunSuite`. I think you're right that the annotation can be removed, though.

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: This is the `forAll` from `PropertyChecks`.

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: Personally, I prefer `should` because I think it's more expressive, and produces more readable failure messages, but I'd be happy to rewrite this to use `assert`.

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: Whoops. Fixed that here: 0192b58e485c191112b447cd8aff57e2a14a2650

linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
hawkw: Hmm. I used `lazy val` since the original code used a `val` and it seemed like the intention was for this value to not change after the config is parsed. Are we expecting that configs will ever be parsed?

hawkw: I think that –  if we're not in the habit of mutating these configs post-parsing – ideally, it would be nice if the config was immutable, and was instantiated by the parser with all the values set, rather than instantiated and then mutated by the parser (this error could have been avoided entirely!). I'm not sure if our parsing code would easily allow that in this instance, though...

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: The main advantage of table-based checks is that they allow naming of columns and make somewhat more readable failure messages, and use the same verbs as generator-based property checks (for consistency). I think in this case, since the table has a single column, this is basically the same thing as iterating over a `Seq`...I could easily replace this with a `Seq` if you'd prefer.

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: Ah, yeah, that definitely shouldn't be there; I haven't written ScalaTest tests for a bit & misunderstood something from the docs. Fixed in d16e001e88abf2882cacb304fa354e9603c00616

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: They should be equivalent; the `for` expression is completely vestigial. I think this started out its life as `for { n <- Arbitrary[Int] if n >= 1} yield n` or something like that, before I rewrote it with `Gen.choose` and forgot to remove the `for`-expression. The map is just doing the identity operation at this point; I'll get rid of it. Thanks for catching that, now I feel silly.

linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
adleong: Ideally the parser would spit out an immutable config, but, between jackson and inheritance, and polymorphism, that's not the world we live in.

Given that `backoff` has to be a var (sad face) I really think this should be a def.  Otherwise a stray early access of this field could lock it to an incorrect value.  In general, lazy vals in scala are kind of dangerous and hard to reason about so I like to avoid them except as a last resort.

hawkw: Okay, if we're limited by Jackson in this case, I'll definitely make it a `def` for now. It would be nice to use a parsing library that generates immutable immutable config classes but that sounds like it would be a major structural change at this point.

I tend to make fairly liberal use of `lazy val`s in my own code, but this is because I generally try to make everything immutable – i've been known to set WartRemover to make `var` into a compile error. In this case, you're right, and I'll change it to a `def`.

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,114 @@
+package io.buoyant.linkerd.failureAccrual
+
+import io.buoyant.config.Parser
+import io.buoyant.linkerd.{ConstantBackoffConfig, FailureAccrualConfig, JitteredBackoffConfig}
+import io.buoyant.test.FunSuite
+import org.scalatest.{Matchers, OptionValues, OutcomeOf}
```
adleong: I think we can drop Matchers and OutcomeOf now

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
```diff
@@ -0,0 +1,114 @@
...
+      new NoneInitializer,
+      new SuccessRateInitializer,
+      new SuccessRateWindowedInitializer
+    )))
+    mapper.readValue[FailureAccrualConfig](yaml)
+  }
+
+  val kinds =
+    Table(
+      ("kind"),
```
adleong: Can drop extra parens

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
adleong: So Table + forAll is essentially Seq + foreach but with nicer test output?  Unclear whether this is worth the cognitive overhead of doing things differently than other tests in the repo.

linkerd/core/src/main/scala/io/buoyant/linkerd/FailureAccrualInitializer.scala
hawkw: Changed: b1f480edd0ed9a452dbdd7950b820afb7158cabd

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConsecutiveFailuresTest.scala
```diff
@@ -2,10 +2,27 @@ package io.buoyant.linkerd.failureAccrual
...
+import com.twitter.conversions.time._
+import io.buoyant.config.Parser
+import io.buoyant.linkerd.{FailureAccrualConfig, FailureAccrualInitializer}
 import io.buoyant.test.FunSuite
+import org.scalacheck.Gen
+import org.scalatest.Matchers
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+
+class ConsecutiveFailuresTest extends FunSuite
+  with Matchers
```
adleong: Lets drop Matchers from this file too

hawkw: There's a couple places in this file where I'm actually still using `Matchers` for testing the structure of collections. I think they're more expressive than assertions & that code would get a bit more complex with just `assert`

linkerd/failure-accrual/src/test/scala/io/buoyant/linkerd/failureAccrual/ConfigTest.scala
hawkw: Ah, yeah, I forgot to remove unused imports. My bad.

Issue #1450, opened 2017-06-28
Automatic env var substitution in linkerd config files


kevholditch:
Support automatic replacement of environment variables in the linkerd.conf file when linkerd starts up.

This is useful in a number of scenarios when parts of the config are not known until runtime especially when running in a docker container ([see my blog post for a real world scenario](https://kevinholditch.co.uk/2017/06/28/running-linkerd-in-a-docker-container-on-aws-ecs/)).

The format of the variable in the config file could be prefixed with a '$' for example:

```
interpreter:
kind: default
    transformers:
    – kind: io.l5d.specificHost
    host: $LOCAL_IP
```

Here $LOCAL_IP would be replaced with the value found at runtime in the environment variable LOCAL_IP.

olix0r:
the way i've done this sort of thing in the past is passing in the config on stdin like:
```
./genconfig.sh |./linkerd -
```

where genconfig.sh has something like:
```sh
#!/bin/sh

set -eu

cat <<EOF
interpreter:
  kind: default
    transformers:
    – kind: io.l5d.specificHost
    host: $LOCAL_IP
EOF
```

wmorgan:
Thanks for submitting this @kevholditch!

redijedi:
@kevholditch I followed your post but I get the following when dropping it in as a replacement for the standard image:

`[dumb-init] /io.buoyant/linkerd/config/config.yaml: Permission denied`

Please advise.

PR #1453, opened 2017-06-28
ec2Host per-host-interpreter to filter requests to the ec2 host you are running on

kevholditch:
The transformer uses the aws meta endpoint on to retrieve the ip of the host and filter based upon this.  This is especially useful if you are running on a docker container in AWS ECS and you want to filter requests to other docker containers running on the same ec2 instance.

To use the interpreter kind is: `io.l5d.ec2Host`.

Example config snippet:

```
interpreter:
    kind: default
    transformers:
    - kind: io.l5d.ec2Host
```

### DIFF
--- a/interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
+++ b/interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
@@ -1,0 +1,36 @@
+package io.buoyant.transformer
+package perHost
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import io.buoyant.namer.{NameTreeTransformer, TransformerConfig, TransformerInitializer}
+import java.net.InetAddress
+import com.twitter.finagle.{Http, Stack, http, Path}
+import com.twitter.util.Await
+
+class Ec2HostTransformerInitializer extends TransformerInitializer {
+  val configClass = classOf[Ec2HostTransformerConfig]
+  override val configId = "io.l5d.ec2Host"
+}
+
+class Ec2HostTransformerConfig extends TransformerConfig {
+
+  @JsonIgnore
+  val defaultPrefix = Path.read("/io.l5d.ec2Host")
+
+  @JsonIgnore
+  override def mk(params: Stack.Params): NameTreeTransformer = {
+
+    val servicePort = 80
+    val serviceHost = "169.254.169.254"
+    val service = Http.client
+      .newService(s"/$$/inet/$serviceHost/$servicePort")
+
+    val req = http.Request("latest/meta-data/local-ipv4")
+    req.method = http.Method.Get
+    val response = Await.result(service.apply(req))
+    val ec2HostIp = response.contentString
+
+    new SubnetLocalTransformer(prefix, Seq(InetAddress.getByName(ec2HostIp)), Netmask("255.255.255.255"))
+  }
+
+}
--- a/interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
+++ b/interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
@@ -1,18 +1,22 @@
 package io.buoyant.transformer
 package perHost
 
-import com.fasterxml.jackson.annotation.JsonIgnore
-import io.buoyant.namer.{NameTreeTransformer, TransformerConfig, TransformerInitializer}
 import java.net.InetAddress
-import com.twitter.finagle.{Http, Stack, http, Path}
-import com.twitter.util.Await
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.{Http, Path, Stack, http}
+import com.twitter.util.Activity
+import io.buoyant.namer.{NameTreeTransformer, TransformerConfig, TransformerInitializer}
 
 class Ec2HostTransformerInitializer extends TransformerInitializer {
   val configClass = classOf[Ec2HostTransformerConfig]
   override val configId = "io.l5d.ec2Host"
 }
 
-class Ec2HostTransformerConfig extends TransformerConfig {
+class Ec2HostTransformerConfig() extends TransformerConfig {
+
+  private val awsMetaAddress = "169.254.169.254"
+  private val port = 80
 
   @JsonIgnore
   val defaultPrefix = Path.read("/io.l5d.ec2Host")
@@ -20,17 +24,13 @@ class Ec2HostTransformerConfig extends TransformerConfig {
   @JsonIgnore
   override def mk(params: Stack.Params): NameTreeTransformer = {
 
-    val servicePort = 80
-    val serviceHost = "169.254.169.254"
     val service = Http.client
-      .newService(s"/$$/inet/$serviceHost/$servicePort")
+      .newService(s"/$$/inet/$awsMetaAddress/$port")
 
     val req = http.Request("latest/meta-data/local-ipv4")
-    req.method = http.Method.Get
-    val response = Await.result(service.apply(req))
-    val ec2HostIp = response.contentString
+    val activity = Activity.future(service.apply(req).map(r => Seq(InetAddress.getByName(r.getContentString()))))
 
-    new SubnetLocalTransformer(prefix, Seq(InetAddress.getByName(ec2HostIp)), Netmask("255.255.255.255"))
+    new FutureSubnetLocalTransformer(prefix, activity, Netmask("255.255.255.255"))
   }
 
 }
--- a/interpreter/per-host/src/main/resources/META-INF/services/io.buoyant.namer.TransformerInitializer
+++ b/interpreter/per-host/src/main/resources/META-INF/services/io.buoyant.namer.TransformerInitializer
@@ -1,3 +1,4 @@
 io.buoyant.transformer.perHost.LocalhostTransformerInitializer
 io.buoyant.transformer.perHost.PortTransformerInitializer
 io.buoyant.transformer.perHost.SpecificHostTransformerInitializer
+io.buoyant.transformer.perHost.Ec2HostTransformerInitializer
--- a/interpreter/subnet/src/main/scala/io/buoyant/transformer/SubnetLocalTransformer.scala
+++ b/interpreter/subnet/src/main/scala/io/buoyant/transformer/SubnetLocalTransformer.scala
@@ -1,7 +1,9 @@
 package io.buoyant.transformer
 
-import com.twitter.finagle.{Address, Path}
 import java.net.InetAddress
+
+import com.twitter.finagle.{Address, Path}
+import com.twitter.util.Activity
 import io.buoyant.namer.FilteringNameTreeTransformer
 
 /**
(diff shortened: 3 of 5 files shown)


adleong:
This is cool!  Can you tell me a little bit more about how it works?  Is this a scenario where each docker container has its own IP or do they all share the host IP?  Looking at the code, it looks like we get the host IP from the aws api and then only allow traffic to that IP.  Could the `io.l5d.localhost` transformer work here?

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
```diff
@@ -0,0 +1,36 @@
...
+class Ec2HostTransformerConfig extends TransformerConfig {
+
+  @JsonIgnore
+  val defaultPrefix = Path.read("/io.l5d.ec2Host")
+
+  @JsonIgnore
+  override def mk(params: Stack.Params): NameTreeTransformer = {
+
+    val servicePort = 80
+    val serviceHost = "169.254.169.254"
```
adleong: Does the aws api always live at this ip address?  I think this should be configurable.

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
```diff
@@ -0,0 +1,36 @@
...
+
+class Ec2HostTransformerConfig extends TransformerConfig {
+
+  @JsonIgnore
+  val defaultPrefix = Path.read("/io.l5d.ec2Host")
+
+  @JsonIgnore
+  override def mk(params: Stack.Params): NameTreeTransformer = {
+
+    val servicePort = 80
```
adleong: Let's make this configurable.

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
```diff
@@ -0,0 +1,36 @@
...
+    val serviceHost = "169.254.169.254"
+    val service = Http.client
+      .newService(s"/$$/inet/$serviceHost/$servicePort")
+
+    val req = http.Request("latest/meta-data/local-ipv4")
+    req.method = http.Method.Get
```
adleong: GET is default, so I think we can omit this.

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
```diff
@@ -0,0 +1,36 @@
...
+    val response = Await.result(service.apply(req))
```
adleong: We always want to avoid blocking the thread with `Await.result`.  Instead, I think you'll need to write a new transformer (perhaps it's possible to subclass SubnetLocalTransformer) that turns this Future into an Activity (using `Activity.future`) and uses that.

kevholditch:
@adleong thanks! The scenario is you are running linkerd on a docker container on AWS ECS.  The linkerd docker container gets its own IP address different from the host EC2 instance.  You have a consul container also running on the same host and a container with your service in.  Your service will be registered in consul with the IP address of the EC2 host.  The ec2Host filter allows you to filter only services running in containers on the same machine in an automated way.

Unfortunately none of the other filters fit the bill as we don't know the hostname or IP of the EC2 host until runtime.  I came up with a workaround for now as detailed on [my blog](https://kevinholditch.co.uk/2017/06/28/running-linkerd-in-a-docker-container-on-aws-ecs/).

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
kevholditch: It could make sense to make the port configurable as it is possible to expose it differently inside your docker container.  On the host it will always be port 80 though.

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
kevholditch: The AWS API always lives at this address.  It is a special endpoint that you can hit on any AWS EC2 box.  I don't believe this is possible to change so I'm not sure it makes sense to make this configurable?

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
kevholditch: I'm going to try and attempt to do this... (wish me luck :) )

interpreter/per-host/src/main/scala/io/buoyant/transformer/perHost/Ec2HostTransformerInitializer.scala
kevholditch: It is always on port 80 so I do not see why we should make this configurable?  It adds more configuration overhead, surely it is better to favour convention over configuration?  This filter is specifically for use on AWS EC2 instances.

adleong:
Thanks for all the context, that's super helpful.  The solution you describe on your blog actually seems quite nice and elegant to me.  Assuming the IP of the host doesn't change over time, fetching it prior to starting linkerd and then passing it into the config with `envsubst` seems like the right thing to do.  Are there shortcomings to this solution that made you want to build this functionality into linkerd itself?

On a related note, we could definitely consider changing our docker image to include gettext so that users who want to envsubst don't have to roll their own.

kevholditch:
@adleong thanks for getting back to me. Yep the shortcomings of this solution is that currently to filter to a specific host when running on AWS ECS you have to jump through the hoops mentioned above. The solution may be ok but every person who wants to run linkerd on AWS ECS and filter on the host is going to run into this problem and have to work out a solution (like I did). It also means that it is necessary to create your own docker container (even if the official one included gettext) as you have to template the linkerd config before starting Linkerd. 

@dario-simonetti (the author) of the specific host filter ran into the same issue and had to solve it using a similar technique.

adleong:
I'd like to get our official docker image to the point where users can volume mount a templated linkerd config and have envsubst run on it without needing to build their own image.  And we should definitely document all of this so that folks don't have to work this out on their own.

kevholditch:
@adleong another way of approaching this would be to build in environment variable substition into linkerd itself. Linkerd could parse its config file and replace environment variables before it parses the config file. I have [raised an issue for this](https://github.com/linkerd/linkerd/issues/1450)

klingerf:
Just wanted to mention that this is already mostly possible with the existing image.

For instance, if I have a shell script that generates a config file and does env var substitution:

```
$ cat gen-config.sh 
#!/bin/sh

cat <<EOF
admin:
  port: $ADMIN_PORT

routers:
- protocol: http
  dtab: /svc => /$/inet/127.1/$ADMIN_PORT;
  servers:
  - port: 4140
    ip: 0.0.0.0
EOF
```

Then I can mount that into the released linkerd image and pipe the output of the script to start the linkerd process:

```
$ docker run --rm -e ADMIN_PORT=9999 -p 4140:4140 -v `pwd`/gen-config.sh:/gen-config.sh --entrypoint='/bin/sh' buoyantio/linkerd:1.1.0 -c '/gen-config.sh | ./bundle-exec -- -'
...
I 0630 18:41:32.733 UTC THREAD1: serving http admin on /0.0.0.0:9999
```

You can see that it starts the admin server on the port specified by the ADMIN_PORT environment variable. And sending a request to the router running on port 4140 forwards to the correct admin port.

```
$ curl 192.168.99.100:4140/admin/ping
pong
```

kevholditch:
@klingerf thanks. The problem is when running on AWS ECS this means having the linkerd templates config file on each ec2 ECS host machine that is running the containers. This means that to change the config you have to coordinate updating the config on every ec2 host and then restart the containers. Due to this constraint practically you have to run your own linkerd container. So when time comes you want to make a change you can just roll that out using AWS ECS by updating the task and not having somehow change files on all of the host machines

klingerf:
@kevholditch Ok, makes sense. Sorry, posted by comment before reading the issue that you had created.

kevholditch:
@adleong Hey Alex, I've made the changes to the filter to use an activity so it no longer blocks waiting for the IP address back from the AWS meta endpoint.

dario-simonetti:
I understand there are ways around this so that this transformer it's not essential for deployments on ECS, but it would be a nice-to-have. I had to do a [similar hack](https://medium.com/p/linkerd-a-service-mesh-for-aws-ecs-937f201f847a) in order to get it working on AWS ECS and I have would liked this transformer to exist in order to make my life a bit easier and not having to go through that mess.

kevholditch:
@adleong where is your documentation repo?  I'd like to update it if possible to include my change as part of the PR

dario-simonetti:
It's in this very repo @kevholditch, see https://github.com/linkerd/linkerd/blob/master/linkerd/docs/config.md

adleong:
Hey @kevholditch, we keep missing each other on Slack so I'll post my thoughts here.  I'm still having a hard time understanding what the drawback is to using a templated config file to accomplish this (as is done in the blogpost).

It sounds like distributing config files to ec2 hosts is difficult and therefore you need to roll the linkerd config into the docker image itself.  That's a bummer but it seems independent of whether the config is templated or not.  Unless I'm missing something, even with this transformer you would still need to distribute linkerd configs by bundling them into a docker image.

So I'm not really sure what this transformer buys you other than not needing to run `envsubst` as part of the Linkerd command.  To me, the task of fetching this kind of configuration information from the AWS api feels more natural as a startup command that feeds into Linkerd's configs than as something Linkerd needs to figure out during runtime.

Please let me know if I've misunderstood or am missing something.

kevholditch:
hey @adleong, that's ok timezones are tough as you said.

You are correct with everything you say.  In that we would have to run our own custom container anyway as it is tough to manage a config file on the ECS host itself.  

The workaround using `envsubst` works but feels quite awkward.  It puts up an obstacle that both I and @dario-simonetti encountered (and probably many more people) when trying to run linkerd on a container on ECS. As my `envsubst` solution is included in this PR I think it seems quite obvious but it took us a few days to come up with it and during that time we actually was starting to think it wasn't possible to run linkerd on a container on ECS.  I'm sure other people will also come up against this issue and may come to that conclusion causing them potentially to look at service mesh alternatives (as we nearly did).

When we can write a solution that makes linkerd play nicely with ECS I'm not sure why we wouldn't include that?  It would give ECS users a much better experience and would save them all having to come up with the same workaround as I did.

adleong:
You folks have done great work figuring out how to run Linkerd on ECS and I definitely want to build on that and make sure people don't have to re-invent the wheel.  What do you think about this:

1. Build gettext into the official Linkerd docker image. (I understand this is less important since ECS users probably have to build their own images anyway in order to include configs, but it's probably still more convenient for it to be in the base image.)
2. Add a Getting Started with ECS page to https://linkerd.io/getting-started/ that includes a sample templated linkerd config and a sample task definition or launch configuration user data script that fetches from the AWS API and uses envsubst.

For what it's worth, we do a somewhat similar thing for Kubernetes by setting the node ip into an environment variable before starting Linkerd.  

Let me know what you think.

kevholditch:
@adleong thanks.  I agree with both of those points.  We could take it one stage further and as well as building in gettext build in envsubst too.  So we could allow the user of the container to pass in the config as an environment variable and we could have a flag in the container that causes it to run env substitution over the config. e.g. a sample `entrypoint.sh` would be:
```
# optionally allow passing of config via env variable
if [ -n "$LINKERD_CONFIG" ]; then
    echo "$LINKERD_CONFIG" > "/opt/linkerd/conf/linkerd.conf"
fi

# optionally replace env vars in config file
if [ "$ENV_SUBST" = "true" ]; then
    envsubst < /opt/linkerd/conf/linkerd.conf > /opt/linkerd/conf/linkerd.conf
fi

./bundle-exec /opt/linkerd/conf/linkerd.conf 
```

This would make the container a lot more flexible.  In our case we would still have to derive our own container to set the ip variable but I think a lot of people would find the above useful.

I'm happy to help with this in any way I can.

siggy:
Hi @kevholditch. Thanks for submitting this PR, it's helped us a ton in understanding what ECS integration looks like. After a bit of discussion, I think we'd like to try to support ECS via configs and examples, minimizing changes inside linkerd, and avoiding custom Docker containers, if possible. To that end, I just put up a pull request that demonstrates ECS+linked+Consul, using `io.l5d.specificHost`:
https://github.com/linkerd/linkerd-examples/pull/169

I'd love your feedback on this, particularly because your PR and blog post had significant influence on these examples (We've credited your blog at the bottom of the README).

kevholditch:
@siggy no worries.  Understand your decision to cover this with examples.  Have looked through your example and looks good to me.

Issue #1430, opened 2017-06-21
Support k8s services of type: ExternalName


esbie:
Istio uses ExternalName services to [direct egress traffic](https://istio.io/docs/tasks/egress.html#configuring-the-external-services). These services don't appear in the SDS (although they DO show up in RDS), so we need another way to handle these in the IstioInterpreter or IstioNamer. 

Tangentially, it looks like the EndpointsNamer doesn't handle externalName correctly either.

esbie:
More investigation on this today:

The services show up in RDS with routes rewritten/redirected to egress cluster. From there I *think* the current the egress cluster routes to the FQDN, which kubedns has a CNAME for that points outside the cluster.

There's no way to get `externalName` from the istio pilot apis, so the only options I see are (1) Talk to the k8s api directly to get this information or (2) blindly rely on kubedns CNAME w/o knowing if the route is external or not. Or (3), whatever alex comes up with :)

adleong:
Do the external clusters exist in SDS?

esbie:
No, they don't... so currently the identifier returns `/svc/dest/externalbin.default.svc.cluster.local/::/http`  but `/#/io.l5d.k8s.istio/reviews.default.svc.cluster.local/::/http` fails to resolve and returns "No hosts"

adleong:
I think (1) and (2) are both valid.  For (1) we would probably have to write an ExternalNameNamer or something like that....

esbie:
Deprioritizing this feature until after mvp. Istio users can already reach external apis w/o external name services

PR #1455, opened 2017-06-29
Fallback to egress cluster for services w/ no cluster

esbie:
Problem:
K8s services of type ExternalName have entries in RDS but not in CDS,
causing negative nametree resolution.

Solution:
Send traffic to egress cluster as a fallback

Verification:
Minikube manual testing

### DIFF
--- a/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
+++ b/interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
@@ -2,14 +2,17 @@ package io.buoyant.interpreter.k8s
 
 import com.twitter.finagle._
 import com.twitter.finagle.naming.NameInterpreter
+import io.buoyant.k8s.SingleNsNamer
 import io.buoyant.k8s.istio.RouteCache
 import io.buoyant.namer.ConfiguredDtabNamer
 import istio.proxy.v1.config.RouteRule
 
 object IstioInterpreter {
   private val istioPfx = "/#/io.l5d.k8s.istio"
+  private val k8sPfx = "/#/io.l5d.k8s.ns"
   private val defaultRouteDtab = Dtab.read(s"""
     |/svc/ext => /$$/inet ;
+    |/svc/dest => $k8sPfx/incoming/istio-egress ;
     |/svc/dest => $istioPfx ;
   """.stripMargin)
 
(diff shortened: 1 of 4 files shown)


interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/IstioInterpreter.scala
```diff
@@ -2,14 +2,17 @@ package io.buoyant.interpreter.k8s
...
+import io.buoyant.k8s.SingleNsNamer
 import io.buoyant.k8s.istio.RouteCache
 import io.buoyant.namer.ConfiguredDtabNamer
 import istio.proxy.v1.config.RouteRule
 
 object IstioInterpreter {
   private val istioPfx = "/#/io.l5d.k8s.istio"
+  private val k8sPfx = "/#/io.l5d.k8s.ns"
   private val defaultRouteDtab = Dtab.read(s"""
     |/svc/ext => /$$/inet ;
```
adleong: based on my updated understanding of egress, I think we should delete this dentry.  `/svc/ext` names are generated when the host header doesn't match any clusters (even "external" ones).  These requests should fail.

esbie: I agree that's how istio currently operates. I think it's sort of a point of contention though, I've seen a few people in the istio/k8s slacks challenging the design choice of only allowing external traffic via services. Maybe we should make it optional. (I don't have a strong opinion, but wanted to bring it up for discussion)

esbie: In person discussion:

Going to leave this as is until we find a good solution for https://github.com/linkerd/linkerd/issues/1430

adleong: I think our updated logic should be:

* If matching route exists: /svc/route (send to target service)
* If cluster exists but no matching route: /svc/dest (send to target service; fallback to istio-egress)
* If cluster does not exist: /svc/ext (send to istio-egress)

adleong: @esbie do you think we should change this line to send /svc/ext names to the egress linkerd?  I think, as is, this will send directly from the daemonset to the outside world.

esbie: good point, will fix

PR #1458, opened 2017-06-30
Docs update and Pointing all egress traffic to istio-egress

esbie:
followup to https://github.com/linkerd/linkerd/pull/1455

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/istio/package.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/istio/package.scala
@@ -1,9 +1,9 @@
 package io.buoyant.k8s
 
 package object istio {
-  val DefaultDiscoveryHost = "istio-manager"
+  val DefaultDiscoveryHost = "istio-pilot"
   val DefaultDiscoveryPort = 8080
 
-  val DefaultApiserverHost = "istio-manager"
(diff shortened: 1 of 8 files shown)


k8s/src/main/scala/io/buoyant/k8s/istio/package.scala
```diff
@@ -1,9 +1,9 @@
 package io.buoyant.k8s
 
 package object istio {
-  val DefaultDiscoveryHost = "istio-manager"
+  val DefaultDiscoveryHost = "istio-pilot"
```
adleong: Isn't still called the manager as of the 1.6 release?  And should we use the fully qualified name in case linkerd is deployed in a different namespace from istio?

esbie: The 0.1.6 I downloaded has it as istio-pilot.
I had the exact opposite fear: that someone would deploy istio+linkerd in a namespace other than `default` and be confused why it doesn't work.

adleong: Your fear is probably more accurate than my fear.

esbie: Fear Driven Development

PR #1461, opened 2017-07-01
Transparent proxy injection

### LINKED ISSUE #1415: Transparent proxying with iptables
Create an init-container that uses iptables to transparently proxy all outgoing traffic to the node-local daemonset linkerd.

rmars:
Transparently proxy all outgoing requests to a daemonset linkerd via iptables rules (supplied by an init container).

This only looks big because of /vendor. Relevant code is in https://github.com/linkerd/linkerd/commit/ee778d59fb9773401f1a632154557c8c92caffe5.

This is based on istioctl's [proxy injection](https://github.com/istio/pilot/blob/master/doc/proxy-injection.md) (they run a sidecar proxy and do transparent proxying via iptables rules).

- Adds a Dockerfile to create an init-container image that we can use to apply iptables rules to the pod. These rules forward outgoing pod traffic to a daemonset linkerd
- Add prepare_proxy.sh which sets up these rules
- Add an inject.go which will add the initContainer to the user's resource file 

All this is based heavily on istio's existing code and ideally this would go with that code somewhere, but I think it is being moved around right now. Would be nice to delete our inject.go in favour of something in that repo after it has settled.

Testing:
Ran inject.go against yamls in https://github.com/istio/pilot/tree/master/platform/kube/testdata.
Used our hello world example, ran it in minikube and in a 3 node GKE cluster.

Notes:
* When the services call each other, the service names need to be resolvable by kubedns so that the packet has an IP and can be routed (hence renaming world-v1 to world).
* I ran into a go dependency mess, see https://github.com/kubernetes/client-go/blob/master/INSTALL.md

Fixes #1415


esbie:
looks good! but I feel pretty strongly that this needs to be in its own repo (similar to namerctl) (or possibly in linkerd-examples)

istio/src/main/inject/docker/prepare_proxy.sh
```diff
@@ -0,0 +1,44 @@
...
+  echo "Please set linkerd port -p"
+  usage
+  exit 1
+fi
+
+# Don't forward local traffic
+iptables -t nat -A OUTPUT -d 127.0.0.1/32 -j RETURN                                       -m comment --comment "istio/bypass-explicit-loopback"
+
+# Forward traffic to the daemonset linkerd router
+iptables -t nat -A OUTPUT -p tcp -j DNAT --to ${L5D_SERVICE_HOST}:${LINKERD_PORT}         -m comment --comment "istio/dnat-to-daemonset-l5d"
```
adleong: I think L5D_SERVICE_HOST is the vip of the L5D kubernetes service which balances over all linkerd pods.  I think you want to get the node name from the downward API and use that.

rmars:
Moved to https://github.com/linkerd/linkerd-inject/pull/1

PR #1463, opened 2017-07-03
Add HTTP Router Loggers

siggy:
Problem
Linkerd lacks the ability to inject arbitrary loggers into the request
stack.

Solution
Add a new `routers` config section for HTTP `loggers`.

Validation
Coming as part of #1435

precursor to #1435

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpLoggerConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpLoggerConfig.scala
@@ -1,0 +1,56 @@
+package io.buoyant.linkerd.protocol
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.{Filter, Service, ServiceFactory, Stack, Stackable}
+import com.twitter.finagle.http.{Request, Response}
+import com.twitter.finagle.stack.nilStack
+import com.twitter.util.Future
+import io.buoyant.config.PolymorphicConfig
+
+abstract class HttpLoggerConfig extends PolymorphicConfig { config =>
+
+  @JsonIgnore
+  def role: Stack.Role
+  @JsonIgnore
+  def description: String
+  @JsonIgnore
+  def parameters: Seq[Stack.Param[_]]
+
+  @JsonIgnore
+  def mk(params: Stack.Params): Filter[Request, Response, Request, Response]
+
+  @JsonIgnore
+  def module = new Stack.Module[ServiceFactory[Request, Response]] {
+    override def role: Stack.Role = config.role
+    override def description: String = config.description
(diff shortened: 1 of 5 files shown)


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpLoggerConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.buoyant.config.PolymorphicConfig
+
+abstract class HttpLoggerConfig extends PolymorphicConfig {
+  @JsonIgnore
+  def mk(params: Stack.Params): Filter[Request, Response, Request, Response]
+}
+
+object HttpLoggerConfig {
+  object param {
+    case class Logger(logger: (Stack.Params) => Filter[Request, Response, Request, Response])
```
adleong: Yes, you're right, what you suggest is much better.

siggy: Spent some time with @adleong trying to implement `override def router(params: Stack.Params): Router`. At this point in the code we don't have access to the stack without plumbing it through a few layers. Is this change warranted? Are there better approaches?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpLoggerConfig.scala
```diff
@@ -0,0 +1,57 @@
...
+  def role: Stack.Role
+  @JsonIgnore
+  def description: String
+  @JsonIgnore
+  def parameters: Seq[Stack.Param[_]]
+
+  @JsonIgnore
+  def mk(params: Stack.Params): Filter[Request, Response, Request, Response]
+
+  def module = new Stack.Module[ServiceFactory[Request, Response]] {
```
olix0r: doesnt' this need a `@JsonIgnore`?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpLoggerConfig.scala
```diff
@@ -0,0 +1,57 @@
...
+    ): Stack[ServiceFactory[Request, Response]] = {
+      val filter = mk(params)
+      Stack.Leaf(role, filter.andThen(next.make(params)))
+    }
+  }
+}
+
+object HttpLoggerConfig {
+  val LoggerNilRole = Stack.Role("LoggerNil")
+  val LoggerNil = Stack.Leaf[ServiceFactory[Request, Response]](LoggerNilRole, null)
```
adleong: Not a big deal but I realized after the fact that `com.twitter.finagle.stack.nilStack` exists and we could probably use that instead.

Issue #1419, opened 2017-06-20
Support redirect and rewrite route-rules


esbie:
Istio allows users to specify `redirect` and `rewrite` in [istio routing-rules](https://istio.io/docs/reference/config/traffic-rules/routing-rules.html#httpredirect). Linkerd should honor those configurations.

PR #1467, opened 2017-07-05
Support istio rewrite routing-rules

esbie:
Implements https://istio.io/docs/reference/config/traffic-rules/routing-rules.html#httprewrite
Related to #1419 

Also updates header matching to accommodate special keywords ("uri", "scheme", "method", etc) as described in https://istio.io/docs/reference/config/traffic-rules/routing-rules.html#matchcondition

Question: how do I get http/https scheme from finagle http requests?

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
@@ -5,6 +5,16 @@ import istio.proxy.v1.config.{HTTPRewrite, MatchCondition, RouteRule, StringMatc
 import istio.proxy.v1.config.StringMatch.OneofMatchType
 
 trait IdentifierPreconditions {
+  case class IstioRequestMeta(
+    uri: String,
+    scheme: String,
+    method: String,
+    authority: String,
+    getHeader: (String) => Option[String]
+  )
+
+  def pathFromUri(uri: String): String = "/" + uri.split("/").drop(1).mkString("/")
+
   def headerMatches(headerValue: String, stringMatch: StringMatch): Boolean = {
     stringMatch.`matchType` match {
       case Some(OneofMatchType.Exact(value)) => headerValue == value
@@ -23,12 +33,15 @@ trait IdentifierPreconditions {
     }
   }
 
-  def matchesAllConditions(getHeader: (String) => Option[String], matchCondition: MatchCondition): Boolean = {
+  def matchesAllConditions(req: IstioRequestMeta, matchCondition: MatchCondition): Boolean = {
     val matchesHeaders = matchCondition.`httpHeaders`.forall {
+      case ("uri", m) => headerMatches(req.uri, m)
+      case ("scheme", m) => headerMatches(req.scheme, m)
+      case ("method", m) => headerMatches(req.method, m)
+      case ("authority", m) => headerMatches(req.authority, m)
       case (headerName, stringMatch) =>
-        // return false if the headerName does not appear in the request's headerMap
-        getHeader(headerName) match {
-          case Some(headerValue) => headerMatches(headerValue, stringMatch)
+        req.getHeader(headerName) match {
+          case Some(a) => headerMatches(a, stringMatch)
           case None => false
         }
     }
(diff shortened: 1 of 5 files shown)


k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
```diff
@@ -23,32 +33,51 @@ trait IdentifierPreconditions {
...
   }
 
-  def filterRules(rules: Map[String, RouteRule], dest: String, getHeader: (String) => Option[String]): Seq[(String, RouteRule)] = rules.filter {
+  def httpRewrite(rule: RouteRule, uri: String, authority: Option[String]): (String, Option[String]) = {
+    rule.`rewrite` match {
+      case Some(HTTPRewrite(url, updatedAuth)) =>
+        val updatedUri = url.map { replacement =>
+          rule.`match`.flatMap(_.`httpHeaders`.get("uri").flatMap(_.`matchType`)) match {
+            case Some(OneofMatchType.Prefix(pfx)) =>
+              uri.replace(pfx, replacement)
```
adleong: should we validate that uri starts with pfx?

k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
```diff
@@ -23,32 +33,51 @@ trait IdentifierPreconditions {
...
+            case _ =>
+              uri.split("/").head + replacement
```
adleong: This is probably similar to pathFromUri but I don't understand it.

k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
```diff
@@ -23,32 +33,51 @@ trait IdentifierPreconditions {
...
+  def filterRules(rules: Map[String, RouteRule], dest: String, req: IstioRequestMeta): Seq[(String, RouteRule)] = rules.filter {
     case (_, r) if r.`destination`.contains(dest) =>
       // return true if no match conditions were defined on the route-rule
-      r.`match`.forall(matchesAllConditions(getHeader, _))
+      r.`match`.forall(matchesAllConditions(req, _))
     case _ => false
   }.toSeq
 
-  def maxPrecedenceRuleName(rules: Seq[(String, RouteRule)]): Option[String] = {
+  def maxPrecedenceRuleName(rules: Seq[(String, RouteRule)]): Option[(String, RouteRule)] = {
```
adleong: We should probably change the name of this method now that it's no longer just returning the name.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
```diff
@@ -19,9 +19,15 @@ class IstioIdentifier(val pfx: Path, baseDtab: () => Dtab, routeCache: RouteCach
       case Some(host) =>
         Future.join(clusterCache.get(host), routeCache.getRules).map {
           case (Some(Cluster(dest, port)), rules) =>
-            val filteredRules = filterRules(rules, dest, req.headerMap.get)
+            //TODO: request scheme?
```
adleong: I don't think this information is available on the HTTP request (unless it's in the request.uri).  We may have to punt on this.

k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
```diff
@@ -23,32 +33,51 @@ trait IdentifierPreconditions {
     }
   }
 
-  def matchesAllConditions(getHeader: (String) => Option[String], matchCondition: MatchCondition): Boolean = {
+  def matchesAllConditions(req: IstioRequestMeta, matchCondition: MatchCondition): Boolean = {
```
esbie: For sure, this is not very clear as is

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/IstioIdentifier.scala
esbie: ok np thanks for verifying

k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
```diff
@@ -1,10 +1,20 @@
...
 trait IdentifierPreconditions {
+  case class IstioRequestMeta(
+    uri: String,
+    scheme: String,
+    method: String,
+    authority: String,
+    getHeader: (String) => Option[String]
+  )
+
+  def pathFromUri(uri: String): String = "/" + uri.split("/").drop(1).mkString("/")
```
esbie: Actually, I don't think this is needed anymore. Originally I thought I'd need to separate out "/my/path" from "foo.bar.com/my/path".

PR #1471, opened 2017-07-06
Support istio rewrite routing-rules

esbie:
https://github.com/linkerd/linkerd/pull/1467 redux

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
@@ -5,6 +5,11 @@ import istio.proxy.v1.config.{HTTPRewrite, MatchCondition, RouteRule, StringMatc
 import istio.proxy.v1.config.StringMatch.OneofMatchType
 
 trait IdentifierPreconditions {
+
+  /**
+   * Defines a request's metadata for istio rules to match against
+   * (normalizes fields between http and h2)
+   */
   case class IstioRequestMeta(
     uri: String,
     scheme: String,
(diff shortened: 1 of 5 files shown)


k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
```diff
@@ -23,32 +36,57 @@ trait IdentifierPreconditions {
...
+   * @param uri
+   * @param authority
+   * @return Tuple of uri and authority to used to rewrite headers on the request
+   */
+  def httpRewrite(rule: RouteRule, uri: String, authority: Option[String]): (String, Option[String]) = {
+    rule.`rewrite` match {
+      case Some(HTTPRewrite(url, updatedAuth)) =>
+        val updatedUri = url.map { replacement =>
+          rule.`match`.flatMap(_.`httpHeaders`.get("uri").flatMap(_.`matchType`)) match {
+            case Some(OneofMatchType.Prefix(pfx)) if uri.startsWith(pfx) => uri.replace(pfx, replacement)
```
adleong: Replace will replace all occurrences, I think we just want to replace the first (ie the prefix):
```
scala> "hotshots".replace("hots", "nots")
res1: String = notsnots

scala> "hotshots".replaceFirst("hots", "nots")
res2: String = notshots
```

k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
```diff
@@ -23,32 +36,57 @@ trait IdentifierPreconditions {
...
+            case _ => replacement
```
adleong: Not sure if the istio spec spells out what this behavior should be if the match type is prefix but the prefix doesn't match.  I would expect the uri to be left unmodified, rather than replaced.

esbie: Ok I can make that more explicit. This is a bit overkill imo since we wouldn't be doing a replace without all matchConditions being met in the first place. Of course that ordering isn't obvious here..

k8s/src/main/scala/io/buoyant/k8s/istio/IdentifierPreconditions.scala
esbie: thanks!

Issue #682, opened 2016-09-19
Detect bad framing


olix0r:
From https://www.mnot.net/blog/2011/07/11/what_proxies_must_do

> Proxies also need to be on the lookout for Content-Length headers that are duplicates, as well as ones that conflict with the use of Transfer-Encoding, and either reject the message or remove the bad headers.
>
> This is because there are [entire classes of attacks](http://projects.webappsec.org/w/page/13246931/HTTP-Response-Splitting) that exploit the differences between how implementations frame messages.
>
> For example, this response:
> ```
> HTTP/1.1 200 OK
> Content-Type: text/html; charset=utf-8
> Content-Length: 45
> Content-Length: 20
> ```
> has an ambiguous length. If a proxy treats it differently than a client, an attacker can inject a response. Likewise, this one:
> ```
> HTTP/1.1 200 OK
> Content-Type: text/html; charset=utf-8
> Content-Length: 200
> Transfer-Encoding: chunked
> ```
> has both a Content-Length and chunked encoding. The chunked encoding has precedence, and the Content-Length has to be removed before forwarding the message.
> 
> See [the spec](http://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-14#section-3.3) for how to do it well.

olix0r:
I suspect that finagle does most of this for us, but we should verify.

hawkw:
I'm working on a [tool to automate this](https://github.com/BuoyantIO/flossy), and so far, I've found at least one case we're not handling correctly.

According to the spec:
>  If a message is received without Transfer-Encoding and with
> either multiple Content-Length header fields having differing
> field-values or a single Content-Length header field having an
> invalid value, then the message framing is invalid and MUST be
> treated as an error to prevent request or response smuggling. 
...
> If this is a response message received by a proxy, the proxy MUST 
> discard the received response, send a 502 (Bad Gateway) status 
> code as its downstream response, and then close the connection.

We are, as you might have guessed, Not Doing This.

Testing with `flossy`, where the upstream server sends first bad response message in the blog post @olix0r quoted above, linkerd happily forwards the bad response, multiple headers intact. 

(N.B. for potential users of `flossy` – it's currently a minimum working proof-of-concept. you probably don't want to use this yet)

hawkw:
I did a manual `telnet` session to confirm (I don't entirely trust `flossy` yet; there's probably several lurking bugs in there):

```
[...]

Issue #1468, opened 2017-07-05
Correct handling of HTTP responses containing multiple `Content-Length` headers


hawkw:
As seen in https://github.com/linkerd/linkerd/issues/682#issuecomment-312730463. 

It looks like we're mishandling badly-framed HTTP 1.1 responses. According to the spec, proxies are supposed to return 502 errors when the upstream sends a response with multiple conflicting `Content-Length` headers. However, linkerd currently just happily forwards these responses, all headers intact.

hawkw:
While I'd eventually like to look at fixing this behavior downstack from us in Finagle, I think for now the Right Thing to Do is just to add a filter that turns these requests into 502 errors. @olix0r, does that sound good to you?

olix0r:
It's fine to add a linkerd-specific solution for the time being.

You'll likely want to write client and server stack modules that install the appropriate filtering logic to fail requests with multiple content-lengths on the server and fail responses with invalid content-lengths on the client.

The tricky part here is figuring out where to install these filters so that we measure success/failures properly.

hawkw:
Closed by #1475

PR #1475, opened 2017-07-07
Add a filter for duplicate `Content-Length` headers in requests

hawkw:
Problem:
---------

The HTTP 1.1 specification for proxies says the following: 
> If a message is received without Transfer-Encoding and with
> either multiple Content-Length header fields having differing
> field-values or a single Content-Length header field having an
> invalid value, then the message framing is invalid and MUST be
> treated as an error to prevent request or response smuggling.

Testing with [`flossy`](https://github.com/BuoyantIO/flossy) indicates that we don't handle this case correctly – we just forward these requests to the downstream service (see https://github.com/linkerd/linkerd/issues/682#issuecomment-312730463).

Solution
--------

I added a `RequestFramingFilter` to the Path stack in `HttpConfig`. This filter detects requests with duplicate `Content-Length` headers and responds with error code 400. The filter is inserted into the stack above `ClassifiedRetries` so that we don't waste time retrying these malformed requests.

In the future, this filter can be expanded to handle more types of bad request framing (such as the case of conflicting `Transfer-Encoding` and `Content-Length` headers, which I _strongly suspect_ we are also not handling correctly).

Validation
----------

I reran `flossy` against the linkerd proxy example, and the corresponding test now passes.

Fixes #1475.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -31,6 +31,7 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .prepend(Headers.Dst.PathFilter.module)
       .replace(StackClient.Role.prepFactory, DelayedRelease.module)
       .prepend(http.ErrorResponder.module)
+      .insertBefore(ClassifiedRetries.role, RequestFramingFilter.module)
     val boundStack = Http.router.boundStack
       .prepend(Headers.Dst.BoundFilter.module)
     val clientStack = Http.router.clientStack
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -31,7 +31,7 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .prepend(Headers.Dst.PathFilter.module)
       .replace(StackClient.Role.prepFactory, DelayedRelease.module)
       .prepend(http.ErrorResponder.module)
-      .insertBefore(ClassifiedRetries.role, RequestFramingFilter.module)
+      .insertAfter(http.ErrorResponder.role, RequestFramingFilter.module)
     val boundStack = Http.router.boundStack
       .prepend(Headers.Dst.BoundFilter.module)
     val clientStack = Http.router.clientStack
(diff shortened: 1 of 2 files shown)


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -31,6 +31,7 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .prepend(Headers.Dst.PathFilter.module)
       .replace(StackClient.Role.prepFactory, DelayedRelease.module)
       .prepend(http.ErrorResponder.module)
+      .insertBefore(ClassifiedRetries.role, RequestFramingFilter.module)
```
adleong: I think this can move even higher in the path stack.  Since the request is invalid, we want to reject it as quickly as possible so that we don't waste resources processing it.  I would say you could just prepend it here to put it at the top of the stack.  Or maybe right below ErrorResponder just in case this filter ever throws an exception (even though it shouldn't).

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RequestFramingFilter.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.twitter.finagle.http.{Request, Response, Status}
+import com.twitter.finagle.buoyant.linkerd.Headers
+import com.twitter.logging.Logger
+import com.twitter.util.Future
+
+/**
+ * A filter that fails badly-framed requests.
+ */
+class RequestFramingFilter extends SimpleFilter[Request, Response] {
+  private[this] lazy val log = Logger.get("RequestFramingFilter")
```
adleong: Doing upfront initialization of the Logger is probably fine and lets us avoid the evil `lazy` keyword.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RequestFramingFilter.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  override def apply(
+    request: Request,
+    service: Service[Request, Response]
+  ): Future[Response] =
+    if (request.headerMap.getAll("Content-Length").length > 1) {
```
adleong: I think this will fail requests with multiple content length headers that all have the same value.  If I'm reading the spec right, I think that should be valid.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/RequestFramingFilter.scala
```diff
@@ -0,0 +1,46 @@
...
+      // if the length of the Content-Length key in the request's header map
+      // is greater than 1, then there are duplicate values.
+      log.error("request with duplicate Content-Length headers", request)
+      val resp = Headers.Err.respond(
+        "Request contained duplicate `Content-Length` headers",
+        Status.BadRequest
+      )
+      Future(resp)
```
adleong: Prefer `Future.value(resp)`

hawkw:
@adleong I thought I'd handle badly-framed responses in a separate PR, since they have a different issue (#1468). Does that seem appropriate to you?

PR #1478, opened 2017-07-07
Add a filter for duplicate `Content-Length` headers in responses

hawkw:
This is essentially just the upstream version of #1475. Per the HTTP 1.1 spec:
> If a message is received without Transfer-Encoding and with
> either multiple Content-Length header fields having differing
> field-values or a single Content-Length header field having an
> invalid value, then the message framing is invalid and MUST be
> treated as an error to prevent request or response smuggling.
> [...]
> If this is a response message received by a proxy, the proxy MUST
> discard the received response, send a 502 (Bad Gateway) status
> code as its downstream response, and then close the connection.

Testing with [`flossy`](https://github.com/BuoyantIO/flossy) indicates that we don't handle this case correctly – we just forward these responses  downstream, with the invalid headers intact (see https://github.com/linkerd/linkerd/issues/682).

I added a `RespomnseFramingFilter` to the Path stack in `HttpConfig`. This filter detects requests with duplicate `Content-Length` headers and responds with error code 400. The filter is inserted into the stack below the `ErrorResponder` so we can just handle the error there.

In the future, this filter can be expanded to handle more types of bad response framing (such as the case of conflicting `Transfer-Encoding` and `Content-Length` headers, which I _strongly suspect_ we are also not handling correctly).

I reran `flossy` against the linkerd proxy example, and the corresponding test now passes.

Fixes #1478.


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseFramingFilter.scala
```diff
@@ -0,0 +1,44 @@
...
+  override def apply(
+    request: Request,
+    service: Service[Request, Response]
+  ): Future[Response] =
+    service(request) flatMap { response =>
+      val numHeaders = response.headerMap.getAll("Content-Length").toSet.size
+      if (numHeaders > 1) {
+        // if the response has multiple Content-Length headers with unique
+        // values, it's invalid
+        Future.exception(FramingException(
```
adleong: Thinking about it some more... we might actually want to return a 503 response directly here instead of returning an exception.  That way the classified retries module can decide if the 503 is retryable or not (based on the request verb, for example) rather than us returning an exception that we must explicitly mark as retryable or not.  What do you think, @olix0r?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseFramingFilter.scala
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * A filter that fails badly-framed responses
+ */
+class ResponseFramingFilter extends SimpleFilter[Request, Response] {
+  override def apply(
+    request: Request,
+    service: Service[Request, Response]
+  ): Future[Response] =
+    service(request) flatMap { response =>
```
hawkw: Whoops, that was IDEA's fault; it looks like the default config in my new install automatically reformats code on commit. Gotta turn that off!

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -31,6 +31,7 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .prepend(Headers.Dst.PathFilter.module)
       .replace(StackClient.Role.prepFactory, DelayedRelease.module)
       .prepend(http.ErrorResponder.module)
+      .insertAfter(http.ErrorResponder.role, ResponseFramingFilter.module)
```
hawkw: @olix0r, Alex & I discussed this, and we weren't sure if this ought to go above or below the failure accrual module in the client stack. What do you think?

olix0r: I think these errors should probably count against failure accrual...  We'd want a load balancer to evict nodes sending these sorts of malformed responses.

hawkw: 👍 yeah, that makes sense...

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseFramingFilter.scala
hawkw: I asked Oliver about it & he says we probably should just add the `FramingException` to the exceptions that the classifier knows about, so I can do that.

adleong: I don't think that works because we want the configured response classifier to make the determination of if the request is retryable.  For example, a GET that results in malframed response might be retryable but a POST that results in a malframed response might not be.

hawkw: Hmm...so, should I just make this return a 503, or wait for @olix0r to see this?

olix0r: I think we should update our classifiers to be aware of this type of error.

How about we do this:
- Introduce a ProtocolError abstract type (in a protocol-agnostic module);
- Update existing classifiers to treat protocol errors like 5xx;
- Make the malformed framing filters throw a type that implements ProtocolError.

I think it's correct to treat this as an exception that should be handled in an error handler: in the router's stack, a 5xx from the remote service should be distinct from linkerd deciding it won't handle the response.

hawkw: @olix0r Okay, that sounds like a plan. While I'm at it, do you think there are other places we ought to refactor to use a new `ProtocolError` type?

hawkw: (I'm calling it `ProtocolException` rather than `ProtocolError`, since IIRC the Java/Scala convention is for `Error`s to be fatal errors, while `Exception`s are recoverable)

hawkw: @olix0r, what do you think about the current iteration of this (824c06ea189b93198feaef36940c521b05487b7)?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -19,6 +20,7 @@ import io.buoyant.linkerd.protocol.http._
 import io.buoyant.router.{ClassifiedRetries, Http, RoutingFactory}
 import io.buoyant.router.RoutingFactory.{IdentifiedRequest, RequestIdentification, UnidentifiedRequest}
 import io.buoyant.router.http.AddForwardedHeader
+
```
adleong: rm blank line please

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ErrorResponderTest.scala
```diff
@@ -10,22 +10,24 @@ import io.buoyant.router.RoutingFactory
...
+import io.buoyant.linkerd.protocol.http.FramingFilter.FramingException
 import org.scalatest.FunSuite
 
 class ErrorResponderTest extends FunSuite with Awaits {
 
   val svc = Service.mk[Request, Response] { _ =>
     Future.exception(RoutingFactory.UnknownDst(Request(), s"foo\nbar"))
   }
-  val stk = ErrorResponder.module.toStack(
+  val stk = ErrorResponder.serverModule.toStack(
```
hawkw: I'd like to rewrite these tests to run with both the `clientModule` and the `serverModule`, but I don't think it's terribly important.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
hawkw: fixed 1bc8676a02a62f1db3187c835ddbe0ada0386c73. And you'll be glad to know I found the IntelliJ pref to turn that off

hawkw:
@adleong I'm not sure why there's an `ErrorResponder` in both stacks, that was there when I got here :)

I think even if we continue having two filters, they could still throw different subclasses of `MalframedWhateverException` and encode the response status that way, rather than based on the error responder's position in that stack – I'd be happy to make that change. @olix0r, does that seem reasonable?

Re: having a single filter, it would definitely be simpler that way; IIRC Oliver suggested I do it this way. I'm open to suggestions.

olix0r:
There's an error responder on the _server stack_ only for convenience (so that errors on the server are handled).  It's conceptually correct for the the error responder to be at the top of the router stack -- exceptions communicating with the client are turned into errors, which the server just transmits.  This way we get a clear indication of where exceptions are being thrown (in the server o rin the client)

PR #1484, opened 2017-07-10
Add a reference to the discourse instance to the linkerd help page

stevej:


linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/HelpPageHandler.scala
```diff
@@ -30,6 +30,7 @@ object HelpPageHandler {
             If you still need help, you can reach us in a couple ways:
           </p>
           <ul>
+              <li>Ask questions about configuration or troubleshooting in the <a href="https://discourse.linkerd.io/">linkerd Discourse instance</a>.</li>
               <li>Chat with us in <a href="http://slack.linkerd.io">the linkerd public Slack</a>.</li>
               <li>Email us at <a href="mailto:help@buoyant.io">help@buoyant.io</a>.</li>
```
klingerf: Would you mind also changing this email address while you're here? I think we should point it to `linkerd-users@googlegroups.com`.

stevej: 👍

PR #1489, opened 2017-07-11
Add Transfer-Encoding to FramingFilter

### LINKED ISSUE #1479: Correct handling of conflicting chunked transfer encoding and content-length headers in requests
Yet more work on #682!

BuoyantIO/flossy@57c500485fd7fb7a55cfebad728b9eb07774fa60 adds a test for requests containing both `Content-Length` and `Transfer-Encoding: chunked` headers. The HTTP 1.1 spec says proxies should handle this by obeying the `Transfer-Encoding` header and removing the `Content-Length` header. We are currently failing to remove that header.

hawkw:
According to the HTTP/1.1 spec, when a proxy receives a message containing both `Content-Length` and `Transfer-Encoding: chunked` headers, the chunked encoding is supposed to take priority, and any `Content-Length` headers should be removed.

I've enhanced the `FramingFilter` I added in #1478 to support this behavior as well. In the process, I also ended up rewriting a lot of the internal implementation of `FramingFilter`; I think the new code is conceptually simpler and (hopefully!) more elegant.

`flossy` currently contains one test for the case of conflicting `Content-Length` and `Transfer-Encoding: chunked` headers, which passes after making these changes. I'd like to add more comprehensive tests for this behavior to `flossy`, though.

Fixes #1479. Closes #1478.


hawkw:
This PR extends on and obsoletes my existing PR #1478, so if we merge this, we should close that PR. Is this the correct convention for obsoleting a PR, and should I add the summary of changes in that PR to this one, so that it shows up in the commit history?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -42,6 +42,9 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .insertAfter(DtabStatsFilter.role, HttpLoggerConfig.module)
       .insertAfter(Retries.Role, http.StatusCodeStatsFilter.module)
       .insertAfter(AddrMetadataExtraction.Role, RewriteHostHeader.module)
+      // ensure the client-stack framing filter is placed below the stats filter
+      // so that any malframed responses it fails are counted as errors
+      .insertAfter(StatusCodeStatsFilter.role, FramingFilter.clientModule)
```
adleong: I think this also should be after the RequestStats module so that these malframed responses show up as failures there and also below FailureAccrual so that malframed responses contribute to failure accrual.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -68,6 +71,9 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       .replace(Headers.Ctx.serverModule.role, Headers.Ctx.serverModule)
       .prepend(http.ErrorResponder.module)
       .prepend(http.StatusCodeStatsFilter.module)
+      // ensure the server-stack framing filter is placed below the stats filter
+      // so that any malframed requests it fails are counted as errors
+      .insertAfter(StatusCodeStatsFilter.role, FramingFilter.serverModule)
```
adleong: Do you know if this is above or below the StatsFilter?  We should make sure this is below both the StatusCodeStatsFilter and the StatsFilter

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ErrorResponder.scala
```diff
@@ -5,10 +5,13 @@ import com.twitter.finagle.http.{Request, Response, Status}
 import com.twitter.finagle.service.RetryPolicy.RetryableWriteException
 import com.twitter.finagle._
 import com.twitter.logging.Logger
+import io.buoyant.linkerd.ProtocolException
 import io.buoyant.router.RoutingFactory
+
```
adleong: rm blank line

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/FramingFilter.scala
```diff
@@ -0,0 +1,91 @@
...
+    // exceptions to an ErrorResponder
+    private[this] val log = Logger.get("FramingFilter.ServerFIlter")
+
+    override def apply(
+      request: Request,
+      service: Service[Request, Response]
+    ): Future[Response] =
+      filterMessage(request)
+        .flatMap(service(_))
+        .rescue { case e: FramingException =>
```
adleong: for clarity I'd move this up above the flatMap since this exception can only come from filterMessage.  I'd also switch this to use handle instead of rescue.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/FramingFilter.scala
```diff
@@ -0,0 +1,91 @@
...
+          log.error("framing error", e)
```
adleong: I think it would be helpful for this error message to indicate that it's a Request that is malframed (as opposed to a Response)

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/FramingFilter.scala
```diff
@@ -0,0 +1,91 @@
...
+  // this is factored out so that the same logic can
+  // be applied to requests from clients and responses from services.
+  private[FramingFilter] def filterMessage[M <: Message](message: M): Future[M] = {
+    val headers = message.headerMap
+    val contentLengths = message.headerMap.getAll("Content-Length")
+
+    if (contentLengths.toSet.size > 1) {
+      // if the length of the Content-Length key in the request/response's
+      // header map is greater than 1, then there are duplicate values.
+      Future.exception(FramingException("conflicting `Content-Length` headers"))
```
adleong: This exception is always the same so we can store it in a val to avoid allocating a new one each time.

hawkw: I can do that, but do you think the exception ought to have information specific to the request added to the message, instead?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/FramingFilter.scala
hawkw: As it's currently written, the `flatMap` has to be before the `rescue`/`handle`, since it changes the type from `Request` to `Response` – that's why changing it broke the build in 7dd76a64c7c1246b1243683a76a212322f7eacb4 (sorry).

hawkw: I think I can rewrite this to just use `transform` though.

adleong: ah yeah, good point

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
hawkw: Okay, fixed 73cba51fe755d9072cd68805cb564406a2de6b5b. (Turns out `FailureAccrual` is lower)

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
hawkw: Okay, looks like `StatusCodeStatsFilter` is actually at the top of the stack, and `RequestStats` towards the bottom. Fixed in 33d05c3c3b255f44fb42b05ff119f88afc51bc19

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/FramingFilter.scala
adleong: It's probably not a big deal.  The response that linkerd responds will have a different status code depending on if it's a malframed request or response.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/FramingFilter.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /**
+   * A filter that fails badly-framed requests.
+   */
+  val serverFilter = new SimpleFilter[Request, Response] {
+    // unlike the client-side filter, this needs its own logger, since
+    // it responds to bad requests directly, rather than bubbling up
+    // exceptions to an ErrorResponder
+    private[this] val log = Logger.get("FramingFilter.ServerFIlter")
```
adleong: typo: ServerFIlter

hawkw: agh

hawkw:
@adleong I manually verified the retries behavior locally, and it looks like if 68701cda69cbb44f7aeec56c4cd4bf1e2a0e9277 builds successfully, this should be done! 🎊

Issue #1488, opened 2017-07-11
improve prometheus telemeter memory allocation


siggy:
User reported OOM occurring in PrometheusTelemeter. The current implementation allocates a new `StringBuilder` and then does possibly thousands of `append`'s. For short `scrape_interval`'s, this is a lot of allocations.

## Relevant code

https://github.com/linkerd/linkerd/blob/master/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala#L25

## OOM error

```
java.lang.OutOfMemoryError: Java heap space
    at java.lang.StringCoding$StringEncoder.encode(StringCoding.java:300)
    at java.lang.StringCoding.encode(StringCoding.java:344)
    at java.lang.String.getBytes(String.java:918)
    at com.twitter.io.Buf$StringCoder.apply(Buf.scala:1014)
    at com.twitter.finagle.http.Message.contentString_$eq(Message.scala:391)
    at io.buoyant.telemetry.prometheus.PrometheusTelemeter.$anonfun$handler$1(PrometheusTelemeter.scala:25)
```

related to https://github.com/linkerd/linkerd-viz/issues/32

PR #1490, opened 2017-07-11
Preallocate telemeter StringBuilders

siggy:
Problem
Influx and Prometheus Telemeter StringBuilders initialize with default
16 byte capacity. Typical Telemeter payloads can grow to 1MB or more.
This causes many unneed reallocations.

Solution
Set initial StringBuilder capacity to 16384.

Validation
Unit tests

Part of #1488

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 ## In the next release...
 
 * We fixed an issue where requests that time out were not being retried.
+* Improved memory allocation in InfluxDb and Prometheus telemeters.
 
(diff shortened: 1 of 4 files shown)


adleong:
Makes sense.  How did you come up with 16384 as the initial capacity.  Have you been able to experimentally measure the amount of allocation before and after this change?

siggy:
@adleong Did not experimentally arrive at 16384. @olix0r said "set it at 16k" and I settled on 2^14.

PR #1493, opened 2017-07-11
Linkerd should retry on timeouts

adleong:
When Linkerd is configured with a requestAttemptTimeoutMs and an individual
request attempt violates that timeout, that request is not retried, even if
the request should be retryable.

This is due to the HeaderRetryable response classifier only being defined when
the response is a Return (rather than a Throw) and the fact that timeouts
propagate as exceptions.  We fix this by making HeaderRetryable defined
whenever the underlying ResponseClassifier is defined.

Added an endToEnd test to validate this behavior.


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ErrorResponder.scala
```diff
@@ -23,12 +23,18 @@ class ErrorResponder extends SimpleFilter[Request, Response] {
...
         case _ =>
-          log.error(e, "service failure")
           val message = e.getMessage match {
             case null => e.getClass.getName
             case msg => msg
           }
-          val rsp = Headers.Err.respond(message, Status.BadGateway)
+          val status = e match {
+            case _: TimeoutException | Failure(Some(_: TimeoutException)) =>
+              Status.ServiceUnavailable
```
olix0r: i think, technically, we should be calling getCause on the exception to unwrap it (until getCause returns null).  Finagle does this some places already

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ResponseClassifiers.scala
```diff
@@ -111,12 +111,12 @@ object ResponseClassifiers {
...
         val rc = classifier(rr)
-        if (rc == ResponseClass.NonRetryableFailure && Headers.Retryable.get(rsp.headerMap)) {
-          ResponseClass.RetryableFailure
-        } else {
-          rc
+        rr match {
+          case ReqRep(req, Return(rsp: Response)) if rc == ResponseClass.NonRetryableFailure && Headers.Retryable.get(rsp.headerMap) =>
+            ResponseClass.RetryableFailure
+          case _ => rc
         }
```
olix0r: i'd check rc before destructuring...

```scala
classifier(rr) match {
  ResponseClass.NonRetryableFailure =>
    rr match {
      case ReqRep(req, Return(rsp)) if Headers.Retryable.get(rsp.headerMap) => ResponseClass.RetryableFailure
      case _=> rc
    }
    case _ => rc
}
```

this should be a bit less work in the common case

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/ErrorResponder.scala
adleong: I was following the pattern from here: https://github.com/twitter/finagle/blob/72cac55fecfd13daf00a4cfed4dad11b926baa8d/finagle-core/src/main/scala/com/twitter/finagle/service/RetryPolicy.scala#L169

Taik:
Hey guys, any ETA when this will be released?

siggy:
@Taik Planning to release later today, stay tuned!

PR #1495, opened 2017-07-11
Add import to BetterHttp2ConnectionEncoder

adleong:
For reasons that I cannot fathom, BetterHttp2ConnectionEncoder fails to compile
when publishing artifacts even though it compiles successfully with the
compile task.  The compilation error is
```
[error] /Users/alex/workspace/linkerd/finagle/h2/src/main/java/io/netty/handler/codec/http2/BetterHttp2ConnectionEncoder.java:324: not found: type Configuration
[error]     public Configuration configuration() {
[error]            ^
```

Adding an explicit import of Configuration seems to fix the issue.

### DIFF
--- a/finagle/h2/src/main/java/io/netty/handler/codec/http2/BetterHttp2ConnectionEncoder.java
+++ b/finagle/h2/src/main/java/io/netty/handler/codec/http2/BetterHttp2ConnectionEncoder.java
@@ -28,6 +28,7 @@ import java.util.ArrayDeque;
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_PRIORITY_WEIGHT;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
+import static io.netty.handler.codec.http2.Http2FrameWriter.Configuration;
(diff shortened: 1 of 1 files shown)


finagle/h2/src/main/java/io/netty/handler/codec/http2/BetterHttp2ConnectionEncoder.java
```diff
@@ -28,6 +28,7 @@
 import static io.netty.handler.codec.http2.Http2CodecUtil.DEFAULT_PRIORITY_WEIGHT;
 import static io.netty.handler.codec.http2.Http2Error.PROTOCOL_ERROR;
 import static io.netty.handler.codec.http2.Http2Exception.connectionError;
+import static io.netty.handler.codec.http2.Http2FrameWriter.Configuration;
```
klingerf: I have a slight preference for omitting this import and changing the reference to this class on line 324 to be:

```
    public Http2FrameWriter.Configuration configuration() {
```

That more consistent with how it is referenced on line 83:

```
        Http2FrameWriter.Configuration config = configuration();
```

adleong: Unless you feel strongly I'd prefer to leave this as is, just because it's annoying to test.

PR #1497, opened 2017-07-12
Add TLS client configuration to Marathon namer (#1445)

### LINKED ISSUE #1445: Linkerd on Mesosphere Enterprise DC/OS strict mode does not work
Linkerd is available from the DC/OS universe.
In the DC/OS 1.8 release, authorization was introduced in the Enterprise Edition of DC/OS. Marathon only has https available in strict mode so the linkerd fails to deploy as http is only available.

The community version of DC/OS is unaffected by this as strict mode authorization is unavailable.

For linkerd to deploy successfully on an Enterprise DC/OS strict mode cluster the package requires https capabilities

mtweten:
For DC/OS clusters running in strict security mode, TLS is enforced for any calls to Marathon API endpoints, using certificates signed by the DC/OS Certificate Authority. Without being able to configure the relevant root certificate to be trusted, the Marathon namer cannot talk to the necessary Marathon endpoints. 

This change introduces the existing [TlsClientConfig](https://linkerd.io/config/1.1.1/linkerd/index.html#client-tls) to the Marathon namer configuration to enable TLS secured communication with Marathon, since it already has everything we need. The only caveat is that the clientAuth config section will remain unused, as DC/OS does not support mutual TLS. I've added a simple test to ensure that the TLS config is parsed from the config file correctly.

Fixes #1445


wmorgan:
@mtweten thank you! if you send us your address & size we'll happily send you a t-shirt and some stickers. william at buoyant.io or DM me on the linkerd slack

Issue #1198, opened 2017-04-12
New interpreter that reads dtab from a k8s configmap


klingerf:
Let's add another interpreter, similar to the [file-system interpreter](https://linkerd.io/config/head/linkerd/index.html#file-system), that reads dtab from a Kubernetes ConfigMap. We should be able to setup a watch on the API object in order to pick up any changes.

klingerf:
I had started on this change awhile ago but didn't finish. You can see my branch here:

https://github.com/linkerd/linkerd/compare/kl/configmap-interpreter

I wanted to go back and refactor the changes in k8s/resources.scala, since I felt like my implementation was really messy. Let me know if you have any questions about it.

hawkw:
Great, thanks @klingerf! I'll let you know if I have any trouble.

hawkw:
Implemented in #1603! 🎉

PR #1506, opened 2017-07-15
Add an interpreter that reads Dtabs from a k8s ConfigMap

hawkw:
See issue #1198. @klingerf started on this a while ago, but wasn't able to merge his code in; I've picked it up where he left off. I've refactored a lot of Kevin's code and added some tests.

Note that I've made a number of changes to the watch code in `io.buoyant.k8s`; `ObjectResource`s are now watchable.


hawkw:
I think I may need to add some additional tests for the `ConfigMapInterpreter` before this is ready to merge; if so, I'll probably save that for tomorrow/Monday.

hawkw:
I forgot to mention yesterday but this should be ready for review as of 50f515368b971bcf51a97b00b51ab7667eae638c.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -28,7 +30,12 @@ trait Resource extends Closable {
  * @tparam O The parent type for all [Object]s served by this version.
  */
 private[k8s] trait Version[O <: KubeObject] extends Resource {
-  val path = s"/$group/$version"
+  // alex, I know you don't like lazy-evaluation, but i thought it was
+  // reasonable here, since the behaviour is fairly predictable, and
+  // it's nice to save us a handful of string allocations if we only
+  // want one or the other
+  override lazy val path = s"/$group/$version"
```
adleong: This isn't in the hot path so allocations of a few strings won't matter.  Begone, fowl lazy keyword!

k8s/src/main/scala/io/buoyant/k8s/v1.scala
```diff
@@ -72,17 +72,49 @@ package object v1 {
...
+  implicit private val configMapType = new TypeReference[ConfigMap] {}
   implicit private val endpointsWatch = new TypeReference[EndpointsWatch] {}
   implicit private val serviceWatch = new TypeReference[ServiceWatch] {}
+  implicit private val configMapWatch = new TypeReference[ConfigMapWatch] {}
 
   case class Api(client: Client) extends Version[Object] {
     def group = v1.group
     def version = v1.version
     override def withNamespace(ns: String) = new NsApi(client, ns)
+//    override def withNamespaceWatch(ns: String) = new NsWatchApi(client, ns)
```
adleong: should this be removed?

k8s/src/main/scala/io/buoyant/k8s/v1.scala
```diff
@@ -91,8 +123,14 @@ package object v1 {
     extends NsVersion[Object](client, v1.group, v1.version, ns) {
     def endpoints = listResource[Endpoints, EndpointsWatch, EndpointsList]()
     def services = listResource[Service, ServiceWatch, ServiceList]()
+    def configMap(name: String) = objectResource[ConfigMap, ConfigMapWatch](name)
   }
 
+//  class NsWatchApi(client: Client, ns: String)
```
adleong: should this be removed?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
hawkw: 😢

interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterInitializer.scala
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @JsonIgnore
+  val nsApi = api.withNamespace(nsOrDefault)
+
+  @JsonIgnore
+  private[this] object Closed extends Throwable
+
+  @JsonIgnore
+  val stateWithClose = Var.async[Activity.State[Dtab]](Activity.Pending) { state =>
```
hawkw: Ask @klingerf – this is his code, untouched by me.

interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterInitializer.scala
```diff
@@ -0,0 +1,101 @@
...
+  def portNum = port.map(_.port)
+
+  @JsonIgnore
+  private[this] val log = Logger()
+
+  @JsonIgnore
+  val nsOrDefault = namespace.getOrElse(DefaultNamespace)
+
+  @JsonIgnore
+  val api = {
```
hawkw: Hmm, not a bad idea, trying to figure out how that would look (once again, this bit of code was written by @klingerf).

interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterInitializer.scala
```diff
@@ -0,0 +1,101 @@
...
+          state.update(Activity.Ok(newState))
+          newState
+        }
+    }
+
+    Closable.make { t =>
+      pending.raise(Closed)
+      Closable.ref(closeRef).close(t)
+    }
+  }
```
klingerf: Also, as an aside, one thing we're _not_ doing here that we are doing for IngressCache and K8sDtabStore, is fetching the current resource version before establishing the watch, and then passing that version as the resourceVersion param on the `.watch` call.

If I remember correctly, that's because when I tested this code on Kubernetes, I was running into an infinite loop where the call to watch with a given resource version would fail if the resource version is "too old" (more than ~1m old), which would trigger resetWatches, which would then re-fetch the same resource version, try to re-establish a watch, and run into the same error.

The reason why we don't encounter the loop with IngressCache and K8sDtabStore is that those are watching list endpoints, and I think it's the case that List endpoints always return the system-wide current resource version. For instance:

```
# curl -s localhost:8001/api/v1/namespaces/test/configmaps | jq '.metadata.resourceVersion'
"52871682"
# curl -s localhost:8001/api/v1/namespaces/test/configmaps | jq '.metadata.resourceVersion'
"52871686"
```

Whereas Read endpoints return the resource version for the object is being read:

```
# curl -s localhost:8001/api/v1/namespaces/test/configmaps/namerd-config | jq '.metadata.resourceVersion'
"45625196"
# curl -s localhost:8001/api/v1/namespaces/test/configmaps/namerd-config | jq '.metadata.resourceVersion'
"45625196"
```

So I think it's the correct behavior here to pass in `resourceVersion = None` to the watch call, but I just wanted to call it out because it is weird. Maybe, to be consistent with the IngressCache and K8sDtabStore implementations, we should be calling a different list endpoint altogether (`/api/v1/namespaces`?), and using the resourceVersion from that endpoint to initiate this watch.

k8s/src/test/scala/io/buoyant/k8s/ServiceNamerTest.scala
```diff
@@ -32,7 +32,7 @@ class ServiceNamerTest extends FunSuite with Awaits {
         rsp.content = Rsps.Init
         Future.value(rsp)
 
-      case req if req.uri == "/api/v1/namespaces/pythonsky/services?watch=true&resourceVersion=11600623" =>
+      case req if req.uri == "/api/v1/watch/namespaces/pythonsky/services?resourceVersion=11600623" =>
```
klingerf: I think we need to make the same update in the IngressIdentifierTest.scala test file as well, here:

https://github.com/linkerd/linkerd/blob/eliza/configmap-interpreter/linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/IngressIdentifierTest.scala#L55

interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterInitializer.scala
hawkw: @klingerf yeah, I think the main reason I didn't touch too much of your code was that I didn't _really_ understand what some of it was doing (c.f. `stateWithClose`, which seems like it uses a bunch of finagle stuff I wasn't really familiar with). I'll see if I can refactor some of this & hopefully make it clearer?

klingerf: Totally, yeah, this stuff is pretty icky. I would take a look at the implementations of [IngressCache](https://github.com/linkerd/linkerd/blob/master/k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala#L81) and [K8sDtabStore], since that's where I cribbed from when putting this branch together in the first place. And don't stress too much if there isn't an easy way to untangle this. I just wanted to give it some attention because it's not super user friendly at the moment.

k8s/src/test/scala/io/buoyant/k8s/ServiceNamerTest.scala
hawkw: Done 4463b6699c70ef2a77da6d0abbbbc3fb0c3dc3cb

klingerf:
Sorry, should have mentioned with my review yesterday -- can you also update `linkerd/docs/interpreter.md` with a description of this interpreter?

interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterInitializer.scala
hawkw: Okay, I've factored out the `Activity` code from `ConfigMapInterpreterInitializer`, `IngressCache`, and `K8sDtabStore` and moved it all to `Watchable`; let me know what you think of this PR as of 14191e996b9621a7468eaf8c3ab8c29fb2d28120?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +160,60 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+
+          val version = if (resourceVersion) {
+            initial.metadata.flatMap(_.resourceVersion)
+          } else None
+
+          val (stream, close) = watch(None, None, version)
+
+          closeRef.set(close)
+          val _ = stream.scanLeft(initialState) { onEvent }
+            .foreach { s => state.update(Activity.Ok(s)) }
```
klingerf: This is super subtle, but I think the activity needs to updated as part of the first iteration over the stream, rather than in a separate iteration. As written, I don't think the activity will ever update, since the stream doesn't terminate.

I tested this branch on Kubernetes and observed that the dtab was never picking up changes to the config map. Switching back to foldLeft fixes the issue, such that changes in the config map are reflected in the dtab as well.

```
          val _ = stream.foldLeft(initialState) { (s, watchEvent) =>
            val newState = onEvent(s, watchEvent)
            state.update(Activity.Ok(newState))
            newState
          }
```

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -84,36 +84,25 @@ class IngressCache(namespace: Option[String], apiClient: Service[Request, Respon
   }
 
   private[this] object Closed extends Throwable
```
klingerf: The Closed object is no longer used in this class; I think you can remove it.

interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterInitializer.scala
```diff
@@ -0,0 +1,82 @@
...
+  val api = {
+    val client = mkClient(Params.empty).configured(Label("configMapInterpreter"))
+    Api(client.newService(dst))
+  }
+
+  @JsonIgnore
+  val nsApi = api.withNamespace(nsOrDefault)
+
+  @JsonIgnore
+  private[this] object Closed extends Throwable
```
klingerf: Looks like this is unused now as well and can be removed.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
```diff
@@ -54,44 +54,27 @@ class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
 
   private[this]type NsMap = Map[String, VersionedDtab]
   private[this] object Closed extends Throwable
```
klingerf: Closed is unused here as well.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: Oh, gosh, that's...really bad! Glad you caught that! 

I think this bug came about because I was expecting Rust "iterator adapter" behavior here – both combinators would be called on every iteration – and I forgot Scala's iterators had different semantics. I rewrote it to use `scanLeft` because I  thought passing `onEvent` points-free was more aesthetically pleasing, but I'll definitely change it back!

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
hawkw: Ah, you're right – I overlooked that.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -198,8 +198,11 @@ private[k8s] abstract class Watchable [
           val (stream, close) = watch(None, None, version)
 
           closeRef.set(close)
-          val _ = stream.scanLeft(initialState) { onEvent }
-            .foreach { s => state.update(Activity.Ok(s)) }
+          val _ = stream.foldLeft(initialState){ (state0, event ) =>
+            val state1 = onEvent(state0, event)
+            state.update(Activity.Ok(state1))
+            state1
```
hawkw: in Haskell, we tend to pronounce this as `state''` 😛

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +160,63 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+   * @return
+   */
+  def activity[T](convert: G => T, resourceVersion: Boolean = true)
+                  (onEvent: (T, W) => T)
+  : Activity[T] =
+    Activity(Var.async[Activity.State[T]](Activity.Pending) { state =>
+      val closeRef = new AtomicReference[Closable](Closable.nop)
+      val pending = get(retryIndefinitely = true)
+        // if the initial GET failed, then the activity is a failure
+        .onFailure { e => state.update(Activity.Failed(e)) }
```
klingerf: I think we should `log.warn` when this happens, to make it easier to track down configuration errors.

interpreter/k8s/src/main/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterInitializer.scala
```diff
@@ -0,0 +1,79 @@
...
+          case ConfigMapError(e) =>
+            log.error("k8s watch error: %s", e)
+            dtab
+        }
+    }
+
+  @JsonIgnore
+  def getDtab(configMap: ConfigMap): Dtab =
+    configMap.data.get(filename) match {
+      case None => Dtab.empty
```
klingerf: Related to my other comment about logging, this case should probably `log.warn` as well, to let users know that they've configured the interpreter with a filename that does not exist in the configmap.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: yeah, that seems correct – I was wondering about that, just wasn't sure if I should change the behavior of the existing code. 👍

hawkw:
@klingerf & @adleong, I think all requested changes have been addressed? Sorry this one took so long.

hawkw:
@klingerf, should we close this in favor of #1532 to merge the same changes into `kibe-refactor`?

klingerf:
@hawkw Sorry for the delay! Yep, am all for closing this in favor of #1532.

PR #1509, opened 2017-07-15
Istio Request Redirect

### LINKED ISSUE #1419: Support redirect and rewrite route-rules

esbie:
Redirects requests based on istio route-rules. Fixes #1419
Adds H2ResponseException to ErrorResetter to match HttpResponseException in Error Responder.
DRYs up identifiers by pushing more logic into IdentifierPreconditions

### DIFF
--- a/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
+++ b/router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
@@ -53,6 +53,8 @@ object RoutingFactory {
 
   }
 
+  abstract class ResponseException[Rsp](rsp: Rsp) extends NoStackTrace
+
   /** The result of attempting to identify a request. */
   sealed trait RequestIdentification[Req]
 
@@ -165,7 +167,11 @@ class RoutingFactory[Req, Rsp](
             Future.exception(UnknownDst(req0, unidentified.reason))
           case Throw(e) =>
             Annotations.Failure.Identification.record(e.getMessage)
-            Future.exception(UnknownDst(req0, e.getMessage))
+            e match {
+              case e: ResponseException[Rsp] => Future.exception(e)
+              case _ => Future.exception(UnknownDst(req0, e.getMessage))
+            }
+
         }
 
         // Client acquisition failures are recorded within the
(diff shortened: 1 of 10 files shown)


router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -165,7 +167,11 @@ class RoutingFactory[Req, Rsp](
             Future.exception(UnknownDst(req0, unidentified.reason))
           case Throw(e) =>
             Annotations.Failure.Identification.record(e.getMessage)
-            Future.exception(UnknownDst(req0, e.getMessage))
+            e match {
+              case e: ResponseException[Rsp] => Future.exception(e)
```
esbie: I could use some feedback here. Without this change, the ResponseException never makes it to the ErrorResponder/ErrorResetter. But with this change, the exception also gets propagated to the default monitor:

```
Exception propagated to the default monitor (upstream address: /172.17.0.1:50971, downstream address: n/a, label: 0.0.0.0/80).
io.buoyant.linkerd.protocol.http.ErrorResponder$HttpResponseException
```
Makes me doubt that ErrorResponder is the right place in the stack to be handling these kinds of responses. Thoughts?

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -53,6 +53,8 @@ object RoutingFactory {
 
   }
 
+  abstract class ResponseException[Rsp](rsp: Rsp) extends NoStackTrace
```
adleong: I think this can just be a trait or abstract class with no type param or constructor args.  The child class will contain the response anyway and we don't have useful information about the type at this level.

router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
adleong: This lead my down a bit of a rabbit hole trying to figure out how Monitors work but I think we can have the ResponseException extend `com.twitter.logging.HasLogLevel` and set the log level to TRACE or something like that.

esbie: Neat that totally worked!

PR #1512, opened 2017-07-18
Annotate thrift client method on native thrift requests

klingerf:
## Problem

The thrift client stack only sets the thrift method name as the Rpc trace annotation if the request is TTwitterThrift.

## Solution

Introduce a separate thrift TracingFilter client module that will set the Rpc trace annotation for both native thrift and TTwitterThrift requests.

Note, this branch includes a few scalariform fixes in a separate commit.

### DIFF
--- a/router/thrift/src/main/scala/io/buoyant/router/thrift/TracingFilter.scala
+++ b/router/thrift/src/main/scala/io/buoyant/router/thrift/TracingFilter.scala
@@ -1,0 +1,42 @@
+package io.buoyant.router.thrift
+
+import com.twitter.finagle.Thrift.{param => tparam}
+import com.twitter.finagle.client.StackClient
+import com.twitter.finagle.thrift.ThriftClientRequest
+import com.twitter.finagle.tracing.Trace
+import com.twitter.finagle._
+import org.apache.thrift.protocol.TProtocolFactory
+import org.apache.thrift.transport.TMemoryInputTransport
+
+object TracingFilter {
+  val role = StackClient.Role.protoTracing
+  val module: Stackable[ServiceFactory[ThriftClientRequest, Array[Byte]]] =
+    new Stack.Module2[param.Tracer, tparam.ProtocolFactory, ServiceFactory[ThriftClientRequest, Array[Byte]]] {
+      val role = TracingFilter.role
+      val description = "Traces HTTP-specific request metadata"
+      def make(_tracer: param.Tracer, _pf: tparam.ProtocolFactory, next: ServiceFactory[ThriftClientRequest, Array[Byte]]) = {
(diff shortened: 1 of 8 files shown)


router/thrift/src/main/scala/io/buoyant/router/thrift/TracingFilter.scala
```diff
@@ -0,0 +1,42 @@
...
+import com.twitter.finagle._
+import org.apache.thrift.protocol.TProtocolFactory
+import org.apache.thrift.transport.TMemoryInputTransport
+
+object TracingFilter {
+  val role = StackClient.Role.protoTracing
+  val module: Stackable[ServiceFactory[ThriftClientRequest, Array[Byte]]] =
+    new Stack.Module2[param.Tracer, tparam.ProtocolFactory, ServiceFactory[ThriftClientRequest, Array[Byte]]] {
+      val role = TracingFilter.role
+      val description = "Traces HTTP-specific request metadata"
```
adleong: description should reference thrift instead of http

klingerf:
@adleong Heh, yes, there is broader context, which I sent to the linkerd-users mailing list, which no longer exists :-/ Here was my email:

> Ah, this looks like an oversight in the native thrift (non-TTwitter) implementation. Good find!

> Linkerd will produce 2 spans for each thrift request: a server span and a client span. The server span's Rpc annotation is the name of the router that served the request, plus the request's identifier. For thrift, the Rpc annotation for a router with default configuration will be `thrift /svc`. However, if you set the thriftMethodInDst option on your thrift router, then the Rpc annotation will be `thrift /svc/<method-name>`.

> For the client span, it looks like we're only setting the Rpc annotation if it's a TTwitter request. But it should be possible to set that field for non-TTwitter requests, since the method name exists in the native thrift payload. I went ahead and put together a branch to fix this:

> https://github.com/linkerd/linkerd/pull/1512

> Thanks for bringing it to our attention!

Basically, even for native thrift, we can still read the method name that's being called (we already do this in the thrift identifier), so it makes sense to use the method name as the Rpc annotation on the client span.

Before this change, a native thrift trace in Zipkin looked like:

screen shot 2017-07-18 at 10 47 57 am

And with this change, it looks like:

screen shot 2017-07-18 at 10 46 41 am

adleong:
Ah, thanks for the context.  Just to make sure I understand, for non-TTwitter thrift, we can emit traces like this from a single Linkerd but we can't join spans from multiple Linkerds?  If that's the case then tracing seems of very limited value.  We might as well fix this though.

klingerf:
@adleong Yeah, your understanding is correct. We're already providing a bunch of annotations anyway (DstTracing, for example), so we might as well fix the Rpc annotation.

PR #1514, opened 2017-07-18
Improve docs for the io.l5d.tracelog level field

klingerf:
This change was originally authored by @sumit-vij-ck in #1373. I cherry-picked his commits and then made the changes requested from that PR.

## Problem

All of the examples in this repo that demonstrate how to configure the `io.l5d.tracelog` telemeter set the `level` field to `TRACE`. Since linkerd's default log level is `INFO`, none of the tracelog output will show up in linkerd's logs unless the log level is adjusted.

## Solution

Update the `level` field's documentation to clarify how that field relates to linkerd's configured log level.

### DIFF
--- a/linkerd/docs/config.md
+++ b/linkerd/docs/config.md
@@ -40,7 +40,6 @@ telemetry:
   sampleRate: 0.01
 - kind: io.l5d.tracelog
   sampleRate: 0.2
-  level: TRACE
 ```
 
 Welcome to the Configuration Reference for linkerd!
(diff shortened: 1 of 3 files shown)


thedebugger:
hi @klingerf somehow this skipped my mind. I think it makes sense to have trace level default to INFO, instead of setting it to TRACE -- https://github.com/linkerd/linkerd/blob/master/linkerd/examples/tracelog.yaml#L16

It is not intutive when someone runs examples/tracelog.yaml, and not see any traces logged when the purpose of examples/tracelog.yaml is primarily to log traces.

klingerf:
@thedebugger Good point -- have updated tracelog.yaml.

Issue #1508, opened 2017-07-15
Advanced Observability for H2


adleong:
Continuation of https://github.com/linkerd/linkerd/issues/1507

Update the stats filter from https://github.com/linkerd/linkerd/issues/1507 to accept an H2ResponseClassifier as a stack param.  H2ResponseClassifier is essentially the type:

```
PartialFunction[(Request, Try[(Response, Try[Frame])]), ResponseClass]
```

Whenever any of the following conditions are met, the stats filter should consult the classifier and increment the `success` or `failures` stat accordingly:
* The response future fails.  `(req, Throw(e))` should be passed to the classifier.
* The response stream fails.  `(req, Return(rsp, Throw(e)))` should be passed to the classifier.
* The response stream completes successfully.  `(req, Return(rsp, Return(f)))` should be passed to the classifier where `f` is the final frame (EOS) of the response stream.

adleong:
An added wrinkle here is that `onEnd` is a `Future[Unit]` which means we don't have access to the final frame of the stream.  Changing the type of `onEnd` to `Future[Frame]` is probably not a good solution either because this would cause the Stream to hold a reference to the final frame, even after the frame is released, which is a violation of flow control.

I recommend creating a `StreamProxy` class that delegates all calls to an underlying stream but also takes a function `onFrame: Try[Frame] => ()` which is called on each read.  By wrapping the response stream in a StreamProxy, we can give it an `onFrame` that increments the appropriate stats on the final frame.

pseudocode would be something like
```
svc(req).transform {
  case Return(rsp0) =>
    val stream1 = new StreamProxy(rsp0.stream) {
      case Return(f) if f.isEnd => // consult classifier and incr stat
      case Return(f) => // do nothing
      case Throw(e) => // consult classifier and incr stat
    }
    Future.value(rsp0.copy(stream = stream1))
  case Throw(e) => 
    // consult classifier and incr stat
    Future.exception(e)
}
```

adleong:
We should also add some reasonable built-in classifiers:
* Successful iff response and response stream are successful
* Successful if response and response stream are successful and response status is < 500

hawkw:
It occurs to me that with the addition of an `onFrame` function, we also have the ability to collect some pretty granular metrics, such as, say, the number of frames per request, . Is this data we're interested in collecting @adleong?

adleong:
That would be "hella dope".  Similarly: histogram of frame sizes.

hawkw:
[...]

PR #1517, opened 2017-07-19
Advanced observability for H2

hawkw:
As described in #1508, I've added a `H2ResponseClassifier` type and modified the `StreamStatsFilter` to accept a `H2ResponseClassifier` as a stack param. The `H2ResponseClassifier` is more or less the same as a `ResponseClassifier`, except that it also requires a `Try[Option[Frame]]` containing the end frame of the H2 stream. 

In order to pass the final frame to the `H2ResponseClassifier` without violating flow control, I've added a `StreamProxy` class which allows an `onFrame: Try[Frame] => Unit` function to be attached to an H2 stream. I've also used this to add the following per-frame metrics:

+ `response/stream/total_bytes`: the total size (in bytes) of all the data frames in the response stream 
+ `request/stream//total_bytes`: tthe total size (in bytes) of all the data frames in the request stream 
I've added unit tests for the new functionality to `StreamStatsFilterTest`.


router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
-    @inline def apply(startT: Stopwatch.Elapsed)(result: Try[_]): Unit = {
+    @inline def onEnd(startT: Stopwatch.Elapsed)(result: Try[_]): Unit = {
       durationMs.add(startT().inMillis)
       result match {
         case Return(_) => successes.incr()
         case Throw(_) => failures.incr()
       }
     }
+
+    @inline def apply(startT: Stopwatch.Elapsed)(result: Try[_]): Unit = onEnd(startT)(result)
```
adleong: Oh, I think I see, you override apply in the subclass.

I would keep this all in apply and then in the subclass you can override it and call out to super.apply.

OR you can embrace "composition over inheritance" and have FrameStats contain a StreamStats instead of extending it.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
+        frameCount.add(streamFrameCount)
+        this.onEnd(startT)(result)
+      }
+      stream1
+    }
+
+    @inline def apply(startT: Stopwatch.Elapsed)(stream: Stream): Stream = onFrame(startT)(stream)
+  }
+
+  object ClassifierStats {
```
adleong: This class level object is pretty confusing.  I think we can get rid of it and move its contents up into the class directly.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
+      case _ => failures.incr()
+    }
   }
 
   //   total number of requests received
   private[this] val reqCount = statsReceiver.counter("requests")
 
   private[this] val reqStreamStats =
-    new StreamStats(statsReceiver.scope("request", "stream"))
+    new StreamStats(statsReceiver.scope("request", "stream")) with FrameStats
```
adleong: Personally I don't love this mixin pattern.  Would prefer to pass arguments into FrameStats' constructor directly.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
+    private[this] val frameSize = stats.stat("data_frame", "bytes")
+    private[this] val frameCount = stats.stat("data_frame", "count")
+
+    def onFrame(startT: Stopwatch.Elapsed)(stream0: Stream): Stream = {
+      var streamFrameCount = 0
+      val stream1 = stream0.onFrame {
+        case Return(frame: Frame.Data) =>
+          frameSize.add(frame.buf.length)
+          streamFrameCount += 1
+        case _ =>
```
adleong: I think we should count trailers towards frame count too.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
 
-    req.stream.onEnd.respond(reqStreamStats(reqT))
-
-    service(req)
+    service(req1)
+      .transform {
+        case Return(rsp0) =>
+          val rspT = Stopwatch.start()
+          val stream = rsp0.stream.onFrame {
+            case Return(frame) if frame.isEnd =>
```
adleong: We have two StreamProxies, one here and then one in rspStreamStats.  Can we move this logic into FrameStats' onFrame?

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
+    val req1 = Request(req0.headers, reqStreamStats(reqT)(req0.stream))
 
-    req.stream.onEnd.respond(reqStreamStats(reqT))
-
-    service(req)
+    service(req1)
+      .transform {
+        case Return(rsp0) =>
+          val rspT = Stopwatch.start()
+          val stream = rsp0.stream.onFrame {
```
adleong: I think we also need to handle the case where the stream isEmpty separately.  I think it should add(0) to the frameCount stat and should call the classifier with None as the final frame.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
adleong: Or, alternatively, move the logic from FrameStats out to here.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
+        case _ =>
+      }
+      val _ = stream1.onEnd.respond { result =>
+        frameCount.add(streamFrameCount)
+        this.onEnd(startT)(result)
+      }
+      stream1
+    }
+
+    @inline def apply(startT: Stopwatch.Elapsed)(stream: Stream): Stream = onFrame(startT)(stream)
```
adleong: it's pretty confusing that this has both an `apply` that takes a stream and an `apply` that takes a Try.  I think this can be clearer/simpler.

router/h2/src/test/scala/io/buoyant/router/h2/StreamStatsFilterTest.scala
```diff
@@ -40,26 +48,38 @@ class StreamStatsFilterTest extends FunSuite with Awaits {
     (stats, service)
   }
 
+  private[this] def readAll(stream: Stream): Future[Unit] =
+    stream.read().flatMap { frame =>
+      val end = frame.isEnd
+      frame.release().flatMap { _ =>
```
adleong: scala golf: `frame.release().before {`

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2ResponseClassifier.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import com.twitter.finagle.service.ResponseClass
+import com.twitter.util.Return
+
+object H2ResponseClassifier {
+
+  /**
+   * Create a [[H2ResponseClassifier]] with the given name for its `toString`.
+   *
+   * @note be careful when `underlying` composes other `H2ResponseClassifiers`
```
hawkw: ...I copied the comment from `ResponseClassifier` & forgot to delete that?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2ResponseClassifier.scala
```diff
@@ -0,0 +1,40 @@
...
+    case H2ReqRep(_, Return((_, Return(frame)))) => ResponseClass.Success
+    case _ => ResponseClass.NonRetryableFailure
+  }
+
+  /**
+   * an [[H2ResponseClassifier]] that classifies responses as successful if
+   * the response and response stream are both successful, and the status code
+   * is less than 500
+   */
+  val Not5xxH2ResponseClassifier: H2ResponseClassifier =
```
hawkw: 👍 I kind of thought so...was hoping you'd confirm in code review.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Wraps an underlying [[Stream]] with an [[StreamProxy.onFrame onFrame]] function.
+ * The `onFrame` function is called for each frame in the stream.
+ * @param underlying the [[Stream]] wrapped by this proxy
+ * @param onFrame function called for each [[Frame]] in the underlying [[Stream]]
+ */
+class StreamProxy(underlying: Stream, onFrame: Try[Frame] => Unit) extends Stream {
+  override def isEmpty: Boolean = underlying.isEmpty
+
+  override def read(): Future[Frame] = underlying.read().respond(frame => onFrame(frame))
```
hawkw: It was point-free for a while & changed because I thought it needed to free the frame after, realized that was wrong, and this never got changed back :)

project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=0.13.15
```
hawkw: probably not – I think IntelliJ makes these, and I've committed it accidentally.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,83 +1,130 @@
...
+          frameSize.add(frame.buf.length)
+          streamFrameCount += 1
+        case _ =>
+      }
+      val _ = stream1.onEnd.respond { result =>
+        frameCount.add(streamFrameCount)
+        this.onEnd(startT)(result)
+      }
+      stream1
+    }
```
olix0r: might be slightly clearer naming stream0 _underlying_ or _inner_ or something.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2ResponseClassifier.scala
```diff
@@ -0,0 +1,37 @@
...
+      def isDefinedAt(reqRep: H2ReqRep): Boolean = underlying.isDefinedAt(reqRep)
+      def apply(reqRep: H2ReqRep): ResponseClass = underlying(reqRep)
+      override def toString: String = name
+    }
+
+  /**
+   * a simple total [[H2ResponseClassifier]] that classifies responses as successful
+   * if and only if the response and response stream are both successful
+   */
+  val ReturnsOnly: H2ResponseClassifier = named("ReturnsOnlyH2ResponseClassifier") {
```
adleong: 🚲 🏡 
What about calling this `ExceptionsAsFailures`

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2ResponseClassifier.scala
```diff
@@ -0,0 +1,37 @@
...
+    case H2ReqRep(_, Return((_, Return(frame)))) => ResponseClass.Success
+    case _ => ResponseClass.NonRetryableFailure
+  }
+
+  /**
+   * an [[H2ResponseClassifier]] that classifies responses as successful if
+   * the response and response stream are both successful, and the status code
+   * is less than 500
+   */
+  val Default: H2ResponseClassifier =
```
adleong: :bike: 🏡 
What about calling this `ServerErrorsAsFailures` (this mirrors its H1 counterpart).

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2ResponseClassifier.scala
```diff
@@ -0,0 +1,37 @@
...
+    named("DefaultResponseClassifier") {
```
adleong: TIOLI this might be clearer as 
```
{
  case H2ReqRep(_, Return((response, Return(_)))) if response.status.code >= 500 => ResponseClass.NonRetryableFailure
}.orElse(ExceptionsAsFailures)
```

hawkw:
@adleong & @olix0r, I've made a bunch of additional changes to this branch – I think I've addressed everything you've requested. It should be ready for review now?

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -41,46 +44,118 @@ class StreamStatsFilter(statsReceiver: StatsReceiver)
         case Throw(_) => failures.incr()
       }
     }
+
   }
 
-  //   total number of requests received
+  class FrameStats(
```
adleong: I think we could simplify by moving the contents of StreamStats into this class, deleting StreamStats, and renaming this to StreamStats.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -41,46 +44,118 @@ class StreamStatsFilter(statsReceiver: StatsReceiver)
...
+      val streamStatsT = streamStats(startT)(_)
+      val stream =
+        if (underlying.isEmpty) {
+          // it would be easier to just put this in `stream.onEnd` and handle
+          // both the empty-stream case and the end of a non-empty stream case
+          // in the same block of code, but `join()`ing `rsp.stream.onEnd` with
+          // the `req` stream's `.onEnd` future in `StreamStatsFilter.apply()`
+          // somehow clobbers this callback, so it never triggers.
+          //
+          // i strongly suspect this is a bug in `Stream.onEnd`, since this
```
adleong: Are you able to put together a standalone repro of this bug?

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -41,46 +44,118 @@ class StreamStatsFilter(statsReceiver: StatsReceiver)
...
 
-    service(req)
+    service(req1)
+      .transform {
+        case Return(response) =>
+          val rspT = Stopwatch.start()
+          val stream = rspStreamStats(rspT, classify(Return(response)))(response.stream)
+          Future.value(Response(response.headers, stream))
+        case Throw(e) =>
+          classify(Throw(e))(Throw(e))
```
adleong: This is weird.  The signature of the method doesn't exactly match what we want to express.  At this point we know the second argument will be discarded so it's confusing that we pass in an irrelevant values (we might as well be passing null).

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -41,46 +44,118 @@ class StreamStatsFilter(statsReceiver: StatsReceiver)
...
+      }
       .respond { result =>
-        rspFutureStats(reqT)(result)
+        reqLatency.add(reqT().inMillis)
         val stream = result match {
-          case Return(rsp) => req.stream.onEnd.join(rsp.stream.onEnd)
-          case Throw(_) => req.stream.onEnd
+          case Return(rsp) =>
+            // this is the call i mentioned above that seems to be wiping out
+            // the `.onEnd.respond { }` callback in `frameStats.apply()`...
```
adleong: that would surprise me.  if there's something improper happening here we should get a repro of it.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ResponseClassifiers.scala
```diff
@@ -1,7 +1,8 @@
 package com.twitter.finagle.buoyant.h2
 
 import com.twitter.finagle.service._
-import com.twitter.util.{NonFatal, Return, Throw, Try}
+import com.twitter.util.{Return, Throw, Try}
+import scala.util.control.NonFatal
 
 object ResponseClassifiers {
```
hawkw: Are these used elsewhere? If not, I can go ahead and delete them.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,38 +1,41 @@
...
   extends SimpleFilter[Request, Response] {
 
-  class StreamStats(
-    stats: StatsReceiver,
-    durationName: Option[String] = None,
-    successName: Option[String] = None
-  ) {
-
+  // the `StreamStats` (and the related `FrameStats`) class is really just
+  // a "stateful function" – essentially a closure, but closing over state
```
hawkw: This is just my thought process out loud, I guess  – iteration one of `StreamStatsFilter` had all of the stats floating around as instance variables, so I was conceptualizing this as sort of being like a closure over them?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ResponseClassifiers.scala
adleong: Yeah, they're settable via the H2 config and they feed into Finagle's retries and metrics filters.  As part of this change we want to rip all of that out and replace it with our stream aware versions.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
hawkw: I'll see what I can do!

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/package.scala
```diff
@@ -0,0 +1,27 @@
...
+   *
+   * @param request  a H2 [[Request]]
+   * @param response a [[Try]] containing either the H2 [[Response]] and the
+   *                 final [[Frame]], or an error. the final frame is contained in
+   *                 both an [[Option]], so that empty streams can pass [[None]], and
+   *                 in a [[Try]], in case the [[Stream]] [[com.twitter.util.Throw Throw]]s.
+   */
+  // TODO: consider changing this type to `(Request, Try[Response], Try[Option[Frame]])`
+  //       so that patterns matching it can be a little bit less gross?
+  case class H2ReqRep(request: Request, response: Try[(Response, Try[Option[Frame]])])
```
hawkw: Yeah, I think it's this way primarily because classifiers can match on `Return(_)` instead of having to have separate cases for `Some(Return(_))` and `None`, but the semantics of having it be like that is a little wonky.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ResponseClassifiers.scala
```diff
@@ -68,41 +68,87 @@ object ResponseClassifiers {
...
 
   /**
    * Classifies 5XX responses and all exceptions as non-retryable
    * failures.
    */
   val NonRetryableServerFailures: ResponseClassifier =
-    ResponseClassifier.named(s"NonRetryableServerFailures") {
-      case ReqRep(_, Throw(_) | Return(Responses.Failure())) => ResponseClass.NonRetryableFailure
+    named(s"NonRetryableServerFailures") {
+      case H2ReqRep(_, Throw(_) | Return((Responses.Failure(), _))) => ResponseClass.NonRetryableFailure
```
adleong: I think we want to classify `(_, Return((_, Throw(_))))` as a failure too.  eg when we get a 200 OK but the response stream fails.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ResponseClassifiers.scala
```diff
@@ -68,41 +68,87 @@ object ResponseClassifiers {
...
+    named("ExceptionsAsFailuresH2ResponseClassifier") {
+      case H2ReqRep(_, Throw(_)) | H2ReqRep(_, Return((_, Some(Throw(_))))) =>
+        ResponseClass.NonRetryableFailure
+    }
+
+  /**
+   * an [[ResponseClassifier]] that classifies responses as failures if
+   * the status code was greater than or equal to 500 (server errors)
+   */
+  val ServerErrorsAsFailures: ResponseClassifier =
```
adleong: I think this is the same as NonRetryableServerFailures

linkerd/core/src/main/scala/io/buoyant/linkerd/SvcConfig.scala
```diff
@@ -18,6 +18,7 @@ import io.buoyant.router.RetryBudgetModule.{param => ev}
  * class as part of config deserialization.
  */
 trait SvcConfig {
+  type ProtocolResponseClassifier = ResponseClassifier
```
adleong: Why this type alias?

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,50 +1,109 @@
...
+    def success(streamDuration: Duration): Unit = {
+      durationMs.add(streamDuration.inMillis)
+      successes.incr()
+    }
+
+    def failure(streamDuration: Duration): Unit = {
+      durationMs.add(streamDuration.inMillis)
+      successes.incr()
+    }
+    private[this] val frameBytes = stats.stat("data_frame", "total_bytes")
```
adleong: This represents the number of bytes of data in the whole stream right? not in a single frame?  I think the name is misleading.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,50 +1,109 @@
...
+
+    def apply(
+      startT: Stopwatch.Elapsed,
+      classifyFrame: Option[Try[Frame]] => Unit = { _ => }
```
adleong: maybe call this `onFinalFrame` or something

linkerd/core/src/main/scala/io/buoyant/linkerd/SvcConfig.scala
hawkw: So the types of some of the defs in this trait can be polymorphic – `H2SvcConfig` overrides this type with `h2.ResponseClassifier`

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
hawkw: Yeah, that's why I changed it from `bytes` to `total_bytes` - I'll probably remove the `data_frame` namespace (the other stat in that namespace was removed so it's kind of meaningless anyway), does that seem clearer?

hawkw:
@adleong, I did a bunch more work on this branch this morning; H2 `ResponseClassifiers` have now been brought much more in line with how protocol-specific `ResponseClassifier`s look elsewhere in Finagle. It should be ready for another round of review – also, do we want to talk about trying to integrate this with your work on retries in #1523 ?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRep.scala
```diff
@@ -0,0 +1,66 @@
...
+ * Like a [[com.twitter.finagle.service.ReqRep]], but for H2
+ *
+ * @param request  a H2 [[Request]]
+ * @param response a [[Try]] containing either the H2 [[Response]] and the
+ *                 final [[Frame]], or an error. the final frame is contained in
+ *                 both an [[Option]], so that empty streams can pass [[None]], and
+ *                 in a [[Try]], in case the [[Stream]] [[com.twitter.util.Throw Throw]]s.
+ */
+class H2ReqRep(override val request: Request, override val response: Try[RepAndFrame])
+  extends ReqRep(request, response) {
```
adleong: Because of this, I think we still need our own H2ResponseClassifier type (or h2.ResponseClassifier or whatever).

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRep.scala
```diff
@@ -0,0 +1,66 @@
...
+  type FinalFrame = Option[Try[Frame]]
+
+  def apply(request: Request, response: Try[RepAndFrame]): H2ReqRep =
+    new H2ReqRep(request, response)
+
+  @inline def apply(request: Request, response: Try[Response], finalFrame: FinalFrame = None): H2ReqRep = H2ReqRep(request, response.map(r => (r, finalFrame)))
+
+  @inline def apply(request: Request, err: Throwable): H2ReqRep = H2ReqRep(request, Throw(err))
+
+  def unapply(reqrep: ReqRep): Option[(Request, Try[RepAndFrame])] =
```
adleong: I don't think we want this method anyway.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRep.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  override def toString: String = s"H2ReqRep($request, $responseOnly, finalFrame=$finalFrame)"
+}
+
+object H2ReqRep {
+  // type aliases to minimise typing
+  type RepAndFrame = (Response, Option[Try[Frame]])
+  type FinalFrame = Option[Try[Frame]]
+
+  def apply(request: Request, response: Try[RepAndFrame]): H2ReqRep =
```
adleong: I'm not sure if these convenience methods actually add much convenience.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRep.scala
```diff
@@ -0,0 +1,66 @@
...
+    new H2ReqRep(request, response)
+
+  @inline def apply(request: Request, response: Try[Response], finalFrame: FinalFrame = None): H2ReqRep = H2ReqRep(request, response.map(r => (r, finalFrame)))
```
adleong: I still find it a bit icky that this signature admits invalid inputs:

```
apply(r, Throw(e), Some(f)) // f is specified as the final frame but is unused
```

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/param.scala
```diff
@@ -139,4 +140,13 @@ package object param {
   object Tracer {
     implicit val param: Stack.Param[Tracer] = Stack.Param(Tracer(DefaultTracer))
   }
+
+  //  case class H2ResponseClassifier(classifier: ResponseClassifier) {
```
adleong: I think we gotta bring this back

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
adleong: Any progress on this?  I'd like to remove this (and related) comment if it's not a real thing.  Or fix it if it is.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -55,31 +117,41 @@ class StreamStatsFilter(statsReceiver: StatsReceiver)
...
+        case Successful(_) => successes.incr()
+        case _ => failures.incr()
+      }
 
-    service(req)
+    service(req1)
+      .transform {
+        case Return(response) =>
+          val rspT = Stopwatch.start()
+          val stream = rspStreamStats(rspT, classify(Return(response)))(response.stream)
```
adleong: I'd explicitly name onFinalFrame since it's sort of an optional parameter:

```
rspStreamStats(rspT, onFinalFrame = classify(Return(response)))
```

This helps clarify that `reqStreamStats` is the same thing but doesn't provide any `onFinalFrame` callback

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRep.scala
```diff
@@ -0,0 +1,66 @@
...
+ *                 in a [[Try]], in case the [[Stream]] [[com.twitter.util.Throw Throw]]s.
+ */
+class H2ReqRep(override val request: Request, override val response: Try[RepAndFrame])
+  extends ReqRep(request, response) {
+
+  /**
+   * the [[Response]] part of `this.response`,
+   * ignoring the final frame of the response stream
+   */
+  def responseOnly: Try[Response] = response.map { case (rsp, _) => rsp }
```
hawkw: I try to avoid using the `._n` syntax for tuples – generally, I feel like it reduces clarity. Though in this case I suppose it's pretty clear.

olix0r: scala golf: val-up bodies to eliminate needless alloc ;)

```scala
private[this] val _toRsp = { case (rsp, _) => rsp }
def responseOnly = response.map(_toRsp)
```

(and everywhere we can do this, we should.  keep in mind that classifers will run multiple times for each req/rep, so we should be as careful as possible wrt allocation

hawkw: @olix0r I (very briefly) had these methods as `lazy val`s instead of `def`s, but I figured it would make Alex mad. Since this is in the hot path, and the parameters are `val`s, can I _pleeeease_? :)

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ResponseClassifierConfig.scala
```diff
@@ -0,0 +1,11 @@
+package io.buoyant.linkerd.protocol.h2
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.buoyant.h2.service.ResponseClassifier
+import io.buoyant.config.{ConfigInitializer, PolymorphicConfig}
+
+abstract class ResponseClassifierInitializer extends ConfigInitializer
+abstract class ResponseClassifierConfig extends PolymorphicConfig {
```
adleong: For most dynamically loaded plugins we list all the plugin types in Linkerd.scala in linkerd-core.  However, we can't do that for this one since it's H2 specific and we don't want linked-core to depend on linkerd-protocol-h2.  The easiest thing to do for now is to explicitly list all subclasses here with a @JsonSubTypes annotation.  This is what we do for LoadBalancerConfig, for example.

linkerd/examples/src/test/scala/io/buoyant/linkerd/examples/ExamplesTest.scala
```diff
@@ -22,10 +22,12 @@ class ExamplesTest extends FunSuite {
         val lines = source.getLines().toSeq
         val firstLine = lines.headOption
         if (!firstLine.contains("#notest")) {
-          val yaml = lines.mkString("\n")
-          val parsed = Linker.parse(yaml)
-          val loaded = parsed.mk()
-          assert(mapper.writeValueAsString(parsed).nonEmpty)
+          withClue(s"example: ${file.getName}") {
```
hawkw: sure, I was just using it to debug which tests were failing

linkerd/core/src/main/scala/io/buoyant/linkerd/SvcConfig.scala
```diff
@@ -33,9 +33,11 @@ trait SvcConfig {
    * responseClassifier categorizes responses to determine whether
    * they are failures and if they are retryable.
    */
-  @JsonProperty("responseClassifier")
   var _responseClassifier: Option[ResponseClassifierConfig] = None
```
hawkw: 👍

linkerd/core/src/main/scala/io/buoyant/linkerd/SvcConfig.scala
```diff
@@ -33,9 +33,11 @@ trait SvcConfig {
...
 
+  @JsonProperty("responseClassifier")
+  def responseClassifier_=(r: ResponseClassifierConfig): Unit = _responseClassifier = Some(r)
```
hawkw: yeah – I was going to add one, but the conventional name for it, `responseClassifier()`, clashes with the `responseClassifier()` def & I didn't want to rename that? I suppose it could be called `getResponseClassifier` but that doesn't seem v. idiomatic?

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
```diff
@@ -172,17 +167,32 @@ class H2StaticSvc(val configs: Seq[H2SvcPrefixConfig]) extends H2Svc with Static
...
 
 trait H2SvcConfig extends SvcConfig {
+  /**
+   * Override the setter for [[SvcConfig]]'s `_responseClassifier` field
+   * so that we can set `JsonIgnore` on it (and rewire [[_h2ResponseClassifier]]
+   * to the `"responseClassifier"` JSON property).
+   * @param r a HTTP [[H1ResponseClassifierConfig]]. Not used.
+   */
+  @JsonIgnore
+  final override def responseClassifier_=(r: H1ResponseClassifierConfig): Unit = {}
```
hawkw: yeah that's probably a good call

linkerd/core/src/main/scala/io/buoyant/linkerd/SvcConfig.scala
adleong: I think `getResponseClassifier` is fine.  We're outside the idiom because of this little hack.  As long as it has the right JsonProperty annotation so that it serializes correctly.

hawkw: What about renaming this field to `responseClassifierConfig`? Although it's a little wordier, I feel like it makes the difference between this field and the `responseClassifier()` method clearer, and allows the getter and setter to have more idiomatic names?

adleong: sure

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ResponseClassifiers.scala
```diff
@@ -39,7 +37,8 @@ class NonRetryable5XXInitializer extends ResponseClassifierInitializer {
 object NonRetryable5XXInitializer extends NonRetryable5XXInitializer
 
 class AllSuccessfulConfig extends ResponseClassifierConfig {
-  def mk: ResponseClassifier = ClassifiedRetries.Default
+  def mk: ResponseClassifier = ??? //ClassifiedRetries.Default
```
hawkw: `ClassifiedRetries.Default` is of type `http.ResponseClassifier`, not `h2.ResponseClassifier`. This could make `ResponseClassifiers.Default` which should have that behaviour?

adleong: ah yeah, `ResponseClassifiers.Default`

olix0r:
Note: it's fine for linkerd's _config_ to continue to reference response classifiers.  i think in the code, though, we should be clear that these are separate APIs

hawkw:
@olix0r I'm on board with the renaming suggestion – it also means we'd be able to refer to `H2StreamClassifier`s and `ResponseClassifier`s in the same file w/o renaming imports, which I'm in favor of. WDYT @adleong?

adleong:
:+1:

PR #1518, opened 2017-07-19
Fix build modules so that mesh-core will be published

Ashald:
Right now it's not published to Maven Central and I get:
```
sbt.ResolveException: unresolved dependency: io.buoyant#mesh-core_2.12;1.1.2: not found
```
when I try to use `interpreter-mesh`.


Ashald:
Thanks!

PR #1519, opened 2017-07-19
Add http identifier endpoint to linkerd

esbie:
Problem:
As identifiers increase in complexity (see IstioIngressIdentifier), it becomes more
difficult to debug them.

Solution:
Adds a config admin option `httpIdentifierPort` that, if set, serves a json endpoint
on that port. Given a request, the endpoint returns a map where the keys are
router labels and the values are identifier responses from the identifiers 
configured for those routers.

Example:
```json
$ curl -H "l5d-name: /hello"  -H "Host: world" localhost:4888 | jq
{
  "netty3": {
    "path": "/svc/hello",
    "dtab": "/svc=>/#/io.l5d.fs"
  },
  "netty4": {
    "path": "/svc/world",
    "dtab": "/svc=>/#/io.l5d.fs"
  }
}
```

I wasn't 100% sure what packages to put this stuff in.
We could do this for h2 as well.

### DIFF
--- a/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/HttpIdentifierHandler.scala
+++ b/linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/HttpIdentifierHandler.scala
@@ -1,0 +1,46 @@
+package io.buoyant.linkerd.admin
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
+import com.twitter.finagle.Service
+import com.twitter.finagle.http._
+import com.twitter.io.Buf
+import com.twitter.util.{Future, Return, Throw, Try}
+import io.buoyant.router.RoutingFactory
+import io.buoyant.router.RoutingFactory.{IdentifiedRequest, UnidentifiedRequest}
+
+case class IdResult(path: Option[String] = None, dtab: Option[String] = None, error: Option[String] = None)
+
+object Json {
+  private[this] val mapper = new ObjectMapper with ScalaObjectMapper
+  mapper.registerModule(DefaultScalaModule)
+
+  def write[T](t: T): Buf =
+    Buf.ByteArray.Owned(mapper.writeValueAsBytes(t))
+}
+
+class HttpIdentifierHandler(
+  identifiersByLabel: Map[String, RoutingFactory.Identifier[Request]]
+) extends Service[Request, Response] {
+
+  def apply(req: Request): Future[Response] = {
+    val results = identifiersByLabel.mapValues { id =>
+      id(req).transform {
+        case Return(IdentifiedRequest(dst, req1)) =>
+          Future.value(IdResult(Some(dst.path.show), Some(dst.dtab.show)))
+        case Return(unidentified: UnidentifiedRequest[Request]) =>
+          Future.value(IdResult(error = Some(unidentified.reason)))
+        case Throw(e) =>
+          Future.value(IdResult(error = Some(e.getMessage)))
+      }
+    }
+
+    Future.collect(results).map { r =>
+      val rsp = Response()
+      rsp.content = Json.write(r)
+      rsp.contentType = MediaType.Json
+      rsp
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -64,10 +64,16 @@ object Main extends App {
...
     linker: Linker
-  ): Closable with Awaitable[Unit] = {
+  ): Seq[Closable with Awaitable[Unit]] = {
     val server = linker.admin.serve(this, LinkerdAdmin(config, linker))
     log.info(s"serving ${linker.admin.scheme} admin on ${server.boundAddress}")
-    server
+
+    config.admin.flatMap(_.httpIdentifierPort) match {
+      case Some(p) =>
+        val idServer = linker.admin.serveHandler(new InetSocketAddress(p.port), LinkerdAdmin.identifierHandler(config, linker))
```
adleong: We probably want to have this serve on the same IP as the admin service (or make the ip configurable)

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/HttpIdentifierHandler.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.twitter.finagle.http._
+import com.twitter.io.Buf
+import com.twitter.util.{Future, Return, Throw, Try}
+import io.buoyant.router.RoutingFactory
+import io.buoyant.router.RoutingFactory.{IdentifiedRequest, UnidentifiedRequest}
+
+case class IdResult(path: Option[String] = None, dtab: Option[String] = None, error: Option[String] = None)
+
+object Json {
+  private[this] val mapper = new ObjectMapper with ScalaObjectMapper
```
adleong: TIOLI: I think there's a setting that makes Jackson not serialize fields with null values.  It might be nice to omit the "error" field if there's no error.

linkerd/admin/src/main/scala/io/buoyant/linkerd/admin/LinkerdAdmin.scala
```diff
@@ -53,6 +54,20 @@ object LinkerdAdmin {
...
 
+  def identifierHandler(lc: Linker.LinkerConfig, linker: Linker): Handler = {
+    linker.routers.head.protocol.configId
+
+    val identifiers = linker.routers.collect {
+      case router if router.protocol.configId == "http" =>
+        val RoutingFactory.DstPrefix(pfx) = router.params[RoutingFactory.DstPrefix]
+        val RoutingFactory.BaseDtab(baseDtab) = router.params[RoutingFactory.BaseDtab]
+        val Http.param.HttpIdentifier(id) = router.params[Http.param.HttpIdentifier]
+        router.label -> id(pfx, baseDtab)
```
adleong: Keep in mind this will create a second identifier.  This is fine but we should be aware of this and try to make sure identifiers don't create expensive resources.

esbie: Ohhh good point I hadn't considered that

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
siggy: sort of related: how about serving on the existing admin ip:port ? could be useful/simpler for the admin ui down the road.

adleong: The issue with putting it on the existing admin port, I think, is that it would have to be served at a particular URL.  But the URL can factor into identification.

siggy: Hmm, what about serving it at a particular URL, consuming the expected path, and then using the rest of the path for identification? From a user's perspective this would feel similar to the delegation endpoints.

My primary concern is when we need this info (via admin ui or curl) our users will need to redeploy with a new port open in their k8s/ELB/whatever, and/or we update linkerd-examples to expose this new port. I guess this comes down to internal complexity vs. user complexity?

esbie: I'm glad you're thinking about UX here. I don't see this as internal vs. user complexity. More like accuracy vs. accessibility. I share your concerns though; I don't really see a path towards integrating this with the admin ui.
Serving from the existing admin port would preclude being able to implement this for h2 identifiers. Multiple ports allow for multiple protocols.
Let's talk through it more though, there could be something I'm not thinking of.

PR #1520, opened 2017-07-19
Copy editing for linkerd and namerd configuration documentation

ctozzi:


linkerd/docs/namer.md
```diff
@@ -699,8 +699,8 @@ would be rewritten to `/pfx/foo/resource/name`
 ```
 
 Rewrites a name of the form "host:ip" as a path with host followed by ip. Does
-not not support IPv6 host IPs (because ipv6 notation doesn't work in Paths as-
-is, due to bracket characters).
+not not support IPv6 host IPs (because ipv6 notation doesn't work in Paths as-is
```
klingerf: I think there's another typo here: "not not". Can you also standardize the capitalization of IPv6?

linkerd/docs/namer.md
```diff
@@ -720,8 +720,8 @@ would be rewritten to `/pfx/host/port/etc`.
 ```
 
 Rewrites a name of the form "host:ip" as a path with ip followed by host. Does
-not not support IPv6 host IPs (because ipv6 notation doesn't work in Paths as-
-is, due to bracket characters).
+not not support IPv6 host IPs (because ipv6 notation doesn't work in Paths as-is
```
klingerf: Same comment here as above.

PR #1523, opened 2017-07-20
H2 retries

adleong:
Finagle's retry mechanisms do not work for H2 because of the presence of request
and response streams. When Linkerd is processing a response from a downstream
service, it may not have enough information to determine if the request was
retryable until the response stream completes. Furthermore, the request stream
has already been (or is already being) streamed to the first downstream endpoint
so we cannot send that same stream to a second downstream endpoint in the case
of a retry.

We add a stream aware ClassifiedRetriesFilter which uses BufferedStream and the
stream aware H2Classifiers. When a request comes in, the request stream is
buffered so that it can be replayed to a second downstream endpoint if a retry
is necessary. When the response comes back, the response stream is buffered as
well and no response object is returned until we can make a determination about
retryability or we give up. Depending on the H2Classifier, we may be able to
determine the retryability immediately (if the H2Classifier's responseClassifier
is defined) or we may need to wait until the response stream is complete and use
the H2Classifier's streamClassifier. If the response buffer becomes full or if a
timeout is exceeded, we give up and return the current response stream from the
buffer. If we determine that a retry should be made, we fork a new request
stream from the request stream buffer and send it to a downstream endpoint.


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
```diff
@@ -0,0 +1,197 @@
...
+ * When the buffer becomes full or when discardBuffer() is called, the buffer will be discarded
+ * and no new child Streams may be created.  However, child Streams that had previously been
+ * created will continue to receive Frames as they are read from the underlying Stream.
+ *
+ * The Frames that are stored in the buffer and offered to the child Streams are refcounted and
+ * the underlying Frame will only be released once each child Stream releases the Frame and the
+ * buffer itself releases the Frame.  Threrefore you should always call discardBuffer on a
+ * BufferedStream before it leaves scope.
+ */
+class BufferedStream(underlying: Stream, bufferCapacity: Long = 8.kilobytes.bytes) {
```
olix0r: technically, according to the h2 spec, the window size may change.  it may be worth modeling capacity as a Var.  but i wouldn't make this a blocker.  Also, i'd default this to 65535 bytes to match the rfc default window size

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ClassifiedRetryFilter.scala
```diff
@@ -0,0 +1,134 @@
...
+    val requestBuffer = new BufferedStream(request.stream, requestBufferSize)
+    val fork = Future.const(requestBuffer.fork())
+    // begin reading the request stream
+    requestBuffer.read()
+
+    def dispatch(reqStream: Stream): Future[Response] = {
+      val req = Request(request.headers.dup(), reqStream)
+      service(req).flatMap { rsp =>
+        // Buffer the response stream so that we can attempt to classify it before returning
+        // or discarding it
```
olix0r: nit, prefer periods at the end of comments to indicate it's a complete thought

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ClassifiedRetryFilter.scala
```diff
@@ -0,0 +1,134 @@
...
+        val responseBuffer = new BufferedStream(rsp.stream, responseBufferSize)
+        // We eagerly create a child response stream since we need something to return in case we
+        // don't want to (or can't) retry.
+        val responseStream = responseBuffer.fork()
+        val onFullyBufferd = responseBuffer.read()
```
olix0r: isn't this really onEnd?  read() doesn't return when it's fully buffered...

adleong:
I'm happy to do a walkthrough if you think that would be helpful.  In many places we call `read()` to kick off reading/buffering from the underlying stream but drop the Future so we continue immediately instead of waiting until the stream is complete.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ClassifiedRetryFilter.scala
olix0r: typo

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ClassifiedRetryFilter.scala
```diff
@@ -0,0 +1,134 @@
+package com.twitter.finagle.buoyant.h2
+
+import com.twitter.conversions.storage._
+import com.twitter.finagle.{Service, SimpleFilter}
+import com.twitter.util.{Future, Return, Throw}
+
+/**
+ * The ClassifiedRetyFilter uses BufferedStreams to implement retries.  The request stream is
```
natemurthy: super minor misspelling: `ClassifiedRetyFilter` => `ClassifiedRetryFilter `

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/ClassifiedRetryFilter.scala
```diff
@@ -0,0 +1,134 @@
...
+ * buffered and a child stream is sent to the service.  This is done so that if it becomes
+ * necessary to retry, a new child of a the request stream can be created.
```
natemurthy: got an extra article here: `...of a the request...`

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
```diff
@@ -37,7 +37,10 @@ class BufferedStream(underlying: Stream, bufferCapacity: Long = 8.kilobytes.byte
 
   def bufferSize: Long = _bufferSize
 
-  def onEnd: Future[Unit] = _onEnd
+  def onEnd: Future[Unit] = if (underlying.isEmpty)
+    Future.Unit
+  else
+    _onEnd
```
olix0r: nit: definitely prefer either all on one line or braces

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
```diff
@@ -0,0 +1,212 @@
...
+      underlying.read().transform {
+        case Return(f: Frame.Data) => synchronized {
+          val refCounted = new RefCountedDataFrame(f)
+          handleFrame(refCounted, f.buf.length)
+          if (f.isEnd) _onEnd.setDone()
+          Future.Unit
+        }
+        case Return(f: Frame.Trailers) => synchronized {
+          val refCounted = new RefCountedTrailersFrame(f)
+          // TODO: Determine size of Trailers Frame to count against buffer capacity
```
olix0r: fwiw, for the purposes of HTTP/2, only data frames are counted against flow control.  so this probably doesn't have to be todone

finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/BufferedStreamTest.scala
```diff
@@ -0,0 +1,174 @@
...
+
+    assert(frame1.buf == Buf.ByteArray(0, 1, 2, 3, 4))
+    assert(frame2.buf == Buf.ByteArray(5, 6, 7, 8, 9))
+    assert(frame3.buf == Buf.ByteArray(10, 11, 12, 13, 14))
+    assert(frame4.buf == Buf.ByteArray(15, 16, 17, 18, 19))
+
+    frame1.release()
+    frame2.release()
+    frame3.release()
+    frame4.release()
```
olix0r: unclear to me if this is tested elsewhere, but should you verify that buffer capacity is reclaimed as frames are released?

adleong: buffer capacity is never reclaimed.  If a child releases a (RefCounted) frame while the frame is still in the buffer, the underlying frame is not released because the buffer holds a reference to it.  If the buffer has been discarded there is no way to recover the beginning of stream so there's no point putting anything in the buffer.  Once it's gone, it's gone.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
adleong: cool

adleong:
Here's what's left:

* With this change, responses are held and not returned until either the buffer is overfull or the response stream completes.  This breaks a few tests which rely on the response returning before the stream is complete.  I could remove the tests that rely on that behavior but I'd prefer to instead wait and merge https://github.com/linkerd/linkerd/pull/1545 first, which will allow early classification in many cases.  We should ALSO probably add some kind of classification timeout where we give up on classification if the response stream does not complete in a certain amount of time.  Otherwise, we could be holding the response indefinitely.
* Fix typos and nits
* Write a commit message
* Use generated load and see what the allocation rate looks like before/after.

router/h2/src/e2e/scala/io/buoyant/router/h2/RouterEndToEndTest.scala
```diff
@@ -147,11 +149,11 @@ class RouterEndToEndTest
       assert(!rspF.isDefined)
 
       dogRspP.setValue(serverLocalStream)
+      serverLocalQ.offer(Frame.Data("boop", eos = false))
       val rsp = await(rspF)
       val reqReadF = reqStream.read()
       val rspReadF = rsp.stream.read()
 
-      rspReadF.raise(new ChannelClosedException())
```
adleong: This seemingly fixed a flaky test failure:
```
Throw(Reset.Closed) did not equal Throw(Reset.Cancel) (RouterEndToEndTest.scala:159)
```
But I don't really understand why this was here in the first place or how it could have non-deterministically caused that failure.

adleong:
Some very preliminary allocation data points:

master (baseline) with 0% failures and 300rps: 17.4 MiB/s
h2-retries with 0% failures and 300rps: 19.9 MiB/s
h2-retries with 10% failures and 270rps: 19.3 MiB/s
h2-retries with 50% failures and 150rps: 16.0 MiB/s

Because I used strest-grpc to generate load, I wasn't able to keep rps constant at different retry levels.  As I increased the amount of retries, the rps from strest-grpc would drop.  What we do see, however, is that allocations per request goes down as we increase the retries.  This makes sense because a retry should be cheaper than a brand new request.

This isn't a comprehensive picture, but I think it demonstrates that we're in the right ballpark, at least for now.

linkerd/docs/protocol-h2.md
```diff
@@ -75,13 +75,27 @@ prior-knowledge requests.
...
 initialStreamWindowBytes | 64KB | Configures `SETTINGS_INITIAL_WINDOW_SIZE` on streams.
 maxConcurrentStreamsPerConnection | unlimited | Configures `SETTINGS_MAX_CONCURRENT_STREAMS` on new streams.
 maxFrameBytes | 16KB | Configures `SETTINGS_MAX_FRAME_SIZE` on new streams.
 maxHeaderListByts | none | Configures `SETTINGS_MAX_HEADER_LIST_SIZE` on new streams.
 
+## HTTP/2 Service Parameters
+
+Key                     | Default Value | Description
+----------------------- | ------------- | -----------
+classificationTimeoutMs | None          | The amount of time to wait for a response stream to complete before determining if it should be retried.
```
olix0r: we should probably have a default

router/h2/src/main/scala/io/buoyant/router/h2/ClassifiedRetries.scala
```diff
@@ -3,35 +3,58 @@ package io.buoyant.router.h2
...
 
+  case class BufferSize(requestBufferSize: Long, responseBufferSize: Long)
+  implicit object BufferSize extends Stack.Param[BufferSize] {
+    override def default =
+      BufferSize(ClassifiedRetryFilter.DefaultBufferSize, ClassifiedRetryFilter.DefaultBufferSize)
+  }
+
+  case class ClassificationTimeout(timeout: Duration)
+  implicit object ClassificationTimeout extends Stack.Param[ClassificationTimeout] {
+    override def default = ClassificationTimeout(Duration.Top)
```
olix0r: we should probably have a better timeout (maybe enforced in config though)

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
```diff
@@ -200,16 +202,30 @@ trait H2SvcConfig extends SvcConfig {
...
+
+  var retryBufferSize: Option[RetryBufferSize] = None
 
   @JsonIgnore
-  override def params(vars: Map[String, String]): Stack.Params = super.params(vars)
-    .maybeWith(h2ResponseClassifier.map(H2StreamClassifier(_)))
+  override def params(vars: Map[String, String]): Stack.Params =
+    super.params(vars)
+      .maybeWith(h2Classifier.map(h2Param.H2Classifier(_)))
+      .maybeWith(classificationTimeoutMs.map { t => ClassificationTimeout(t.millis) })
```
olix0r: this should have a default -- ALSO we should ensure that it fits in the timer granularity ( i think has to be greater than 10ms?)

linkerd/docs/protocol-h2.md
```diff
@@ -75,13 +75,27 @@ prior-knowledge requests.
...
+retryBufferSize         | see below     | A RetryBufferSize object describing the size of the buffers for request and response streams used for retries.
+
+#### RetryBufferSize
+
+Key           | Default Value   | Description
+------------- | --------------- | -----------
+requestBytes  | `65535`         | If the request stream exceeds this value, the request cannot be retried.
+responseBytes | `65535`         | If the response stream exceeds this value, the request cannot be retried.
```
olix0r: This should _really_ be the same value of the window size configurations (i.e. we can buffer up to an h2 server's window size if a request stream, and up to an h2 client's window size for the response stream.

If we can't enforce this in code, it should be clearly documented.

adleong: is that `initialStreamWindowBytes`?  TODO would be using the current window size

Issue #1511, opened 2017-07-17
Don't require port to be an Int in HostColonPort


adleong:
The HostColonPort matcher used by the hostportPfx and porthostPfx namers requires that the port be an Int.  For example `google.com:80`.  If we lift this restriction, then it would be able to accept port names like `foo.default:http`.  This would allow us to write dtabs that let users specify the destination port name in the Host/Authority header.  Ideally, we could update the standard k8s service mesh dtab to something like:

```
        /ph  => /$/io.buoyant.rinet ;                     # /ph/80/google.com -> /$/io.buoyant.rinet/80/google.com
        /k8s => /#/io.l5d.k8s.http ;                      # /k8s/default/http/foo -> /#/io.l5d.k8s.http/default/http/foo
        /host => /#/nsSvcPortToK8s ;                      # /host/default/foo/http -> /k8s/default/http/foo
        /hp  => /host/default ;                           # /hp/foo/http -> /host/default/foo/http
        /hp  => /$/io.buoyant.http.domainToPathPfx/host ; # /hp/foo.default/http -> /host/default/foo/http
        /ph  => /#/phToHp ;                               # /ph/http/foo.default -> /hp/foo.default/http
        /svc => /ph/80 ;                                  # /svc/foo.default -> /ph/80/foo.default
        /svc => /$/io.buoyant.porthostPfx/ph ;            # /svc/foo.default:http -> /ph/http/foo.default
```

Which allows users to specify both the destination port and destination namespace `foo.<namespace>:<port>` (and defaults to `default` and `80` for namespace and port respectively).

hawkw:
@adleong what's a reasonable token to parse the port with? alphanumerics-only?

adleong:
This is the regex that Kubernetes uses for port names: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L40

PR #1526, opened 2017-07-20
Don't require port to be an Int in HostColonPort 

hawkw:
As @adleong suggested in #1511, allowing Strings rather than Ints in the `HostColonPort` would allow us to accept port names like `:http`, so that ports can be specified in Host/Authority headers, rather than in dtabs.

I've rewritten the `HostColonPort` to use a regex that matches valid RFC 1035/1123 "label" strings. 

I added tests to `HostPortTest` to test the matcher with non-numeric port names.

### DIFF
--- a/namer/core/src/main/scala/io/buoyant/hostport.scala
+++ b/namer/core/src/main/scala/io/buoyant/hostport.scala
@@ -17,7 +17,7 @@ private object HostColonPort {
     * this is based on the `DNS_LABEL` regex in Kubernetes:
     * https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L40-L43
     */
-  private[this] val DnsLabel  = """[a-z0-9]([-a-z0-9]*[a-z0-9])?""".r
+  private[this] val DnsLabel  = """([a-z0-9][-a-z0-9]*[a-z0-9]?)""".r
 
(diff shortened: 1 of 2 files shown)


namer/core/src/main/scala/io/buoyant/hostport.scala
```diff
@@ -9,15 +9,20 @@ import com.twitter.finagle.Path
...
 
-  def unapply(s: String): Option[(String, Int)] = s.split(":") match {
-    case Array(host, PortStr(port)) => Some((host, port))
+  def unapply(s: String): Option[(String, String)] = s.split(":") match {
+    case Array(host, DnsLabel(port))
+      // DNS labels may not exceed 63 characters in length
+      if port.length <= 63 => Some((host, port))
     case _ => None
   }
 }
```
adleong: I think this change allows you to delete a bunch of `.toString` calls on port later in this file.

hawkw: Oh, you're probably right – I'll remove those.

hawkw: Done 6460c2e2d3f0dffc9a32298e3aaa16ff1932d017 – I think this should be merge-able whenever?

PR #1527, opened 2017-07-20
Make Kubernetes Objects watchable

hawkw:
Previously, only Kubernetes list resources were watchable, since our code for establishing watches on Kubernetes resources used the `?watch=true` path param.

I've rewritten our code to establish watches using the `$API/$NS/watch/$RESOURCE` endpoint, instead, thus making the `KubeObject` type watchable as well as `KubeList`. I've also factored out some commonly repeated code for creating an `Activity` to watch a Kubernetes resource, and put that in the `Watchable` trait.

The unit tests in`EndpointsNamerTest`, `ServiceNamerTest`, and `ApiTest` have all been updated to cover the new `Watchable` implementation. Additionally, I believe @klingerf has tested this somewhat on a live Kubernetes cluster.


k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -50,6 +38,37 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+    fieldSelector: Option[String] = None,
+    resourceVersion: Option[String] = None,
+    retryIndefinitely: Boolean = false,
+    watch: Boolean = false
+  ): Future[G] = {
+    val req = Api.mkreq(
+      http.Method.Get,
+      if (watch) watchPath else path,
+      None,
+      "labelSelector" -> labelSelector,
```
olix0r: tioli:

i tend to constify strings -- with, somewhere on an object:
```scala
   private[pkgname] val LabelSelector = "labelSelector"
```

only because typos become compile errors.

we'd want to use these static keys throughout

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +156,64 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
 
     (_watch(resourceVersion), Closable.ref(close))
   }
+
+  /**
+   * Convert this Watchable into an [[Activity]]
+   * @param resourceVersion whether or not to send the initial resource
+   *                        version of the watched resource. this is a
+   *                        special case due to errors with the
+   *                        ConfigMap interpreter
```
klingerf: I think more generally you want to supply the resourceVersion when calling Watch List endpoints, but you don't want to include it when calling the Watch (object) endpoints, since it may be too old. In fact, maybe there's a way to enforce that in the code as well?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +156,64 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+   * @param onEvent function called on each [[Watch]] event
+   * TODO: i wish this returned a [[Activity.State]] so that the caller
+   *       of this function could turn a [[Watch.Error]] into an
+   *       [[Activity.Failed]] if they wanted watch errors to fail this
+   *       [[Activity]], although none of our current code exhibits this
+   *       use-case...
+   *         - eliza, 7/18/2017
+   * @return
```
klingerf: Missing return description here?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +156,64 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+      val closeRef = new AtomicReference[Closable](Closable.nop)
+      val pending = get(retryIndefinitely = true)
+        // if the initial GET failed, then the activity is a failure
+        .onFailure { e =>
+          log.warning(s"k8s failed to get resource at $path: $e")
+          state.update(Activity.Failed(e))
+        }
+        // otherwise, update the activity with the initial state, and
+        // apply the onEvent function to each successive watch event in
+        // the stream
```
olix0r: Minor note, I prefer when comments end in punctuation so it's clear that it wasn't half a sentence.  I write lots of incomplete comments, so using proper capitalization and punctuation help me quickly get a sense for how complete my comments are.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -28,7 +30,9 @@ trait Resource extends Closable {
  * @tparam O The parent type for all [Object]s served by this version.
  */
 private[k8s] trait Version[O <: KubeObject] extends Resource {
-  val path = s"/$group/$version"
+
+  override val path = s"/$group/$version"
+  override val watchPath = s"/$group/$version/watch"
```
klingerf: TIOLI -- could write this as `watchPath = s"$path/watch"`.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -91,9 +106,11 @@ trait ThirdPartyVersion[O <: KubeObject] extends Version[O] {
   /** version within owner domain, i.e. "v1" */
   def ownerVersion: String
 
-  def version: String = ThirdPartyVersion.version(owner, ownerVersion)
-  def basePath: String = s"/${ThirdPartyVersion.group}/$version"
-  val group = ThirdPartyVersion.group
+  override val version: String = ThirdPartyVersion.version(owner, ownerVersion)
+  override val path: String = s"/${ThirdPartyVersion.group}/$version"
+  override val watchPath: String = s"/${ThirdPartyVersion.group}/$version/watch"
```
klingerf: Same comment here: `s"$path/watch"`.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -113,51 +130,49 @@ class NsThirdPartyVersion[O <: KubeObject](client: Client, owner: String, ownerV
...
-      (list.items.map(od.toWatch), list.metadata.flatMap(_.resourceVersion))
-    }
+    get(
+      labelSelector,
+      fieldSelector,
+      None,
+      retryIndefinitely = true,
+      watch = true
+    )
+      .map { list =>
```
klingerf: Tiny style nit -- I'd collapse this onto one line:

```
).map { list =>
```

Same goes for line 235 in this file.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -174,19 +189,22 @@ private[k8s] class NsListResource[O <: KubeObject: TypeReference, W <: Watch[O]:
 }
 
 private[k8s] class NsObjectResource[O <: KubeObject: TypeReference, W <: Watch[O]: TypeReference](
-  val client: Client,
-  listPath: String,
+  parent: Resource,
   objectName: String,
+  maybeListName: Option[String] = None,
```
klingerf: I have a slight preference for just calling this `listName`; that's a bit more consistent with how we name option params elsewhere, but definitely not formalized in any way.

k8s/src/main/scala/io/buoyant/k8s/v1.scala
```diff
@@ -72,12 +72,43 @@ package object v1 {
     def toWatch(e: Service) = ServiceModified(e)
   }
 
+  @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
```
klingerf: I'd include the changes in this file with your next branch, instead of this one.

k8s/src/test/scala/io/buoyant/k8s/v1/ApiTest.scala
```diff
@@ -383,4 +387,144 @@ class ApiTest extends FunSuite with Awaits with Exceptions {
     } finally await(closable.close())
     if (failure != null) throw failure
   }
+
+  val configMap0 =
```
klingerf: And likewise would move these tests to the next branch as well.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: Okay, I'll keep that in mind for the future, thanks.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: good call!

k8s/src/main/scala/io/buoyant/k8s/resources.scala
hawkw: Yeah, it was like that for a while, but then I changed these to `lazy val`s based on the observation that we probably wouldn't have to allocate both strings. @adleong quashed that & I never changed it back...

k8s/src/main/scala/io/buoyant/k8s/resources.scala
hawkw: I think Scalariform decided it should be like that?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
hawkw: I think I called it that because I wanted to use the identifier `listName` somewhere else, but that may be vestigial – I'll check.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +156,64 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+   *                        version of the watched resource. this is a
+   *                        special case due to errors with the
+   *                        ConfigMap interpreter
+   * @param onEvent function called on each [[Watch]] event
+   * TODO: i wish this returned a [[Activity.State]] so that the caller
+   *       of this function could turn a [[Watch.Error]] into an
+   *       [[Activity.Failed]] if they wanted watch errors to fail this
+   *       [[Activity]], although none of our current code exhibits this
+   *       use-case...
+   *         - eliza, 7/18/2017
```
adleong: I don't think you need to sign and date comments, git blame has your back

k8s/src/main/scala/io/buoyant/k8s/resources.scala
klingerf: IntelliJ's formatting tends to do this, but I think scalariform should be ok with having them on the same line.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +156,64 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
   }
+
+  /**
+   * Convert this Watchable into an [[Activity]]
+   * @param resourceVersion whether or not to send the initial resource
+   *                        version of the watched resource. this is a
+   *                        special case due to errors with the
+   *                        ConfigMap interpreter
+   * @param onEvent function called on each [[Watch]] event
+   * TODO: i wish this returned a [[Activity.State]] so that the caller
```
adleong: probably would want it to return Try.  You never want callers to be able to return Activity.State.Pending here or Bad Things will happen.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -138,10 +156,64 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+   *       [[Activity]], although none of our current code exhibits this
+   *       use-case...
+   *         - eliza, 7/18/2017
+   * @return
+   */
+  def activity[T](convert: G => T, resourceVersion: Boolean = true)(onEvent: (T, W) => T): Activity[T] =
+    Activity(Var.async[Activity.State[T]](Activity.Pending) { state =>
+      val closeRef = new AtomicReference[Closable](Closable.nop)
+      val pending = get(retryIndefinitely = true)
+        // if the initial GET failed, then the activity is a failure
```
adleong: is this even possible with `retryIndefiniately = true`?

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -174,19 +189,22 @@ private[k8s] class NsListResource[O <: KubeObject: TypeReference, W <: Watch[O]:
...
-      Api.parse[O](rsp)
-    }
+)(implicit od: ObjectDescriptor[O, W])
+  extends Watchable[O, W, O]
+  with Resource {
+
+  override val client: Client = parent.client
+  override val (path, watchPath) = {
+    // the value of listName can be dropped when we're done with it,
+    // so allocate it in a scope.
```
adleong: This is not in a hot path so the goal should be to maximize clarity not minimize allocations.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -202,4 +220,19 @@ private[k8s] class NsObjectResource[O <: KubeObject: TypeReference, W <: Watch[O
...
+    fieldSelector: Option[String] = None
+  ): Future[(Seq[W], Option[String])] =
+    get(
+      labelSelector,
+      fieldSelector,
+      None,
+      retryIndefinitely = true,
+      watch = true
+    )
+      .map { obj =>
```
adleong: intellij formatter: r u ok?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: probably not, but who knows?

PR #1528, opened 2017-07-20
H2 IstioIngressIdentifier and IstioLogger

### LINKED ISSUE #1480: GRPC in Istio
We still need an h2 IstioLogger, IstioIngressIdentifier, and lots of testing in order to support grpc.

esbie:
These are largely copies of their http counterparts. 
Adds the concept of loggers to the h2 protocol. Fixes #1480 

Validation: Manually tested against hello world grpc.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/istio/IstioLoggerBase.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/istio/IstioLoggerBase.scala
@@ -1,0 +1,46 @@
+package io.buoyant.k8s.istio
+
+import com.twitter.finagle._
+import com.twitter.util.Duration
+
+trait IstioLoggerBase {
+  val unknown = "unknown"
+
+  // expected istioPath
+  // Path(%,io.l5d.k8s.daemonset,default,incoming,l5d,#,io.l5d.k8s.istio,reviews.default.svc.cluster.local,az:us-west::env:prod::version:v1,http)
+  val pathLength = 10
+  val pathServiceIndex = 7
+  val pathLabelsIndex = 8
+
+  def targetService(istioPath: Option[Path]): Option[String] = istioPath.map { path =>
+    path.showElems(pathServiceIndex)
+  }
+
+  def version(istioPath: Option[Path]): Option[String] = istioPath match {
+    case Some(path) =>
+      path.showElems(pathLabelsIndex).split("::").find {
+        e => e.startsWith("version:")
+      }.map { label => label.split(":").last }
+    case _ => None
+  }
+
+  // note that this should really come from the "app" label of the target pod.
+  // in most cases this will be the same value, but eventually the correct solution
+  // should be to get this directly from the label or the uid.
+  def targetLabelApp(targetService: Option[String]): Option[String] = targetService.flatMap(_.split('.').headOption)
+
+  def mixerClient: MixerClient
+
+  def report(istioPath: Option[Path], responseCode: Int, path: String, duration: Duration) = {
+    val tService = targetService(istioPath)
+    mixerClient.report(
+      responseCode,
+      path,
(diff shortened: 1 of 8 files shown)


k8s/src/main/scala/io/buoyant/k8s/istio/IstioLoggerBase.scala
```diff
@@ -0,0 +1,46 @@
+package io.buoyant.k8s.istio
+
+import com.twitter.finagle._
+import com.twitter.util.Duration
+
+trait IstioLoggerBase {
```
siggy: nice!

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/IstioIngressIdentifier.scala
```diff
@@ -0,0 +1,128 @@
...
+              val dst = Dst.Path(path, baseDtab(), Dtab.local)
+              new IdentifiedRequest(dst, req)
+            }
+          case _ => Future.value(new UnidentifiedRequest(s"ingress path ${ingressPath.svc}:${ingressPath.port} does not match any istio vhosts"))
+        }
+    }
+  }
+
+  def redirectRequest(redir: HTTPRedirect, req: Request): Future[Nothing] = {
+    val resp = Response(Status.Found, Stream.empty(0))
```
adleong: I can't find where Stream.empty(Int) is defined.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2LoggerConfig.scala
```diff
@@ -0,0 +1,56 @@
+package io.buoyant.linkerd.protocol.h2
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.buoyant.h2.{Request, Response}
+import com.twitter.finagle.{Filter, ServiceFactory, Stack, Stackable}
+import com.twitter.finagle.stack.nilStack
+import io.buoyant.config.PolymorphicConfig
+
+abstract class H2LoggerConfig extends PolymorphicConfig { config =>
```
siggy: you may have already looked at this, but any thoughts on sharing code with `HttpLoggerConfig`? they're pretty similar, though i imagine the `Request`, `Response` types could be a pain to abstract.

same question for the `IstioLogger` classes.

esbie: I looked into it briefly, but let me squint at it some more

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/IstioIngressIdentifier.scala
esbie: Bleh, yeah thanks for looking at this, it's a different Stream. Will update to h2.Stream

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2LoggerConfig.scala
esbie: `HttpLoggerConfig`/`H2LoggerConfig`: `new Stack.Module` itself already does good job of reducing to the essentials.

`IstioLoggerConfig`: I could pull out host, port, client, etc into it's own trait (`IstioLoggerConfig extends HttpConfig with IstioLoggerConfigBase`), but having overrides in all three places was unnecessarily confusing. 


In general I only focus on DRYing things up when there are 3 copies of something, and didn't find a lot of value in sharing code between configs.

PR #1541, opened 2017-07-27
Fix incorrect retries stat in router summary of admin dashboard

rmars:
Problem: we were only fetching stats for services named 'svc'
Solution: remove hardcoded service name from path stats accessor

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/src/router_summary.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/src/router_summary.js
@@ -89,23 +89,23 @@ define([
     }
 
     function processResponses(data, routerName, metrics) {
-      function process(metric) {
+      function summarize(metric) {
         var m = metrics[metric];
         var datum = _(data).get(m.accessor);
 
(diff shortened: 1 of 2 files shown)


admin/src/main/resources/io/buoyant/admin/js/src/router_summary.js
```diff
@@ -94,14 +93,10 @@ define([
...
-          return _.get(datum, m.metricAccessor) || 0;
-        } else {
-          return _.reduce(datum, function(mem, entityData) {
-            mem += _.get(entityData, m.metricAccessor) || 0;
-            return mem;
-          }, 0);
-        }
+        return _.reduce(datum, function(mem, entityData) {
+          mem += _.get(entityData, m.metricAccessor) || 0;
+          return mem;
```
siggy: nit, and realize this was already here, but, should this be `value` or `stat` or something, not `mem`?

rmars: totally can name this better :)

PR #1542, opened 2017-07-27
Stream-aware classifiers for H2 client stats

### LINKED ISSUE #1537: H2 client stats should use H2StreamClassifier
We should do an audit of the stats modules in the H2 client stack and make sure everything is wired up correctly.  I think we currently have:

* StreamStatsFilter 👌 
* PerDstPathStatsFilter 
  * This filter (which lives in the client stack) records some stats (success, failures, and latency, I think) scoped per service.  This gives us a breakdown of the client metrics by the services that called them.  We should replace this filter with one dervied from StreamStatsFilter so that the stats that it reports are stream aware also.

Furthermore, we should have the client stack use the H2StreamClassifier that is set in the service config.  The commit where we added this for general finagle response classifiers is here: https://github.com/linkerd/linkerd/commit/a7939a65e1e771fd4e238c3893a2e2bebd17109a

We now need to do the same thing for H2StreamClassifiers.  Basically, we need a module in the path stack that replaces `ResponseClassifierCtx.Setter.module` and sets the H2StreamClassifier into a local context.  Then we need something analogous to the `LocalClassifierStatsFilter` that passes the H2StreamClassifier from the local context into a StreamStatsFilter.  Similarly for `PerDstPathStatsFilter`.

Then we should do a round of manual or e2e testing to check that both the service and client stats (and per-service client stats) are correct for the various H2 response classifiers.  (although I guess there are effectively only 2 as they pertain to stats: one that classifies 5XX as success and one that classifies 5XX as failure.)

hawkw:
Although I added stream-based response classification to the H2 router's `StreamStatsFilter` in  b81efb74a48aecda7a064af20aec1362a6e7b6f7, the `LocalClassifierStatsFilter` and `PerDstPathStatsFilter` were still wired up to use standard finagle `ResponseClassifier`s. In order to make client stats also stream-aware, these filters need to be rewritten to use `H2StreamClassifier`s as well.

I've added a `StreamClassifierCtx` `LocalKey` to the H2 path stack that functions analogously to the `ResponseClassifierCtx` in other protocols, and I've added `PerDstPathStreamStatsFilter` and `LocalClassifierStreamStatsFilter` types that operate similarly to their non-stream-aware counterparts, but based on `StreamStatsFilter` rather than on the standard `StatsFilter`.

I've added a unit test class, `PerDstPathStreamStatsFilterTest`, and end-to-end test class, `StreamClassificationEndToEndTest`. These tests are both based their non-stream-aware counterparts, `PerDstPathStatsFilterTest` and `ResponseClassificationEndToEndTest`, but modified to test the new stats filters. Note that some of the tests in the end-to-end test are currently ignored, since they test for classified retries functionality which isn't currently available; they should be un-ignored when #1523 lands.

Closes #1537

### DIFF
--- a/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
+++ b/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
@@ -15,21 +15,22 @@ object StreamTestUtils {
     * @param stream the [[Stream]] to read to the end
     * @return a [[Future]] that will finish when the whole stream is read
     */
-  final def readAll(stream: Stream): Future[Unit] =
+  final def readToEnd(stream: Stream): Future[Unit] =
     stream.read().flatMap { frame =>
       val end = frame.isEnd
       frame.release().before {
-        if (end) Future.Done else readAll(stream)
+        if (end) Future.Done else readToEnd(stream)
       }
     }
 
   /**
-    * Enhances a [[Stream]] by providing the [[readAll()]] function in the
+    * Enhances a [[Stream]] by providing the [[readToEnd()]] function in the
     * method position
+ *
     * @param stream the underlying [[Stream]]
     */
   implicit class ReadAllStream(val stream: Stream) extends AnyVal {
-    @inline def readAll: Future[Unit] = StreamTestUtils.readAll(stream)
+    @inline def readToEnd: Future[Unit] = StreamTestUtils.readToEnd(stream)
   }
 
 }
--- a/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
+++ b/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
@@ -16,14 +16,13 @@ object StreamTestUtils {
    * @return a [[Future]] that will finish when the whole stream is read
    */
   final def readToEnd(stream: Stream): Future[Unit] =
-    if (stream.isEmpty) { Future.Done }
-    else {
+    if (stream.isEmpty) Future.Unit
+    else
       stream.read().flatMap { frame =>
         val end = frame.isEnd
         frame.release().before {
-          if (end) Future.Done else readToEnd(stream)
+          if (end) Future.Unit else readToEnd(stream)
         }
-      }
     }
 
   /**
--- a/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
+++ b/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
@@ -26,7 +26,6 @@ object StreamTestUtils {
       }
     }
 
-
   /**
    * Enhances a [[Stream]] by providing the [[readToEnd()]] function in the
    * method position
--- a/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
+++ b/finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
@@ -5,16 +5,16 @@ import com.twitter.util.Future
 
 object StreamTestUtils {
   /**
-    * Read a [[Stream]] to the end, [[Frame.release release()]]ing each
-    * [[Frame]] before reading the next one.
-    *
-    * The value of each frame is discarded, but assertions can be made about
-    * their contents by attaching an [[Stream.onFrame onFrame()]] callback
-    * before calling `readAll()`.
-    *
-    * @param stream the [[Stream]] to read to the end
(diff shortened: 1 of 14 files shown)


hawkw:
Oh, I've also added a time interval stack param to `StreamStatsFilter`, since the regular `StatsFilter` has one. Functionality should remain unchanged by default.

router/h2/src/main/scala/io/buoyant/router/h2/LocalClassifierStreamStatsFilter.scala
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Like [[io.buoyant.router.LocalClassifierStatsFilter]],
+ * but specialized for H2 streams.
+ */
+object LocalClassifierStreamStatsFilter {
+
+  def module: Stackable[ServiceFactory[Request, Response]] =
+    new Stack.Module2[param.Stats, StreamStatsFilter.Param, ServiceFactory[Request, Response]] {
+      val role: Stack.Role = PerDstPathStatsFilter.role
+      val description = "Report request statistics for each logical destination"
```
adleong: should this be "Report request statistics using local H2 stream classifier"

router/h2/src/main/scala/io/buoyant/router/h2/LocalClassifierStreamStatsFilter.scala
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Like [[io.buoyant.router.LocalClassifierStatsFilter]],
+ * but specialized for H2 streams.
+ */
+object LocalClassifierStreamStatsFilter {
+
+  def module: Stackable[ServiceFactory[Request, Response]] =
+    new Stack.Module2[param.Stats, StreamStatsFilter.Param, ServiceFactory[Request, Response]] {
+      val role: Stack.Role = PerDstPathStatsFilter.role
```
adleong: Should this be `StatsFilter.role`?  Or possibly `LocalClassifierStatsFilter.role`?  (the choice is academic, they have the same value)

router/h2/src/main/scala/io/buoyant/router/h2/LocalClassifierStreamStatsFilter.scala
```diff
@@ -0,0 +1,48 @@
...
+            def mkClassifiedStatsFilter(path: Path): SimpleFilter[Request, Response] = {
+              val H2StreamClassifier(classifier) =
+                StreamClassifierCtx.current.getOrElse(H2StreamClassifier.param.default)
+              new StreamStatsFilter(stats, classifier, timeUnit)
+            }
+
+            val filter = new PerDstPathFilter(mkClassifiedStatsFilter _)
+            filter.andThen(next)
+
+          // can this actually be null? the HTTP1 `PerDstPathStatsFilter`
```
adleong: isNull checks to see if the StatsReceiver is the NullStatsReceiver.  In that case there's no point adding the StatsFilter so we just skip it.

namer/core/src/main/scala/io/buoyant/hostport.scala
```diff
@@ -10,19 +10,18 @@ import com.twitter.finagle.Path
...
   /**
-    * regex for capturing strings that conform to the definition
-    * of a "label" in RFCs 1035 and 1123, used for the `port` part
-    * of a [[HostColonPort]].
-    *
-    * this is based on the `DNS_LABEL` regex in Kubernetes:
-    * https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L40-L43
-    */
-  private[this] val DnsLabel  = """([a-z0-9][-a-z0-9]*[a-z0-9]?)""".r
+   * regex for capturing strings that conform to the definition
```
hawkw: whoops I did not mean to commit this

router/h2/src/main/scala/io/buoyant/router/h2/LocalClassifierStreamStatsFilter.scala
hawkw: Ah, okay, so that's not actually checking for `null`, gotcha. Good to know

finagle/h2/src/test/scala/io/buoyant/test/h2/StreamTestUtils.scala
```diff
@@ -0,0 +1,39 @@
+package io.buoyant.test.h2
+
+import com.twitter.finagle.buoyant.h2.{Frame, Stream}
+import com.twitter.util.Future
+
+object StreamTestUtils {
```
hawkw: :D

hawkw:
@adleong, I think 438780e36fcb22e33dceb7555e83b40d7fa7c434 addresses all your comments?

PR #1543, opened 2017-07-27
Add ExceptionStatsHandler to StreamStatsFilters

hawkw:
While working on tests for #1542, I noticed that the standard `StatsFilter` also has an `ExceptionStatsHandler` stack param, which `StreamStatsFilter` lacks. This param is used for counting the number of occurrences of specific exceptions that occur.

I thought we might want to have that for stream stats as well (and @adleong confirmed). This PR adds the `ExceptionStatsHandler` to both `StreamStatsFilter` and the `LocalClassifier` and `PerDstPath` stream stats filter derivatives added in #1542. 

The only aspect of this PR that I think is complicated enough to comment on in detail is that I've added a public re-export of `ResponseClassificationSyntheticException` to `com.twitter.finagle.buoyant`. This exception appears to be used by the `ExceptionHandler` for cases where the classifier determined that there was a failure, but there was no associated exception thrown. Unfortunately, it's declared as `private[finagle]`, so in order to use it in `StreamStatsFilter` (which lives in the `io.buoyant.router.h2` package), I needed to re-export it from somewhere that lives inside the `finagle` package.

I've re-enabled the tests in `PerDstPathStreamStatsFilterTest` that expect per-exception stats counters. These were commented out when I modified that test class from the non-stream-aware version to test #1542, since we weren't currently generating those stats. Now that we are, they can be re-enabled.


hawkw:
(trying to figure out why this broke on CI, it builds in dev. retrying the build & if that fails, I'll do some additional digging)

hawkw:
Okay, this is real weird: the test appears to be failing when run from SBT, but not when run from IDEA (which is why I committed this)...

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,64 +1,104 @@
...
+      failures.incr()
+      exceptionStats.record(stats, e)
     }
 
     private[this] val frameBytes = stats.stat("data_bytes")
 
     def apply(
       startT: Stopwatch.Elapsed,
-      onFinalFrame: Option[Try[Frame]] => Unit = { _ => }
+      onFinalFrame: Option[Try[Frame]] => Try[Unit] => Unit = { _ => _ => }
```
adleong: I don't think we want to pass an addition `Try[Unit]` in here to capture the potential exception.  Even if we did, `Option[Throwable]` is probably a better type than `Try[Unit]`.  See below for suggestions on how to eliminate this.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -1,64 +1,104 @@
...
     )(underlying: Stream): Stream = {
       val streamFrameBytes = new AtomicLong(0)
       val stream = if (underlying.isEmpty) {
         // if the stream is empty, we still need to call the response
         // classifier with `None`, so add that to the `onEnd` callback.
-        underlying.onEnd.respond { _ => onFinalFrame(None) }
+        underlying.onEnd.respond { onFinalFrame(None)(_) }
```
adleong: I believe an empty Stream always has `onEnd` satisfied.  I think we can just immediately and synchronously call `onFinalFrame(None)`.

router/h2/src/main/scala/io/buoyant/router/h2/StreamStatsFilter.scala
```diff
@@ -108,10 +150,14 @@ class StreamStatsFilter(statsReceiver: StatsReceiver, classifier: H2StreamClassi
...
 
-    @inline def classify(rsp: Try[Response])(frame: Option[Try[Frame]]): Unit =
+    @inline def classify(rsp: Try[Response])(frame: Option[Try[Frame]])(e: Try[Unit] = Return(())): Unit =
       classifier(H2ReqRep(req1, rsp.map((_, frame)))) match {
         case Successful(_) => successes.incr()
-        case _ => failures.incr()
+        case _ =>
+          exceptionStats.record(
+            statsReceiver,
+            e.asScala.failed.getOrElse(StreamStatsFilter.SyntheticException)
```
adleong: `e` should be purely derivable from `rsp` and `frame`.

```
// something like...
def failed[T](t: Try[T]): Option[Throwable] = t match { case Return(_) => None; case Throw(e) => Some(e) }
val e = failed(rsp).orElse(frame.flatMap(failed)).getOrElse(SSF.SE)
```

hawkw: ah, yeah, that's much better, thanks!

hawkw:
Okay, finally fixed the failing test – it turns out the tests were expecting exceptions to be in the `io.buoyant.router` package and not `io.buoyant.router.h2`, which I totally overlooked. Now I feel silly.

adleong:
I think latest master needs to be merged here.  The diff should be much smaller now, given the overlap with #1542

hawkw:
@adleong does this seem right, now?

PR #1544, opened 2017-07-27
Add H2StreamClassifier to ClassifierFilter

### LINKED ISSUE #1538: H2 server stats should use H2StreamClassifier
The H2 server StreamStatsFilter should use the response classifier configured on the service.  

For HTTP, `ClassifierFilter` on the path stack classifies the ReqRep and then sets the response class onto a response header.  Then the server stack is configured with a response classifier that looks at the value of that header.

For H2 we probably have to do things differently.  We will need an H2 version of `ClassifierFilter` that uses `StreamProxy` and classifies the H2ReqRep on the final frame.  If that final frame is a Trailers frame, then we can set the response class into the Trailers.  Then we should configure the server stack with a H2StreamClassifier that looks at this response class in the trailers.

We should then do a round of manual or e2e testing to ensure that the server stats are consistent with the classifier set on the service.

hawkw:
Previously, the server stack `ClassifierFilter` did not use a stream-aware `H2StreamClassifier`, and wasn't able to set the `l5d-success-class` header in the trailers frame of H2 streams. The `ClassifierFilter` needs to be able to make classification decisions based on the final frame of a `Stream` as well as the status code, and if the final frame is a Trailers frame, the response class header should be set into the trailers frame.

In this PR, I've rewritten the H2 `ClassifierFilter` to use `H2StreamClassifier` as a stack parameter, and to write response class headers into Trailers frames, if they are present. Otherwise, the header is written to the response headers.

I've modified `StreamClassificationEndToEndTest` to include several e2e tests for these changes.

Closes #1538


router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -11,17 +15,17 @@ object ClassifierFilter {
...
+      val description = "Sets the stream classification into a header"
+      def make(classifierP: param.H2StreamClassifier, next: ServiceFactory[Request, Response]) = {
+        val param.H2StreamClassifier(classifier) = classifierP
         new ClassifierFilter(classifier).andThen(next)
       }
     }
 
-  val successClassClassifier: ResponseClassifier = {
-    case ReqRep(req, Return(rep: Response)) if rep.headers.contains(SuccessClassHeader) =>
+  val successClassClassifier: H2StreamClassifier = {
```
adleong: it looks like frame is unused here.  ie if the response class is written into the trailers, we never check it.  https://github.com/linkerd/linkerd/pull/1545 would actually make this easier, allowing us to classify based on the response class header if it exists, and then based on the response class trailer, if it exists.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -33,18 +37,30 @@ object ClassifierFilter {
...
-class ClassifierFilter(classifier: ResponseClassifier) extends SimpleFilter[Request, Response] {
+class ClassifierFilter(classifier: H2StreamClassifier) extends SimpleFilter[Request, Response] {
 
   private[this] val successClass = classifier.andThen(_.fractionalSuccess).lift
 
   def apply(req: Request, svc: Service[Request, Response]): Future[Response] = {
-    svc(req).map { rep =>
-      successClass(ReqRep(req, Return(rep))) match {
-        case Some(success) =>
+    svc(req).map {
```
adleong: https://github.com/linkerd/linkerd/pull/1545 would probably help with this too.  If we can early classify the response, then we should write the response class into a header.  If not, we should wait until the final frame, classify, and write the response class into the final frame if it's a trailers.

The one case that doesn't get covered is if we need the final frame for classification but it's not a Trailers frame.  It's a pretty weird case and I'm okay with punting on it.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -33,18 +37,30 @@ object ClassifierFilter {
...
+        rep
+      case rep =>
+        val stream = rep.stream.onFrame {
+          case Return(f: Trailers) =>
+            successClass(H2ReqRep(req, Return(rep), Some(Return(f)))).foreach { success =>
+              f.set(ClassifierFilter.SuccessClassHeader, success.toString)
+            }
+          case Return(f) if f.isEnd =>
+            successClass(H2ReqRep(req, Return(rep), Some(Return(f)))).foreach { success =>
+              rep.headers.set(ClassifierFilter.SuccessClassHeader, success.toString)
```
adleong: This doesn't work.  By the time we get to the final frame, the response has already been transmitted so mutating the headers here will have no effect.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
hawkw: I like your work on #1545 and I think that seems right – I'll happily make that change when we land #1545.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
hawkw: Yeah, you're right, that won't work, I wasn't thinking. I'm unsure what to do if the final frame is not a Trailers frame – you're right that it's a pretty weird case, but I'd like to cover it nonetheless.  Unfortunately, we can't just add a Trailers frame from inside of the `onFrame` method...

hawkw: I'm wondering if we can add a method that is to `onFrame` as `map` is to `foreach`...

adleong: or flatMap even.  But I dont know if we should be adding Trailers frames to responses that don't have them to begin with

hawkw: Yeah, I just can't think of any other way to handle that case?

adleong: keep in mind that classifiers are partial functions.  in this case I think we just fail to classify

hawkw: Will failing to set the `l5d-success-class` header cause problems elsewhere?

adleong: We just need to make sure the responseClassClassifier can handle this value being absent.  And likewise we need to make sure everything that uses H2Classifiers can handle the case where the PF is undefined.  We may want a "unclassified" stats counter in addition to "success" and "failures".

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -11,40 +14,71 @@ object ClassifierFilter {
...
-          rep.headers.set(ClassifierFilter.SuccessClassHeader, success.toString)
-        case None =>
+    svc(req).map { rep: Response =>
+      if (rep.stream.isEmpty) {
+        // classify early - response class goes in headers
+        val success =
+          classifier.responseClassifier
+            .lift(H2ReqRep(req, Return(rep)))
+            // the responseClassifier is non-total, it won't classify successes
+            .getOrElse(ResponseClass.Success)
```
adleong: dont want to do a `getOrElse` here, we can just check above that the pf is defined here.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -11,40 +14,71 @@ object ClassifierFilter {
...
+        // late classification on the last frame in the response stream
+        val stream = rep.stream.onFrame {
+          case Return(frame: Trailers) =>
+            val success =
+              classifier
+                .streamClassifier(H2ReqRepFrame(req, Return(rep), Some(Return(frame))))
+            frame.set(SuccessClassHeader, successHeader(success))
+          case _ =>
+        }
+        Response(rep.headers, stream)
```
adleong: add `te: trailers` header.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -11,40 +14,71 @@ object ClassifierFilter {
...
+      } else {
+        // if the early classification attempt is not defined, attempt
+        // late classification on the last frame in the response stream
+        val stream = rep.stream.onFrame {
+          case Return(frame: Trailers) =>
+            val success =
+              classifier
+                .streamClassifier(H2ReqRepFrame(req, Return(rep), Some(Return(frame))))
+            frame.set(SuccessClassHeader, successHeader(success))
+          case _ =>
```
adleong: in this case we want to add a Trailers frame to the stream.  To do this we probably need something like

```
class StreamFlatMap(underlying: Stream, f: Frame => Seq[Frame]) extends Stream { 
  val q = new QueueOfSomeKind
  def read(): Future[Frame] = {
    if (q.nonEmpty) Future.value(q.pop)
    else underlying.read().map(f).map { fs => q.push(fs) ; q.pop }
  }
}
```

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifiers.scala
```diff
@@ -127,21 +131,29 @@ object H2Classifiers {
...
 
   case object AllSuccessful extends H2Classifier {
-    override val responseClassifier = {
+    override val responseClassifier: PartialFunction[H2ReqRep, ResponseClass] = {
       case H2ReqRep(_, Throw(_)) => ResponseClass.NonRetryableFailure
+      case H2ReqRep(_, Return(_)) => ResponseClass.Success
     }
-    override val streamClassifier = {
+    override val streamClassifier: PartialFunction[H2ReqRepFrame, ResponseClass] = {
+      case H2ReqRep(reqRep: H2ReqRep) if responseClassifier.isDefinedAt(reqRep) =>
```
hawkw: The `H2ReqRep` extractor may not be the clearest way to handle this, you're right. I wanted `streamClassifier` to have a way to fall back to the `responseClassifier` if it gets a `H2ReqRepFrame` with an empty stream, which was happening...somewhere (can't remember precisely atm) and causing some of our existing tests to break.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifiers.scala
```diff
@@ -127,21 +131,29 @@ object H2Classifiers {
...
+        ) => ResponseClass.NonRetryableFailure
       case _ => ResponseClass.Success
     }
   }
 
   case object AllSuccessful extends H2Classifier {
-    override val responseClassifier = {
+    override val responseClassifier: PartialFunction[H2ReqRep, ResponseClass] = {
       case H2ReqRep(_, Throw(_)) => ResponseClass.NonRetryableFailure
+      case H2ReqRep(_, Return(_)) => ResponseClass.Success
```
hawkw: This case is for handling responses with no response stream – otherwise, a bunch of responses that should be considered successful were being marked as failures.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -11,40 +14,71 @@ object ClassifierFilter {
...
-          Try(value.toDouble).toOption
-        }.getOrElse(0.0)
-      if (success > 0.0)
-        ResponseClass.Successful(success)
-      else
-        ResponseClass.Failed(false)
+  private[this] object ResponseSuccessClass {
+    @inline def unapply(headers: Headers): Option[ResponseClass] =
+      headers.get(SuccessClassHeader).map { value =>
+        val success = value.toDouble
```
hawkw: I was considering that, just wasn't sure what the correct way to handle the error was in this case.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -11,40 +14,71 @@ object ClassifierFilter {
...
+  private[this] val successHeader: ResponseClass => String =
+    _.fractionalSuccess.toString
   def apply(req: Request, svc: Service[Request, Response]): Future[Response] = {
-    svc(req).map { rep =>
-      successClass(ReqRep(req, Return(rep))) match {
-        case Some(success) =>
-          rep.headers.set(ClassifierFilter.SuccessClassHeader, success.toString)
-        case None =>
+    svc(req).map { rep: Response =>
+      if (rep.stream.isEmpty) {
```
hawkw: I'm not sure if this will work like that – in order to call `responseClassifier.isDefinedAt`, I'd have to take the `Response` and wrap it in a `H2ReqRep`, and whether or not the response also has a non-empty stream, the `responseClassifier` will be defined...

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
hawkw: ☝️see above

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -11,40 +14,71 @@ object ClassifierFilter {
...
+    }
 
-  private[this] val successClass = classifier.andThen(_.fractionalSuccess).lift
+    override val responseClassifier: PartialFunction[H2ReqRep, ResponseClass] = {
+      case H2ReqRep(_, Return(rsp)) if rsp.headers.contains(SuccessClassHeader) =>
+        rsp.headers.get(SuccessClassHeader).map { value =>
+          val success = value.toDouble
+          if (success > 0.0) ResponseClass.Successful(success)
+          else ResponseClass.Failed(false)
+        }.get
```
hawkw: You mean like [this](https://github.com/linkerd/linkerd/pull/1544/commits/0e517f7a1ea2b76f0e6949430632d1b3ba102d46#diff-89dd41f75da774153feac265bfc1574aR29)?

hawkw: I think I only removed that because it seemed overengineered at the time, but there *might* have been a reason it didn't work that I'm forgetting. I can possibly see about re-adding it.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifiers.scala
adleong: this will match all responses, regardless of if the stream is empty or not.

if the stream is empty, this should be undefined and we should call the stream classifier with `Return(req, Return((rep, None)))` which should fall into the `_ => ResponseClass.Success` case

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifiers.scala
adleong: theoretically, if the responseClassifier is defined for this H2ReqRep, we would never even call the streamClassifier...

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
adleong: See above, the responseClassifier should be undefined in this case.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
adleong: Response is not a Headers so I don't see how `case H2ReqRep(_, Return(ResponseSuccessClass(c)))` could match.  But yeah, if you add an extractor for Response, I think it would work.

hawkw: ag, you're right, that's why this didn't work – `Response` _has_ a `Headers`, and I must've somehow gotten it into my head that it _is_ a `Headers`. 🙄

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -1,23 +1,53 @@
...
+ * @param underlying the [[Stream]] wrapped by this proxy
+ * @param onFrame function called for each [[Frame]] in the underlying [[Stream]]
+ */
+// TODO: consider renaming `onFrame` to `foreach`?
+class StreamOnFrame(underlying: Stream, onFrame: Try[Frame] => Unit) extends StreamProxy(underlying) {
+  override def read(): Future[Frame] = underlying.read().respond(onFrame)
   override def toString: String = s"StreamProxy($underlying, onFrame=$onFrame)"
 }
+
+class StreamFlatMap(underlying: Stream, f: Try[Frame] => Seq[Frame]) extends StreamProxy(underlying) {
```
adleong: I think we want `f` to be of type `Frame => Seq[Frame]`.  In the case where we fail when reading a Frame off of the underlying stream, I think we want to also fail the resulting stream.  

To put this in context: if we read part of a response stream and then hit an error, I think we want to transmit that error to the caller (or retry the whole request).  I don't think we want to tack a Trailers frame onto the end of the incomplete stream.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,99 @@
...
+        .map { success =>
+          // classify early - response class goes in headers
+          rep.headers.set(SuccessClassHeader, successHeader(success))
+          rep
+        }
+        .getOrElse {
+          // if the early classification attempt is not defined, attempt
+          // late classification on the last frame in the response stream
+          val stream = rep.stream.flatMap {
+            case Throw(e) =>
```
adleong: I think we just want to pass the error along (see above comment on flatMap)

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,99 @@
...
+              // if the final frame is a Return, but not a Trailers,
+              // then we need to send the final frame followed by a new
+              // Trailers frame
+              val success =
+                successHeader(classifier
+                  .streamClassifier(H2ReqRepFrame(req, Return(rep), Some(Return(frame)))))
+              Seq(frame, Trailers(SuccessClassHeader -> success))
+            case Return(frame) => Seq(frame)
+          }
+          Response(rep.headers, stream)
```
adleong: I think we need to add the "te: trailers" header here.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,99 @@
...
+      classifier.responseClassifier
+        .lift(H2ReqRep(req, rep))
+        .map { success =>
+          // classify early - response class goes in headers
+          rep.headers.set(SuccessClassHeader, successHeader(success))
+          rep
+        }
+        .getOrElse {
+          // if the early classification attempt is not defined, attempt
+          // late classification on the last frame in the response stream
```
adleong: We need to handle the case where rep.stream.isEmpty here

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -34,11 +34,14 @@ trait Stream {
...
+   * Wraps this [[Stream]] with a [[StreamOnFrame]] that calls the provided function on each frame after [[read()]]
+   *
    * @param onFrame the function to call on each frame.
-   * @return a [[StreamProxy]] wrapping this [[Stream]]
+   * @return a [[StreamOnFrame]] wrapping this [[Stream]]
    */
-  def onFrame(onFrame: Try[Frame] => Unit): Stream = new StreamProxy(this, onFrame)
+  def onFrame(onFrame: Try[Frame] => Unit): Stream = new StreamOnFrame(this, onFrame)
+
+  def flatMap(f: Try[Frame] => Seq[Frame]): Stream = new StreamFlatMap(this, f)
```
adleong: Let's add some scaladoc here including the restrictions on `f` (namely that it has to give back the input frame in the output Seq or take ownership of releasing it)

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
hawkw: I wrote it like this so that we could tack a trailers frame on the end of a failing stream, but if we don't want to do that, you're definitely right.

linkerd/protocol/h2/src/e2e/scala/io/buoyant/linkerd/protocol/h2/StreamClassificationEndToEndTest.scala
```diff
@@ -266,11 +282,165 @@ class StreamClassificationEndToEndTest extends FunSuite with Awaits {
...
+    val q = new AsyncQueue[Frame]()
+    val downstream =
+      Downstream("ds", Service.mk { req =>
+        q.offer(Frame.Data("aaaaaaaa", eos =
+          false))
+        q.fail(StreamError.Local(new Exception))
+        val rsp = Response(Status.Ok, Stream(q))
+
+
+      Future.value(rsp
```
hawkw: scalariform did this

adleong: r... really?? 😱

hawkw: well, I certainly don't remember doing this. at least not on purpose

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -1,23 +1,84 @@
...
+ *       [[scala.collection.TraversableLike.flatMap() flatMap]] function on
+ *       Scala collections, but with the limitation that the `flatMap`ped
+ *       function may not change the type of the mapped element (since H2
+ *       [[Stream]]s always consist of [[Frame]]s). To be properly monadic,
+ *       we would have the `flatMap`ped function return a new [[Stream]]
+ *       which would be spliced into this stream, but constructing streams
+ *       is much more complex (you have to make an
+ *       [[com.twitter.concurrent.AsyncQueue AsyncQueue]] and offer frames
+ *       to it one by one...), so we cheat a little by having the
+ *       `flatMap`pped function yield a [[Seq]] of frames instead.
```
adleong: `flatMap`pppppppped

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -1,23 +1,84 @@
...
+ * @note that in order to avoid violating flow control, `f` must either
+ *       take ownership over the frame and release it, or return it in
```
adleong: s/release/eventually release

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -1,23 +1,84 @@
...
+class StreamOnFrame(underlying: Stream, onFrame: Try[Frame] => Unit)
+  extends StreamProxy(underlying) {
+  override def read(): Future[Frame] = underlying.read().respond(onFrame)
   override def toString: String = s"StreamProxy($underlying, onFrame=$onFrame)"
 }
+
+/**
+ * Wraps an underlying [[Stream]] with a function `f` that is called on each
+ * frame in the stream. The sequence of [[Frame]]s yielded by `f` will be
+ * inserted into the stream in order at the current position.
```
adleong: Might reword this slightly to indicate that this returns a new stream which is backed by this stream with f applied on it.  imo "inserted" isn't really correct.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -1,23 +1,84 @@
...
+ * @param underlying the [[Stream]] wrapped by this proxy.
+ * @param f function called on each [[Frame]] in the underlying [[Stream]].
+ *
+ */
+class StreamFlatMap(underlying: Stream, f: Frame => Seq[Frame])
+  extends StreamProxy(underlying) {
+  private[this] var q = new mutable.Queue[Frame]()
+
+  /**
+   * @note that this stream is only empty if both the underlying stream
```
adleong: I'm a little unclear on the semantics of `isEmpty` but I'm pretty sure that this should be isEmpty iff the underlying isEmpty (regardless of q).

router/h2/src/main/scala/io/buoyant/router/H2.scala
```diff
@@ -90,8 +90,10 @@ object H2 extends Router[Request, Response]
 
     private val serverResponseClassifier =
       // TODO: insert H2 classified retries here?
```
adleong: unrelated, but we can rm this comment

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,155 @@
...
+   *       lacking this header.
+   */
+  object SuccessClassClassifier extends H2Classifier {
+    override val streamClassifier: PartialFunction[H2ReqRepFrame, ResponseClass] = {
+      case H2ReqRepFrame(_, Return((_, Some(Return(GetSuccessClass(c)))))) => c
+      case H2ReqRepFrame(_, Return((GetSuccessClass(c), Some(Return(_))))) => c
+    }
+
+    override val responseClassifier: PartialFunction[H2ReqRep, ResponseClass] = {
+      case H2ReqRep(_, Return(GetSuccessClass(c))) => c
```
adleong: I think if the response is a throw we can classify as nonretryable failure.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,155 @@
...
+   * An `H2Classifier` that classifies responses/streams based on the
+   * l5d-success-class` header.
+   *
+   * @note that this classifier is undefined for responses or streams
+   *       lacking this header.
+   */
+  object SuccessClassClassifier extends H2Classifier {
+    override val streamClassifier: PartialFunction[H2ReqRepFrame, ResponseClass] = {
+      case H2ReqRepFrame(_, Return((_, Some(Return(GetSuccessClass(c)))))) => c
+      case H2ReqRepFrame(_, Return((GetSuccessClass(c), Some(Return(_))))) => c
```
adleong: I think this case should get caught in early classification (but I guess there's no harm in having it here too)

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,155 @@
...
+    }
```
adleong: I think we need a catchall case that goes to non-retryable failure.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,155 @@
...
+                val trailers = Trailers(
+                  SuccessClassHeader -> success,
+                  "te" -> "trailers"
+                )
+                Seq(frame2, trailers)
+              case frame =>
+                // if the current frame is not an end frame, just keep going...
+                Seq(frame)
+            }
+            Response(rep.headers, stream)
```
adleong: I think we need to add `te: trailers` to the headers here.  At this point we're committed to adding a trailers frame (if one doesn't already exist).

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,155 @@
...
+            // flatMap over the frames in the response stream, since we may
+            // need to add a new Trailers frame to the end of the stream.
+            val stream = rep.stream.flatMap {
+              case frame: Trailers =>
+                // if the current frame is a Trailers frame, we've reached
+                // the end of the stream. just add the  success class header
+                // to it.
+                val success = classifyStream(req, rep, Some(frame))
+                frame.set(SuccessClassHeader, success)
+                frame.set("te", "trailers")
```
adleong: want to set this on headers, not trailers.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -1,50 +1,155 @@
...
+                // since this frame has the end of stream flag, we need to
+                // replace it with a new data frame without that flag.
+                // otherwise, the trailers we add to the stream will never
+                // be reached.
+                val frame2 = Frame.Data(frame.buf, eos = false)
+                // release the old data frame, since it will be replaced.
+                frame.release()
+                val trailers = Trailers(
+                  SuccessClassHeader -> success,
+                  "te" -> "trailers"
```
adleong: rm

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
hawkw: Hmm...I decided to override `isEmpty` because it seems to me that a `Stream` probably shouldn't claim to be empty if `read()`ing from it will yield more frames? If `q` is non-empty, `read()` on this stream will yield frames, even if the `underlying` stream `isEmpty`. Which seems...surprising, if not actually incorrect?

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
hawkw: Oops.

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
hawkw: Nevermind, I was completely wrong, this is removable.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
adleong: agree that this is confusing but we should be consistent with the existing semantics.  If you pass a non-empty queue to Stream.empty(q), `isEmpty` is `true`.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -1,23 +1,83 @@
...
+   * @note that if the [[StreamFlatMap]] proxy's internal queue of [[Frame]]s
+   *       is non-empty, a frame will be dequeued from it, rather than read
+   *       from the underlying stream. if the queue is empty, `f` will be
+   *       called on a frame read from the underlying stream before returning
+   *       it, and  additional frames may be enqueued to be read before that
+   *       frame.
+   */
+  override def read(): Future[Frame] = synchronized {
+    if (q.nonEmpty) Future.value(q.dequeue())
+    else underlying.read().map(f).map { fs =>
```
adleong: TIOLI: you could use 1 fewer map here if you want

```
read().map { frame => 
  q.enqueue(f(frame): _*)
```

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
hawkw: okay, will change this to just document that a stream may be "empty" but still return frames.

linkerd/protocol/h2/src/e2e/scala/io/buoyant/linkerd/protocol/h2/StreamClassificationEndToEndTest.scala
```diff
@@ -285,17 +373,59 @@ class StreamClassificationEndToEndTest extends FunSuite with Awaits {
 
     val stats = new InMemoryStatsReceiver
     val linker = Linker.load(config).configured(param.Stats(stats))
-    val router = linker.routers.head.initialize()
-    val server = router.servers.head.serve()
+    val router = linker.routers .head.initialize()
```
adleong: weird space

hawkw: Oh, huh, don't know how that happened.

PR #1545, opened 2017-07-28
Allow early H2 classification

adleong:
The H2StreamClassifier type requires that a response stream be complete before
the response can be classified.  This is so that classifiers may use
information from the Trailers, such as gRPC status code, to inform
classification.  However, for classifiers that only use information from the
response Headers, classification is unnecessary delayed until the stream is
complete.  Furthermore, once retries are introduced, responses will be held
and buffered until the reponse can be classified or the buffer is overfull.
Holding the response until the response stream completes introduces an
unnecessary delay when enough information exists in the response Headers to
classify that response.

Rename H2StreamClassifer to H2Classifer and enrich it to include two partial
functions: responseClassifier and streamClassifier.  When the response future
completes, the responseClassifier should be consulted.  If the
responseClassifier partial function is not defined for this input, the
streamClassifier partial function should be consulted once the response stream
is complete.  This allows classifiers to classify "early" when there is enough
information in the response headers to make a classification.


adleong:
@hawkw: I'm very interested in your feedback for this proposal

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRepFrame.scala
```diff
@@ -32,7 +33,7 @@ case class H2ReqRep(request: Request, response: Try[RepAndFrame]) {
   override def toString: String = s"H2ReqRep($request, $responseOnly, finalFrame=$finalFrame)"
 }
 
-object H2ReqRep {
+object H2ReqRepFrame {
   // type aliases to minimise typing
   type RepAndFrame = (Response, FinalFrame)
   type FinalFrame = Option[Try[Frame]]
```
hawkw: If we have separate `H2ReqRep` and `H2ReqRepFrame` types, is there any reason for this to still be an `Option`?

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
```diff
@@ -198,22 +196,15 @@ trait H2SvcConfig extends SvcConfig {
     )
 
   @JsonProperty("responseClassifier")
-  var _h2ResponseClassifier: Option[H2StreamClassifierConfig] = None
+  var _h2ResponseClassifier: Option[H2ClassifierConfig] = None
 
   @JsonIgnore
-  def h2BaseResponseClassifier = H2StreamClassifiers.Default
+  def h2ResponseClassifier: Option[H2Classifier] =
```
hawkw: This field should probably be renamed to `h2Classifier` if the class is being renamed...

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2ClassifiersTest.scala
```diff
@@ -24,10 +23,10 @@ class H2StreamClassifiersTest extends FunSuite {
     val key = H2ReqRep(req, status.map { Response(_, Stream.empty()) })
     classification match {
       case None =>
-        assert(!classifier.isDefinedAt(key))
+        assert(!classifier.responseClassifier.isDefinedAt(key))
       case Some(classification) =>
-        assert(classifier.isDefinedAt(key))
-        assert(classifier(key) == classification)
+        assert(classifier.responseClassifier.isDefinedAt(key))
+        assert(classifier.responseClassifier(key) == classification)
```
hawkw: We probably want to include tests for the stream classifiers as well, right?

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2ClassifiersTest.scala
```diff
@@ -167,19 +166,4 @@ class H2StreamClassifiersTest extends FunSuite {
       }
     }
   }
-
-  test("NonRetryableStream: makes RetryableFailures NonRetryableFailures when chunked") {
```
hawkw: Is there a reason these tests were removed instead of being rewritten to use stream classifiers?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifier.scala
```diff
@@ -0,0 +1,23 @@
...
+ * on the final Frame of the Response stream.  It is H2 specific.  When a Response future
+ * completes, the responseClassifier PartialFunction can be checked.  If responseClassifier is
+ * defined then the response can be immediately classified, just as with Finagle's
+ * ResponseClassifier.  If responseClassifier is not defined at that input, we must wait until the
+ * final Frame of the response stream has been read and then use the streamClassifier
+ * PartialFunction to classify the response.
+ */
+trait H2Classifier { self =>
+  def streamClassifier: PartialFunction[H2ReqRepFrame, ResponseClass] = PartialFunction.empty
+  def responseClassifier: PartialFunction[H2ReqRep, ResponseClass] = PartialFunction.empty
```
hawkw: TIOLI: how do you feel about overloading `apply(H2ReqRep)` and `apply(H2ReqRepFrame)` so we don't have to explicitly call `streamClassifier` and `responseClassifier`?

hawkw: (IMO that better leverages the type-level distinction between ReqRep and ReqRepFrame)...

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRepFrame.scala
```diff
@@ -1,9 +1,10 @@
 package com.twitter.finagle.buoyant.h2
 package service
 
-import com.twitter.finagle.buoyant.h2.service.H2ReqRep.{FinalFrame, RepAndFrame}
-import com.twitter.finagle.service.ReqRep
-import com.twitter.util.{Return, Throw, Try}
+import com.twitter.finagle.buoyant.h2.service.H2ReqRepFrame.{FinalFrame, RepAndFrame}
+import com.twitter.util.Try
+
+case class H2ReqRep(request: Request, response: Try[Response])
```
hawkw: Curious to know if there's any particular rationale for not just using (or extending) `ReqRep` here?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifier.scala
hawkw: Any reason the `responseClassifier` def can't just have the type `ResponseClassifier`? That might let us re-use some existing code for classifying without streams?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRepFrame.scala
adleong: I think so.  The responseClassifier only looks at the request and response.  The stream classifier looks at the final frame as well, which may or may not exist.  Technically, once the response is complete we'll be able to tell if the stream is empty and could fold the empty stream case into early classification but... conceptually it feels cleaner to handle the empty stream classification as part of stream classification.  By doing this, responseClassifiers never have to look at response.stream at all.

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2ClassifiersTest.scala
adleong: probably

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2ClassifiersTest.scala
adleong: NonRetryableStream went away.  We can do retries on streams!  Or we will be able to soon.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifier.scala
adleong: The difference between H2ReqRep and H2ReqRepFrame is subtle and the two are easily confused.  I think being explicit about which classifier you're using helps with clarity.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2ReqRepFrame.scala
adleong: ReqRep isn't type parameterized.  Given that we don't want to reuse any of finagle's classification machinery, better to have our own type.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifier.scala
adleong: This lets us be more specific with the types and I think we want to explicitly prevent ourselves from accidentally using any of the finagle classification machinery.  There are semantic differences that are subtle.  For example, finagle might classify a 200 response as successful whereas we want the responseClassifier for a 200 response to be undefined so that we can wait and see if the response stream fails.

hawkw: Fair enough!

hawkw:
@adleong thanks for answering all of my questions; I'm totally open to merging this whenever you're ready!

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/service/H2Classifier.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  def orElse(next: H2Classifier): H2Classifier = new H2Classifier {
+    override def streamClassifier: PartialFunction[H2ReqRepFrame, ResponseClass] = self.streamClassifier.orElse(next.streamClassifier)
+    override def responseClassifier: PartialFunction[H2ReqRep, ResponseClass] = self.responseClassifier.orElse(next.responseClassifier)
```
olix0r: tioli: perhaps worth (lazy?) valing these, since they'll be called multiple times per request

olix0r: we should consider (lazy?) valing these up, since this will be run multiple times per request

hawkw: lazy val! lazy val! lazy val! 😴

hawkw:
🎉

PR #1546, opened 2017-07-28
Do not allow metrics.json to modify the MetricsTree

adleong:
Calling metrics.json with the q parameter for a subtree that does not exist
creates the subtree.  For example:

```
$ curl 'http://localhost:9990/admin/metrics.json?tree=1&q=jvm/mem/allocations/eden/bytes/foo/bar/bad'
{}
$ curl 'http://localhost:9990/admin/metrics.json?tree=1&q=jvm/mem/allocations/eden/bytes'
{"gauge":3.808624E8,"foo":{"bar":{"bad":{}}}}
```

We introduce `tryResolve` as an alternative to `resolve` that does not modify
the MetricsTree and use that when rendering metrics.json.


telemetry/admin-metrics-export/src/main/scala/io/buoyant/telemetry/admin/AdminMetricsExportTelemeter.scala
```diff
@@ -33,17 +33,23 @@ class AdminMetricsExportTelemeter(
...
-      response.withOutputStream(writeFlatJson(_, subtree, pretty))
+    subtree match {
+      case Some(st) =>
+        if (tree)
+          response.withOutputStream(writeJsonTree(_, st))
+        else
+          response.withOutputStream(writeFlatJson(_, st, pretty))
+      case None =>
+        response.status = Status.NotFound
+        response.contentString = s""""No such subtree: $q""""
```
esbie: shouldn't this content string be a json object?

adleong: Maybe?  Currently it's a json string, but I'm open to suggestions.

esbie: I would like to suggest `{"error": "No such subtree: $q"}`

adleong: thank you! will update.

PR #1548, opened 2017-08-02
Add H2 Classified Failure Accrual

adleong:
Finagle's FailureAccrualFactory does not use the configured H2Classifier.

Replace FailureAccrualFactory with a new H2FailureAccrualFactory that uses
the context local H2Classifier.

### DIFF
--- a/router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
+++ b/router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
@@ -168,70 +168,76 @@ class H2FailureAccrualFactory(
     }
   }
 
-  private[this] def classifyResponse(classifier: H2Classifier, h2ReqRep: H2ReqRep): Try[Response] =
-    classifier.responseClassifier.lift(h2ReqRep) match {
-      case Some(ResponseClass.Successful(_)) =>
-        didSucceed()
-        h2ReqRep.response
-      case Some(ResponseClass.Failed(_)) =>
-        didFail()
-        h2ReqRep.response
-      case None =>
-        h2ReqRep.response match {
-          case Return(rep) =>
-            if (rep.stream.isEmpty) {
-              classifyStream(classifier, H2ReqRepFrame(h2ReqRep.request, Return((rep, None))))
-              Return(rep)
-            } else {
-              val stream = rep.stream.onFrame {
-                case Return(f) if f.isEnd =>
-                  classifyStream(classifier, H2ReqRepFrame(h2ReqRep.request, Return((rep, Some(Return(f))))))
-                case Throw(e) =>
-                  classifyStream(classifier, H2ReqRepFrame(h2ReqRep.request, Return((rep, Some(Throw(e))))))
-                case _ => ()
-              }
-              Return(Response(rep.headers, stream))
-            }
-          case Throw(e) =>
-            classifyStream(classifier, H2ReqRepFrame(h2ReqRep.request, Throw(e)))
-            Throw(e)
-        }
-    }
-
-  private[this] def classifyStream(classifier: H2Classifier, h2ReqRepFrame: H2ReqRepFrame): Unit =
-    classifier.streamClassifier(h2ReqRepFrame) match {
-      case ResponseClass.Successful(_) => didSucceed()
-      case ResponseClass.Failed(_) => didFail()
-    }
-
-  def apply(conn: ClientConnection) = {
+  def apply(conn: ClientConnection): Future[Service[Request, Response]] = {
     underlying(conn).map { service =>
       // N.B. the reason we can't simply filter the service factory is so that
       // we can override the session status to reflect the broader endpoint status.
-      new Service[Request, Response] {
-        def apply(request: Request): Future[Response] = {
-          val param.H2Classifier(classifier) =
-            H2ClassifierCtx.current.getOrElse(param.H2Classifier.param.default)
-          // If service has just been revived, accept no further requests.
-          // Note: Another request may have come in before state transitions to
-          // ProbeClosed, so > 1 requests may be processing while in the
-          // ProbeClosed state. The result of first to complete will determine
-          // whether the factory transitions to Alive (successful) or Dead
-          // (unsuccessful).
-          stopProbing()
-
-          service(request).transform { rep =>
-            Future.const(classifyResponse(classifier, H2ReqRep(request, rep)))
-          }
-        }
-
-        override def close(deadline: Time): Future[Unit] = service.close(deadline)
-        override def status: FStatus = FStatus.worst(
-          service.status,
-          H2FailureAccrualFactory.this.status
-        )
-      }
+      new FailureAccrualService(service)
     }.onFailure(onServiceAcquisitionFailure)
+  }
+
+  private[this] class FailureAccrualService(
+    underlying: Service[Request, Response]
+  ) extends Service[Request, Response] {
+    def apply(request: Request): Future[Response] = {
+      // We want to use the H2Classifier configured on the path stack so we load the classifier
+      // from the request local context.
+      val param.H2Classifier(classifier) =
+        H2ClassifierCtx.current.getOrElse(param.H2Classifier.param.default)
+      // If service has just been revived, accept no further requests.
+      // Note: Another request may have come in before state transitions to
+      // ProbeClosed, so > 1 requests may be processing while in the
+      // ProbeClosed state. The result of first to complete will determine
+      // whether the factory transitions to Alive (successful) or Dead
+      // (unsuccessful).
+      stopProbing()
+
+      underlying(request).transform { rep =>
+        Future.const(classifyResponse(classifier, H2ReqRep(request, rep)))
+      }
+    }
+
+    override def close(deadline: Time): Future[Unit] = underlying.close(deadline)
+    override def status: FStatus = FStatus.worst(
+      underlying.status,
+      H2FailureAccrualFactory.this.status
+    )
+
+    private[this] def classifyResponse(classifier: H2Classifier, h2ReqRep: H2ReqRep): Try[Response] =
+      classifier.responseClassifier.lift(h2ReqRep) match {
+        case Some(ResponseClass.Successful(_)) =>
+          didSucceed()
+          h2ReqRep.response
+        case Some(ResponseClass.Failed(_)) =>
+          didFail()
+          h2ReqRep.response
+        case None =>
+          h2ReqRep.response match {
+            case Return(rep) =>
+              if (rep.stream.isEmpty) {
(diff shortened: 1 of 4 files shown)


router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,276 @@
...
+import com.twitter.finagle.client.Transporter
+import com.twitter.finagle.liveness.{FailureAccrualFactory, FailureAccrualPolicy}
+import com.twitter.finagle.service.ResponseClass
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.{Status => FStatus, param => fParam, _}
+import com.twitter.logging.Level
+import com.twitter.util._
+import io.buoyant.router.context.h2.H2ClassifierCtx
+
+/* Modified from https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/liveness/FailureAccrualFactory.scala */
```
adleong: This file is identical to FailureAccrualFactory except for the addition of `classifyResponse` and `classifyStream` and a modified `apply` method.

router/h2/src/test/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactoryTest.scala
```diff
@@ -0,0 +1,107 @@
...
+import com.twitter.finagle.service.ResponseClass
+import com.twitter.finagle.stats.InMemoryStatsReceiver
+import com.twitter.finagle.{FactoryToService, Service, ServiceFactory, Status => FStatus}
+import com.twitter.util.{Duration, Future, MockTimer, Return}
+import io.buoyant.router.context.h2.H2ClassifierCtx
+import io.buoyant.test.FunSuite
+import java.util
+import scala.{Stream => SStream}
+
+class H2FailureAccrualFactoryTest extends FunSuite {
```
hawkw: Are these tests based on Finagle's tests for `FailureAccrualFactory`?

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
hawkw: It would be really nice if there was a way of doing this other than completely duplicating `FailureAccrualFactory`, but because `didFail` is `private[this]`, I get that there may not be an easy option...

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,276 @@
...
+  // Alive        |
+  // |  ,---ProbeClosed
+  // ∨  ∨         ^
+  // Dead         |
+  //  `---> ProbeOpen
+  protected[finagle] sealed trait State
+  protected[finagle] object Alive extends State
+  protected[finagle] object Dead extends State
+  protected[finagle] object ProbeOpen extends State
+  protected[finagle] object ProbeClosed extends State
```
hawkw: Do you think we should just reuse the stat definitions in `FailureAccrualFactory`?

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,276 @@
...
+          case Return(rep) =>
+            if (rep.stream.isEmpty) {
+              classifyStream(classifier, H2ReqRepFrame(h2ReqRep.request, Return((rep, None))))
+              Return(rep)
+            } else {
+              val stream = rep.stream.onFrame {
+                case Return(f) if f.isEnd =>
+                  classifyStream(classifier, H2ReqRepFrame(h2ReqRep.request, Return((rep, Some(Return(f))))))
+                case Throw(e) =>
+                  classifyStream(classifier, H2ReqRepFrame(h2ReqRep.request, Return((rep, Some(Throw(e))))))
```
hawkw: wondering if there's an easy way to collapse these match arms...probably there isn't.

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,276 @@
...
+import io.buoyant.router.context.h2.H2ClassifierCtx
+
+/* Modified from https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/liveness/FailureAccrualFactory.scala */
+object H2FailureAccrualFactory {
+
+  val role = FailureAccrualFactory.role
+
+  def module: Stackable[ServiceFactory[Request, Response]] =
+    new Stack.ModuleParams[ServiceFactory[Request, Response]] {
+      val role: Role = FailureAccrualFactory.role
```
hawkw: Very minor quibble, but it seems to me this should be `H2FailureAccrualFactory.role`, _just in case_ the parent's role changes?

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
adleong: I'm launching a background thread to make some changes to FAF to make it possible for us to subclass, but that will take a while to land in finagle and be released.  Hopefully this whole-cloth duplication is a temporary measure.

router/h2/src/test/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactoryTest.scala
adleong: Nope.  The FAF tests are quite a bit more extensive but obviously don't cover the stream classifier aspect.

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,276 @@
...
+ * This uses a H2Classifier to inform failure accrual at either response time, or when the stream
+ * is complete.
+ */
+class H2FailureAccrualFactory(
+  underlying: ServiceFactory[Request, Response],
+  policy: FailureAccrualPolicy,
+  timer: Timer,
+  statsReceiver: StatsReceiver
+)
+  extends ServiceFactory[Request, Response] { self =>
```
adleong: I think this would clobber `state` since we can read theirs but not update it.

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,276 @@
...
+    state match {
+      case ProbeOpen =>
+        probesCounter.incr()
+        state = ProbeClosed
+      case _ =>
+    }
+  }
+
+  private[this] def classifyResponse(classifier: H2Classifier, h2ReqRep: H2ReqRep): Try[Response] =
+    classifier.responseClassifier.lift(h2ReqRep) match {
```
adleong: personally I find `match` easier to read and understand than `.map.getOrElse`.

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+  def apply(conn: ClientConnection) = {
+    underlying(conn).map { service =>
+      // N.B. the reason we can't simply filter the service factory is so that
+      // we can override the session status to reflect the broader endpoint status.
+      new Service[Request, Response] {
+        def apply(request: Request): Future[Response] = {
+          val param.H2Classifier(classifier) =
+            H2ClassifierCtx.current.getOrElse(param.H2Classifier.param.default)
```
olix0r: Why does this classifier need to be computed on each request?  Couldn't this be computed once on the parent class?

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
```diff
@@ -0,0 +1,254 @@
...
+          }
+        }
+
+        override def close(deadline: Time): Future[Unit] = service.close(deadline)
+        override def status: FStatus = FStatus.worst(
+          service.status,
+          H2FailureAccrualFactory.this.status
+        )
+      }
+    }.onFailure(onServiceAcquisitionFailure)
```
olix0r: slight pref for capturing a temp val here -- the body is large enough that i've lost the context of what this failure pertains to.

```scala
val factory = underlying(conn).map {
  ???
}
factory.onFailure(???)
```

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
adleong: because the classifier comes from the request local context.  In other places we memoize this keyed off the service Path with the assumption that the classifier is static for each service but.... we're gonna have to rip that all out when we violate that assumption with dynamic service policy.

olix0r: ah right.  i'd add a comment there because apparently "Ctx.current" wasn't enough of a clue for me...  just something like "This must be computed on each request to get the the classifier from the request-local context" etc

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
hawkw: that's fair, it also (probably) compiles to more performant bytecode?

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
hawkw: I was wondering about this myself for a hot second.

router/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2FailureAccrualFactory.scala
adleong: Instead of splitting factory into a val, I changed the Service into a named inner class.  Let me know if you think this improves readability.

PR #1550, opened 2017-08-02
Add gRPC H2StreamClassifiers

### LINKED ISSUE #1539: gRPC response classifier
Write 2 gRPC H2StreamClassifiers: one that assumes the request is idempotent and one that assumes the request is not idempotent.

The first should return RetryableFailure for all non-OK gRPC status codes except for ones that explicitly indicate the request should not be retried.

The second should return NonretryableFailure for all non-OK gRPC status codes except for ones that explicitly indicate that the request is safe to be retried.

Relevant background:
https://github.com/grpc/proposal/blob/master/A6-client-retries.md#retryable-status-codes
https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

hawkw:
This PR adds four `H2StreamClassifier`s which classify response streams based on the `grpc-status` code in the trailers frame.

+ `io.l5d.h2.grpc.alwaysRetryable`, which marks all gRPC failure status codes as retryable.
+ `io.l5d.h2.grpc.neverRetryable`, which never marks gRPC failures as retryable
+ `io.l5d.h2.grpc.default`, which marks UNAVAILABLE as retryable, and all other gRPC failures as non-retryable
+ `io.l5d.h2.grpc.retryableStatusCodes`, which accepts a user-configurable list of status codes to mark as retryable, and marks all other failures as non-retryable. Right now, these are parsed as integer literals, although I'd eventually like to add support for using the symbolic names of these statuses.

I've added a test class, [GrpcClassifierTest](https://github.com/linkerd/linkerd/blob/9bf050a6b17e1948d5ee6ec9f7ea92fe8dec538d/linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/GrpcClassifierTest.scala) which includes tests for classifier behavior and for parsing the classifiers from YAML.

Closes #1539


hawkw:
I think before this can be merged, I also should add it to the docs, but I'm opening this PR so it can be reviewed.

adleong:
I think you need to merge master here.

hawkw:
@adleong  try it now?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/GrpcStatus.scala
```diff
@@ -52,9 +74,21 @@ object GrpcStatus {
     case 16 => Unauthenticated(msg)
     case code => Other(code, msg)
   }
+  private[this] val StatusKey = "grpc-status"
+  private[this] val MessageKey = "grpc-message"
 
   def unapply(s: GrpcStatus): Option[(Int, String)] = Some((s.code, s.message))
 
+  def unapply(frame: h2.Frame): Option[GrpcStatus] =
```
adleong: I'm on the fence about extractors like this.  Code like

```
frame match {
  case GrpcStatus(status) =>
  case _ =>
}
```

would be surprising to me because frame is not a GrpcStatus.  Or maybe more to the point, a method `GrpcStatus.apply(status: GrpcStatus): Frame` wouldn't make sense, so inverting it to an unapply maybe doesn't make sense either.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/GrpcStatus.scala
```diff
@@ -52,9 +74,21 @@ object GrpcStatus {
...
+    frame match {
+      case trailers: h2.Frame.Trailers =>
+        for { headerValue <- trailers.get(StatusKey)
```
adleong: formatting nit: I prefer:

```
for {
  headerValue <- ...
  code <- ...
  status = ...
} yield status
```

or even

```
for {
  headerValue <- ...
  code <- ...
} yield GrpcStatus(...)
```

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
```diff
@@ -0,0 +1,94 @@
...
+
+trait GrpcClassifier extends H2Classifier {
+  /**
+    * @inheritdoc
+    * Since GRPC sends status codes in the
+    * [[com.twitter.finagle.buoyant.h2.Frame.Trailers Trailers]] frame of an H2
+    * stream, we can never attempt early classification
+    */
+  override val responseClassifier: PartialFunction[H2ReqRep, ResponseClass] =
+    PartialFunction.empty
```
adleong: We can classify Throws here as non-retryable failures (there will be no stream).

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
```diff
@@ -0,0 +1,94 @@
...
+}
+
+
```
adleong: rm extra newline

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
```diff
@@ -0,0 +1,94 @@
...
+    * codes as [[ResponseClass.RetryableFailure]]
+    */
+  object AlwaysRetryable extends GrpcClassifier {
+
+    override val streamClassifier: PartialFunction[H2ReqRepFrame, ResponseClass] = {
+      case H2ReqRepFrame(_, Return((_, Some(Return(GrpcStatus(Ok(_))))))) =>
+        ResponseClass.Success
+      case H2ReqRepFrame(_, Return((_, Some(Return(GrpcStatus(_)))))) =>
+        ResponseClass.RetryableFailure
+    }
```
adleong: I think we need to add a `_` case that leads to Non-retryable failure.  This covers the cases where the stream is empty and where the stream is a throw and where the stream doesn't have a grpc status.

adleong: ditto for other classifiers here

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifierConfig.scala
```diff
@@ -0,0 +1,72 @@
...
+  new Type(
+    value = classOf[DefaultConfig],
+    name = "io.l5d.h2.grpc.default"
+  ),
+  new Type(
+    value = classOf[RetryableStatusCodesConfig],
+    name = "io.l5d.h2.grpc.retryableStatusCodes"
+  )
+))
+abstract class GrpcClassifierConfig extends H2ClassifierConfig {
```
adleong: It's kinda nice that this refines the return type of `mk` down to `GrpcClassifier` but this ends up adding a lot of boilerplate.  I think its fine to just have the individual grpc classifier configs extend H2ClassifierConfig directly.

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2ClassifiersTest.scala~HEAD
```diff
@@ -0,0 +1,227 @@
+package io.buoyant.linkerd.protocol.h2
```
adleong: errant file got included here

adleong: and errant rename of H2ClassifiersTest

router/h2/src/main/scala/io/buoyant/router/H2.scala
```diff
@@ -12,6 +12,7 @@ import com.twitter.finagle.service.StatsFilter
 import io.buoyant.router.context.ResponseClassifierCtx
 import io.buoyant.router.context.h2.H2ClassifierCtx
 import io.buoyant.router.h2.{ClassifiedRetries => H2ClassifiedRetries, _}
+import io.buoyant.router.h2.{LocalClassifierStreamStatsFilter, PerDstPathStreamStatsFilter, StreamStatsFilter}
```
adleong: is this needed?

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
hawkw: oh, good point, hadn't thought of that.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifierConfig.scala
hawkw: I mean, it's about three lines of boilerplate, right? Or is there something I'm missing?

adleong: You can get rid of the JsonSubTypes annotation

hawkw: Oh, yeah, I guess you're right.

hawkw: @adleong isn't the annotation still needed, though, but on `H2ClassifierConfig` instead?

adleong: oh yeah, that's true.  I forgot that these aren't plugins.

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2ClassifiersTest.scala
```diff
@@ -1,227 +0,0 @@
-package io.buoyant.linkerd.protocol.h2
```
adleong: I think you accidentally deleted this file.

hawkw:
@adleong [5586](https://circleci.com/gh/linkerd/linkerd/5586) certainly _looks_ like a spurious failure to me, restarted it and we'll see...

PR #1556, opened 2017-08-03
Only emit a grpc-message trailer when a message is provided

olix0r:
Previously, grpc status could include `null` message strings (especially when created from an exception's getMessage call).  This can cause NullPointerExceptions when trailers are handled (esp when writing trailers into netty).

Now, the grpc-message trailer is only included when the grpc status object has a non-empty message.

Furthermore, all headers with null keys or values are ignored when constructing netty-friendly headers types.


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4Message.scala
```diff
@@ -70,7 +70,11 @@ private[h2] object Netty4Message {
       case orig: Headers => orig.underlying
       case orig =>
         val headers = new DefaultHttp2Headers
-        for ((k, v) <- orig.toSeq) headers.add(k, v)
+        for ((k, v) <- orig.toSeq) {
+          if (k != null && v != null) {
```
adleong: Should we log.error when we get a null key or value?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/GrpcStatus.scala
```diff
@@ -82,10 +82,13 @@ object GrpcStatus {
...
   }
 
-  private def toTrailers(status: GrpcStatus): h2.Frame.Trailers =
-    h2.Frame.Trailers(
-      "grpc-status" -> status.code.toString,
-      "grpc-message" -> status.message
-    )
-
+  private def toTrailers(status: GrpcStatus): h2.Frame.Trailers = {
+    var pairs = Seq("grpc-status" -> status.code.toString)
```
adleong: TIOLI:
```
if (status.message != null && status.message.nonEmpty) {
  Seq(... both items ...)
} else {
  Seq(... status only...)
}
```
Avoids smelly `var` and might be fewer allocations at the expense of a small amount of code duplication.

PR #1559, opened 2017-08-04
Use request local classifier for failure accrual

adleong:
FailureAccrualFactory uses the response classifier configured on the stack
instead of the response classifier from the request local context.  This means
that circuit breakers will trip on different success criteria than is used for
stats reporting.

Replace Finagle's FailureAccrualFactory with a subclass which uses the response
classifier from the request local context.

### DIFF
--- a/router/core/src/main/scala/com/twitter/finagle/buoyant/FailureAccrualFactory.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/buoyant/FailureAccrualFactory.scala
@@ -11,6 +11,10 @@ import com.twitter.logging.Level
 import com.twitter.util.Timer
 import io.buoyant.router.context.ResponseClassifierCtx
 
+/**
+ * A replacement for Finagle's FailureAccrualFactory that reads the respones classifier from the
+ * request local context.
+ */
 object FailureAccrualFactory {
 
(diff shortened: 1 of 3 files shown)


router/core/src/main/scala/com/twitter/finagle/buoyant/FailureAccrualFactory.scala
```diff
@@ -0,0 +1,86 @@
...
+import com.twitter.finagle.liveness.{FailureAccrualPolicy, FailureAccrualFactory => FFailureAccrualFactory}
+import com.twitter.finagle.liveness.FailureAccrualFactory.Param
+import com.twitter.finagle.service.{ReqRep, ResponseClass, ResponseClassifier}
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.{ServiceFactory, Stack, Stackable, param}
+import com.twitter.logging.Level
+import com.twitter.util.Timer
+import io.buoyant.router.context.ResponseClassifierCtx
+
+object FailureAccrualFactory {
```
olix0r: it'd be nice to have a comment here like "A replacement for finagle's FailureAccrualFactory that ..."

PR #1563, opened 2017-08-04
Remove deprecated sbt operators

hawkw:
The key dependency operators (`<<=`, `<+=`, `<++=`) and tuple enrichments for `TaskKey`s and `SettingKey`s are [deprecated in sbt 0.13.13](http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html). Building Linkerd with sbt 0.13.13 currently generates a bunch of warnings due to our use of deprecated features. Furthermore, according to the sbt documentation, these features are slated for removal in sbt 1.0, so I thought we ought to remove them.

I've rewritten the parts of our build file that use these deprecated features to use the `.value` DSL instead.

I've ran our tests with the new build file, and everything appears to work correctly. Furthermore, I ran the `docker` task on my branch and compared the tags of the generated Docker images with those generated by the `docker` task on master, to ensure that the tag was still being generated correctly after my changes.

### DIFF
--- a/project/Base.scala
+++ b/project/Base.scala
@@ -101,9 +101,9 @@ class Base extends Build {
           <url>https://buoyant.io/</url>
         </developer>
       </developers>,
-    publishTo <<= version { (v: String) =>
+    publishTo := {
       val nexus = "https://oss.sonatype.org/"
-      if (v.trim.endsWith("SNAPSHOT"))
+      if (version.value.trim.endsWith("SNAPSHOT"))
         Some("snapshots" at nexus + "content/repositories/snapshots")
       else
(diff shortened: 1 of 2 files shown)


hawkw:
N.B. that the `Build` trait [is also deprecated](http://www.scala-sbt.org/0.13/docs/Migrating-from-sbt-012x.html#Migrating+from+the+Build+trait), so we will also want to migrate to `build.sbt` before sbt 1.0 is released; I thought that ought to be a separate PR.

hawkw:
@olix0r yup, everything appears to be nominal!

Issue #1278, opened 2017-05-05
Add dynamic file-backed client and service configs


adleong:
Add `client` and `service` config kinds for loading client and service parameters dynamically from a file or directory.

PR #1565, opened 2017-08-04
Add io.l5d.fs service config

esbie:
Problem:
Per prefix service configuration is static. Users need to restart linkerd in order for it to pick up changes.

Solution:
Adds `FileSvc` that uses file watcher to generate service configs
Adds `DynamicServiceFactory` that recreates the stack for each activity update.
Adds dynamic params to `PerPathParams`

Validation:
Tested with service-options.yaml
Unit test

TODO: make sure this doesn't cause any performance regressions
Related to #1278


router/core/src/main/scala/io/buoyant/router/Router.scala
```diff
@@ -286,7 +291,7 @@ object StackRouter {
 
     case class PathParams(prefix: PathMatcher, mk: Map[String, String] => Stack.Params)
 
-    case class PerPathParams(params: Seq[PathParams]) {
+    case class PerPathParams(params: Seq[PathParams], dynamicParams: Option[Activity[Seq[PathParams]]]) {
```
adleong: It's worth considering if we can just have a `Activity[Seq[PathParams]]`.  The static and global configs could simply return an `Activity.value(...)` that never changes.

linkerd/core/src/main/scala/io/buoyant/linkerd/Svc.scala
```diff
@@ -58,9 +62,42 @@ trait StaticSvc { self: Svc =>
...
+
+  @JsonIgnore
+  private[this] lazy val watcher = Watcher(path.getParent)
+
+  @JsonIgnore
+  private[this] def configsAct: Activity[Seq[SvcPrefixConfig]] = {
+    watcher.children.flatMap { children =>
+      children.get(path.getFileName.toString) match {
+        case Some(file: Watcher.File.Reg) => file.data
+        case _ => Activity.pending
```
adleong: This can put the Activity back into a pending state if the file is deleted.  Instead, I think we probably want to either go into a failed state or use the default SvcPrefixConfig or something.

linkerd/core/src/main/scala/io/buoyant/linkerd/Svc.scala
```diff
@@ -58,9 +62,42 @@ trait StaticSvc { self: Svc =>
...
+      }
+    }.map {
+      case Buf.Utf8(dtab) =>
```
adleong: not a dtab

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -143,6 +145,8 @@ class HttpStaticSvc(val configs: Seq[HttpSvcPrefixConfig]) extends HttpSvc with
 
 class HttpSvcPrefixConfig(prefix: PathMatcher) extends SvcPrefixConfig(prefix) with HttpSvcConfig
 
+class HttpFsSvc(val serviceFile: File) extends HttpSvc with FileSvc
```
adleong: probably need this for all other protocols as well, I think

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
```diff
@@ -0,0 +1,47 @@
...
+      val Param(dynamicParams) = params[Param]
+      val sf = dynamicParams.map { p =>
+        next.make(params ++ p)
+      }
+      Stack.Leaf(role, new DynamicServiceFactory(sf))
+    }
+  }
+}
+
+class DynamicServiceFactory[Req, Rep](sf: Activity[ServiceFactory[Req, Rep]]) extends ServiceFactory[Req, Rep] {
```
adleong: This could probably use a bunch of scaladoc.

adleong: whenever the Activity updates, we also need to call `close()` on the previous ServiceFactory.

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
```diff
@@ -0,0 +1,47 @@
...
+  val dedupSf = new Activity(Var(Activity.Pending, sf.states.dedup))
+  private lazy val obs = dedupSf.states.respond(_ => ())
+  obs.hashCode() // start observing the Activity
+
+  override def apply(conn: ClientConnection): Future[Service[Req, Rep]] = {
+    dedupSf.values.toFuture.flatMap(Future.const).flatMap(_.apply(conn))
+  }
+
+  override def close(deadline: Time): Future[Unit] = obs.close(deadline)
```
adleong: We also want to close the current ServiceFactory as well as closing the observation.

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
```diff
@@ -0,0 +1,47 @@
...
+      }
+      Stack.Leaf(role, new DynamicServiceFactory(sf))
+    }
+  }
+}
+
+class DynamicServiceFactory[Req, Rep](sf: Activity[ServiceFactory[Req, Rep]]) extends ServiceFactory[Req, Rep] {
+  val dedupSf = new Activity(Var(Activity.Pending, sf.states.dedup))
+  private lazy val obs = dedupSf.states.respond(_ => ())
+  obs.hashCode() // start observing the Activity
```
adleong: I think we can just make this non-lazy and remove the `hashCode()` call.

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
adleong: We might want some stats in here.  Frequency of updates probably?  Not sure what else...

olix0r: as for this, i'd just track a counter (rather than a distribution of idle times or something fancy like that)

router/core/src/main/scala/io/buoyant/router/Router.scala
esbie: Cool I will try out the idea. Originally I thought io.l5d.fs could have config defined in BOTH a file and also statically, but dropped that idea due to its complexity. So I bet separating them is no longer needed.

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+ */
+class DynamicServiceFactory[Req, Rep](sf: Activity[ServiceFactory[Req, Rep]]) extends ServiceFactory[Req, Rep] {
+  @volatile var closable: Closable = Closable.nop
+  // don't rebuild the ServiceFactory unless we absolutely need to
+  val dedupSf = new Activity(Var(Activity.Pending, sf.states.dedup))
+  // keep activity open until the ServiceFactory is closed explicitly
+  private val obs = dedupSf.states.respond {
+    case Activity.Ok(sf) =>
+      // close the previous service factory
+      closable.close()
```
adleong: I think the body of this case needs to be in a synchronized block

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * DynamicServiceFactory wraps a ServiceFactory Activity.
+ * @param sf A ServiceFactory Activity that updates when it's underlying Stack.Params update.
+ */
+class DynamicServiceFactory[Req, Rep](sf: Activity[ServiceFactory[Req, Rep]]) extends ServiceFactory[Req, Rep] {
+  @volatile var closable: Closable = Closable.nop
+  // don't rebuild the ServiceFactory unless we absolutely need to
+  val dedupSf = new Activity(Var(Activity.Pending, sf.states.dedup))
+  // keep activity open until the ServiceFactory is closed explicitly
+  private val obs = dedupSf.states.respond {
```
adleong: I think we want to move the dedup up to the dynamicParams so that we don't build a new stack unnecessarily

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+    case Activity.Ok(sf) =>
+      // close the previous service factory
+      closable.close()
+      closable = sf
+    case _ =>
```
adleong: in the error case we probably still want to close the previous sf.

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Proxy the ClientConnection to the underlying ServiceFactory
+   */
+  override def apply(conn: ClientConnection): Future[Service[Req, Rep]] = {
+    val toFuture = dedupSf.values.toFuture.flatMap(Future.const)
+    toFuture.flatMap(_.apply(conn))
+  }
+
+  override def close(deadline: Time): Future[Unit] = {
+    closable.close(deadline)
```
adleong: I think we want to combine the closables with Closable.all or something

adleong: or maybe closable.sequence... we may want to close the obs first so that we don't create any new SFs before closing the current one.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -133,7 +134,8 @@ trait HttpClientConfig extends ClientConfig {
 )
 @JsonSubTypes(Array(
   new JsonSubTypes.Type(value = classOf[HttpDefaultSvc], name = "io.l5d.global"),
-  new JsonSubTypes.Type(value = classOf[HttpStaticSvc], name = "io.l5d.static")
+  new JsonSubTypes.Type(value = classOf[HttpStaticSvc], name = "io.l5d.static"),
+  new JsonSubTypes.Type(value = classOf[HttpFsSvc], name = "io.l5d.fs")
```
esbie: Oh weird I thought I did this, but I didn't. Must have been a dream.

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
esbie: Wouldn't we want to keep the previous service factory open even if we like, fail to read the updated service config file? So that we can continue to serve requests?

adleong: I don't think that will work the way this is written.  If the params activity goes to error, then the sf activity will go to error and the toFuture will return an exception.  If we get into this state we've lost the previous sf anyway so we should probably close it.

We might be able to solve this at the FsSvcConfig layer by not updating the activity if the file cannot be parsed, or something like that.

router/core/src/main/scala/io/buoyant/router/DynamicServiceFactory.scala
esbie: I don't know why, but moving the dedup up to the dynamicParams causes the stack to be rebuilt on every request

adleong:
I think the issue here is that each path stack has a fixed terminal SF, even if the stack is rebuilt.  Looking the `pathMk` method in Router.scala we see there is a single sf (which is DynBoundFactory in this case) that is used as the Endpoint.  Even if this path stack is rebuilt, it will still have the same DynBoundFactory on the bottom.  This means that when the path SF is closed, it will close the DynBoundFactory.  When the path stack is rebuilt, it will get the same (now closed) DynBoundFactory.

One possible solution would be to have DynamicServiceFactory install a new RefCountedServiceFactory module immediately above the Endpoint in the `next` stack.  When DynamicServiceFactory builds a stack it calls `open` on the RefCountedServiceFactory which increments a counter.  When RefCountedServiceFactory gets `close` it decrements the counter and it only calls `close` on the underlying SF when the counter goes to zero.

esbie:
Update: added `HaltClosePropagationFactory` to the stack which swallows calls to `close` unless they come from `DynParamsFactory` explicitly. Tested manually for correctness and now running load through it. Will attempt to write unit tests. Some previous feedback still needs addressing.

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+  implicit object Param extends Stack.Param[Param] {
+    override def default: Param = Param(Activity.value(Stack.Params.empty))
+  }
+
+  def module[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] = new finagle.Stack.Module[ServiceFactory[Req, Rep]] {
+    override def role: Stack.Role = DynParamsFactory.role
+    override def description: String = "Dynamically reconfigure the stack"
+    override def parameters = Seq(implicitly[Stack.Param[Param]])
+
+    def insertBeforeLeaf(stack: Stack[ServiceFactory[Req, Rep]], module: Stackable[ServiceFactory[Req, Rep]]): Stack[ServiceFactory[Req, Rep]] = stack match {
```
adleong: dope

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+import com.twitter.finagle
+import com.twitter.finagle.Stack.{Leaf, Node}
+import com.twitter.finagle._
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util._
+
+object HaltClosePropagationFactory {
+  def role: Stack.Role = Stack.Role("HaltClose")
+
+  case class Param(closer: Promise[Time])
```
adleong: use Future instead of Promise

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
+package io.buoyant.router
+
+import com.twitter.finagle
+import com.twitter.finagle.Stack.{Leaf, Node}
+import com.twitter.finagle._
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util._
+
+object HaltClosePropagationFactory {
```
adleong: Need some scaladoc here explaining what this is.

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+    override def description: String = "Stop service factory closes from propagating down the stack"
+    override def make(
+      params: Param,
+      next: ServiceFactory[Req, Rep]
+    ): ServiceFactory[Req, Rep] =
+      new HaltClosePropagationFactory[Req, Rep](next, params.closer)
+  }
+}
+
+class HaltClosePropagationFactory[Req, Rep](underlying: ServiceFactory[Req, Rep], closer: Promise[Time]) extends ServiceFactory[Req, Rep] {
```
adleong: TIOLI: I think we might save a tiny bit of boilerplate by extending ServiceFactoryProxy

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    override def make(
+      params: Stack.Params,
+      next: Stack[ServiceFactory[Req, Rep]]
+    ): Stack[ServiceFactory[Req, Rep]] = {
+      val param.Stats(stats) = params[param.Stats]
+      val Param(dynamicParams) = params[Param]
+      val dynStats = stats.scope("dynparams")
+      val nextWithHaltClose = insertBeforeLeaf(next, HaltClosePropagationFactory.module[Req, Rep])
```
adleong: This took us a while to figure out so it's definitely worth explaining in a comment.

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+class DynParamsFactory[Req, Rep](
+  dynamicParams: Activity[Stack.Params],
+  params: Stack.Params,
+  next: Stack[ServiceFactory[Req, Rep]],
+  dynStats: StatsReceiver
+) extends ServiceFactory[Req, Rep] {
+  val updatesCounter = dynStats.counter("updates")
+  val closesCounter = dynStats.counter("closes")
+
+  @volatile var closablePromise: Promise[Time] = new Promise[Time]
```
adleong: This is not technically a closable.  I might call it `closeLeaf` or something.  Would also make it private[this]

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+      updatesCounter.incr()
+      closablePromise = new Promise[Time]
+      // each time the params are updated, make a new ServiceFactory to replace
+      // the one currently in use
+      next.make(params ++ p + HaltClosePropagationFactory.Param(closablePromise))
+    }
+  }
+
+  // don't rebuild the ServiceFactory unless we absolutely need to
+  val dedupSf = new Activity(Var(Activity.Pending, sf.states.dedup))
```
adleong: I think to avoid rebuilding ServiceFactories we have to dedup the params.  ie
```
val dedupParams = ... // dedup dynamicParams
val sf = dedupParams.map { ...
```

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  @volatile var closableUnderlying: Closable = Closable.nop
+  // keep activity open until the ServiceFactory is closed explicitly
+  private val obs = dedupSf.states.respond {
+    case Activity.Ok(sf) =>
+      closesCounter.incr()
```
adleong: I think it makes sense to handle this in the same block as we are building the stacks.  We probably still need a (maybe nop) observation to keep the activity open.

esbie:
putting this on hold until the finagle 7 upgrade

router/core/src/main/scala/io/buoyant/router/DynParamsFactory.scala
esbie: I tried doing this but it only seemed to increase the number of updates. Now they happen on every request instead of every file change, so I am going to revert it. Not sure what's going wrong there

PR #1567, opened 2017-08-07
Grpc-Status code may come in H2 headers

adleong:
The grpc response classifiers assume that the grpc-status code will appear in
a trailers frame but this is not always the case.  If a response has no data
frames, the grpc-status may appear in the headers frame and the stream will be
empty.  The grpc response classifiers fail to correctly classify in this case.

Change the grpc response classifiers to check the headers frame for a
grpc-status code as part of early classification.

### DIFF
--- a/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
+++ b/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
@@ -7,6 +7,9 @@ import io.buoyant.grpc.runtime.GrpcStatus
 import io.buoyant.grpc.runtime.GrpcStatus.{Ok, Unavailable}
 
 trait GrpcClassifier extends H2Classifier {
+
+  def retryable(status: GrpcStatus): Boolean
+
   /**
    * @inheritdoc
    * Since GRPC sends status codes in the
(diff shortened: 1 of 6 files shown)


linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifierConfig.scala
```diff
@@ -17,7 +15,7 @@ class NeverRetryableInitializer extends H2ClassifierInitializer {
 object NeverRetryableInitializer extends NeverRetryableInitializer
 
 class AlwaysRetryableConfig extends H2ClassifierConfig {
-  override def mk: H2Classifier = GrpcClassifiers.NeverRetryable
+  override def mk: H2Classifier = GrpcClassifiers.AlwaysRetryable
```
hawkw: agh, can't believe I missed that.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
```diff
@@ -16,19 +19,19 @@ trait GrpcClassifier extends H2Classifier {
    */
   override val responseClassifier: PartialFunction[H2ReqRep, ResponseClass] = {
     case H2ReqRep(_, Throw(_)) => ResponseClass.NonRetryableFailure
+    case H2ReqRep(_, Return(GrpcStatus(Ok(_)))) => ResponseClass.Success
+    case H2ReqRep(_, Return(GrpcStatus(status))) =>
+      if (retryable(status)) ResponseClass.RetryableFailure
```
hawkw: this is much clearer, I should have written it that way :)

PR #1574, opened 2017-08-09
1.1.3

adleong:
The 1.1.3 release of Linkerd is mostly focused on improving our HTTP/2 support,
including better support for gRPC.  Linkerd now supports automatic retries in
HTTP/2 for retryable requests.

* HTTP/2
  * Cleaned up spurious errors messages in the Linkerd log output.
  * Added a number of gRPC response classifiers that use the `grpc-status` code
    to determine if the response was successful and if it should be retried.
    See [the docs](https://linkerd.io/config/1.1.3/linkerd/index.html#grpc-response-classifiers) for details.
  * Added support for failure accrual and automatic retries to HTTP/2.
  * Fixed a memory leak related to messages with only a headers frame.
* Istio
  * Added HTTP/2 support to the Istio integration: the `io.l5d.k8s.istio`
    identifier can now be used in H2 router configs.
  * Added support for HTTPRedirect Route Rules.
* The Linkerd and Namerd admin sites can now be configured to require HTTPS.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,24 @@
+## 1.1.3 2017-08-09
+
+The 1.1.3 release of Linkerd is mostly focused on improving our HTTP/2 support,
+including better support for gRPC.  Linkerd now supports automatic retires in
+HTTP/2 for retryable requests, even when there are (small) request or response
+bodies.
+
+* HTTP/2
+  * Cleaned up spurious errors messages in the Linkerd log output.
+  * Added a number of gRPC response classifiers that use the `grpc-status` code
+    to determine if the response was successful and if it should be retried.
+    See [the docs](https://linkerd.io/config/1.1.3/linkerd/index.html#grpc-response-classifiers)
+    for details.
+  * Added support for failure accrual and automatic retries to HTTP/2.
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,24 @@
+## 1.1.3 2017-08-09
+
+The 1.1.3 release of Linkerd is mostly focused on improving our HTTP/2 support,
+including better support for gRPC.  Linkerd now supports automatic retires in
```
rmars: s/retires/retries :)

CHANGES.md
```diff
@@ -1,3 +1,23 @@
+## 1.1.3 2017-08-09
+
+The 1.1.3 release of Linkerd is mostly focused on improving our HTTP/2 support,
+including better support for gRPC.  Linkerd now supports automatic retries in
+HTTP/2 for retryable requests, even when there are (small) request or response
+bodies.
```
klingerf: The last part of this sentence isn't really clear to me. Does this mean that automatic retries only happen for requests or responses of a certain size? What's the size threshold after which a request is not retried?

adleong: Good point.  I don't really want to get into the nitty-gritty so I removed this bit.

PR #1576, opened 2017-08-09
Remove scaladoc links

adleong:
For some reason, when publishing scaladoc artifacts, scaladoc links cannot be
resolved and cause publishing of the project to fail.

Remove all scaladoc links.


hawkw:
Oh, wow, it looks like most of these links were added by me, sorry! 

I vaguely remember there being a fix for this besides removing links, I'll see if I can find any information on that.

PR #1579, opened 2017-08-11
Fix for memory leak when using JsonStreamParser

mejran:
This fixes https://github.com/linkerd/linkerd/issues/1361. It's been tested in a live environment against kube-system and shows no memory leak (the original version hit 300mb usage within 30 minutes on the same environment).

### DIFF
--- a/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
+++ b/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
@@ -111,6 +111,7 @@ class JsonStreamParser(mapper: ObjectMapper with ScalaObjectMapper) {
     AsyncStream.fromFuture(read).flatMap {
       //Fake None element to get around scanLeft being one behind bug
       //Tracked in https://github.com/twitter/util/issues/195
+      //Can be removed once 195 is fixed
       case Some(buf) => Some(buf) +:: None +:: fromReaderJson(r, chunkSize)
       case None => AsyncStream.empty[Option[Buf]]
     }
--- a/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
+++ b/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
@@ -109,7 +109,8 @@ class JsonStreamParser(mapper: ObjectMapper with ScalaObjectMapper) {
     }
 
     AsyncStream.fromFuture(read).flatMap {
-      //Fake None element to get around scanLeft being one behind
+      //Fake None element to get around scanLeft being one behind bug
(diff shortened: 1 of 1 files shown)


mejran:
Does anyone know a better way to get AsyncStream.scanLeft to not be behind by one element? I'm currently adding two elements to the input stream (one real, one fake) and then ignoring the second one. It works but it's definitely not the cleanest.

adleong:
This is really great!  scanLeft seems much more elegant than what we had before.  I'm not sure why your scanLeft ends up one element behind.  In my experiments with scanLeft it seems to do what I would expect.  I'll look at the code more closely and see if I can figure it out.

mejran:
Thanks Alex. 

I've been able to create a more minimal example of the scanLeft failing:
```
    val p = Promise[Int]
    val p2 = Promise[Int]
    def s: AsyncStream[Int] = AsyncStream.fromFuture(p)
      .flatMap(_ +:: AsyncStream.fromFuture(p2)
        .flatMap(_ +:: AsyncStream.empty[Int]))
    val ss = s.scanLeft(0)(_ + _).filter(_ >= 1)
    println(ss.uncons.isDefined)
    p.setValue(1)
    //p2.setValue(1)
    println(ss.uncons.isDefined)
```

The second println should be true however it is false. If you uncomment the p2.setValue it becomes true. If you change almost anything else in the example it also becomes true (for example remove the flatMap or remove the  AsyncStream.empty). So basically we're hitting an edge case of scanLeft which doesn't behave correctly.

The best I can figure is that the flatMap causes the AsyncStream to be wrapped in an AsyncStream.Embed which triggers the first case of the scanLeft case match which blocks. The first case however doesn't output the previous result but rather indirectly calls the Cons case to return it (which would happen as soon as the Embed future is defined). Still not fully sure of the flow or how to fix it however.

adleong:
This is a fascinating bug.  I've filed: https://github.com/twitter/util/issues/195

It's not even clear to me how to fix it.  Perhaps someone from the twitter-util team will have some suggestions.

adleong:
I think we can proceed with the workaround here with the intention of removing it once the bug in AsyncStream has been fixed.  Would you mind adding a comment that links to the AsyncStream issue and notes that the workaround can be removed once the AsyncStream issue is fixed?

In the meantime @klingerf is validating that performance improvement of this branch on our end.  Once that's done I think we should be able to merge this.

Thank you!

mejran:
Comment added.

mejran:
My boss I believe has already accepted the CLA on behalf of the company, Iown.

klingerf:
Great, thanks @mejran. Will merge this now.

Issue #1515, opened 2017-07-19
io.l5d.specificHost does not render host param in config.json admin endpoint


siggy:
given the following config block:

```yaml
routers:
- protocol: http
  interpreter:
    kind: default
    transformers:
    - kind: io.l5d.specificHost
      host: 192.168.99.100
```

the `/config.json` endpoint does not render the value of `host`:

```bash
curl -s 192.168.99.100:9990/config.json|jq .routers[1].interpreter.transformers
[
  {
    "kind": "io.l5d.specificHost"
  }
]
```

siggy:
Fixed in #1580, thanks @cponomaryov !

PR #1580, opened 2017-08-13
io.l5d.specificHost does not render host param in config.json admin endpoint #1515

cponomaryov:
This is a fix for #1515

### DIFF
--- a/test-util/src/main/scala/io/buoyant/test/Json.scala
+++ b/test-util/src/main/scala/io/buoyant/test/Json.scala
@@ -1,0 +1,66 @@
+package io.buoyant.test
+
+import com.fasterxml.jackson.databind.SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS
+import com.fasterxml.jackson.databind.node.TextNode
+import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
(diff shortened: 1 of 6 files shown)


test-util/src/main/scala/io/buoyant/test/Json.scala
```diff
@@ -0,0 +1,79 @@
...
+      _arrayIndenter = DefaultIndenter.SYSTEM_LINEFEED_INSTANCE
+    })
+  }
+  private lazy val sortingMapper = (new ObjectMapper with ScalaObjectMapper).configure(ORDER_MAP_ENTRIES_BY_KEYS, true)
+
+  def assertJsonEquals(actualJson: String, expectedJson: String, verbose: Boolean = true): Unit = {
+    val actualJsonNode = tryJsonNodeParse(actualJson)
+    val expectedJsonNode = tryJsonNodeParse(expectedJson)
+    if (actualJsonNode != expectedJsonNode) {
+      println("JSON DIFF FAILED!")
```
adleong: Since this is intended to be mixed into tests, we should use a scalatest primitive such as `assert` for output instead of println

adleong: (This comment applies to all println statements in this change)

cponomaryov:
I made a fix according to code review, but it seems that there is an unrelated test failure. Could you please rerun the build because I have no permission to do this.

PR #1584, opened 2017-08-15
Make k8s Ingress class used by Linkerd configurable  (#1433)

pcalcado:
As per #1433 , sometimes one has many different Ingress Resource and Ingress Controller deployed. One common example seems to be when you have one Controller provided by a cloud (e.g. GCE), another one for Linkerd, and a third one provided by Istio. Linkerd, like most controllers,  will ignore any Ingress Resource not meant for it, and the way to indicate which one is meant for Linkerds is by using the `kubernetes.io/ingress.class` annotation.

Initially, Linkerd would always look for any Ingress annotated as `linkerd`. This change allows for users to  configure what annotation value Linkerd should look for.

### DIFF
--- a/k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
+++ b/k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
@@ -1,11 +1,333 @@
 package io.buoyant.k8s
 
+import com.twitter.finagle.Service
+import com.twitter.io.Buf
+import com.twitter.util.Future
 import io.buoyant.test.Awaits
 import org.scalatest.FunSuite
+import com.twitter.finagle.http.{Request, Response}
 
 class IngressCacheTest extends FunSuite with Awaits {
   val host = Some("myhost")
   val ns = Some("ns")
+
+  val ingressResourceListWithManyAnnotatedIngresses =
+    """
+{
+  "kind": "IngressList",
+  "apiVersion": "extensions/v1beta1",
+  "metadata": {
+    "selfLink": "/apis/extensions/v1beta1/ingresses",
+    "resourceVersion": "58101"
+  },
+  "items": [
+    {
+      "metadata": {
+        "name": "istio-ingress",
+        "namespace": "default",
+        "selfLink": "/apis/extensions/v1beta1/namespaces/default/ingresses/istio-ingress",
+        "uid": "07900e6b-813d-11e7-b89b-080027a996b8",
+        "resourceVersion": "58098",
+        "generation": 1,
+        "creationTimestamp": "2017-08-14T22:07:57Z",
+        "annotations": {
+          "kubernetes.io/ingress.class": "istio"
+        }
+      },
+      "spec": {
+        "rules": [
+          {
+            "http": {
+              "paths": [
+                {
+                  "path": "/istio-only",
+                  "backend": {
+                    "serviceName": "httpbin",
+                    "servicePort": 6060
+                  }
+                },
+                {
+                  "path": "/shared",
+                  "backend": {
+                    "serviceName": "httpbin",
+                    "servicePort": 6061
+                  }
+                }
+              ]
+            }
+          }
+        ]
+      },
+      "status": {
+        "loadBalancer": {}
+      }
+    },
+    {
+      "metadata": {
+        "name": "linkerd-ingress",
+        "namespace": "default",
+        "selfLink": "/apis/extensions/v1beta1/namespaces/default/ingresses/linkerd-ingress",
+        "uid": "07948e8e-813d-11e7-b89b-080027a996b8",
+        "resourceVersion": "58099",
+        "generation": 1,
+        "creationTimestamp": "2017-08-14T22:07:57Z",
+        "annotations": {
+          "kubernetes.io/ingress.class": "linkerd"
+        }
+      },
+      "spec": {
+        "rules": [
+          {
+            "http": {
+              "paths": [
+                {
+                  "path": "/linkerd-only",
+                  "backend": {
+                    "serviceName": "echo",
+                    "servicePort": 7070
+                  }
+                },
+                {
+                  "path": "/shared",
+                  "backend": {
+                    "serviceName": "echo",
+                    "servicePort": 7071
+                  }
+                }
+              ]
+            }
+          }
+        ]
+      },
+      "status": {
+        "loadBalancer": {}
+      }
+    },
+    {
+      "metadata": {
+        "name": "nginx-ingress",
+        "namespace": "default",
+        "selfLink": "/apis/extensions/v1beta1/namespaces/default/ingresses/nginx-ingress",
+        "uid": "078c9288-813d-11e7-b89b-080027a996b8",
+        "resourceVersion": "58096",
+        "generation": 1,
+        "creationTimestamp": "2017-08-14T22:07:57Z",
+        "annotations": {
+          "kubernetes.io/ingress.class": "nginx",
+          "ingress.kubernetes.io/rewrite-target": "/"
+        }
+      },
+      "spec": {
+        "rules": [
+          {
+            "http": {
+              "paths": [
+                {
+                  "path": "/nginx-only",
+                  "backend": {
+                    "serviceName": "nginx",
+                    "servicePort": 8080
+                  }
+                },
+                {
+                  "path": "/shared",
+                  "backend": {
+                    "serviceName": "nginx",
+                    "servicePort": 8081
+                  }
+                }
+              ]
+            }
+          }
+        ]
+      },
+      "status": {
+        "loadBalancer": {}
+      }
+    }
+  ]
+}"""
+
+  val ingressResourceListWithMoreThanOneLinkerdIngresses =
+    """
+{
+  "kind": "IngressList",
+  "apiVersion": "extensions/v1beta1",
+  "metadata": {
+    "selfLink": "/apis/extensions/v1beta1/ingresses",
+    "resourceVersion": "58101"
+  },
+  "items": [
+    {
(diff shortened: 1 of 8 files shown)


k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
```diff
@@ -1,12 +1,228 @@
 package io.buoyant.k8s
 
+import com.twitter.finagle.Service
+import com.twitter.io.Buf
+import com.twitter.util.Future
 import io.buoyant.test.Awaits
 import org.scalatest.FunSuite
+import com.twitter.finagle.http.{Request => FRequest, Response => FResponse}
```
hawkw: TIOLI, but as far as I can tell, the renaming import shouldn't be necessary here?

pcalcado: Indeed. I'm happy with keeping the name, but something for consideration: Because there are so many types called things like `Request`, `Response`, `Logger`, and `Service` in the classpath, at SoundCloud we had a convention that every time we imported one of those we'd rename it so that we could identify at a glance where it came from. e.g. we'd always do

```scala
import com.twitter.finagle.http.{Request => FFInagleRequest, Response => FInagleResponse}
import com.soundcloud.ourinternalstuff.http.{Request => ScRequest, Response => ScResponse}
```

But if we don't follow anything like this convention I'm happy to simplify.

hawkw: In general, I think our convention is to rename imports only when necessary. If you were importing both `finagle.http.Request` and a type from Linkerd called `Request`, then I think we would rename the type from Finagle. But in most of our code that I've seen where we don't have to deal with namespace clashes, we tend not to rename imports by default.

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
```diff
@@ -1,12 +1,228 @@
...
+  def mkIngressApiServiceReturning(response: String) = Service.mk[Request, Response] {
+    case req if req.uri.contains("/apis/extensions/v1beta1/ingresses") =>
+      val rsp = Response()
+      rsp.content = Buf.Utf8(response)
+      Future.value(rsp)
+    case req =>
+      fail(s"unexpected request for [${req.uri}]: $req")
+  }
+
+  test("builds cache with the existing ingress if only one configured") {
```
adleong: Are these test names backwards?  This test has many ingresses configured, right?

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
```diff
@@ -1,12 +1,228 @@
...
+    assert(await(cache.matchPath(host, "/linkerd-only")).get.svc == "echo")
+  }
+
+  test("builds cache with only valid ingresses when more than one configured") {
+    val service = mkIngressApiServiceReturning(ingressResourceListWithOneIngress)
+    val cache = new IngressCache(None, service, annotationClass)
+    assert(await(cache.matchPath(host, "/some-path")).get.svc == "echo")
+    assert(await(cache.matchPath(host, "/non-existing-path")).isEmpty)
+  }
+
```
adleong: should we also test the case where there are multiple ingresses with the linkerd class configured?  Or is that tested elsewhere?

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
```diff
@@ -1,12 +1,228 @@
...
+      Future.value(rsp)
+    case req =>
+      fail(s"unexpected request for [${req.uri}]: $req")
+  }
+
+  test("builds cache with the existing ingress if only one configured") {
+    val service = mkIngressApiServiceReturning(ingressResourceListWithManyAnnotatedIngresses)
+    val cache = new IngressCache(None, service, annotationClass)
+    assert(await(cache.matchPath(host, "/istio-only")).isEmpty)
+    assert(await(cache.matchPath(host, "/ngix-only")).isEmpty)
```
esbie: I'm a little confused why this path isn't matched, since I don't see a "kubernetes.io/ingress.class" specified for nginx-ingress, and we should still use the ingress if no ingress.class is specified.

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
```diff
@@ -1,12 +1,228 @@
...
+    assert(await(cache.matchPath(host, "/non-existing-path")).isEmpty)
+    assert(await(cache.matchPath(host, "/linkerd-only")).get.svc == "echo")
+  }
+
+  test("builds cache with only valid ingresses when more than one configured") {
```
esbie: I'm a bit confused about these test names. It seems like this is the test that tests a single configured ingress, whereas the test above is the one with "more than one [ingress class] configured".

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
pcalcado: Uh... good catch. I'll add the test.

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
pcalcado: My understanding is that the way this should work is:

1. If there is only one ingress resource, we use it
2. If there are more than one ingress resources, we only use the one annotated with 'linkerd' (or whatever we configure it to be)

Do you think that on (2) we should also add `and any other ingress resources without annotations`?

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
pcalcado: Yeah, I've flipped these by mistake :P

hawkw: Can't believe I missed that...thanks for fixing this, @pcalcado !

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
esbie: I don't see anything in the ingress resource spec that dictates controller annotation behavior. In lieu of a good spec, I try to follow other ingress controllers' behavior. So for traefik https://docs.traefik.io/user-guide/kubernetes/#excluding-an-ingress-from-trfik
>You can control which ingress Træfik cares about by using the kubernetes.io/ingress.class annotation. By default if the annotation is not set at all Træfik will include the ingress. If the annotation is set to anything other than traefik or a blank string Træfik will ignore it.

and nginx https://github.com/kubernetes/ingress/tree/master/controllers/nginx
>Setting the annotation kubernetes.io/ingress.class to any value other than "nginx" or the empty string, will force the NGINX Ingress controller to ignore your Ingress. Do this if you wish to use one of the other Ingress controllers at the same time as the NGINX controller.

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
```diff
@@ -1,12 +1,331 @@
...
+      Future.value(rsp)
+    case req =>
+      fail(s"unexpected request for [${req.uri}]: $req")
+  }
+
+  test("when multiple ingress resources, only adds one with desired annotation class") {
+    val service = mkIngressApiServiceReturning(ingressResourceListWithManyAnnotatedIngresses)
+    val cache = new IngressCache(None, service, annotationClass)
+    assert(await(cache.matchPath(host, "/istio-only")).isEmpty)
+    assert(await(cache.matchPath(host, "/ngix-only")).isEmpty)
```
esbie: looking at this line for another 10 seconds, I see there's a typo in ngi*n*x

esbie:
If you're interested in adding tests for the config, I'd recommend writing a ConfigTest. We have a bunch of examples of this, eg https://github.com/linkerd/linkerd/blob/a9c1627d389fab206faac44d1f674695207ff5cd/namerd/storage/consul/src/test/scala/io/buoyant/namerd/storage/consul/ConsulConfigTest.scala

k8s/src/test/scala/io/buoyant/k8s/IngressCacheTest.scala
```diff
@@ -1,12 +1,332 @@
...
+  test("when multiple ingress resources, only adds one with desired annotation class") {
+    val service = mkIngressApiServiceReturning(ingressResourceListWithManyAnnotatedIngresses)
+    val cache = new IngressCache(None, service, annotationClass)
+    assert(await(cache.matchPath(host, "/istio-only")).isEmpty)
+    assert(await(cache.matchPath(host, "/nginx-only")).isEmpty)
+    assert(await(cache.matchPath(host, "/non-existing-path")).isEmpty)
+    assert(await(cache.matchPath(host, "/linkerd-only")).get.svc == "echo")
+  }
+
+  test("when multiple ingress resources with desired annotation class, only adds the first one") {
```
esbie: This does not describe the behavior I expect to see. Linkerd will add as many ingress resources as have correct annotations. Whether linkerd chooses to match to echo1 or echo2 is undefined in this case

pcalcado: Aha! Now I understand what you meant earlier, thanks! I've changed it to match specific routes and to be more liberal on the shared one, given which Controller would be picked is undefined. wdyt?

esbie: This looks great, thanks for all the updates!

PR #1586, opened 2017-08-15
add ISSUE_TEMPLATE to route to discourse/slack

siggy:
Landing on the New Issue page, it's not clear to users that they should
start by asking questions in discourse and slack.

This change puts instructions directly into the description of a New
Issue.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,30 @@
+# Thanks for your help improving the project! #
+
+## Getting Help ##
+
+Github issues are for bug reports and feature requests. For questions about
+linkerd, how to use it, or debugging assistance, start by
+[asking a question in the forums](https://discourse.linkerd.io/) or join us on
+[Slack](https://slack.linkerd.io/).
+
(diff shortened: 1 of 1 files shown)


esbie:
oh snap! I was about to do this also ha. I really like k8s's https://raw.githubusercontent.com/kubernetes/kubernetes/e123311d8a45e4c1abc2cbc9a407f75dec25341f/.github/ISSUE_TEMPLATE.md

1) makes it clear this is not the place to ask for help
2) directs the filer on how to write better bug reports/feature requests

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+
+**How to reproduce it (as minimally and precisely as possible)**:
+
+**Anything else we need to know?**:
+
+**Environment**:
+- linkerd/namerd version, config files:
+- Platform, version, and config files (Kubernetes, DC/OS, etc):
+- Cloud provider or hardware configuration:
+- OS (e.g. from /etc/os-release):
```
adleong: I think these last 4 bullets are rarely relevant.  I think we can ask for these if we need them.

PR #1587, opened 2017-08-15
Refactor EndpointsNamer to watch individual endpoints, rather than entire namespace

### LINKED ISSUE #1534: Switch k8s namers to watch individual endpoints, rather than all endpoints in namespace
Right now, the `io.l5d.k8s` and `io.l5d.k8s.ns` namers setup a watch on the List endpoints API, which streams update events for all of the endpoints in a given namespace. This can potentially lead to receiving a lot of events for endpoints that Linkerd is not trying to actually route to. It can also lead to a memory leak, in the case that Linkerd is watching a namespace with very frequent updates (e.g. `kube-system` / #1361).

As of PR #1527, we now have the ability to watch individual objects in the Kubernetes API. We should considering the Kubernetes namers to only watch the endpoint objects to which they're trying to route, instead of watching all endpoints in a particular namespace. This can also potentially clean up a bunch of the caching code that exists in EndpointsNamer. This change will need to be rigorously tested to make sure it doesn't introduce any regressions in the namer.

### LINKED ISSUE #1575: Switch k8s namers to watch individual services, rather than all services in namespace
Follow-on from #1534.

Right now, the `io.l5d.k8s` and `io.l5d.k8s.ns` namers setup a watch on the List services API, which streams update events for all of the services in a given namespace. This can potentially lead to receiving a lot of events for services that Linkerd is not trying to actually route to.

As of PR #1527, we now have the ability to watch individual objects in the Kubernetes API. We should considering the Kubernetes namers to only watch the service objects to which they're trying to route, instead of watching all services in a particular namespace. This can also potentially clean up a bunch of the caching code in `ServiceCache`. This change will need to be rigorously tested to make sure it doesn't introduce any regressions in the namer.

hawkw:
Right now, the `io.l5d.k8s` and `io.l5d.k8s.ns` namers setup a watch on the List endpoints API, which streams update events for all of the endpoints in a given namespace. This can potentially lead to receiving a lot of events for endpoints that Linkerd is not trying to actually route to. It can also lead to a memory leak, in the case that Linkerd is watching a namespace with very frequent updates (e.g. `kube-system`/ #1361).

My PR #1527 added the ability to watch individual Kubernetes API objects, rather than watching every object in a namespace. This pull request rewrites the `EndpointsNamer` and `ServiceNamer`s to use the new watch code added in #1527, both for the `/namespaces/{ns}/endpoints/{service}` API endpoint and for the `/namespaces/{ns}/services/{service}` endpoint that we use for resolving numbered port mappings.

The `EndpointsNamer` caches watches on each (namespace, service name) combination and reuses existing watches so that we don't create a whole bunch of threads all watching the same API response. Furthermore, the watch on `.../services/...` is only established if it is needed (i.e. the namer was given a numbered rather than named port). I've also attempted to simplify the caches used in `EndpointsNamer` as much as possible.

Note that I've also made a change to the `Watchable.get()` method signature; it now returns a `Future[Option[G]]` rather than `Future[G]`. The test suite for` ServiceNamer` dictates that an unknown service name should return a negative `NameTree`, _but_ watch the service name in case it is created later. In order to fix this, I modified `Watchable.get()`to return `None` if the requested path returns 404. `Watchable.activity()` still starts the watch if `None` is returned. This required some modification to other existing classes, but appears to not have broken anything.

I've updated the existing unit tests in `EndpointsNamerTest`and `ServiceNamerTest` and confirmed they all work in dev. I've also modified other Kubernetes API tests to reflect the change to `Watchable.get()` I mentioned above.

Closes #1534 
Closes #1575


hawkw:
(n.b. that CI builds on this branch are breaking because I messed with the git history too much, and not for any real reason)

adleong:
"This branch has conflicts that must be resolved" :sob:

hawkw:
@adleong try it now?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
-    }
+    labelSelector: Option[String]
+  ): Activity[EndpointsCache] = synchronized {
+    def mkCache() =
+      mkApi(nsName)
+      .endpoints(serviceName)
+      .activity(
+        EndpointsCache.fromEndpoints(serviceName, nsName),
+        labelSelector = labelSelector
+      ) { case ((cache, event)) => cache.update(event) }
```
adleong: I think the destructuring here doesn't need the double-parens.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
+  ): Activity[NameTree[Name]] = {
+    val cache = endpointsCache(nsName, serviceName, labelSelector)
+    val unstable = Try(portName.toInt).toOption match {
+      case Some(portNumber) =>
+        // if `portName` was successfully parsed as an `int`, then
+        // we are dealing with a numbered port. we will thus also
+        // need the port mappings from the `Service` API response,
+        // so join its activity with the endpoints cache activity.
+        cache.join(numberedPortRemappings(nsName, serviceName, labelSelector))
+          .map { case ((endpoints, ports)) =>
```
adleong: don't need double-parens

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
-            service.port(portName).map {
-              case Some(addr) =>
-                log.debug("k8s ns %s service %s port %s found + %s", nsName, serviceName, portName, residual.show)
-                Activity.Ok(NameTree.Leaf(Name.Bound(addr, idPrefix ++ id, residual)))
-              case None =>
-                log.debug("k8s ns %s service %s port %s missing", nsName, serviceName, portName)
-                Activity.Ok(NameTree.Neg)
+    labelSelector: Option[String]
+  ) : Activity[NumberedPortMap] = synchronized {
+    def _getPortMap() =
```
adleong: should this be private?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
+      mkApi(nsName)
+        .service(serviceName)
+        .activity(_.portMappings, labelSelector = labelSelector) {
+          case ((oldMap, v1.ServiceAdded(service))) =>
```
adleong: similar comments as below: can probably clean and/or DRY this up by factoring out the logging.  also don't nee double-parens here.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
+)(implicit timer: Timer = DefaultTimer)
+  extends Namer with Stabilize {
 
   import EndpointsNamer._
+  protected[this] val variablePrefixLength: Int
 
-  def lookup(path: Path): Activity[NameTree[Name]]
+  // memoize port remapping watch activities so that we don't have to
+  // create multiple watches on the same `Services` API object.
+  private[this] val portRemappings =
```
adleong: TIOLI: use com.twitter.util.Memoize instead of HashMap.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
-  backoff: Stream[Duration] = Backoff.exponentialJittered(10.milliseconds, 10.seconds)
-)(implicit timer: Timer = DefaultTimer) extends EnumeratingNamer {
+  mkApi: String => v1.NsApi,
+  labelName: Option[String] = None,
+  backoff: Stream[Duration] = EndpointsNamer.DefaultBackoff
+)(implicit timer: Timer = DefaultTimer)
+  extends Namer with Stabilize {
 
   import EndpointsNamer._
+  protected[this] val variablePrefixLength: Int
```
adleong: as far as I can tell, this isn't used.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
-        portNumber.map(port)
-      }
-    }
-
-    def port(portNumber: Int): Var[Addr] =
-      endpointsState.map { endpoints =>
-        val addrs: Set[Address] = endpoints.map { endpoint =>
-          val isa = new InetSocketAddress(endpoint.ip, portNumber)
-          Address.Inet(isa, endpoint.nodeName.map(Metadata.nodeName -> _).toMap)
+  private[EndpointsNamer] case class EndpointsCache(
```
adleong: Or `ServiceEndpoints`

k8s/src/main/scala/io/buoyant/k8s/Ns.scala
```diff
@@ -5,9 +5,9 @@ import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.tracing.Trace
 import com.twitter.finagle.util.DefaultTimer
 import com.twitter.util._
-import io.buoyant.k8s.Ns.ObjectCache
+import io.buoyant.k8s.Ns.ListCache
 
-abstract class Ns[O <: KubeObject: Manifest, W <: Watch[O]: Manifest, L <: KubeList[O]: Manifest, Cache <: ObjectCache[O, W, L]](
+abstract class Ns[O <: KubeObject : Manifest, W <: Watch[O] : Manifest, L <: KubeList[O] : Manifest, Cache <: ListCache[O, W, L]](
```
adleong: I think if we move ServiceNamer off of Ns, then we can delete this file entirely.

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -5,10 +5,7 @@ import com.twitter.finagle.{Service => _, _}
 import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.util.DefaultTimer
 import com.twitter.util._
-import io.buoyant.k8s.Ns.ObjectCache
 import io.buoyant.k8s.v1._
-import java.net.InetSocketAddress
-import scala.collection.mutable
 
 /**
  * Accepts names in the form:
```
adleong: I think this class needs to be refactored to be very similar to the EndpointsNamer.  That is, it should use `NsApi.service(serviceName).activity` instead of ServiceCache.  Ideally, ServiceCache can be deleted entirely.

k8s/src/main/scala/io/buoyant/k8s/Stabilize.scala
```diff
@@ -0,0 +1,60 @@
+package io.buoyant.k8s
+
+import com.twitter.util.{Activity, Updatable, Var}
+
+trait Stabilize {
```
adleong: I'm not a huge fan of providing utility methods as mixins like this.  I'd prefer to put these utility methods in a static `object` or on the package object.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -167,36 +167,49 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+        labelSelector = labelSelector,
+        fieldSelector = fieldSelector,
+        retryIndefinitely = true)
         // since we're retrying the GET request forever, this `onFailure`
         // should probably never fire. but who knows?
         .onFailure { e =>
-          log.warning(s"k8s failed to get resource at $path: $e")
-          state.update(Activity.Failed(e))
-        }
+        log.warning(s"k8s failed to get resource at $path: $e")
```
adleong: weird indentation

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -167,36 +167,49 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+    labelSelector: Option[String] = None,
+    fieldSelector: Option[String] = None
+  )(onEvent: (T, W) => T): Activity[T] =
     Activity(Var.async[Activity.State[T]](Activity.Pending) { state =>
       val closeRef = new AtomicReference[Closable](Closable.nop)
-      val pending = get(retryIndefinitely = true)
+      val pending = get(
+        labelSelector = labelSelector,
+        fieldSelector = fieldSelector,
+        retryIndefinitely = true)
```
adleong: closing paren on its own line please

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -167,36 +167,49 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+        log.warning(s"k8s failed to get resource at $path: $e")
+        state.update(Activity.Failed(e))
+      }
         // otherwise, update the activity with the initial state, and
         // apply the onEvent function to each successive watch event in
         // the stream.
         .onSuccess { initial =>
-          val initialState = convert(initial)
-          state.update(Activity.Ok(initialState))
+        val initialState = convert(initial)
```
adleong: this seems like maybe intellij got some weird IDEAs in it's head about formatting (see what I did there?)

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
         }
-        Addr.Bound(addrs)
-      }
-
-    def ports: Var[Map[String, Int]] = portsState
 
-    def update(subsets: Option[Seq[v1.EndpointSubset]]): Unit = {
-      val newEndpoints = getEndpoints(subsets)
-      val newPorts = getPorts(subsets)
+    def port(portName: String): Option[Set[Address]] =
```
hawkw: I went back and forth on that for a while and ended up using the name from the previous code. Can definitely rename it.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: I was wondering if something like that existed, but all I could find was stuff from scalaz which I figured you *probably* wouldn't approve of us depending on 🙃

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: aren't nested defs always restricted to the scope they're created in, or am I missing something?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: `ServiceEndpoints` is a nice solution to the problem of having a whole million different things in this file whose names are some variation of "service" or "endpoints" 😛

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
-        if (newPorts != oldPorts) portsState() = newPorts
+    def port(portNumber: Int): Set[Address] =
+      for {
+        Endpoint(ip, nodeName) <- endpoints
+        isa = new InetSocketAddress(ip, portNumber)
+      } yield Address.Inet(isa, nodeName.map(Metadata.nodeName -> _).toMap)
+        .asInstanceOf[Address]
+
+    def update(event: v1.EndpointsWatch): EndpointsCache =
+      // TODO: i wish this logging code was less ugly...
```
hawkw: was considering that, too lazy. will add.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +103,277 @@ class SingleNsNamer(
...
+            for { name <- ports.keySet if !newPorts.contains(name) } yield {
+              log.debug(s"k8s ns $nsName service $serviceName deleted port $name")
+              name
+            }
+          this.copy(
+            endpoints = endpoints -- deletedEndpoints,
+            ports = ports -- deletedPorts
+          )
+        case v1.EndpointsError(error) =>
+          log.debug(s"k8s ns $nsName service $serviceName endpoints watch error $error")
```
hawkw: 👍

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
hawkw: yeah, my initial plan was to save that for a separate PR?

k8s/src/main/scala/io/buoyant/k8s/Stabilize.scala
hawkw: okay, I can change it – for my education, are there any particular reasons why this is considered harmful?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: hey, that was a pun, I understood that! and yeah, I blame IntelliJ

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: the concrete namer implementations [use it](https://github.com/linkerd/linkerd/pull/1587/files/6b648dbf421303b3f483e0f843aaf4d50c77116e#diff-6e1db8d78e58df54b2256d8d193c5000R37), I moved it to the `EndpointsNamer` class because I was planning on trying to factor out more of `lookup` from the concrete classes.

hawkw:
@adleong, I think I've addressed all your feedback as of 9928679?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
adleong: No need to have it in the superclass if it's not used there, though.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
adleong: oh, yup.  I misread.

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
adleong: It might be easier to this all at once.  I'll leave it up to you.

k8s/src/main/scala/io/buoyant/k8s/Stabilize.scala
adleong: Inheritance is a powerful tool that we don't actually need here.  `stabilize` is just a static function.  This is the first step down a dark path that eventually leads to the cake pattern.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: that's fair....would still like to factor out some stuff from `lookup` but it is probably less possible than I originally thought.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +106,324 @@ class SingleNsNamer(
...
+            case (oldMap, v1.ServiceAdded(service)) =>
+              val newMap = service.portMappings
+              logEvent.addition(newMap -- oldMap.keySet)
+              oldMap ++ newMap
+            case (oldMap, v1.ServiceModified(service)) =>
+              val newMap = service.portMappings
+              logEvent.addition(newMap -- oldMap.keySet)
+              logEvent.deletion(oldMap -- newMap.keySet)
+              val keptPorts = newMap.keySet &~ oldMap.keySet
+              val remappedPorts =
```
adleong: isn't this just 
```
keptPorts.filter { p => newPorts.get(p) != ports.get(p) }
```

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +106,324 @@ class SingleNsNamer(
...
-    val endpoints = mutable.Set.empty[Endpoint]
-    for {
-      subset <- subsets.getOrElse(Seq.empty)
-      addresses <- subset.addresses
-      addrs <- addresses
-    } {
-      endpoints += Endpoint(InetAddress.getByName(addrs.ip), addrs.nodeName)
-    }
-    endpoints.toSet
+  private trait Loggable {
```
adleong: Lots of changes to the logging logic in this branch.  Do you have any ideas about the best way to test these changes?

adleong: Maybe creating a StringLogger and using Logger.withLoggers in the test and validating the log output

k8s/src/main/scala/io/buoyant/k8s/ServiceCache.scala
```diff
@@ -0,0 +1,128 @@
+package io.buoyant.k8s
+
+import java.net.InetSocketAddress
+import com.twitter.finagle.Address
+import com.twitter.util.{Updatable, Var}
+import io.buoyant.k8s.Ns.ListCache
+import io.buoyant.k8s.v1._
+import scala.collection.mutable
+
+class ServiceCache(namespace: String)
```
adleong: I'm intentionally not reviewing this file for now, as it will change.

k8s/src/test/scala/io/buoyant/k8s/EndpointsNamerTest.scala
```diff
@@ -3,65 +3,705 @@ package io.buoyant.k8s
...
+        Future.value(rsp)
+      case req if Rsps.Inits.contains(req.uri) =>
+        val rsp = Response()
+        rsp.content = Rsps.Inits(req.uri)
+        doInit before Future.value(rsp)
+      case req if Rsps.Services.Responses.contains(req.uri) =>
+        val rsp = Response()
+        rsp.content = Rsps.Services.Responses(req.uri)
+        Future.value(rsp).onSuccess { _ => val _ = didServices.setDone() }
+      //      case req if req.uri.startsWith(s"$NonWatchPath$ServicesPath") =>
```
adleong: uncomment or rm

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +106,324 @@ class SingleNsNamer(
...
+          .activity(_.portMappings, labelSelector = labelSelector) {
+            case (oldMap, v1.ServiceAdded(service)) =>
+              val newMap = service.portMappings
+              logEvent.addition(newMap -- oldMap.keySet)
+              oldMap ++ newMap
+            case (oldMap, v1.ServiceModified(service)) =>
+              val newMap = service.portMappings
+              logEvent.addition(newMap -- oldMap.keySet)
+              logEvent.deletion(oldMap -- newMap.keySet)
+              val keptPorts = newMap.keySet &~ oldMap.keySet
```
hawkw: whoops, my bad

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: Oh, yeah, that's a _lot_ clearer, thanks.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -83,20 +87,19 @@ class SingleNsNamer(
...
         val labelSelector = Some(s"$label=$labelValue")
-        val cache = endpointNs.get(nsName, labelSelector)
-        val serviceCache = serviceNs.get(nsName, labelSelector)
-        lookupServices(nsName, portName, serviceName, cache, serviceCache, id, residual)
+        lookupServices(nsName, portName, serviceName, id, residual, labelSelector)
 
       case (id@Path.Utf8(portName, serviceName), Some(label)) =>
-        log.debug("k8s lookup: ns %s service %s label value segment missing for label %s", nsName, serviceName, portName, label)
+        log.debug(
+          "k8s lookup: ns %s service %s label value segment missing for label %s",
```
adleong: ditto

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +108,283 @@ class SingleNsNamer(
...
+        val logEvent = EventLogger(nsName, serviceName)
+        mkApi(nsName)
+          .service(serviceName)
+          .activity(
+            _.map(_.portMappings).getOrElse(Map.empty),
+            labelSelector = labelSelector
+          ) {
+              case (oldMap, v1.ServiceAdded(service)) =>
+                val newMap = service.portMappings
+                logEvent.addition(newMap -- oldMap.keySet)
```
adleong: `oldMap.keys` (and similar throughout this file)

k8s/src/main/scala/io/buoyant/k8s/EventLogger.scala
```diff
@@ -0,0 +1,72 @@
+package io.buoyant.k8s
+
+private[k8s] case class EventLogger(nsName: String, serviceName: String) {
+  import EventLogger._
+  def addition[A: Loggable](additions: Iterable[A]): Unit =
```
adleong: You were right: I _do_ think this is over-engineered.  I love typeclasses as much as the next person but I think it's overkill here given that:
* There are only two implementations of Loggable
* EventLogger is package private
I think it would be simpler and easier to understand to simply have 2 overloaded versions of `addition` (and each other method) that take the concrete types (Endpoint and PortMap).  

Sorry.

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -24,26 +23,135 @@ class ServiceNamer(
...
+          .service(serviceName)
+          .activity(
+            Svc.fromResponse(_),
+            labelSelector = labelSelector
+          ) { case (svc, event) => svc.update(eventLogger, event) }
+    }
 
-    case _ =>
-      Activity.value(NameTree.Neg)
+  private[this] def service(
```
adleong: Is there a purpose for this method?  Could callers just call `servicesMemo` instead?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -59,7 +59,10 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
     )
     val retry = if (retryIndefinitely) infiniteRetryFilter else Filter.identity[http.Request, http.Response]
     val retryingClient = retry andThen client
-    Trace.letClear(retryingClient(req)).flatMap(Api.parse[G])
+    Trace.letClear(retryingClient(req)).flatMap {
+      case rep if rep.status.code == 404 => Future.value(None)
```
adleong: Might be more idiomatic to use `rep.status` and `Status.NotFound`.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
```diff
@@ -53,7 +53,15 @@ class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
   }
 
   private[this]type NsMap = Map[String, VersionedDtab]
-  private[this] val dtabListToNsMap: DtabList => NsMap = _.items.map(toDtabMap)(breakOut)
+  private[this] val dtabListToNsMap: Option[DtabList] => NsMap = {
```
adleong: is this related to the rest of the PR?

k8s/src/main/scala/io/buoyant/k8s/EventLogger.scala
hawkw: 😢 
(but you're right)

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
hawkw: Yes, the changes in 5bea97b65964b5b301f91ffbb99642741bef743a (which are necessary to make one of the tests for `ServiceNamer` pass) broke this.

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
hawkw: Just remembered that `Function.untupled` is a thing, so, yeah, there's no reason to have this.

klingerf:
Oh and I verified that this branch fixes the memory leak described in #1361, which was the original motivation for this change. 🎉

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +108,283 @@ class SingleNsNamer(
...
-              Var(Some(svc.port(targetPortNumber)))
-            case None =>
-              // target port is a name and may or may not exist
-              svc.port(target)
-          }
-        }
-      case None =>
-        Var(None)
+    labelSelector: Option[String]
+  ) = portRemappingsMemo((nsName, serviceName, labelSelector))
```
hawkw: oh, yeah, definitely.  do we always prefer explicit type annotations on private functions?

klingerf: My tendency is to annotate all defs, but skip it for single line vals. Not sure how consistently that's applied in the codebase though.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -57,9 +56,18 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
       FieldSelectorKey -> fieldSelector,
       ResourceVersionKey -> resourceVersion
     )
-    val retry = if (retryIndefinitely) infiniteRetryFilter else Filter.identity[http.Request, http.Response]
+
+    val retry =
+      if (retryIndefinitely)
+        infiniteRetryFilter
+      else
+        Filter.identity[http.Request, http.Response]
```
olix0r: nit: prefer braces or no newlines

either
```scala
  if (retryIndefinitely) infiniteRetryFilter
  else Filter.identity[http.Request, http.Response]
```

or strongly prefer braces or no newlines

either
```scala
  if (retryIndefinitely) {
    infiniteRetryFilter
  } else {
    Filter.identity[http.Request, http.Response]
  }
```

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -167,10 +175,18 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
    * @return an [[Activity]]`[T]` updated by [[Watch]] events on this object,
    *         where `T` is the return type of the `onEvent` function
    */
```
olix0r: this comment probably needs updating, since the signature has changed significantly

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -230,7 +231,12 @@ private[k8s] class NsObjectResource[O <: KubeObject: TypeReference, W <: Watch[O
...
       watch = true
-    ).map { obj =>
-      (Seq(od.toWatch(obj)), obj.metadata.flatMap(_.resourceVersion))
+    ).map { maybeObj =>
+      (maybeObj.toSeq.flatMap { obj => Seq(od.toWatch(obj)) },
+        for {
+          obj <- maybeObj
+          meta <- obj.metadata
+          version <- meta.resourceVersion
+        } yield version)
```
olix0r: this would be easier to read with temp vals

```scala
val w = maybeObj.toSeq.flatMap { obj => Seq(od.toWatch(obj)) }
val v = for {
  obj <- maybeObj
  meta <- obj.metadata
  version <- meta.resourceVersion
} yield version
(w, v)
```

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -105,270 +109,277 @@ class SingleNsNamer(
...
+              case (oldMap, v1.ServiceError(error)) =>
+                log.warning(
+                  "k8s ns %s service %s watch error %s",
+                  nsName, serviceName, error
+                )
+                oldMap
+            }
+    })
+
+  private[this] val serviceEndpoints =
```
olix0r: tioli; but this may benefit from an explicit type annotation (very hard to know what the signature of this is by sight)

```scala
 private[this] val serviceEndpoints: String => Option[String =
  ...
```
(or whatever it is)

Issue #1578, opened 2017-08-10
Remove or improve statsd telemeter


adleong:
The current statsd telementer makes a network call for every stat update which has serious performance problems.  This can be reduced by sampling, but this causes issues with the quality of data.  We should improve this by removing the ability to sample and replace it with some kind of buffering or we should remove this telemeter entirely in favor of the influxdb telemeter.

hawkw:
Okay, now that  #1589 is merged, what else is necessary to consider this issue finished for 1.2.0?

PR #1589, opened 2017-08-15
Return existing statsd metrics if already exist

siggy:
The StatsD telemeter allows for the creation of counters and histograms
with duplicate names. This may result in memory leaks or incorrect
metrics depending on client behavior.

Modify the StatsD telemeter to check for existing metrics by name and
return them if they already exist.

### DIFF
--- a/telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
+++ b/telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
@@ -24,25 +24,36 @@ private[telemetry] class StatsDStatsReceiver(
   val repr: AnyRef = this
 
   private[statsd] def flush(): Unit = {
-    gauges.values.foreach(_.send)
+    gauges.values.asScala.foreach(_.send)
   }
   private[statsd] def close(): Unit = statsDClient.stop()
 
-  private[this] val gauges = new ConcurrentHashMap[String, Metric.Gauge].asScala
+  private[this] val counters = new ConcurrentHashMap[String, Metric.Counter]
+  private[this] val gauges = new ConcurrentHashMap[String, Metric.Gauge]
+  private[this] val stats = new ConcurrentHashMap[String, Metric.Stat]
 
   protected[this] def registerGauge(name: Seq[String], f: => Float): Unit = {
+    deregisterGauge(name)
+
     val statsDName = mkName(name)
-    gauges(statsDName) = new Metric.Gauge(statsDClient, statsDName, f)
+    val _ = gauges.put(statsDName, new Metric.Gauge(statsDClient, statsDName, f))
   }
 
   protected[this] def deregisterGauge(name: Seq[String]): Unit = {
     val _ = gauges.remove(mkName(name))
   }
 
-  def counter(name: String*): Counter =
-    new Metric.Counter(statsDClient, mkName(name), sampleRate)
+  def counter(name: String*): Counter = {
+    val statsDName = mkName(name)
+    val newCounter = new Metric.Counter(statsDClient, statsDName, sampleRate)
+    val counter = counters.putIfAbsent(statsDName, newCounter)
+    if (counter != null) counter else newCounter
+  }
 
-  def stat(name: String*): Stat =
-    new Metric.Stat(statsDClient, mkName(name), sampleRate)
-
+  def stat(name: String*): Stat = {
+    val statsDName = mkName(name)
+    val newStat = new Metric.Stat(statsDClient, statsDName, sampleRate)
+    val stat = stats.putIfAbsent(statsDName, newStat)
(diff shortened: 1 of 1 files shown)


siggy:
@adleong just fyi, i had consider making this part of https://github.com/linkerd/linkerd/issues/1578 but this fix does stand alone and is already in production use in the wild, so thought it prudent to get it merged.

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
```diff
@@ -28,21 +28,36 @@ private[telemetry] class StatsDStatsReceiver(
   }
   private[statsd] def close(): Unit = statsDClient.stop()
 
+  private[this] val counters = new ConcurrentHashMap[String, Metric.Counter].asScala
```
adleong: might be more useful to keep these as ConcurrentHashMaps instead of `asScala`ing them so we get CHM's atomic methods

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
```diff
@@ -28,21 +28,36 @@ private[telemetry] class StatsDStatsReceiver(
...
   }
 
-  def counter(name: String*): Counter =
-    new Metric.Counter(statsDClient, mkName(name), sampleRate)
-
-  def stat(name: String*): Stat =
-    new Metric.Stat(statsDClient, mkName(name), sampleRate)
+  def counter(name: String*): Counter = {
+    val statsDName = mkName(name)
+    if (!counters.contains(statsDName)) {
```
adleong: can use CHM.putIfAbsent here to avoid race conditions

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
```diff
@@ -28,21 +28,36 @@ private[telemetry] class StatsDStatsReceiver(
...
+      counters(statsDName) = new Metric.Counter(statsDClient, statsDName, sampleRate)
+    }
+    counters(statsDName)
+  }
 
+  def stat(name: String*): Stat = {
+    val statsDName = mkName(name)
+    if (!stats.contains(statsDName)) {
```
adleong: ditto

telemetry/statsd/src/main/scala/io/buoyant/telemetry/statsd/StatsDStatsReceiver.scala
```diff
@@ -24,25 +24,40 @@ private[telemetry] class StatsDStatsReceiver(
...
-
-  def stat(name: String*): Stat =
-    new Metric.Stat(statsDClient, mkName(name), sampleRate)
+  def counter(name: String*): Counter = {
+    val statsDName = mkName(name)
+    counters.putIfAbsent(
+      statsDName,
+      new Metric.Counter(statsDClient, statsDName, sampleRate)
+    )
+    counters.get(statsDName)
```
adleong: I think putIfAbsent returns the value.  so no need for this line

siggy: i think if it was absent, it returns null, from:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html
```
Returns:
the previous value associated with the specified key, or null if there was no mapping for the key
```

adleong: whoops, apparently I can't read.  sorry.  what about:
```
val newCounter = new Metric.counter(...)
val counter = ...putIfAbsent(..., newCounter)
if (counter != null) counter else newCounter
```

or something like that

siggy: i suppose that will work. it seems like we're trading a map.get on every call vs. an unnecessary allocation on some calls? i'd assume map.get is cheaper than allocation? what's the intent here?

adleong: well, we're already doing the potentially unnecessary allocation anyway: putIfAbsent takes arguments by value, not by name.  To avoid this we could switch to computeIfAbsent but that's a bit more of a pain.  Unclear if the optimization is worth it.

The thing that I care about here is atomicity.  As written, the value in the hashmap could change after putIfAbsent is called but before get is called.

siggy: is see, makes sense.

PR #1597, opened 2017-08-17
Improve Consul failure handling

### LINKED ISSUE #1593: Improve handling of Consul failures
## Filing a linkerd issue ##

Issue Type:

- [x] Bug report
- [x] Feature request

**What happened**:
Upon Consul failure (in this case, unexpected 403 response) Namerd produced Addr.Fail and wasn't able to resolve any names. Presumably Namerd saw valid responses from Consul before Consul entered failed state.

1) Produce a human-friendly message about namer/dtab-store failures in logs/admin-ui rather than just stacktraces
2) Stabilize Consul-related activities so that once there was a successful result never return failure again

**Environment**:
Linkerd/Namerd `v1.1.3`

hawkw:
Upon Consul failure Namerd produces `Addr.Fail` and isn't able to resolve any names. If it has seen valid responses from Consul before entering the failed state, it should instead continue to resolve names based on the observed known-good state.

I've  modified the`SvcAddr` in `io.buoyant.namer.consul` to cache the last observed good state, and fall back to that state when polling Consul fails. If no good state was previously observed, `SvcAddr` will still produce `Addr.Failed`. I've also added log messages to all failure cases in `SvcAddr`: it will log at the `WARNING` level when an error occurred but it fell back to a previous state, and it will log at the `ERROR` level when an error occurred but no fallback was possible (i.e. `Addr.Fail` was produced). 

I've also added calls to `Activity.stabilize()` on the Consul observation activities in `ConsulDtabStore`. This should prevent these activities from entering the `Failed` state if they have previously observed a good state, and fall back to that state on observing an error instead. Failures that are flagged as interruptions should still close the `Activity`. I've also added added error logging to these activities.

I've added some additional tests in `ConsulNamerTest` to ensure that the namer will fall back to the previous good state on errors, that the state the namer falls back to is the most recent observed good state, and that the namer will resume responding to good updates after falling back.

Closes #1593

### DIFF
--- a/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
+++ b/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
@@ -711,9 +711,5 @@ class ConsulNamerTest extends FunSuite with Awaits {
         "namer did not update after falling back"
       )
     }
-
-
-
-
   }
 }
--- a/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
+++ b/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
@@ -633,4 +633,87 @@ class ConsulNamerTest extends FunSuite with Awaits {
       }
 
   }
+  test("Namer falls back to most recent observed good state on serviceNodes failure") {
+    val scaleUp = new Promise[Unit]
+    val doFail = new Promise[Unit]
+    val scaleToEmpty = new Promise[Unit]
+    @volatile var didFail = false
+    class TestApi extends CatalogApi(null, "/v1") {
+      override def serviceMap(
+        datacenter: Option[String] = None,
+        blockingIndex: Option[String] = None,
+        consistency: Option[ConsistencyMode] = None,
+        retry: Boolean = false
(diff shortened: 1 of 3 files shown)


adleong:
@hawkw I believe that https://github.com/linkerd/linkerd/issues/1593 refers to errors with the consul namer, not the consul dtab store.  In particular, I think the problem stems from the way that errors are handled in SvcAddr.scala

hawkw:
@adleong oh, whoops, I can look into that. Sorry, my bad.

hawkw:
Okay @adleong and @ashald, I've updated this PR; hopefully it now makes the correct changes.

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -581,4 +581,139 @@ class ConsulNamerTest extends FunSuite with Awaits {
...
+      assert(
+        state == Activity.Ok(NameTree.Neg),
+        "namer did not update after falling back"
+      )
+    }
+
+
+
+
+  }
```
esbie: nit: spaces

hawkw:
Okay, have tested this against https://github.com/linkerd/linkerd-examples/tree/master/consul and nothing appears to be obviously broken; going to go ahead and merge

PR #1598, opened 2017-08-18
Ensure frames are released after classification timeout

### LINKED ISSUE #1613: gRPC memory leak
Refer to https://github.com/linkerd/linkerd/pull/1598#issuecomment-323488710. It looks like @adleong's commit 9deb5f7 revealed a memory leak in long-running gRPC streams. This leak was probably already there, but wasn't being reached because the flow-control issue fixed in 9deb5f7 was preventing us from reaching a state in which it could occur.

@klingerf did some investigation of a heap dump (see https://github.com/linkerd/linkerd/pull/1598#issuecomment-325006851) and it looks like unbounded `AsyncQueue` creation might be the cause of this leak. I talked it over w/ Alex & we suspect that long streams probably cause a future chain that gets longer and longer, probably  in `StreamProxy` or `BufferedStream`.

adleong:
After a classification timeout in ClassifiedRetryFilter, an exception is
raised on the classification future to cancel it.  This has the effect that
any frames that had been read by the classification future but not yet released
when the future was cancelled would get dropped and remain un-released forever.
This would cause the flow control window to slowly fill up and eventually
cause frames to stop being sent.

Instead, we no longer use Future.raise and instead implement our own timeout
logic where, even in the case that a timeout exception is returned, processing
continues on the stream and ensures that all frames are released.

Closes #1613.

### DIFF
--- a/router/h2/src/main/scala/io/buoyant/router/h2/ClassifiedRetryFilter.scala
+++ b/router/h2/src/main/scala/io/buoyant/router/h2/ClassifiedRetryFilter.scala
@@ -25,7 +25,7 @@ class ClassifiedRetryFilter(
   classifier: H2Classifier,
   backoffs: SStream[Duration],
   budget: RetryBudget,
-  classificationTimeout: Duration = Duration.Top,
+  classificationTimeout: Duration = 100.millis,
   requestBufferSize: Long = ClassifiedRetryFilter.DefaultBufferSize,
   responseBufferSize: Long = ClassifiedRetryFilter.DefaultBufferSize
 )(implicit timer: Timer) extends SimpleFilter[Request, Response] {
@@ -160,9 +160,9 @@ class ClassifiedRetryFilter(
     responseBuffer.fork() match {
       case Return(s) =>
         // Attempt to determine retryability based on the final frame.  Completes when the stream is
-        // fully buffered.
-        val fullyBuffered = retryable(req, rsp, s)
-          .raiseWithin(classificationTimeout)
+        // fully buffered.  Will always consume all of s, even if the classification deadline is
+        // exceeded.
+        val fullyBuffered = retryable(req, rsp, s, Time.now + classificationTimeout)
           .handle {
             case _: TimeoutException =>
               classificationTimeoutCounter.incr()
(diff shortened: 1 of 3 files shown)


router/h2/src/main/scala/io/buoyant/router/h2/ClassifiedRetryFilter.scala
```diff
@@ -208,27 +212,57 @@ class ClassifiedRetryFilter(
...
       }
+
+      deadline match {
+        case Some(deadline) if deadline > now =>
+          // This is a sort of "safe" timeout.  If the classification deadline passes, we want to
+          // return a classificationTimeoutException but we _don't_ want to cancel the read Future.
+          // Cancelling the read or discarding the result of the read could result in reading a frame
+          // but never releasing it.  If we initiate a read, we must always capture the resulting
+          // frame and ensure it is released.
+          Future.selectIndex(IndexedSeq(Future.sleep(deadline - now), frameF)).flatMap {
```
hawkw: I feel like this is the only thing in this PR that isn't immediately clear to me, but I'm not sure if I can think of any clearer way to express it.

adleong: Future.selectIndex runs both Futures in parallel and returns the index of the one that completes first.  The first Future is scheduled to return at the deadline.  So if we hit the deadline first, we go to the first case and if we read the final frame before hitting the deadline, we go the second case.  Does that help?

hawkw: Yeah, I was able to figure out what was going on here, I just meant that the control flow with `selectIndex` is a little hard to follow at a glance?

hawkw: How would you feel about something like
```scala
val timeoutF = 
    // Classification timeout.  Immediately return a classificationTimeoutException but
    // ensure the final frame is eventually released.
    Future.sleep(deadline - now).onSuccess { 
              frameF.onSuccess { f => f.foreach(_.release()); () }
              Future.exception(classificationTimeoutException)
    }
timeoutF.or(frameF)

klingerf:
I talked this over with @adleong and it looks like with this change gRPC streaming is suffering from a memory leak, either as a result of this change, or due to a different issue that didn't manifest until we fixed the flow control issue. I think we should track that down before merging.

klingerf:
Ok, I merged this branch with latest master following the Kubernetes API refactor branch merging, and the memory leak is still happening. I grabbed a heap dump, and it contains one leak suspect, as follows:

screen shot 2017-08-25 at 11 44 25 am

Looks like one or more unbounded AsyncQueue objects. Am happy to share the heap dump if folks are interested in looking at it.

hawkw:
This is great, thanks @klingerf – I'll look into this in a moment?

hawkw:
@klingerf just to confirm, do we know for sure that this leak did not exist before 9deb5f7?

klingerf:
@hawkw Yeah, as far as I can tell, this is a new leak introduced with this branch. It's a bit hard to be 100% sure, however, since without the fix from this branch, linkerd just stops processing all streaming requests after a short period of time.

hawkw:
@adleong, I've updated your PR summary to note that this fixes #1613 (I also fixed a typo 😛)

Issue #1403, opened 2017-06-16
Support Mixer precondition checking in the IstioIdentifier


esbie:
Mixer provides a [CheckRequest api](https://istio.io/docs/reference/api/mixer/mixer-service.html#checkrequest) that linkerd can use to enforce ACLs, quotas, etc.

hawkw:
Moved from 1.2.0 milestone to 1.2.1 so we can have additional time for beta testing this feature without holding back 1.2.0

wmorgan:
Is this closed by #1606 ?

PR #1606, opened 2017-08-23
Support for Istio Mixer's Conditions Check (#1403) 

pcalcado:
This implementation builds on the [original Istio integration work](https://github.com/linkerd/linkerd/pull/1473/files#diff-ec7d844511058df1c9ecb3491112b988), extending our Mixer integration to not only send reporting data but also checking for preconditions.

The initial design assumed that this check could be done at the `Identifier`, but it turns out that to get the `target.service` attribute we need it to be much later in the processing chain. We then refactored what was known as the `IstioLogger` into an `RequestAuthorization` that performs such check.

The actual check added:  https://github.com/linkerd/linkerd/commit/54ba9ad2ba1b4d3af6fb08b340d2deb9deb77376#diff-1d26b6a6caa59bb04f9815bbca98dd76R22

### Design Rationale
Istio allows for operators to define preconditions that must be met before a request is sent to a service. These preconditions are defined based on [attributes](https://istio.io/docs/concepts/policy-and-control/attributes.html). As per #1403, we want to add this feature to an Istio + Linkerd deployment.

The way this is supposed to work is that the proxy should send the request attributes to the [check request API](https://istio.io/docs/reference/api/mixer/mixer-service.html#checkrequest) and, depending on the result, either forward the request to the underlying service or reject the connection altogether.

The way Istio's vanilla proxy implements the check is [by first checking against a cache of pre-computed results](https://github.com/istio/mixerclient/blob/13460a96b4b3aa792ad47817d4fbf529b63c25e2/src/check_cache.cc#L78). The cache is indexed by [a hash of each attribute from the request](https://github.com/istio/mixerclient/blob/13460a96b4b3aa792ad47817d4fbf529b63c25e2/src/signature.cc#L30). The cache is only used to prevent calls that are known to *fail* the check, and not enabled by default. Our implementation currently doesn't cache anything.

### DIFF
--- a/istio/src/main/scala/io/buoyant/k8s/istio/CurrentIstioPath.scala
+++ b/istio/src/main/scala/io/buoyant/k8s/istio/CurrentIstioPath.scala
@@ -6,26 +6,106 @@ import com.twitter.logging.Logger
 import io.buoyant.router.context.DstBoundCtx
 
 object CurrentIstioPath {
-  val log = Logger(this.getClass.getName)
-  private val pathLength = 10
+  val logger = Logger(this.getClass.getName)
 
   def apply(maybeBound: Option[Dst.Bound]): Option[Path] = {
     maybeBound.flatMap { bound =>
       bound.id match {
-        case path: Path if (path.elems.length == pathLength) => Some(path)
-        case _ => None
+        case path: Path => Some(path)
+        case other =>
+          logger.warning("Expected path, got %s", other)
+          None
       }
     }
   }
 
   /**
-   * Resolves a valid Istio [[Path]], if any exists for the current [[DstBoundCtx]]
-   */
+    * Returns the target service requested based on the supplied [[Path]].
+    *
+    * The service is expected to be specified in the third last element in the path. For example,
+    * in the path:
+    *
+    * `%,io.l5d.k8s.daemonset,default,incoming,l5d,#,io.l5d.k8s.istio,reviews.default.svc.cluster.local,az:us-west::env:prod::version:v1,http`
+    *
+    * The service is `reviews.default.svc.cluster.local`.
+    *
+    *
+    * @param istioPath
+    * @return
+    */
+  def targetServiceIn(istioPath: Path): Option[String] = {
+    val pathServiceIndex = istioPath.elems.size - 3
+    if (pathServiceIndex > 0) {
+      Some(istioPath.showElems(pathServiceIndex))
+    } else {
+      logger.warning(
+        "Tried parsing target service at index %s for Invalid Istio Path %s",
+        pathServiceIndex,
+        istioPath
+      )
+      None
+    }
+  }
+
+  /**
+    *
+    * Returns the target version of the app requested based on the supplied [[Path]].
+    *
+    * The version of the app is expected to be specified in the second last element in the path. For
+    * example, in the path:
+    *
+    * `%,io.l5d.k8s.daemonset,default,incoming,l5d,#,io.l5d.k8s.istio,reviews.default.svc.cluster.local,az:us-west::env:prod::version:v1,http`
+    *
+    * The version is contained in the segment `az:us-west::env:prod::version:v1`, and its value is `v1`.
+    *
+    * @param istioPath
+    * @return
+    */
+  def targetVersionIn(istioPath: Path): Option[String] = {
+    val pathLabelsIndex = istioPath.elems.size - 2
+    if (pathLabelsIndex >= 0) {
+      istioPath
+        .showElems(pathLabelsIndex).split("::")
+        .find { e => e.startsWith("version:") }
+        .map { label => label.split(":").last }
+    } else {
+      logger.warning(
+        "Tried parsing target version at index %s for Invalid Istio Path %s",
+        pathLabelsIndex,
+        istioPath
+      )
+      None
+    }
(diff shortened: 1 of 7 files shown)


k8s/src/main/scala/io/buoyant/k8s/istio/MixerApiRequests.scala
```diff
@@ -0,0 +1,96 @@
...
+    duration: Duration
+  ): ReportRequest = {
+
+    val customWords = Seq(
+      "app",
+      "version",
+      requestPath,
+      targetService,
+
+      sourceLabelApp,
```
pcalcado: @adleong In the original code, this would send the _values_ of these things as part of the dictionary. My understanding is that the dictionary should only contain the _keys_ to be resolved by Mixer. Is this intentional?

pcalcado:
Testing has shown that this initial approach won't work as we dont have an `istioPath` set at that stage yet. I'm working on an alternative approach.

istio-proto/src/main/protobuf/google/rpc/README.md
```diff
@@ -0,0 +1,5 @@
+# Google RPC
```
hawkw: The "g" in gRPC apparently [no longer stands for Google](https://grpc.io/faq/) :)

pcalcado: This was copied verbatim from https://github.com/googleapis/googleapis/blob/master/google/rpc/README.md#google-rpc so I think we should keep it but... it had actually never occurred to me that the `g` stood for "Google", I always thought it was something like "generalised" :P

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
```diff
@@ -0,0 +1,121 @@
...
+      sourceLabelApp,
+      sourceLabelVersion
+    ) ++ sourceLabel.value.keySet ++ targetLabel.value.keySet
+    customWords
+  }
+
+  private def mkAttributesUpdate(customAttributeNames: Seq[String], attributes: Seq[IstioAttribute[_]]) = {
+
+    // TODO: eventually we should not have to send istioGlobalDict over the wire,
+    // and instead use positive index integers
```
ZackButcher: I'm ignorant about Scala: when you retrieve indices in dictionary with `indexOf`, where're you flipping the index around to be negative? You want them to be negative since you're indexing into the local dictionary instead of the global one.

pcalcado: Thanks so much, Zack!

1. Uhm.. the comment was about `mixerclient`, I probably misread something on the code. This iteration doesn't have a cache yet, but I will make sure to follow the same logic when we add it 👍 
2. This is brilliant, thank you. As I understand it, our first integration code was tested against an early alpha release and as everything was moving so fast we decided to send a lot of stuff just to make sure we didn't miss on anything that would become important or mandatory. I'll change it to the list you provided!
3. Our current integration runs against an older [schema that didn't have the negative index logic](https://github.com/istio/api/blob/ae97de1806d1802b65b6a0846f5498b7575f669e/mixer/v1/attributes.proto). I'll update the schemas as we move our integration to 0.2.x

istio/src/main/scala/io/buoyant/k8s/istio/CurrentIstioPath.scala
```diff
@@ -6,26 +6,65 @@ import com.twitter.logging.Logger
...
+        case other =>
+          logger.warning("Expected path, got %s", other)
+          None
       }
     }
   }
 
+  def targetServiceIn(istioPath: Path): Option[String] = {
+    val pathServiceIndex = istioPath.elems.size - 3
+    if (pathServiceIndex > 0) {
```
adleong: Should this be `>= 0`?  I mean, element 0 should always be `$` or `#` but I assume this is intended to just avoid index out of bounds.

istio/src/main/scala/io/buoyant/k8s/istio/CurrentIstioPath.scala
```diff
@@ -6,26 +6,65 @@ import com.twitter.logging.Logger
...
-        case _ => None
+        case path: Path => Some(path)
+        case other =>
+          logger.warning("Expected path, got %s", other)
+          None
       }
     }
   }
 
+  def targetServiceIn(istioPath: Path): Option[String] = {
```
adleong: It would be helpful to have a comment with the expected structure of path so that these offsets are easier to understand.

istio/src/main/scala/io/buoyant/k8s/istio/CurrentIstioPath.scala
pcalcado: This is pretty much a panic situation so I'm not sure if it matters, but I will add the `=`

istio/src/main/scala/io/buoyant/k8s/istio/CurrentIstioPath.scala
pcalcado: Will do

pcalcado:
This should be merge-able, waiting for the final 👍  to merge

Issue #1601, opened 2017-08-21
Consul: query service by specific health states


blacktoe:
The consul namer currently allows filtering nodes by a binary health status via the `useHealthCheck` parameter. Consul however has the concept of `passing`, `warning` and `critical` health statuses, the consul API can also be queried with the `any` state where no filtering is desired. 

A new parameter (like `healthCheckStatuses`?) could be added to the consul namer to allow more precise filtering of consul nodes by one or more health statuses, passed as a comma separated list.

We are particularly interested in this feature so that we can implement a fall back pool for a service which includes only nodes in the warning state (and not the critical state). See also https://discourse.linkerd.io/t/configuring-a-fall-back-consul-namer/243/3. 

The [consul API](https://www.consul.io/api/health.html) does not currently allow for listing a services nodes directly by health status, but does include the health statuses of each node in a separate top level array which could be used to filter results. Such an approach is used in [consul template](https://github.com/hashicorp/consul-template/blob/master/dependency/health_service.go) for example.

adleong:
This sounds pretty reasonable to me!  Are you interested in submitting a PR for this?

amitsaha:
Cool, we will work on it and submit a PR. Thanks @adleong

amitsaha:
hey @adleong do you mind taking a look at the in progress PR please?

hawkw:
Hi @amitsaha, @adleong is currently out of the office. I'm running point on the 1.2.0 release & would be happy to go over your PR if you like. When it's ready for review, we might want to loop @esbie in as well.

hawkw:
And #1608 is merged! Thanks to @blacktoe for all the work on this!

amitsaha:
Hi @hawkw thanks for the merge! Are we looking at this being in the upcoming 1.2.0 release?

hawkw:
@amitsaha yes, that's correct! We're looking at cutting 1.2.0 by September 14th, if not before.

amitsaha:
Thanks, perfect.

PR #1608, opened 2017-08-24
Add option to filter consul nodes by health statuses

blacktoe:
Adding the healthStatuses parameter to linkerd consul api #1601.

The only concern I have is the way the `healthStatuses` are used to filter the nodes are passed to `HealthApi`, I could move the parameter into the serviceNodes method but that would require also modifying `CatalogueApi` and a number of intermediate calls. There may be another approach that I am missing though.

A few other things are worth highlighting:

- Each consul node can have more than one health check associated with it, for this reason the checks for an individual node are aggregated into a single health status before the filtering is applied. This is the same approach as taken in [consul template](https://github.com/hashicorp/consul/blob/master/api/health.go) (in particular see line 42.)

- There is some conflict with the existing `useHealthChecks` parameter, for backwards compatibility my implementation currently makes `useHealthChecks = True` override the healthStatuses parameter to be just `passing`. There may be a better approach to this, such as throwing an exception if both are defined or deprecating `useHealthChecks` in favour of `healthStatuses`.

- If `healthStatuses` is passed as just `passing` (or if `useHealthChecks = True`) then we could 
still use the `passing` parameter of the consul api to filter on the server side. However this would complicate the code for potentially little/no real gain.

- While `healthStatuses` takes an array of statuses it also allows any `any` parameter for convenience, which is also the default parameter. If no value is provided to `healthStatuses` then
it defaults to `any` which is equivalent to the existing behaviour.

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
@@ -82,12 +82,12 @@ class CatalogApi(
 }
 
 object HealthApi {
-  def apply(c: Client, statuses: Option[Set[HealthStatus.Value]]): HealthApi = new HealthApi(c, statuses, s"/$versionString")
+  def apply(c: Client, statuses: Set[HealthStatus.Value]): HealthApi = new HealthApi(c, statuses, s"/$versionString")
 }
 
 class HealthApi(
   override val client: Client,
-  val statuses: Option[Set[HealthStatus.Value]] = None,
+  val statuses: Set[HealthStatus.Value],
   override val uriPrefix: String,
   override val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
   override val stats: StatsReceiver = DefaultStatsReceiver
@@ -116,7 +116,7 @@ class HealthApi(
       "index" -> blockingIndex,
       "dc" -> datacenter,
       "tag" -> tag,
-      "passing" -> Some("false")
+      "passing" -> Some((statuses == Set(HealthStatus.Passing)).toString)
     )
     executeJson[Seq[ServiceHealth]](req, retry).map { indexed =>
       val result = indexed.value.map { health =>
@@ -135,10 +135,10 @@ class HealthApi(
           checks.reduceOption(HealthStatus.worstCase)
         )
       }
-      val nodes = statuses match {
-        case None => result
-        case Some(x) if x contains HealthStatus.Any => result
-        case Some(x) => result.filter(x contains _.Status.getOrElse(HealthStatus.Passing))
+      val nodes = if (statuses == Set(HealthStatus.Passing)) {
+        result
+      } else {
+        result.filter(statuses contains _.Status.getOrElse(HealthStatus.Passing))
       }
       Indexed[Seq[ServiceNode]](nodes, indexed.index)
     }
--- a/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
@@ -116,12 +116,19 @@ class HealthApi(
       "index" -> blockingIndex,
       "dc" -> datacenter,
       "tag" -> tag,
+      // if passing=true only nodes with all health statuses in passing state are returned,
+      // if passing=false no server side filtering is performed.
       "passing" -> Some((statuses == Set(HealthStatus.Passing)).toString)
     )
     executeJson[Seq[ServiceHealth]](req, retry).map { indexed =>
       val result = indexed.value.map { health =>
         val service = health.Service
-        val checks = health.Checks.getOrElse(List()).flatMap(_.Status).map(HealthStatus.withNameSafe(_))
+        val checks =
+          for {
+            checks <- health.Checks.toSeq
+            check <- checks
+            status <- check.Status
+          } yield HealthStatus.withNameSafe(status)
         val node = health.Node
 
         ServiceNode(
@@ -138,7 +145,7 @@ class HealthApi(
       val nodes = if (statuses == Set(HealthStatus.Passing)) {
         result
       } else {
-        result.filter(statuses contains _.Status.getOrElse(HealthStatus.Passing))
+        result.filter(x => statuses.contains(x.Status.getOrElse(HealthStatus.Passing)))
       }
       Indexed[Seq[ServiceNode]](nodes, indexed.index)
     }
--- a/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
@@ -82,11 +82,12 @@ class CatalogApi(
 }
 
 object HealthApi {
-  def apply(c: Client): HealthApi = new HealthApi(c, s"/$versionString")
+  def apply(c: Client, statuses: Option[Set[HealthStatus.Value]]): HealthApi = new HealthApi(c, statuses, s"/$versionString")
 }
 
 class HealthApi(
   override val client: Client,
+  val statuses: Option[Set[HealthStatus.Value]],
   override val uriPrefix: String,
   override val backoffs: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds),
   override val stats: StatsReceiver = DefaultStatsReceiver
@@ -115,12 +116,14 @@ class HealthApi(
       "index" -> blockingIndex,
       "dc" -> datacenter,
       "tag" -> tag,
-      "passing" -> Some("true")
+      "passing" -> Some("false")
     )
     executeJson[Seq[ServiceHealth]](req, retry).map { indexed =>
       val result = indexed.value.map { health =>
         val service = health.Service
+        val checks = health.Checks.getOrElse(List()).flatMap(_.Status).map(HealthStatus.withNameSafe(_))
         val node = health.Node
+
         ServiceNode(
           node.flatMap(_.Node),
           node.flatMap(_.Address),
@@ -128,10 +131,16 @@ class HealthApi(
           service.flatMap(_.Service),
           service.flatMap(_.Tags),
           service.flatMap(_.Address),
-          service.flatMap(_.Port)
+          service.flatMap(_.Port),
+          checks.reduceOption(HealthStatus.worstCase)
         )
       }
-      Indexed[Seq[ServiceNode]](result, indexed.index)
+      val nodes = statuses match {
+        case None => result
+        case Some(x) if x contains HealthStatus.Any => result
+        case Some(x) => result.filter(x contains _.Status.getOrElse(HealthStatus.Passing))
+      }
+      Indexed[Seq[ServiceNode]](nodes, indexed.index)
     }
   }
 }
@@ -151,7 +160,12 @@ case class Service_(
 
 case class ServiceHealth(
   Node: Option[Node],
-  Service: Option[Service_]
+  Service: Option[Service_],
+  Checks: Option[Seq[Check]]
+)
+
+case class Check(
+  Status: Option[String]
 )
 
 case class ServiceNode(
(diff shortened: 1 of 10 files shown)


consul/src/test/scala/io/buoyant/consul/v1/HealthStatusTest.scala
```diff
@@ -0,0 +1,21 @@
...
+  test("health statuses can be deserialized from lowercase names") {
+    val yaml = "[passing, warning, critical, maintenance, any]"
+    val mapper = Parser.objectMapper(yaml, Iterable.empty)
+    val modes = mapper.readValue[Seq[HealthStatus.Value]](yaml)
+    assert(modes == Seq(HealthStatus.Passing, HealthStatus.Warning, HealthStatus.Critical, HealthStatus.Maintenance, HealthStatus.Any))
+  }
+
+  test("health statuses can be aggregated by worst case") {
+    val statuses = Seq(HealthStatus.Passing, HealthStatus.Warning, HealthStatus.Critical)
+    assert(statuses.reduce(HealthStatus.worstCase) == HealthStatus.Critical)
```
hawkw: Would be nice if this test was more comprehensive...

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
```diff
@@ -161,5 +175,6 @@ case class ServiceNode(
   ServiceName: Option[String],
   ServiceTags: Option[Seq[String]],
   ServiceAddress: Option[String],
-  ServicePort: Option[Int]
+  ServicePort: Option[Int],
+  Status: Option[HealthStatus.Value]
```
hawkw: It looks like this addition is causing compile errors in `ConsulNamerTest` (which is what's currently breaking [the CI build](https://circleci.com/gh/linkerd/linkerd/5794)), since the test creates some `ServiceNode`s without passing an argument for the `Status` field. 

We'll want to add the `Status` field to the `ServiceNode`s here 
https://github.com/linkerd/linkerd/blob/ef38c6ad20071a57294bae863923aae99572b5dd/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala#L15-L32
so the test can compile.

consul/src/main/scala/io/buoyant/consul/v1/HealthStatus.scala
```diff
@@ -0,0 +1,45 @@
+package io.buoyant.consul.v1
+
+import com.fasterxml.jackson.core.{JsonGenerator, JsonParser}
+import com.fasterxml.jackson.databind.{DeserializationContext, SerializerProvider}
+import io.buoyant.config.{ConfigDeserializer, ConfigSerializer}
+
+object HealthStatus extends Enumeration {
```
hawkw: I'm not against using `Enumeration` and a custom deserializer here, but just for reference, I think we've generally represented JSON enums using a sealed class and `JsonSubTypes`, as in: 
https://github.com/linkerd/linkerd/blob/a9c1627d389fab206faac44d1f674695207ff5cd/namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/kubernetes/DtabWatch.scala#L7-L23

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
hawkw: Ah, I see you've already mentioned that in your PR, sorry – haven't had my coffee yet.

consul/src/main/scala/io/buoyant/consul/v1/HealthStatus.scala
blacktoe: I was mostly copying from the consistency mode enum also in consul, only I used enum instead of a bunch of case classes so that they get an implicit ordering for the worst case method. It seems like there are also a bunch of custom deserialisers in the [config types package](https://github.com/linkerd/linkerd/tree/a9c1627d389fab206faac44d1f674695207ff5cd/config/src/main/scala/io/buoyant/config/types). Happy to change to a case class with a value or a JsonSubType etc if it's preferable though.

https://github.com/linkerd/linkerd/blob/a9c1627d389fab206faac44d1f674695207ff5cd/consul/src/main/scala/io/buoyant/consul/v1/ConsistencyMode.scala#L9-L29

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
blacktoe: Fixed and passing

consul/src/main/scala/io/buoyant/consul/v1/HealthStatus.scala
hawkw: I think Enumeration is probably the right call in this case, just wanted to put that out there.

adleong:
Here's my suggestion on how to reconcile `useHealthChecks` with `healthStatuses`:

`useHealthChecks` indicates whether filtering should take place at all with a default of `false`.  `healthStatuses` should be a list of statuses to accept with the default being `[passing]`.  In this way, we get backwards compatibility with the current behavior but can also override the `healthStatuses` property to be something else.  This also means we don't need a special `Any` value; we can just set `useHealthChecks: false` to disable filtering altogether.

As for the implementation, I see the `passing` API parameter as an optimization.  If `healthStatuses = [passing]` then we can use the `passing` parameter to do server side filtering.  Otherwise, we have to do client side filtering.  

What do you think?

blacktoe:
Yeah, that makes sense to me. I have update the diff accordingly, should be good to go.

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
```diff
@@ -115,23 +116,31 @@ class HealthApi(
       "index" -> blockingIndex,
       "dc" -> datacenter,
       "tag" -> tag,
-      "passing" -> Some("true")
+      "passing" -> Some((statuses == Set(HealthStatus.Passing)).toString)
```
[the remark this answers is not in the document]
blacktoe: `passing=false` means skip filtering. From the (somewhat unclear) [Consul API docs](https://www.consul.io/api/health.html#list-nodes-for-service):

> passing (bool: false) - Specifies that the server should return only nodes with all checks in the passing state. This can be used to avoid additional filtering on the client side.

linkerd/docs/namer.md
```diff
@@ -186,7 +189,8 @@ prefix | `io.l5d.consul` | Resolves names with `/#/<prefix>`.
 host | `localhost` | The Consul host.
 port | `8500` | The Consul port.
 includeTag | `false` | If `true`, read a Consul tag from the path.
-useHealthCheck | `false` | If `true`, exclude app instances that are failing Consul health checks. Even if `false`, linkerd's built-in resiliency algorithms will still apply.
+useHealthCheck | `false` | If `true`, exclude app instances that do not match one of the provided `healthStatuses`. Even if `false`, linkerd's built-in resiliency algorithms will still apply.
+healthStatuses | `passing` | List of statuses to used to filter Consul app instances by. Possible values are `passing`, `warning`, `critical`, `maintenance` and the wildcard status `any`. Note that if a service defines more than one health check per app instance then the most representative statuses is used (`maintenance` > `critical` > `warning` > `passing`). If `useHealthCheck` is `false` then this parameter has no effect. Regardless of the statuses used to filter, linkerd's built-in resiliency algorithms will still apply.
```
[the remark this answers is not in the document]
blacktoe: Whoops, should be fixed now

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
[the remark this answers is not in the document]
hawkw: Can we add a comment here indicating that please?

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
```diff
@@ -135,10 +135,10 @@ class HealthApi(
...
         )
       }
-      val nodes = statuses match {
-        case None => result
-        case Some(x) if x contains HealthStatus.Any => result
-        case Some(x) => result.filter(x contains _.Status.getOrElse(HealthStatus.Passing))
+      val nodes = if (statuses == Set(HealthStatus.Passing)) {
+        result
+      } else {
+        result.filter(statuses contains _.Status.getOrElse(HealthStatus.Passing))
```
hawkw: Our preferred code style tends to discourage the use of whitespace-delimited function calls. Prefer:
```scala
result.filter(statuses.contains(_.Status.getOrElse(HealthStatus.Passing)))
```

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
```diff
@@ -115,23 +116,31 @@ class HealthApi(
       "index" -> blockingIndex,
       "dc" -> datacenter,
       "tag" -> tag,
-      "passing" -> Some("true")
+      "passing" -> Some((statuses == Set(HealthStatus.Passing)).toString)
     )
     executeJson[Seq[ServiceHealth]](req, retry).map { indexed =>
       val result = indexed.value.map { health =>
         val service = health.Service
+        val checks = health.Checks.getOrElse(List()).flatMap(_.Status).map(HealthStatus.withNameSafe(_))
```
hawkw: TIOLI:
```scala
val checks = 
  for {
    checks <- health.Checks.toSeq
    check <- checks
    status <- check.Status
 } yield HealthStatus.withNameSafe(status)
```

consul/src/main/scala/io/buoyant/consul/v1/HealthStatus.scala
```diff
@@ -0,0 +1,44 @@
...
+  type HealthStatus = Value
+  val Passing, Warning, Critical, Maintenance = Value
+  def withNameSafe(name: String): Value =
+    values.find(_.toString.toLowerCase == name.toLowerCase()).getOrElse(Passing)
```
hawkw: rm parens from `name.toLowerCase()` here please – parens on a zero-argument function [should be used to indicate that the function is executed for side effects](https://docs.scala-lang.org/style/method-invocation.html#arity-0)

consul/src/test/scala/io/buoyant/consul/v1/HealthApiTest.scala
```diff
@@ -49,4 +55,53 @@ class HealthApiTest extends FunSuite with Awaits {
...
+    assert(responsePassing.head.ServiceName == Some("hosted_web"))
+    assert(responsePassing.head.Node == Some("Sarahs-MBP-2"))
+
+    val apiCritical = HealthApi(service, Set(HealthStatus.Critical))
+    val responseCritical = await(apiCritical.serviceNodes("hosted_web")).value
+
+    assert(responseCritical.size == 0)
+  }
+
+  test("HealthApi supports filtering by health status `warning`") {
```
hawkw: Take it or leave it, but I feel like these three tests ("HealthApi supports filtering by health status {`warning`, `passing`, `critical` }") could be made DRYer, maybe like this:
```scala
def filterTest(status: HealthStatus, name: String) = 
  test(s"HealthApi supports filtering by health status `$name`") {
    val service = stubService(nodesWithChecksBuf)
    val api = HealthApi(service, Set(status))
    
    val response = await(api.serviceNodes("hosted_web")).value
    assert(response.size == 1)
    assert(response.head.ServiceName == Some("hosted_web"))
    assert(response.head.Node == Some(s"node-$name"))
    assert(response.head.Status == Some(status))
  }
```

hawkw:
I'd like to hear back from @adleong to confirm, but I think this is now ready to merge!

wmorgan:
@blacktoe thanks for the PR! if you send us your mailing address and t-shirt size we'll send a shirt your way. Email my first name @ buoyant.io or DM me on the Linkerd slack.

blacktoe:
No worries, thanks @hawkw and @adleong, my first time writing Scala (or Java for that matter) so I especially appreciate the help and feedback.

Issue #1549, opened 2017-08-02
K8s: Fallback issue if there are no running pods in service


ievgen-kolomiiets:
I have two services in Kubernetes: **service-a** and **service-b**.
Dtab is configured next way:
```
/service => /service-a | /service-b
```
The issue is that if Kubernetes restarts pods in **service-a** and all pods become unavailable traffic doesn’t go to **service-b** anyway.

wmorgan:
@ievgen-kolomiiets thanks for filing this! We'll take a look.

hawkw:
I think before trying to find a fix for this, we'll want to ensure that the issue persists after #1603 is merged. That branch rewrites enough of the Kubernetes namers that it's possible this issue was fixed as well.

klingerf:
Hey @ievgen-kolomiiets, thanks for reporting this, and apologies for the delay in responding.

I tested this a bit, and in my tests linkerd did fallback from `service-a` to `service-b` when all instances of `service-a` disappeared. That said, I think a better way to go about this would be to remove the `service-a` kubernetes service object prior to the deploy, which would force the fallback to `service-b` explicitly. Then you could deploy `service-a` and restore the service object to shift traffic back to `service-a`.

In investigating this, however, I realized that there's currently a bug in linkerd where it does not properly recognize when a service has removed, and it continues to send traffic to instances of the service rather than falling back. I've described this behavior in more detail in #1612, and we're going to work on fixing that as part of the 1.2.0 release.

ievgen-kolomiiets:
Hi @klingerf.
Let me describe my case. `service-a` and `service-b` are services in different availability zones.  
For zone `a` routes look like `/service => /service-a | /service-b` and for `b` - `/service => /service-b | /service-a`. This allows us to avoid cross-zone requests as much as possible.
So if something bad happens in zone `a` and all pods are getting restarted or unavailable I want traffic to go to zone `b`. But what I saw in namerd code that fallback starts working only if there is a `Neg` resolution but instead empty address list is returned for service with no pods.

klingerf:
Great, thanks for the additional info @ievgen-kolomiiets. Your setup makes sense to me, and what you describe is how I expect it to work: the fallback should happen on `Neg`, but won't happen on `Empty`. But as I described in #1612, I think that's not actually how it works at present, and the behavior changed recently. Which version of linkerd are you running? We might actually want to fallback on `Empty` as well, but I think the fact that Linkerd behaves that way now was not a conscientious decision.

[...]

PR #1611, opened 2017-08-25
add Namer for DNS SRV records

### LINKED ISSUE #1610: DNS SRV-record service discovery plugin
## Issue Type:

- [ ] Bug report
- [x] Feature request

This is a followup from [the forum](https://discourse.linkerd.io/t/dns-srv-record-service-discovery-plugin/177) where I got feedback from @olix0r that linkerd may be interested in a Namer that uses SRV records for service discovery.

**What happened**:

My organisation uses SRV records for service discovery. I was not able to do this with existing linkerd releases because none of the existing service discovery backends implement this.

I expected to be able to use SRV records to discover instances of my service.

**Environment**:
- linkerd/namerd version, config files: n/a
- Platform, version, and config files (Kubernetes, DC/OS, etc): Custom service discovery mechanism based on DNS SRV records.
- Cloud provider or hardware configuration: n/a

ccmtaylor:
This adds a plugin that allows linkerd to use SRV records for service discovery. A `linkerd.yaml` file using the plugin might contain something like this:

```yaml
routers:
- protocol: http
  dtab: |
    /dnssrv => /#/io.l5d.dnssrv;
    /svc/myservice =>
               /dnssrv/myservice.srv.example.org &
               /dnssrv/myservice2.srv.example.org;
    /svc/other => 
               /dnssrv/other.srv.example.org;

namers:
- kind: io.l5d.dnssrv
  experimental: true
```

Fixes #1610

### DIFF
--- a/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
+++ b/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
@@ -1,17 +1,17 @@
 package io.buoyant.namer.dnssrv
 
-import com.twitter.finagle.{Addr, Name, NameTree, Path}
 import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.finagle.{Addr, Name, NameTree, Path}
 import com.twitter.util.{Duration, NullTimer}
-import io.buoyant.test.FunSuite
+import io.buoyant.namer.RichActivity
+import io.buoyant.test.{Awaits, FunSuite}
 import org.scalatest.Matchers
 import org.xbill.DNS
 
-class DnsSrvNamerIntegrationTest extends FunSuite with Matchers {
+class DnsSrvNamerIntegrationTest extends FunSuite with Awaits with Matchers {
   test("can resolve some public SRV revord") {
     val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, new NullTimer, Duration.Zero, new NullStatsReceiver)
-    val result = namer.lookupSrv("_http._tcp.mxtoolbox.com.", Path.read("/foo"), Path.empty)
-    result.get().simplified match {
+    await(namer.lookup(Path.read("/_http._tcp.mxtoolbox.com.")).toFuture) match {
       case NameTree.Leaf(Name.Bound(varAddr)) => varAddr.sample() match {
         case Addr.Bound(addrs, _) => addrs should not be empty
         case addr => fail(s"unexpected addr: $addr")
--- a/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
+++ b/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
@@ -10,7 +10,7 @@ import org.xbill.DNS
 class DnsSrvNamerIntegrationTest extends FunSuite with Matchers {
   test("can resolve some public SRV revord") {
     val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, new NullTimer, Duration.Zero, new NullStatsReceiver)
-    val result = namer.lookupSrv("_http._tcp.mxtoolbox.com.", Path.read("/foo"), null)
+    val result = namer.lookupSrv("_http._tcp.mxtoolbox.com.", Path.read("/foo"))
     result.get().simplified match {
       case NameTree.Leaf(Name.Bound(varAddr)) => varAddr.sample() match {
         case Addr.Bound(addrs, _) => addrs should not be empty
--- a/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
+++ b/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
@@ -10,7 +10,7 @@ import org.xbill.DNS
 class DnsSrvNamerIntegrationTest extends FunSuite with Matchers {
   test("can resolve some public SRV revord") {
     val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, new NullTimer, Duration.Zero, new NullStatsReceiver)
-    val result = namer.lookupSrv("_http._tcp.mxtoolbox.com.", Path.read("/foo"))
+    val result = namer.lookupSrv("_http._tcp.mxtoolbox.com.", Path.read("/foo"), Path.empty)
     result.get().simplified match {
       case NameTree.Leaf(Name.Bound(varAddr)) => varAddr.sample() match {
         case Addr.Bound(addrs, _) => addrs should not be empty
--- a/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
+++ b/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
@@ -1,0 +1,22 @@
+package io.buoyant.namer.dnssrv
+
+import com.twitter.finagle.{Addr, Name, NameTree, Path}
+import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.util.{Duration, NullTimer}
+import io.buoyant.test.FunSuite
+import org.scalatest.Matchers
+import org.xbill.DNS
+
+class DnsSrvNamerIntegrationTest extends FunSuite with Matchers {
+  test("can resolve some public SRV revord") {
+    val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, new NullTimer, Duration.Zero, new NullStatsReceiver)
+    val result = namer.lookupSrv("_http._tcp.mxtoolbox.com.", Path.read("/foo"), null)
+    result.get().simplified match {
+      case NameTree.Leaf(Name.Bound(varAddr)) => varAddr.sample() match {
+        case Addr.Bound(addrs, _) => addrs should not be empty
+        case addr => fail(s"unexpected addr: $addr")
+      }
+      case other => fail(s"unexpected result: $other")
+    }
+  }
+}
--- a/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
+++ b/namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
@@ -1,8 +1,9 @@
 package io.buoyant.namer.dnssrv
 
 import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.finagle.util.DefaultTimer
 import com.twitter.finagle.{Addr, Name, NameTree, Path}
-import com.twitter.util.{Duration, NullTimer}
+import com.twitter.util.{Duration, FuturePool}
 import io.buoyant.namer.RichActivity
 import io.buoyant.test.{Awaits, FunSuite}
 import org.scalatest.Matchers
@@ -10,7 +11,7 @@ import org.xbill.DNS
 
 class DnsSrvNamerIntegrationTest extends FunSuite with Awaits with Matchers {
   test("can resolve some public SRV revord") {
-    val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, new NullTimer, Duration.Zero, new NullStatsReceiver)
+    val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, Duration.Zero, new NullStatsReceiver, FuturePool.unboundedPool)(DefaultTimer)
     await(namer.lookup(Path.read("/_http._tcp.mxtoolbox.com.")).toFuture) match {
       case NameTree.Leaf(Name.Bound(varAddr)) => varAddr.sample() match {
         case Addr.Bound(addrs, _) => addrs should not be empty
--- a/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
+++ b/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
@@ -15,22 +15,23 @@ class DnsSrvNamer(prefix: Path, resolver: DNS.Resolver, timer: Timer, refreshInt
   override def lookup(path: Path): Activity[NameTree[Name]] = memoizedLookup(path)
 
   private val log = Logger.get("dnssrv")
-  private val memoizedLookup: (Path) => Activity[NameTree[Name]] = Memoize {
-    case path@Path.Utf8(address, _) =>
-      val id = path.take(1)
-      Activity(Var.async[State[NameTree[Name]]](Activity.Pending) { state =>
-        timer.schedule(refreshInterval) {
-          val next = lookupSrv(address, prefix ++ id) match {
-            case Return(nameTree) => Activity.Ok(nameTree)
-            case Throw(e) => Activity.Failed(e)
+  private val memoizedLookup: (Path) => Activity[NameTree[Name]] = Memoize { path =>
+    path.take(1) match {
+      case id@Path.Utf8(address) =>
+        Activity(Var.async[State[NameTree[Name]]](Activity.Pending) { state =>
+          timer.schedule(refreshInterval) {
+            val next = lookupSrv(address, prefix ++ id, path.drop(1)) match {
+              case Return(nameTree) => Activity.Ok(nameTree)
+              case Throw(e) => Activity.Failed(e)
+            }
+            state.update(next)
           }
-          state.update(next)
-        }
-      })
-    case _ => Activity.value(NameTree.Neg)
+        })
+      case _ => Activity.value(NameTree.Neg)
+    }
   }
 
-  private[dnssrv] def lookupSrv(address: String, id: Path): Try[NameTree[Name]] = Try {
+  private[dnssrv] def lookupSrv(address: String, id: Path, residual: Path): Try[NameTree[Name]] = Try {
     val question = DNS.Record.newRecord(
       DNS.Name.fromString(address),
       DNS.Type.SRV,
@@ -62,7 +63,7 @@ class DnsSrvNamer(prefix: Path, resolver: DNS.Resolver, timer: Timer, refreshInt
           NameTree.Neg
         } else {
           log.trace("got %d results for %s", srvRecords.length, address)
-          NameTree.Leaf(Name.Bound(Var.value(Addr.Bound(srvRecords: _*)), id))
+          NameTree.Leaf(Name.Bound(Var.value(Addr.Bound(srvRecords: _*)), id, residual))
         }
       case code =>
         log.warning("unexpected RCODE: %s for %s", DNS.Rcode.string(code), address)
--- a/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
+++ b/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
@@ -12,7 +12,13 @@ import com.twitter.util.Activity.State
 import com.twitter.util._
 import org.xbill.DNS
 
-class DnsSrvNamer(prefix: Path, resolver: DNS.Resolver, refreshInterval: Duration, stats: StatsReceiver, pool: FuturePool)(implicit val timer: Timer)
+class DnsSrvNamer(
+  prefix: Path,
+  resolver: DNS.Resolver,
+  refreshInterval: Duration,
+  stats: StatsReceiver,
+  pool: FuturePool
+)(implicit val timer: Timer)
   extends Namer {
 
   override def lookup(path: Path): Activity[NameTree[Name]] = memoizedLookup(path)
--- a/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
+++ b/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
@@ -2,6 +2,7 @@ package io.buoyant.namer.dnssrv
 
 import java.io.IOException
 import java.net.InetSocketAddress
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.TimeUnit
 
 import com.twitter.finagle._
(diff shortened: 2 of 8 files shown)


hawkw:
Hi @ccmtaylor, this is great, thank you for your contribution!

I'll find some reviewers for this PR; in the mean time, would you mind making sure you've signed our [Contributor License Agreement](https://buoyant.io/cla/) so we can merge this?

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,80 @@
...
+          // valid DNS entry, but no instances.
+          // for some reason, NameTree.Empty doesn't work right
+          log.trace("empty response for %s", address)
+          NameTree.Neg
+        } else {
+          log.trace("got %d results for %s", srvRecords.length, address)
+          NameTree.Leaf(Name.Bound(Var.value(Addr.Bound(srvRecords: _*)), id))
+        }
+      case code =>
+        log.trace("unexpected RCODE: %d for %s", code, address)
```
hawkw: I think anything that makes `NameTree.Fail` probably ought to be logged at the `warning` level.

namer/dnssrv/src/test/scala/io/buoyant/namer/dnssrv/DnsSrvNamerTest.scala
```diff
@@ -0,0 +1,53 @@
...
+                  |- localhost
+      """.stripMargin
+
+    val mapper = Parser.objectMapper(yaml, Iterable(Seq(DnsSrvNamerInitializer)))
+    val curator = mapper.readValue[NamerConfig](yaml).asInstanceOf[DnsSrvNamerConfig]
+    assert(curator.refreshIntervalSeconds === Some(60))
+    assert(curator.dnsHosts === Some(Seq("localhost")))
+  }
+
+  test("can resolve some public SRV revord") {
```
hawkw: I'd feel somewhat more comfortable if tests that create external network connections went in integration rather than unit tests.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,80 @@
...
+import com.twitter.util._
+import org.xbill.DNS
+
+import scala.collection.concurrent.TrieMap
+
+class DnsSrvNamer(prefix: Path, resolver: DNS.Resolver, timer: Timer, refreshInterval: Duration, stats: StatsReceiver)
+  extends Namer {
+
+  private val log = Logger.get("dnssrv")
+  private val cache = TrieMap.empty[Path, Var[State[NameTree[Name]]]]
```
hawkw: Take it or leave it: consider [com.twitter.util.Memoize](https://twitter.github.io/util/docs/com/twitter/util/Memoize$.html#apply[A,B](f:A=>B):A=>B) as an alternative to `map.getOrElseUpdate`.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamerInitializer.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  @JsonIgnore
+  override def newNamer(params: Params): Namer = {
+    import org.xbill.DNS
+    val stats = params[param.Stats].statsReceiver.scope(prefix.show.stripPrefix("/"))
+    val resolver = dnsHosts match {
+      case Some(hosts) => new DNS.ExtendedResolver(hosts.toArray)
+      case None => new DNS.ExtendedResolver()
+    }
+    resolver.setEDNS(0, 2048, 0, Collections.EMPTY_LIST)
```
hawkw: nit: maybe put the level, payload size, etc, in a constant?

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -9,36 +9,28 @@ import com.twitter.util.Activity.State
...
           }
+          state.update(next)
         }
-      )
-      Activity(states)
+      })
     case _ => Activity.value(NameTree.Neg)
   }
 
-  private[dnssrv] def lookupSrv(address: String, id: Path, residual: Path): Try[NameTree[Name]] = Try {
```
ccmtaylor: I've seen other namers hand around a `residual` path, but I'm not sure how to use it, or why they would.

ccmtaylor:
@hawkw thaks for the quick review! I've addressed your comments; let me know if things look alright.

> would you mind making sure you've signed our Contributor License Agreement so we can merge this?

SoundCloud recently signed a CLA. I used my SC email in the commits, but if you prefer, I can close this PR and re-submit from a fork under github.com/soundcloud.

hawkw:
@ccmtaylor:
> SoundCloud recently signed a CLA. I used my SC email in the commits, but if you prefer, I can close this PR and re-submit from a fork under github.com/soundcloud.

Oh, okay; I wasn't aware of that and "please sign the CLA" is just in my boilerplate response for all first-time contributors. You're fine, then!

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,72 @@
...
+            hosts.get(srv.getTarget) match {
+              case Some(inetAddress) => Address(new InetSocketAddress(inetAddress, srv.getPort))
+              case None => Address(srv.getTarget.toString, srv.getPort)
+            }
+        }
+        if (srvRecords.isEmpty) {
+          // valid DNS entry, but no instances.
+          // for some reason, NameTree.Empty doesn't work right
+          log.trace("empty response for %s", address)
+          NameTree.Neg
```
adleong: It's debatable if you want NameTree.Neg here or a leaf with Addr.Bound containing an empty set.  The former will consider the name invalid and will fall back to any alternatives in the NameTree.  The later will be considered a valid name but any requests will fail because there are no addresses to send to.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,72 @@
...
+        } else {
+          log.trace("got %d results for %s", srvRecords.length, address)
+          NameTree.Leaf(Name.Bound(Var.value(Addr.Bound(srvRecords: _*)), id))
```
adleong: you want to set the residual as the `path` of the Name.Bound here

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,72 @@
...
+        }
+      case code =>
+        log.warning("unexpected RCODE: %s for %s", DNS.Rcode.string(code), address)
+        NameTree.Fail
```
adleong: I think this should be a `Throw(...)` instead of a `Return(NameTree.Fail)`

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamerInitializer.scala
```diff
@@ -0,0 +1,50 @@
...
+object DnsSrvNamerInitializer extends DnsSrvNamerInitializer
+
+case class DnsSrvNamerConfig(refreshIntervalSeconds: Option[Int], dnsHosts: Option[Seq[String]]) extends NamerConfig {
+
+  @JsonIgnore
+  override def defaultPrefix: Path = Path.read("/io.l5d.dnssrv")
+
+  @JsonIgnore
+  override def newNamer(params: Params): Namer = {
+    import org.xbill.DNS
```
adleong: move imports to top of file, please

namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
```diff
@@ -0,0 +1,22 @@
...
+import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.util.{Duration, NullTimer}
+import io.buoyant.test.FunSuite
+import org.scalatest.Matchers
+import org.xbill.DNS
+
+class DnsSrvNamerIntegrationTest extends FunSuite with Matchers {
+  test("can resolve some public SRV revord") {
+    val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, new NullTimer, Duration.Zero, new NullStatsReceiver)
+    val result = namer.lookupSrv("_http._tcp.mxtoolbox.com.", Path.read("/foo"))
```
adleong: I think this would be a better black-box test if you use namer.lookup instead.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,72 @@
...
+import org.xbill.DNS
+
+class DnsSrvNamer(prefix: Path, resolver: DNS.Resolver, timer: Timer, refreshInterval: Duration, stats: StatsReceiver)
+  extends Namer {
+
+  override def lookup(path: Path): Activity[NameTree[Name]] = memoizedLookup(path)
+
+  private val log = Logger.get("dnssrv")
+  private val memoizedLookup: (Path) => Activity[NameTree[Name]] = Memoize {
+    case path@Path.Utf8(address, _) =>
```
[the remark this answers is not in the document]
ccmtaylor: ok. I actually wrote it that way first, but thought it looked a bit nicer like this because it's one fewer level of indentation (though I guess `Utf8.unapply()` might need to do more work.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
ccmtaylor: a major linkerd use-case for us is to load-balance across two or more deployments of a service (with different SRV records), i.e. the dtab looks like this:
```
    /svc/myservice =>
               /dnssrv/myservice1.srv.example.org &
               /dnssrv/myservice2.srv.example.org;
```
if `myservice1` returns `Addr.Bound(Set.empty)`, would linkerd still call `myservice2`?

namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
ccmtaylor: I agree, but I couldn't figure out how to do something like `Future.await()` for an `Activity`. The closest I could find is `Activity.sample`, but I'm not sure if that would race if the lookup is too slow. I'll push something, but please take a look if that's correct.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,72 @@
...
+            case Return(nameTree) => Activity.Ok(nameTree)
+            case Throw(e) => Activity.Failed(e)
+          }
+          state.update(next)
+        }
+      })
+    case _ => Activity.value(NameTree.Neg)
+  }
+
+  private[dnssrv] def lookupSrv(address: String, id: Path): Try[NameTree[Name]] = Try {
```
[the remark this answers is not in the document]
ccmtaylor: `resolver.send(query)` throws `IOException`. I'll scope the `Try{}` to that call.

ccmtaylor:
@hawkw @adleong: what kind of instrumentation would you expect on namers? I'm passing in a `StatsReceiver`, but not using it atm. I couldn't see anything consistent across the other namers, but I'm not sure if there exist some metrics across all kinds of namers (e.g. lookup latency, counts)?

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
ccmtaylor: ok. I'm a little unclear on the behaviour of the different `NameTree.*` variants. When would I use `NameTree.Fail`?

namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.twitter.finagle.{Addr, Name, NameTree, Path}
+import com.twitter.util.{Activity, Duration, NullTimer}
+import io.buoyant.test.{Awaits, FunSuite}
+import org.scalatest.Matchers
+import org.xbill.DNS
+
+class DnsSrvNamerIntegrationTest extends FunSuite with Awaits with Matchers {
+  test("can resolve some public SRV revord") {
+    val namer = new DnsSrvNamer(Path.empty, new DNS.ExtendedResolver, new NullTimer, Duration.Zero, new NullStatsReceiver)
+    Activity.sample(namer.lookup(Path.read("/_http._tcp.mxtoolbox.com."))) match {
```
ccmtaylor: as mentioned in a previous comment, I'd appreciate a +1 that this shouldn't and doesn't just race on the `Activity` being in `Pending` state.

namer/dnssrv/src/test/scala/io/buoyant/namer/dnssrv/DnsSrvNamerTest.scala
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  test("service registration") {
+    assert(LoadService[NamerInitializer]().exists(_.isInstanceOf[DnsSrvNamerInitializer]))
+  }
+
+  test("parse config") {
+    val yaml = s"""
+                  |kind: io.l5d.dnssrv
+                  |experimental: true
```
ccmtaylor: should I make this required? If so, where?

hawkw: Set `experimentalRequired = true` in `DnsSrvNamerInitializer`. See (for example) `K8sNamerInitializer`:
https://github.com/linkerd/linkerd/blob/5d22b56489c4e646672e5d7fce6ef799d7574397/namer/k8s/src/main/scala/io/buoyant/namer/k8s/K8sExternalInitializer.scala#L34-L35

hawkw:
@ccmtaylor we don't currently collect any metrics from namers, although we probably _should_.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
adleong: If you use an empty address set then half of the time you'll pick the first branch of the union and fail the request and half of the time you'll pick the second branch of the union and use myservice2

If you use NameTree.Neg then you'll skip to the myservice2 every time.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
adleong: It can be useful if you want to artificially halt evaluation of a name tree, but it's rarely used in practice.

namer/dnssrv/src/integration/scala/io/buoyant/namer/dnssrv/DnsSrvNamerIntegrationTest.scala
adleong: if you import `io.buoyant.namer.RichActivity` then you can do `await(namer.lookup(...).toFuture)`

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
ccmtaylor: sounds like `NameTree.Neg` is the right choice then :). In general, why would linkerd _not_ fallback/load-balance for empty sets? I see this has come up in #1612 and #1549, too.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  private def lookupSrv(address: String, id: Path, residual: Path): Try[NameTree[Name]] = {
+    val question = DNS.Record.newRecord(
+      DNS.Name.fromString(address),
+      DNS.Type.SRV,
+      DNS.DClass.IN
+    )
+    val query = DNS.Message.newQuery(question)
+    log.debug("looking up %s", address)
+    Try(resolver.send(query)) flatMap { message =>
```
olix0r: I imagine that resolver.send is a potentially blocking call?  If so, then I think DnsSrvNamer should probably be constructed with a com.twitter.util.FuturePool so that these calls don't block the timer thread.

Furthermore, it may be a good idea to record a stat around the lookup times.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,76 @@
...
+    path.take(1) match {
+      case id@Path.Utf8(address) =>
+        Activity(Var.async[State[NameTree[Name]]](Activity.Pending) { state =>
+          timer.schedule(refreshInterval) {
+            val next = lookupSrv(address, prefix ++ id, path.drop(1)) match {
+              case Return(nameTree) => Activity.Ok(nameTree)
+              case Throw(e) => Activity.Failed(e)
+            }
+            state.update(next)
+          }
```
olix0r: just to confirm i understand this... timer.schedule returns a closable and the timer is canceleld when the Var observation is released?

ccmtaylor: yes, that is correct

ccmtaylor:
lifting this to a top-level comment, so it doesn't get lost in revisions:

> @olix0r: I imagine that resolver.send is a potentially blocking call? If so, then I think `DnsSrvNamer` should probably be constructed with a com.twitter.util.FuturePool so that these calls don't block the timer thread.

very good point, thanks @olix0r! I tried (and failed) to do this today. It's straight forward to make `lookupSrv` async with a `FuturePool` by doing

```scala
futurePool(resolver.send(query)) flatMap { message =>
  //...
  Future.value(NameTree.whatever)
  //...
  Future.exception(e)
}
```

but I can't figure out how to fit Futures into the Activity/Var/Timer APIs :(. Here's what I tried:

* I found `Activity.future(Future[T])`, but that seems to run a future once, and then stop.
* I tried to use `val (act, witness) = Activity[NameTree[Name]()` like this:
  ```scala
  // with lookupSrv modified to return Future[Nametree[Name]] as outlined above
  case id@Path.Utf8(address) =>
    val (act, witness) = Activity[NameTree[Name]()
    def loop(): Future[NameTree[Name]] = lookupSrv(address, prefix ++ id, path.drop(1)).respond { result =>
      witness.notify(result)
      timer.doLater(refreshInterval)(loop())
    }
    act
  ```
  but I'm left with no place to start and evaluate the loop -- calling `Await.result(loop())` would miss the point :)

I'm pretty fluent in using Futures, but the Activity and Var APIs are unfamiliar to me, so I'd appreciate any hints.

olix0r:
@ccmtaylor The Var/Activity apis are a bit nuanced and are difficult at times for all of us ;)  If you poke around some other namers you may find an idiom like the following:

```scala

implicit private[this] val timer = DefaultTimer.twitter

// ...
  Activity(Var.async[NameTree[Name]](NameTree.Pending) { state =>

      def loop(): Future[Unit] = {
        if (stopped) Future.Unit
        else pool(resolve(name)).transform { result =>
            result match {
              Ok(addrs) =>
                state() = NameTree.Leaf(Addr.Bound(addrs))

              Throw(e) =>
                log.error(e, "resolution error: %s", name)
                // it may be appropriate to fail the resolution (i.e. Addr.Failed) if there was no good state before
            }

            // Wait a TTL before resolving again.
            Future.sleep(ttl).before(loop())
        }
      }

      val pending = loop()
      Closable.make { _ =>
        stopped = true

        // Cancel the timer notification or pending requests
        pending.raise(...)

        Future.Unit
      }
  })
...

... or something roughly like that.

**Comment by ccmtaylor** [comment]
@olix0r thanks for the feedback, that worked great! I found `Future.whileDo`, which is a nice wrapper around the `loop`, and I used an `AtomicBoolean` as shown in the [twitter-util cookbook](https://twitter.github.io/util/guide/util-cookbook/futures.html#future-recursion).

File: namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala [hunk 17]
```diff
@@ -0,0 +1,105 @@
...
+import java.util.concurrent.TimeUnit
+
+import com.twitter.finagle._
+import com.twitter.finagle.stats.{Stat, StatsReceiver}
+import com.twitter.logging.Logger
+import com.twitter.util.Activity.State
+import com.twitter.util._
+import org.xbill.DNS
+
+class DnsSrvNamer(prefix: Path, resolver: DNS.Resolver, refreshInterval: Duration, stats: StatsReceiver, pool: FuturePool)(implicit val timer: Timer)
```
olix0r: nit: prefer to break this very long line into a line per argument

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,105 @@
...
+  private def lookupSrv(address: String, id: Path, residual: Path): Future[NameTree[Name]] = {
+    val question = DNS.Record.newRecord(
+      DNS.Name.fromString(address),
+      DNS.Type.SRV,
+      DNS.DClass.IN
+    )
+    val query = DNS.Message.newQuery(question)
+    log.debug("looking up %s", address)
+    pool {
+      val message = Stat.time(latency, TimeUnit.SECONDS)(resolver.send(query))
```
olix0r: probably better to measure in milliseconds?

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -0,0 +1,105 @@
...
+
+class DnsSrvNamer(prefix: Path, resolver: DNS.Resolver, refreshInterval: Duration, stats: StatsReceiver, pool: FuturePool)(implicit val timer: Timer)
+  extends Namer {
+
+  override def lookup(path: Path): Activity[NameTree[Name]] = memoizedLookup(path)
+
+  private val success = stats.counter("lookup_successes_total")
+  private val failure = stats.counter("lookup_failures_total")
+  private val zeroResults = stats.counter("lookup_zero_results_total")
+  private val latency = stats.stat("request_duration_seconds")
```
olix0r: as mentioned below, this should probably be measured in millis (and have suffix `_ms` like other finagle metrics)

hawkw:
Hi @ccmtaylor, thanks for those last commits yesterday! I'm going to go ahead and merge this, and it'll show up in Linkerd 1.2.0 (which we're expecting to release by September 14th, if not before). Thanks again for your contribution!

ccmtaylor:
Thanks for your time and feedback, @hawkw, @adleong and @olix0r! I just noticed that I made an editing mistake while writing the docs. #1623  fixes it, sorry!

hawkw:
Thanks for the fix, @ccmtaylor!

PR #1616, opened 2017-08-30
Move main Istio code to its own module

pcalcado:
Istio and Kubernetes are different software layers, and our module system should reflect that.


project/LinkerdBuild.scala
```diff
@@ -127,7 +133,7 @@ object LinkerdBuild extends Base {
       .withTests()
 
     val k8s = projectDir("namer/k8s")
-      .dependsOn(LinkerdBuild.k8s, core)
+      .dependsOn(LinkerdBuild.k8s, LinkerdBuild.istio, core)
```
hawkw: I think rather than making the `namer/k8s` project depend on `istio`, it might make more sense to move the Istio namers to a new `namer/istio` project?

project/LinkerdBuild.scala
```diff
@@ -440,7 +446,7 @@ object LinkerdBuild extends Base {
       .withTests()
 
     val k8s = projectDir("interpreter/k8s")
-        .dependsOn(Namer.core, LinkerdBuild.k8s, perHost, subnet)
+        .dependsOn(Namer.core, LinkerdBuild.k8s, LinkerdBuild.istio, perHost, subnet)
```
hawkw: Same comment as above.

project/LinkerdBuild.scala
pcalcado: Oh yeah, great point. I didn't want to change too much, hence the k8s->istio dep, but this looks like an easy one and maybe we'll be able to break it completely in one go.

hawkw:
@pcalcado it looks like 3a843eb has broken the `examples/istio.yml` test on CI, so I think your changes to `io.buoyant.namer.NamerInitializer` have prevented the Istio namer from being loaded.

I also think it seems odd that the (non-Istio) K8s namer & interpreter initializers are now being provided by the `istio` module? I'd like to see all of the non-Istio-specific k8s services provided by the `k8s` module, so that it could (theoretically) be used without the `istio` module.

pcalcado:
@hawkw I was thinking "how the hell is this working on my box..." and then realised I hadn't reloaded SBT -_-

Should be fixed now, including the initializers

project/LinkerdBuild.scala
```diff
@@ -474,7 +488,7 @@ object LinkerdBuild extends Base {
     object Protocol {
 
       val h2 = projectDir("linkerd/protocol/h2")
-        .dependsOn(core, Router.h2, k8s, Finagle.h2 % "test->test;e2e->test")
+        .dependsOn(core, Router.h2, istio, Finagle.h2 % "test->test;e2e->test")
```
adleong: I wonder if the istio identifiers should be in a separate project somewhere?  
istio-identifiers depends on [
  * linkerd-protocol-h2
  * linkerd-protocol-http
  * istio
]

Something like that?

hawkw: Yeah, I think I'm inclined to agree w/ @adleong on this – I feel like just replacing every dependency on `k8s` with `istio` doesn't _really_ accomplish the goal of this changeset (i.e. to decouple our Kubernetes and Istio modules)?

pcalcado: This makes sense. Because this branch is a dependency for the mixer work I'll make an issue for it and tackle once that work is done

hawkw:
🎉

PR #1617, opened 2017-08-31
Make PullState more object-oriented

hawkw:
Take it or leave it, but I think this is prettier^H^H^H^H^H^H^H^Hbetter separation of concerns or something

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
@@ -42,6 +42,27 @@ class BufferedStream(underlying: Stream, bufferCapacity: Long = 8.kilobytes.byte
   else
     _onEnd
 
+  private[this] class Fork(
+    override protected[this] val frameQ: AsyncQueue[Frame]
+  ) extends AsyncQueueReader { child =>
+
+    @inline def readFromQueue(): Future[Frame] = super.read()
+
+    override def read(): Future[Frame] = bufferedStream.synchronized {
+      // If the queue is empty, pull a Frame from the underlying Stream (which will be
+      // offered to all child Streams) before polling the queue.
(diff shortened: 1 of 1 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
```diff
@@ -96,9 +87,22 @@ class BufferedStream(underlying: Stream, bufferCapacity: Long = 8.kilobytes.byte
...
   }
 
-  private[this] sealed trait PullState
-  private[this] case object Idle extends PullState
-  private[this] case class Pulling(f: Future[Unit]) extends PullState
+  private[this] sealed trait PullState {
+    def pullFrame(): Future[Frame]
+  }
+  private[this] case object Idle extends PullState {
+    override def pullFrame(): Future[Frame] = {
```
adleong: Probably worth commenting that this is NOT thread-safe and must be externally synchronized (as it is in fork::read)

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
```diff
@@ -61,51 +82,46 @@ class BufferedStream(underlying: Stream, bufferCapacity: Long = 8.kilobytes.byte
...
+     */
+    def pullFrame(child: Fork): Future[Frame]
+  }
+
+  private[this] case object Idle extends PullState {
+    override def pullFrame(child: Fork): Future[Frame] = {
+      val f = bufferedStream.pull()
+      pullState = Pulling(f)
+      f.ensure { pullState = Idle }
+      child.readFromQueue()
```
adleong: Now that we have to pass child in here, I don't think this really accomplishes any separation of concerns.  Since the flow of the program bounces from Fork to here and then back to Fork, I think this is actually more confusing.

Issue #1472, opened 2017-07-06
Strange class cast exception


dereulenspiegel:
Hello,

we are using linkerd successfully for about 2 months now. Strangely since a few days we get lot (like hundreds per hour) of these exceptions:
```
java.lang.ClassCastException: Transport.cast failed. Expected type io.netty.handler.codec.http2.Http2Frame but found io.netty.handler.codec.http.DefaultHttpRequest
at com.twitter.finagle.transport.Transport$$anon$3.$anonfun$readFn$1 (Transport.scala:321)
at com.twitter.util.Future.$anonfun$flatMap$1 (Future.scala:1089)
at com.twitter.util.Promise$Transformer.liftedTree1$1 (Promise.scala:107)
at com.twitter.util.Promise$Transformer.k (Promise.scala:107)
at com.twitter.util.Promise$Transformer.apply (Promise.scala:117)
at com.twitter.util.Promise$Transformer.apply (Promise.scala:98)
at com.twitter.util.Promise$$anon$1.run (Promise.scala:421)
at com.twitter.concurrent.LocalScheduler$Activation.run (Scheduler.scala:200)
at com.twitter.concurrent.LocalScheduler$Activation.submit (Scheduler.scala:158)
at com.twitter.concurrent.LocalScheduler.submit (Scheduler.scala:272)
at com.twitter.concurrent.Scheduler$.submit (Scheduler.scala:108)
at com.twitter.util.Promise.runq (Promise.scala:406)
at com.twitter.util.Promise.updateIfEmpty (Promise.scala:801)
at com.twitter.util.Promise.update (Promise.scala:775)
at com.twitter.util.Promise.setValue (Promise.scala:751)
at com.twitter.concurrent.AsyncQueue.offer (AsyncQueue.scala:123)
at com.twitter.finagle.netty4.transport.ChannelTransport$$anon$1.channelRead (ChannelTransport.scala:181)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead (AbstractChannelHandlerContext.java:362)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead (AbstractChannelHandlerContext.java:348)
at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead (AbstractChannelHandlerContext.java:340)
at io.netty.channel.ChannelInboundHandlerAdapter.channelRead (ChannelInboundHandlerAdapter.java:86)
[...]

PR #1618, opened 2017-08-31
Reduce scariness of log message when a H2 router receives a H1 request

hawkw:
Previously, when a HTTP/2 router received a HTTP/1 request, a `ClassCastException` was thrown, resulting in a "strange" stack trace (#1472). 

I've added a line to match these exceptions & log a more meaningful warning.

Tested this by `curl`ing the H2 router started with `./sbt linkerd-examples/h2:run`. While sending a request to the router's port with `curl` used to produce:
```
I 0831 21:34:14.309 UTC THREAD80: serving h2 on localhost/127.0.0.1:4142
I 0831 21:34:14.370 UTC THREAD80: initialized
E 0831 21:34:41.961 UTC THREAD95: [S L:/127.0.0.1:4142 R:/127.0.0.1:55072] dispatcher failed
java.lang.ClassCastException: Transport.cast failed. Expected type io.netty.handler.codec.http2.Http2Frame but found io.netty.handler.codec.http.DefaultHttpRequest
```
now it produces:
```
I 0831 22:03:50.611 UTC THREAD125: serving h2 on localhost/127.0.0.1:4142
I 0831 22:03:50.634 UTC THREAD125: initialized
W 0831 22:04:02.467 UTC THREAD138: [S L:/127.0.0.1:4142 R:/127.0.0.1:54985] HTTP/2 router could not handle HTTP/1 request!
```

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
@@ -1,19 +1,18 @@
 package com.twitter.finagle.buoyant.h2
 package netty4
 
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 import com.twitter.finagle.stats.StatsReceiver
-import com.twitter.finagle.{ChannelClosedException, Failure}
 import com.twitter.finagle.transport.Transport
+import com.twitter.finagle.{ChannelClosedException, Failure}
 import com.twitter.logging.Logger
 import com.twitter.util._
 import io.netty.handler.codec.http2.{Http2Frame, Http2GoAwayFrame, Http2StreamFrame}
-import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 
 trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
-  import Netty4DispatcherBase._
 
   protected[this] def log: Logger
   protected[this] def prefix: String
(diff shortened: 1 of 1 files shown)


hawkw:
### Alternative Solutions
+ Ideally, we'd like to actually perform a protocol upgrade negotiation (see #841), but that isn't supported yet.
+ I suggested sending  the H1 client either a 429 ("upgrade required") error or a 400 error with a message stating that the H2 router can't handle H1 requests at this time, but @adleong said he'd prefer not to see H2 routers returning any H1 responses...

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -110,7 +111,11 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
       case Throw(e) if closed.get =>
         log.debug("[%s] dispatcher closed: %s", prefix, e)
         Future.exception(e)
-
+      // if we attempted to cast a Http2Frame as a H1 frame, log a message & kill the
+      // connection.
+      case Throw(e: ClassCastException) if e.getMessage == H1RequestError =>
```
adleong: I'm worried about this being quite brittle but... I can't really see a better way

adleong: Perhaps remove the message equality check and make the log more generic?  "HTTP/2 router could not handle non-HTTP/2 request!"?  WDYT?

hawkw: I considered that, but it seemed unwise to assume _all_ `ClassCastException`s were due to a mismatched request protocol. If a `ClassCastException` was thrown for some other reason, logging the protocol error message would obscure that failure...

What about a test that the message contains `"Transport.cast failed."`?

adleong: :+1:

PR #1620, opened 2017-09-05
add myself to maintainers

wmorgan:
Apparently the CNCF sends important email to the people in this file, so I need to be in it.


wmorgan:
Thank you team

PR #1621, opened 2017-09-05
Document disableValidation/clientAuth incompatibility

### LINKED ISSUE #1581: Linkerd Docker image fails to use BoringSSL on Linux
As initially reported [on Discourse](https://discourse.linkerd.io/t/linkerd-h2-router-rejects-requests-from-go-client/216/3?u=eliza), proxying H2 TLS connections with the the `linkerd:latest` Docker image fails with the message "JDK provider does not support NPN_AND_ALPN protocol". This message appears to indicate that BoringSSL is not being used and Linkerd is falling back to the JDK-provided SSL implementation. 

Curiously, this issue only seems to exist when the Docker image is run from a Linux host – I originally [attempted to reproduce the issue on macOS](https://discourse.linkerd.io/t/linkerd-h2-router-rejects-requests-from-go-client/216/6?u=eliza) and everything worked fine. After trying the same configuration on Linux in GCE, I was able to reproduce the issue.

hawkw:
As @adleong said in https://github.com/linkerd/linkerd/issues/1581#issuecomment-327284668, disabling TLS validation will force the use of the JDK SSL provider is thus incompatible with clientAuth.

I've confirmed this in the `linux-test` GCE instance I set up to reproduce issue #1581 – if `disableValidation` is set to `false`, everything works fine.

I've updated the docs to reflect this.

Closes #1581

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/ClientConfig.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/ClientConfig.scala
@@ -3,10 +3,9 @@ package io.buoyant.linkerd
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.twitter.conversions.time._
 import com.twitter.finagle.Stack
-import com.twitter.finagle.buoyant.{ClientAuth, PathMatcher, TlsClientConfig => FTlsClientConfig}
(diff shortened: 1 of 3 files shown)


hawkw:
Would like to add validation for this as well.

hawkw:
Update: as of https://github.com/linkerd/linkerd/pull/1621/commits/8f57a53d8381bd270aee246265c419438caea0cb, `TlsClientConfig` now validates configs – it now `require`s that if `disableValidation` is set to `true`, `clientAuth` must not be present.

PR #1622, opened 2017-09-05
Handle K8s Delete Events Correctly

adleong:
When Linkerd receives a DELETE event from the Kubernetes API, it treats the
contents of the event as the new state.  However, the semantics of the API are
that the contents of the event are the previous state and the resource should
be deleted entirely.  This causes Linkerd to get stuck in a bad state when
watching a resource as it is being deleted.

Treat DELETE events as total deletions and ignore the contents of the event.

Manually tested on Minikube.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
@@ -158,10 +158,9 @@ abstract class EndpointsNamer(
                 logEvent.deletion(oldMap -- newMap.keys)
                 logEvent.modification(oldMap, newMap)
                 oldMap ++ newMap
-              case (oldMap, v1.ServiceDeleted(service)) =>
-                val newMap = service.portMappings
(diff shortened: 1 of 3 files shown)


wjwinner:
I deploy an app in k8s which's endpoints ip are '192.168.10.141' and '192.168.10.144';
then I delete the app and redeploy it,  and new  endpoints  is 
```
world-v2-service   192.168.10.145:7778,192.168.10.146:7778
```
but when I visit the ingress url , the log show 
```
E 0907 08:45:51.882 UTC THREAD33: service failure: Failure(Invalid argument: /192.168.10.141:7778 at remote address: /192.168.10.141:7778. Remote Info: Not Available, flags=0x09) with RemoteInfo -> Upstream Address: /192.168.12.192:55967, Upstream Client Id: Not Available, Downstream Address: /192.168.10.141:7778, Downstream Client Id: #/io.l5d.k8s/default/http/world-v2-service, Trace Id: cd10d485ed2ff694.cd10d485ed2ff694<:cd10d485ed2ff694
```
the linkerd is still search for old endpoints.

PR #1628, opened 2017-09-07
Deprecate StatsD telemeter

### LINKED ISSUE #1578: Remove or improve statsd telemeter

hawkw:
Added a warning message when the StatsD telemeter is initialized. The warning message links to [this thread](https://discourse.linkerd.io/t/deprecating-the-statsd-telemeter/268?u=eliza) on Discourse.

Closes #1578


adleong:
Please update docs in linkerd/docs/telemetry.md as well

hawkw:
@adleong done 4deb53a

PR #1630, opened 2017-09-07
Remove scaladoc links

adleong:
Scaladoc links cause issues when releasing to sonatype.

Remove them.


hawkw:
it keeps happening!

PR #1631, opened 2017-09-08
Fix formatting issue with linkerd docs

klingerf:
The DNS SRV Records documentation had a few issues that were causing our API docs to render improperly. Also applying a few other misc fixes while I'm here.

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -37,7 +37,7 @@ namers:
 
 > Then reference the namer in the dtab to use it:
 
-```
+```yaml
 dtab: |
   /svc => /#/io.l5d.fs
 ```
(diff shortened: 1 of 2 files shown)


linkerd/docs/namer.md
```diff
@@ -610,18 +611,19 @@ prefix | `io.l5d.dnssrv` | Resolves names with `/#/<prefix>`.
...
 dnsHosts | `<empty list>` | If specified, linkerd will use these DNS servers to perform SRV lookups. If not specified, linkerd will use the default system resolver.
- 
+
 ### DNS-SRV Path Parameters
 
 > Dtab Path Format
 
-```
+```yaml
 /#/<prefix>/<address>
```
klingerf: Did you know? Having a code block without a format param is usually ok, but in this case totally breaks because of the `#` sign in side the block. So we're using `yaml` here and elsewhere in the document, even though the contents aren't yaml.

PR #1637, opened 2017-09-13
dnssrv: ensure that DNS names in SRV queries are absolute

ccmtaylor:
the target of SRV queries must be an absolute DNS name (ending
with a `.`). If the dtab specifies a relative name, treat it as
absolute.

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -585,6 +585,7 @@ namers:
 - kind: io.l5d.dnssrv
   experimental: true
   refreshIntervalSeconds: 5
+  domain: srv.dc-1.example.org
   dnsHosts:
   - ns0.example.org
   - ns1.example.org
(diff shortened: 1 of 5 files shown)


ccmtaylor:
I can think of two improvements:

1) add an optional search domain setting to the plugin (default to `.`), and resolve relative records relative to that. This would have the advantage of of removing some duplication, e.g. if my org serves SRV records under `srv.internal.example.com`, we could use something like (cf. the example from #1611):
```yaml
routers:
- protocol: http
  dtab: |
    /dnssrv => /#/io.l5d.dnssrv;
    /svc/myservice =>
               /dnssrv/myservice &
               /dnssrv/myservice;
    /svc/other => 
               /dnssrv/other;

namers:
- kind: io.l5d.dnssrv
  experimental: true
  searchDomain: srv.internal.example.org.
```

2) resolve relative names according to the host's DNS settings (e.g. /etc/resolv.conf on linux). The advantage here would be that it's "least surprise" from a DNS/networking point of view. OTOH, hosts often have multiple search domains, so this might involve multiple DNS queries. In addition, the system resolver is OS-specific (e.g. macOS provides a resolv.conf for compat, but doesn't use it internally; and windows does something completely different), and I'd like to avoid re-implementing it.

ccmtaylor:
I've thought about this some more; I quite like option (1) from my comment above. I'll push it as a separate commit in this PR, let me know what you think (or if I should submit it as a separate PR).

adleong:
I'm going to go ahead and merge this, but option 1 above sounds good to me.

PR #1644, opened 2017-09-15
Simplify K8s event handling

hawkw:
Since the semantics of Kubernetes `ADDED` and `MODIFIED` events call for us
to completely replace the currently-held state with the new state
contained in the event (rather than only making more granular changes to
our current state based on the event type), we can collapse the handling
of these events into a single case.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/Api.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Api.scala
@@ -181,17 +181,6 @@ object Watch {
     def resourceVersion = None
   }
 
-  /**
-   * Extractor allowing `Added` and `Modified` events --- which contain a new
-   * state --- to be handled in one match arm.
-   */
-  object NewState {
-    def unapply[O <: KubeObject](event: Watch[O]): Option[O] = event match {
-      case added: Added[O] => Some(added.`object`)
-      case modified: Modified[O] => Some(modified.`object`)
-      case _ => None
-    }
-  }
 }
 
 case class Status(
--- a/k8s/src/main/scala/io/buoyant/k8s/Api.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Api.scala
@@ -171,6 +171,7 @@ object Watch {
     @JsonIgnore
     def resourceVersion = `object`.metadata.flatMap(_.resourceVersion)
   }
+
   trait Added[O <: KubeObject] extends WithObject[O]
   trait Modified[O <: KubeObject] extends WithObject[O]
   trait Deleted[O <: KubeObject] extends WithObject[O]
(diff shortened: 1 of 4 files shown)


k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -179,6 +180,18 @@ object Watch {
     @JsonIgnore
     def resourceVersion = None
   }
+
+  /**
+   * Extractor allowing `Added` and `Modified` events --- which contain a new
+   * state --- to be handled in one match arm.
+   */
+  object NewState {
```
adleong: I think extractors are useful when it's extremely obvious from context or naming what they do.  In this case, I don't think it's immediately obvious and I end up jumping around the code to figure out what's going on.  Instead, how about factoring the common match arm bodies into a method and calling that method from both arms?  This avoids duplication while also making it clear that there are two match possibilities that both map to the same behavior.

hawkw: Sure!

hawkw:
@adleong, I think I've made all the changes you've requested?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -151,6 +151,8 @@ abstract class EndpointsNamer(
               case (oldMap, v1.ServiceAdded(service)) =>
                 val newMap = service.portMappings
                 logEvent.addition(newMap -- oldMap.keys)
```
adleong: This is mostly unrelated to this change but it just now occurs to me that in large clusters these are potentially expensive set operations.  Might be good to factor this so that the set operations happen inside of a ifTrace block.

adleong: (here and elsewhere)

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
```diff
@@ -83,19 +83,16 @@ class IngressCache(namespace: Option[String], apiClient: Service[Request, Respon
...
-            mkIngress(m)
-              .map { item => ingresses.filterNot(isNameEqual(_, item)) :+ item }
-              .getOrElse(ingresses)
-          case v1beta1.IngressDeleted(_) =>
-            Seq.empty
-          case v1beta1.IngressError(e) =>
-            log.error("k8s watch error: %s", e)
-            ingresses
-        }
+      case (oldIngresses, v1beta1.IngressAdded(a)) => oldIngresses ++ mkIngress(a)
```
adleong: Is this just a cosmetic/style change?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: Yeah, I've wanted to rewrite the logging code for a bit, thanks for reminding me!

k8s/src/main/scala/io/buoyant/k8s/IngressCache.scala
hawkw: Yeah, this class was changed to use the `NewState` extractor, and then it was changed back, but I had restructured the pattern match slightly. Can just revert.

PR #1646, opened 2017-09-19
Remove watch on service list endpoint from Consul namer

hawkw:
As reported [on Discourse](https://discourse.linkerd.io/t/linkerd-cpu-usage-climbs-over-time/274), users running Linkerd with the Consul namer have observed a steady increase of CPU usage. The current implementation of the Consul namer establishes a watch on the `/catalog/services` Consul endpoint, which returns a list of all services in the Consul datacenter. This means that Linkerd will always be notified of changes to the Consul services list, regardless of whether or not it cares about those services. When there are a great deal of services in Consul, tracking updates to this list becomes very demanding. 

We only use the services list endpoint to determine the existence of a service. However, the Consul API endpoint to look up the list of nodes for a service will return an empty list if the service does not exist, which can also be used to determine the existence of a service. I've simplified the Consul namer to use only the `/catalog/service/:service` endpoint instead of the `/catalog/services` endpoint. This means that we only watch the services we care about and are not notified on all changes. Furthermore, this allows the Consul namer to be made much simpler – the entire `DcServices` class can now be removed.

I've reproduced the performance issue by using nginx to simulate Consul. I configured nginx to statically serve a service list JSON file with 1000 services with 10 tags each. Nginx will always respond instantly, which is equivalent to watch events constantly firing. A single HTTP request to Linkerd 1.2.1 causes it to open a watch on this service list, sending CPU usage to hit 100% almost immediately. Against my branch, this no longer happens.

I had to make some changes to the unit tests in `ConsulNamerTest.scala` to reflect the new behavior of the namer. Of note is that we no longer expect the `dc/` set of stats to be incremented, as these stats were related to the watch on the services list endpoint, which no longer happens.


hawkw:
I'd like to deploy this to the test environment for a while before merging to master, in order to make sure the namer is properly exercised.

hawkw:
Deployed to `test11`.

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -21,6 +23,45 @@ private[consul] class LookupCache(
...
+          case None => s"${key.name}.service.${dc}.${domain}"
+        }
+        Addr.Metadata(Metadata.authority -> authority)
+    }
+
+  private[this] val cachedLookup: (String, SvcKey, Path, Path) => Activity[NameTree[Name]] =
+    untupled(Memoize[(String, SvcKey, Path, Path), Activity[NameTree[Name]]] {
+      case (dc, key, id, residual) =>
+        resolveDc(dc).join(domain).flatMap { case ((dcName, domainOption)) =>
+          val meta = mkMeta(key, dcName, domainOption)
```
adleong: unused?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -21,6 +23,45 @@ private[consul] class LookupCache(
   private[this] val localDcMoniker = ".local"
 
   private[this] val lookupCounter = stats.counter("lookups")
+  private[this] val serviceStats = SvcAddr.Stats(stats.scope("service"))
+
+  private[this] def mkMeta(key: SvcKey, dc: String, domain: Option[String]) =
```
adleong: duplicated from elsewhere, and effectively unused?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -21,6 +23,45 @@ private[consul] class LookupCache(
...
+            preferServiceAddress = preferServiceAddress,
+            serviceStats
+          )
+          log.debug("consul ns %s service %s found + %s", dc, key, residual.show)
+
+          val stateVar: Var[Activity.State[NameTree[Name.Bound]]] = addr.map {
+            case Addr.Neg => Activity.Ok(NameTree.Neg)
+            case Addr.Pending => Activity.Pending
+            case Addr.Failed(why) => Activity.Failed(why)
+            case _ =>
```
adleong: Prefer to explicitly list the last case instead of wildcarding it

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -21,6 +23,45 @@ private[consul] class LookupCache(
...
+            dcName,
+            key,
+            domainOption,
+            consistency = consistency,
+            preferServiceAddress = preferServiceAddress,
+            serviceStats
+          )
+          log.debug("consul ns %s service %s found + %s", dc, key, residual.show)
+
+          val stateVar: Var[Activity.State[NameTree[Name.Bound]]] = addr.map {
```
adleong: This looks scary but the Activity de-dupping that Linkerd does means that the NameTree won't get rebuilt on addr changes.
I'm just saying this to reassure myself.

hawkw:
Btw, this has been running on test11 [for a while](http://104.198.158.133/dashboard/file/tests.json?var-testrun=test11&from=1505804071452&to=1505861671452) with no signs of error so far

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
hawkw: Ah, yeah, I thought I removed that. Whoops...

linkerd/examples/consul.yaml
```diff
@@ -1,13 +1,15 @@
 # Do service discovery lookups against the Consul service catalog.
 namers:
 - kind: io.l5d.consul
-  includeTag: true
+  includeTag: false
   useHealthCheck: false
+  host: 0.0.0.0
+  port: 8300
```
hawkw: ah, I did not mean to commit this - I had made some changes locally for testing.

hawkw:
This has been running in the test env for about 24 hours and there don't seem to have been any issues so far.

adleong:
I tried this out locally and it seems to work great.

@Ashald: I know you maintain your own fork of the consul namer.  I'd love to hear your feedback on this change.

Ashald:
@adleong we actually do not - we have only a custom Dtab store plugin.

Speaking about this PR - it's definitely interesting one and I have few thoughts on it.

In our experience we used to assume that Consul is rock-solid and stable and we can rely on it a lot. Nevertheless we did appreciate Consul namer's ability to "cache" the state from Consul so that in case of Consul outage we can not only keep processing requests to services that were requested previously but we also can send requests to services that haven't been used before. So instead of being completely down we can keep operating in some degraded mode.

IMO this is definitely a feature and I wouldn't just throw it away. It might cause issues in large setups but when it can be afforded I think it provides a lot of value.

Speaking about huge number of services - this is clearly might be an issue. In our setup we use Consul namer in 2 ways:
1) Namerd discovers services that Linkerd will use
2) Linkerd discovers Namerd

In (1) we obviously want to see all the services, but in (2) we used Consul ACLs to hide all services but Namerd from Consul thus reducing the load. While such an approach is not applicable in all cases I believe ACLs might help with performance in some of them. 

On the other hand I think we also noticed high CPU utilization with cluster growth to the degree that we might be willing to sacrifice "caching" for performance.

TL;DR of all of this is that IMO it'd be better to keep both old a new namer or add an option to switch between different operational modes. Speaking of performance, I wonder if it's possible to identify what exactly takes so much of CPU and whether we can deal with that by adding some throttling when it's OK to sacrifice speed of updates.

adleong:
@Ashald I don't think the caching behavior of the consul namer will be changed by this branch.  It should still be the case that Linkerd will use last known information for any of the consul services that it's routing to in the case of a consul outage.

hawkw:
> It should still be the case that Linkerd will use last known information for any of the consul services that it's routing to in the case of a consul outage.

That's correct. @Ashald, I believe the caching you're referring to is occurring [here](https://github.com/linkerd/linkerd/blob/c52ab5fd88a70dec93fba6eeb2885516e931e31e/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala#L52-L118) – which should be unchanged in this PR.

Ashald:
I think we're talking about slightly different things. Previously Namerd used to "watch all services" so in case of Consul being unavailable it was able to resolve even those names that weren't looked up properly. As far as I understand this cache now goes away.

adleong:
@Ashald Previously Namerd would keep an up-to-date list of all services/tags, but not the actual address sets for them.  So if Consul was unreachable, it would not be possible to route to a service that had never been routed to before.

Ashald:
I see. Yeah, this way it seems that behavior remains the same.

Thanks for details! Turns out I was assuming wrong thing about Consul namer. Good to know. :)

adleong:
@Ashald Let me know if you would be willing to test this feature out and we can get you a snapshot image.  We're doing out best to test the performance of this on our side but there's no substitute for a real cluster.

Ashald:
@adleong I'd love to try that out but not sure whether we will have tome for that this week. As we're not using containers to run Linkerd, just a binary would be enough. Alternatively, we can build it on our own.

PR #1648, opened 2017-09-19
Upgrade to finagle 7.1

adleong:


linkerd/docs/protocol-http.md
```diff
@@ -485,31 +485,6 @@ Key | Default Value | Description
 mixerHost | `istio-mixer` | Hostname of the Istio Mixer server.
 mixerPort | `9091` | Port of the Mixer server.
 
-<a name="http-engines"></a>
-## HTTP Engines
```
hawkw: do you think this will necessitate a 1.3 release (since it's breaking configs?)

adleong: Yeah

PR #1649, opened 2017-09-20
Workaround for K8s watch events with incorrect status codes

### LINKED ISSUE #1636: HTTP 504 Gateway timeouts after upgrading to linkerd 1.2.0
## Filing a linkerd issue ##

Issue Type: Bug Report

**What happened**:
We have a rails application which serves web pages and it was running as a Kubernetes deployment. Every time a deployment is made and new pods come up, Linkerd gives 504 gateway timeout and when I checked Linkerd logs I can see that it was still making requests to the old endpoints. (Not sure if this is a config issue). It get's fixed on its own after some time.

Endpoints should be instantly updated whenever there is a deployment.

Rails Puma Server serving web pages and making requests to the service just after a deployment.

Our router config:
```
    - protocol: http
      label: webapp-external
      identifier:
        kind: io.l5d.header.token
        header: Host
      interpreter:
        kind: io.l5d.namerd
        dst: /$/inet/namerd.linkerd.svc.cluster.local/4100
        namespace: external
        transformers:
        - kind: io.l5d.k8s.daemonset
          namespace: linkerd
          port: webapp-ingress
          service: linkerd-internal
      servers:
      - port: 4143
        ip: 0.0.0.0
      client:
        kind: io.l5d.global
        loadBalancer:
          kind: ewma
          enableProbation: false
          maxEffort: 5
          decayTimeMs: 10
        failureAccrual:
          kind: io.l5d.consecutiveFailures
          failures: 5
      service:
        kind: io.l5d.global

    - protocol: http
      label: webapp-ingress
      identifier:
        kind: io.l5d.header.token
        header: Host
      interpreter:
        kind: io.l5d.namerd
        dst: /$/inet/namerd.linkerd.svc.cluster.local/4100
        namespace: external
        transformers:
        - kind: io.l5d.k8s.localnode
      servers:
      - port: 4145
        ip: 0.0.0.0
      client:
        kind: io.l5d.global
        loadBalancer:
          kind: ewma
          enableProbation: false
          maxEffort: 5
          decayTimeMs: 10
        failureAccrual:
          kind: io.l5d.consecutiveFailures
          failures: 5
      service:
        kind: io.l5d.global
```

**Environment**:
- linkerd/namerd version, config files: 1.2.0/1.2.0
- Platform, version, and config files (Kubernetes, DC/OS, etc): Kubernetes
- Cloud provider or hardware configuration: AWS

hawkw:
Due to a regression in some versions of Kubernetes (kubernetes/kubernetes#35068), the "resource version too old" watch event sometimes has HTTP status code 200, rather than status code 410. This event is not a fatal error and simply indicates that a watch should be restarted – k8s will fire this event for any watch that has been open for longer than thirty minutes. In Linkerd,`Watchable` currently detects this event by matching the HTTP status code of the watch event, and restarts the watch when it occurs. However, when the event is fired with the incorrect status code, the error is not handled in `Watchable` and passed downstream – in the case of issue #1636, to `EndpointsNamer`, which does not know how to handle this event. This leads to namers intermittently failing to resolve k8s endpoints.

Although this issue seems to have been fixed upstream in kubernetes/kubernetes#25369, many users of Linkerd are running versions of Kubernetes where it still occurs. Therefore,  I've added a workaround in `Watchable` to detect "resource version too old" events with status code 200 and restart the watch rather than passing these events downstream. When this occurs, Linkerd now logs a warning indicating that, although the error was handled, Kubernetes behaved erroneously. 

I've added a test to `v1.ApiTest` that replicates the Kubernetes bug.

Fixes #1636

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -149,6 +149,7 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
             // 3. The watch fails
             //
             // We need to reload the initial information to ensure we are "caught up," and then watch again.
+            log.info("k8s restarting watch on %s, resource version %s was too old", watchPath, resourceVersion)
             AsyncStream.fromFuture {
               restartWatches(labelSelector, fieldSelector).map {
                 case (ws, ver) => AsyncStream.fromSeq(ws) ++ _watch(ver)
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -110,6 +110,24 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
         Future.Unit
       })
 
+      @inline
+      def _resourceVersionTooOld() = {
+        // Gone is returned by k8s to indicate the requested resource version is too old to watch. A common scenario
+        // that exhibits this error is:
+        //
+        // 1. Kubernetes kills a connection or the connection is lost
+        // 2. We try to re-establish the watch from the last received version, but because we are only watching a
+        //    subset of resources, that version was a while ago
+        // 3. The watch fails
+        //
+        // We need to reload the initial information to ensure we are "caught up," and then watch again.
+        log.trace("k8s restarting watch on %s, resource version %s was too old", watchPath, resourceVersion)
+        AsyncStream.fromFuture {
+          restartWatches(labelSelector, fieldSelector).map {
+            case (ws, ver) => AsyncStream.fromSeq(ws) ++ _watch(ver)
+          }
+        }.flatten
+      }
       AsyncStream.fromFuture(initialState).flatMap { rsp =>
         rsp.status match {
           // NOTE: 5xx-class statuses will be retried by the infiniteRetryFilter above.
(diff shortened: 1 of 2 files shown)


k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -123,6 +141,23 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
             })
             val watchStream = Json.readStream[W](rsp.reader, Api.BufSize)
+              // it would be nice to not have to flat map over the watch stream, but this is
+              // necessary to handle the K8s bug where erorrs have the incorrect status code.
+              // hopefully this hack can eventually be removed...
+              .flatMap {
+                // special case to handle Kubernetes bug where "too old resource version" errors
+                // are returned with status code 200 rather than status code 410.
+                // see https://github.com/kubernetes/kubernetes/issues/35068 for details.
+                case e: Watch.Error[O] if e.status.message.exists(_.startsWith("too old resource version")) =>
```
adleong: Can we check `e.status.code` for `410` instead of inspecting the message?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -123,6 +141,23 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+                  log.warning(
```
adleong: Is it true that this k8s bug has no adverse effects now that we're detecting "too old" events with status code 200?  If so, I don't think we need to log.  (we will continue to log at trace level that a watch was restarted)

hawkw: Hmm. I thought it was worthwhile to encourage users to upgrade their Kubernetes version, so that we can eventually remove this workaround? I suspect it adds at least a minor performance hit.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: That's definitely better than string comparison from a performance perspective. I felt like inspecting the message might be more future-proof – what if Kubernetes were to change the status code returned for this event, or added additional events with status code 410 that should be handled differently – but if you don't think that's as big a concern, I can change this?

adleong: I think that looking for a code is actually more future proof than looking at a human readable string.  Plus it's already the case that we look for the HTTP status code 410.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
adleong: I suspect this is pretty cheap, actually.  The objects have already been parsed, after all.  And we will never be able to assume that all users have upgrade off of the affected K8s version.  The best we can do is eventually drop support for certain k8s versions.

hawkw:
aaaah my build finally succeeded! 🙌

PR #1650, opened 2017-09-20
Refactor Istio code before adding Mixer precondition checks (#1403)

pcalcado:
This consolidates some of the Istio-specific logic previously duplicated across the H2 and HTTP modules and increases test coverage in preparation for https://github.com/linkerd/linkerd/pull/1606

Reviewers will likely be interested in:
- The refactoring of the `io.buoyant.k8s.istio` package in an attempt to reduce duplication and improve (unit-)testability: https://github.com/linkerd/linkerd/tree/8d8634105dee7744908ad93346943fa1def4bde8/istio/src/main/scala/io/buoyant/k8s/istio and the HTTP and H2 Identifiers
- The new tests for the above: https://github.com/linkerd/linkerd/pull/1606/commits/10f416dd08ae8dfbb280e980c1d372d097143d63#diff-9bf8c4461d3ac3933b6a57b134a2f492R1


istio/src/main/scala/io/buoyant/k8s/istio/ClusterCache.scala
```diff
@@ -3,8 +3,13 @@ package io.buoyant.k8s.istio
...
 
-class ClusterCache(client: DiscoveryClient) extends Closable {
-  import ClusterCache._
+case class Cluster(dest: String, port: String)
+
+trait ClusterCache extends Closable {
+  def get(domain: String): Future[Option[Cluster]]
+}
+
+class ClusterCacheBackedByApi(client: DiscoveryClient) extends ClusterCache {
```
adleong: This feels a bit verbose.  How do you feel about something like `ApiClusterCache`?

istio/src/main/scala/io/buoyant/k8s/istio/IstioConfigurator.scala
```diff
@@ -0,0 +1,52 @@
...
+import com.twitter.finagle.Stack.Params
+import com.twitter.finagle.buoyant.H2
+import com.twitter.finagle.param.Label
+import com.twitter.finagle.{Address, Name}
+import com.twitter.logging.Logger
+import io.buoyant.config.types.Port
+import io.buoyant.k8s.ClientConfig
+import io.buoyant.k8s.istio.mixer.MixerClient
+
+trait IstioConfigurator {
```
adleong: As far as I can tell, nothing there depends on the current object's state.  Could these methods just be static method instead?  (ie turn this trait into an object)

I tend to shy away from the mixin pattern when possible.

istio/src/main/scala/io/buoyant/k8s/istio/IstioDataFlow.scala
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * An abstraction applicable to both Requests and Responses in an Istio service mesh.
+ *
+ * In an Istio setup, both requests and responses are augmented with [[IstioAttribute]]s that can
+ * be used in various ways. While some attributes are specific to either requests or responses, others
+ * are common in both types of objects.
+ *
+ */
+trait IstioDataFlow {
```
adleong: All of the information this trait provides is derived from the DstBoundCtx and not from any underlying request or response object.  So it seems a bit misleading to attach this to IstioRequest and IstioResponse objects.  It is also confusing that the `vals` on this trait are computed from the DstBoundCtx and saved at constructor time.  This means that the exact position in the stack where this object is constructed will determine this values, which is pretty non-intuitive.  

I would try to keep IstioDataFlow separate from the Request/Response objects and try to make it clear that constructing an IstioDataFlow takes a snapshot of the current DstBoundCtx.

istio/src/main/scala/io/buoyant/k8s/istio/IstioLoggerBase.scala
```diff
@@ -1,46 +1,22 @@
...
-      responseCode,
-      path,
-      tService.getOrElse(unknown), // target in prom (~reviews.default.svc.cluster.local)
-      unknown, // source in prom (~productpage)
-      targetLabelApp(tService).getOrElse(unknown), // service in prom (~reviews)
-      version(istioPath).getOrElse(unknown), // version in prom (~v1)
-      duration
+      response.responseCode,
+      request.requestedPath,
+      response.targetService,
```
adleong: Values from the IstioDataFlow trait are derived from the local context so `response.targetService` could just as easily be `request.targetService` or even `new IstioDataFlow {}.targetService`

istio/src/main/scala/io/buoyant/k8s/istio/IstioResponse.scala
```diff
@@ -0,0 +1,11 @@
+package io.buoyant.k8s.istio
+
+import com.twitter.util.Duration
+
+case class IstioResponse(statusCode: Int, duration: Duration) extends IstioDataFlow {
```
adleong: Any particular reason why IstioRequest is parameterized on the underlying request type and contains the underlying request but IstioResponse is not parameterized and does not include the underlying response?

istio/src/main/scala/io/buoyant/k8s/istio/RouteCache.scala
```diff
@@ -10,7 +10,12 @@ import io.buoyant.k8s.istio.ApiserverClient.RouteRuleConfig
 import io.buoyant.namer.RichActivity
 import istio.proxy.v1.config.RouteRule
 
-class RouteCache(api: ApiserverClient) extends Closable {
+trait RouteCache extends Closable {
+  def getRules: Future[Map[String, RouteRule]]
+  val routeRules: Activity[Map[String, RouteRule]]
+}
+
+class RouteCacheBackedByApi(api: ApiserverClient) extends RouteCache {
```
adleong: Name feels a bit verbose.

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
```diff
@@ -0,0 +1,121 @@
...
+
+import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
+import com.google.local.DurationProto.{Duration => GDuration}
+import com.twitter.util.Duration
+import io.buoyant.k8s.istio._
+import istio.mixer.v1._
+
```
adleong: rm blank line between imports

istio/src/test/scala/io/buoyant/k8s/istio/ApiserverClientTest.scala
```diff
@@ -6,7 +6,7 @@ import com.twitter.finagle.http.{Request, Response}
 import com.twitter.util.Future
 import io.buoyant.k8s.istio.ApiserverClient.RouteRuleConfig
 import io.buoyant.test.{Awaits, Exceptions}
-import istio.proxy.v1.config.{DestinationWeight, RouteRule}
+import _root_.istio.proxy.v1.config.{DestinationWeight, RouteRule}
```
adleong: what is the _root_ prefix?

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/istio/IstioIdentifier.scala
```diff
@@ -0,0 +1,66 @@
...
+  baseDtab: () => Dtab,
+  routeCache: RouteCache,
+  clusterCache: ClusterCache,
+  mixerClient: MixerClient
+)
+  extends Identifier[Request] {
+  val istioIdentifier = new InternalTrafficIdentifier[Request](pfx, baseDtab, routeCache, clusterCache, mixerClient, new H2IstioRequestHandler)
+
+  override def apply(originalRequest: Request): Future[RequestIdentification[Request]] = {
+    val req = H2IstioRequest(originalRequest)
```
adleong: At this point the DstBoundCtx context has not been populated so attempting to call any of the methods from IstioDataFlow on this IstioRequest will fail.

istio/src/main/scala/io/buoyant/k8s/istio/ClusterCache.scala
```diff
@@ -3,8 +3,13 @@ package io.buoyant.k8s.istio
 import com.twitter.logging.Logger
 import com.twitter.util.{Closable, Future, Time}
 
-class ClusterCache(client: DiscoveryClient) extends Closable {
-  import ClusterCache._
+case class Cluster(dest: String, port: String)
+
+trait ClusterCache extends Closable {
```
adleong: I assume this is factored this way for test-ability.  

Instantiating a ClusterCacheBackedByApi with a test client that returns mock data would test more of the codepath than implementing a test ClusterCache.  But it would also be a lot less ergonomic.  I just wanted to bring it up so we can consider the tradeoff.

hawkw: Yeah, I'm not really a big fan of the `trait X`/`class XBackedByApi` pattern that shows up a couple times in this PR. Elsewhere in Linkerd, when we need to mock an external API for testing, I think the general pattern is to inject a mock API with overrides, like [this](https://github.com/linkerd/linkerd/blob/80b320244571ca3f66ae4673de05a438c7e02def/interpreter/k8s/src/test/scala/io/buoyant/interpreter/k8s/ConfigMapInterpreterTest.scala#L163-L167). Like Alex said, this would exercise more of the implementation, and I think it also results in a less complex non-test codebase?

istio/src/main/scala/io/buoyant/k8s/istio/IstioAttribute.scala
```diff
@@ -0,0 +1,99 @@
...
+
+case class SourceServiceIstioAttribute(val value: String) extends IstioAttribute[String]("source.service", StringAttribute, "The fully qualified name of the service that the client belongs to.	redis-master.my-namespace.svc.cluster.local")
+
+case class SourceNameIstioAttribute(val value: String) extends IstioAttribute[String]("source.name", StringAttribute, "The short name part of the source service.	redis-master")
+
+case class SourceNamespaceIstioAttribute(val value: String) extends IstioAttribute[String]("source.namespace", StringAttribute, "The namespace part of the source service.	my-namespace")
+
+case class SourceDomainIstioAttribute(val value: String) extends IstioAttribute[String]("source.domain", StringAttribute, "The domain suffix part of the source service, excluding the name and the namespace.	svc.cluster.local")
+
+case class SourceUidIstioAttribute(val value: String) extends IstioAttribute[String]("source.uid", StringAttribute, "Platform-specific unique identifier for the client instance of the source service.	kubernetes://redis-master-2353460263-1ecey.my-namespace")
```
hawkw: this might be overly nitpicky, but can we insert some line breaks here? I think lines this long are hard to read especially when my editor wraps them weirdly.

istio/src/main/scala/io/buoyant/k8s/istio/IstioResponse.scala
```diff
@@ -0,0 +1,11 @@
+package io.buoyant.k8s.istio
+
+import com.twitter.util.Duration
+
+case class IstioResponse[Resp](statusCode: Int, duration: Duration, resp: Option[Resp]) extends IstioAttributesHolder {
+
+  def responseCode: ResponseCodeIstioAttribute = ResponseCodeIstioAttribute(statusCode)
+
+  def responseDuration: ResponseDurationIstioAttribute = ResponseDurationIstioAttribute(duration)
```
hawkw: once again, I'd really like to see some more line breaks in this file. I'd find something like
```scala
package io.buoyant.k8s.istio

import com.twitter.util.Duration

case class IstioResponse[Resp](
  statusCode: Int,
  duration: Duration,
  resp: Option[Resp]
) extends IstioAttributesHolder {

  def responseCode: ResponseCodeIstioAttribute = 
    ResponseCodeIstioAttribute(statusCode)

  def responseDuration: ResponseDurationIstioAttribute = 
    ResponseDurationIstioAttribute(duration)

}
```
much more readable. 

Note that this is entirely a matter of personal taste & I won't hold it against you :)

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
```diff
@@ -0,0 +1,121 @@
...
+  private def mkAttributesUpdate(customAttributeNames: Seq[String], attributes: Seq[IstioAttribute[_]]) = {
+
+    // TODO: eventually we should not have to send istioGlobalDict over the wire,
+    // and instead use positive index integers
+    val dictionaryToUse = (istioGlobalDict ++ customAttributeNames).toSet.toSeq
+
+    val allStringAttributes = attributes.filter(_.attributeType == StringAttribute).map(_.asInstanceOf[IstioAttribute[String]])
+    val allInt64Attributes = attributes.filter(_.attributeType == Int64Attribute).map(_.asInstanceOf[IstioAttribute[Long]])
+    val allStringMapAttributes = attributes.filter(_.attributeType == StringMapAttribute).map(_.asInstanceOf[IstioAttribute[Map[String, String]]])
+    val allDurationAttributes = attributes.filter(_.attributeType == DurationAttribute).map(_.asInstanceOf[IstioAttribute[Duration]])
```
hawkw: Is there any way to do this w/o casting?

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  def mkCheckRequest(istioREquest: IstioRequest[_]) = {
+    val TODO = Some(1L)
+    val dictionary = mkDictionary(istioREquest.requestedPath, istioREquest.targetService, istioREquest.sourceLabel, istioREquest.targetLabel)
+    val requestAttributes = Seq(istioREquest.requestedPath, istioREquest.sourceLabel, istioREquest.targetLabel, istioREquest.targetService)
+    val attributesUpdate = mkAttributesUpdate(dictionary, requestAttributes)
+    CheckRequest(TODO, Some(attributesUpdate))
+  }
+
+  private def mkDictionary(
```
hawkw: This whole function seems really convoluted, is there any way it can be simplified or broken apart?

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
```diff
@@ -0,0 +1,121 @@
...
+    duration: ResponseDurationIstioAttribute
+  ): ReportRequest = {
+
+    val dictionary = mkDictionary(requestPath, targetService, sourceLabel, targetLabel)
+
+    val updated = mkAttributesUpdate(dictionary, Seq(responseCode, requestPath, targetService, sourceLabel, targetLabel, duration))
+    ReportRequest(attributeUpdate = Some(updated))
+  }
+
+  def mkCheckRequest(istioREquest: IstioRequest[_]) = {
```
hawkw: is there a reason for the capitalization in `istioREquest`?

hawkw: Also, TIOLI, but it seems to me that a lot of this function might be better represented as a series of functions on `IstioRequest`?

istio/src/test/scala/io/buoyant/k8s/istio/ApiserverClientTest.scala
pcalcado: Intellij being silly with protobuf imports -_- fixed

istio/src/main/scala/io/buoyant/k8s/istio/ClusterCache.scala
pcalcado: Interesting, the idea behind the use of an interface was to increase ergonomics. My thinking here is that the usual pattern of implementing a fake `Service[Req,Resp]` on tests have a few drawbacks:
1) I think that it leaks a lot of "infrastructure" types into higher level tests. These tests shouldn't have any reason to even know about Finagle or any other piece of infrastructure
2) Maybe more important, I also think that it also creates coupling between the consumer of a type and its implementation. For example, consider https://github.com/linkerd/linkerd/pull/1650/files#diff-bcd0445a689106c21b39b7304f665801R36 . If we follow the usual pattern of implementing a fake `Service[Req,Resp]` there then the test for `InternalTrafficIdentifier` would make assumptions about the API used by `ClusterCacheBackedByApi`. If the next version of the API used by this class changes the URL or any other implementation detail we would have to change not only the code and tests for `ClusterCacheBackedByApi` but also the tests for `InternalTrafficIdentifier`. This feels like accidental coupling.
3) I am not sure if I agree that this would exercise the codebase more, at least in any useful manner. One rule of thumb that I like to follow when it comes to testing is _"don't mock what you don't own"_, and in this case I think that we might be trying to use unit tests for what should be integration tests. Instead, I would suggest that we keep as much as possible of our own logic independent from these dependencies (which drives back to the 1st point) and use integration tests for those instead of mocking APIs that aren't under our control.
4) The concept of ergonomics or complex code that we are discussing here are subjective, but I have the impression that one of the reasons for what I perceive to be low test coverage around these boundary classes is because it is actually a lot of work to have those fake `Service[Req,Resp]`, so much so that we tend to have a single "fixture" reused across multiple tests, which requires tests to use special requests to know which part of the fixture will be hit.

But with all that said, Linkerd is an existing code base and I am happy to comply with the existing testing strategy if we collectively think that's the way to go.

istio/src/main/scala/io/buoyant/k8s/istio/ClusterCache.scala
pcalcado: I tend to prefer more descriptive names, but I am happy with the suggestion. The only thing I'd suggest is that we keep the same naming pattern around things, do we usually go with `Qualifier` + `Thing`, as in `Api` `ClusterCache` or do we go the other way around? Looking around the code, it seems like so far we have relied on packages to namespace different implementations but keep the same name, should I do this here?

hawkw: third option: `ClusterCache with Api` (Alex will _hate_ this :D)

pcalcado: :P I actually find the cake pattern super cute, but had way too many problems debugging badly wired software 😤

istio/src/main/scala/io/buoyant/k8s/istio/IstioConfigurator.scala
pcalcado: Hm.. that's a good point. I think this had some template pattern logic at first but was gradually removed. Let me explore it, brb

pcalcado: Done!

istio/src/main/scala/io/buoyant/k8s/istio/IstioDataFlow.scala
pcalcado: You are absolutely right, and I had felt some of this already. Unfortunately this branch didn't pick the changes, I've added them: https://github.com/linkerd/linkerd/pull/1650/files#diff-042aed18959e3a7caabf44f4294034ef

I am still not happy with `CurrentIstioPath`, but I ran out of ideas here.

istio/src/main/scala/io/buoyant/k8s/istio/IstioLoggerBase.scala
pcalcado: Yup, fixed with the same merge as the one above

istio/src/main/scala/io/buoyant/k8s/istio/IstioResponse.scala
pcalcado: Fixed on the same as above

istio/src/main/scala/io/buoyant/k8s/istio/RouteCache.scala
pcalcado: Waiting for resolution on similar comment

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
pcalcado: Fixed. Interestingly enough, this seems to be a default on IntelliJ. Maybe we should upload a common IntelliJ formatting configuration file?

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/istio/IstioIdentifier.scala
pcalcado: Yup, this was fixed when we moved it to where `Logger`  was. Current `Identifier` has no `DstBoundCtx`: https://github.com/linkerd/linkerd/pull/1650/files#diff-64056934f2ceb2827c6a08c5026e0ddeR20

istio/src/main/scala/io/buoyant/k8s/istio/IstioAttribute.scala
pcalcado: Can you let me know what you think of: https://github.com/linkerd/linkerd/pull/1650/files#diff-0a76be1349b81581e237715901125ebcR35 ?

I don't have a particular preference—I usually go with the Editor's default.n Like I mentioned before, it would probably be a good idea if we all shared a single formatter configuration file.

istio/src/main/scala/io/buoyant/k8s/istio/IstioResponse.scala
pcalcado: This one is similar to the one above, but different in one important way. As I mentioned there, I don't really have a personal preference, all I would lobby for is for us to share a single formatter configuration—my general philosophy is that I usually don't care about these things as long as there's _one_ way to do it and we all follow, whatever this way might be.

All that said, my editor right now wraps at 120 columns and the Effective Scala guide mentions 100  columns as the standard: 
> Indent by two spaces. Try to avoid lines greater than 100 columns in length. Use one blank line between method, class, and object definitions.

What rule should we follow, wrap at 100, always wrap...?

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
pcalcado: Yup, the reason is that I am still not used to the new mac's keyboard and keep fat fingering everything -_-

Fixed it, thanks :)

On the second comment, I think that most of this stuff is only relevant in a `Mixer API` context, and I would not like to create any coupling between the request object and Mixer. What did you have in mind?

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
pcalcado: I think the current version might be a little bit better? https://github.com/linkerd/linkerd/pull/1650/files#diff-00e114bad2c9fdd832ebe5854d4a1a4fR50

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
pcalcado: I looked at this for a bit and couldn't think of anything, due to `attributes: Seq[IstioAttribute[_]]``. Did I miss something?

hawkw: Hmm. What about a function on `IstioAttribute` taking an attribute type and optionally returning that attribute if it is of the given type?

hawkw: Another potential concern here – which might be a premature optimization – is that the repeated calls to `filter` mean that we traverse the sequence of attributes 4 times. 

As an alternative, consider using mutable collections (gasp!) and traversing the sequence a single time and appending the attribute to a different collection for each attribute type? If this is in the hot path, it might be worth using mutability.

hawkw: Alternatively, if you really want to avoid mutable collections, could always do something like this
```scala
val (stringAttrOpts, int64AttrOpts, stringMapAttrOpts, durationAttrOpts) = 
  attributes.map { 
    case attr: IstioAttribute[String] => (Some(attr), None, None, None)
    case attr: IstioAttribute[Long] => (None, Some(attr), None, None)
    case attr: IstioAttribute[Map[String, String]] => (None, None, Some(attr), None)
    case attr: IstioAttribute[Duration] => (None, None, None, Some(attr))
  }.unzip
// the resultant collections are now `Seq[Option[IstioAttribute[T]]]`, which you'd then have to `.flatten`...
```
which is kind of ugly, but is purely functional!

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
```diff
@@ -0,0 +1,118 @@
...
+      responseCode,
+      requestPath,
+      targetService,
+      sourceLabel,
+      targetLabel,
+      duration
+    )
+    log.trace("MixerClient.report: %s", reportRequest)
+    client.report(Stream.value(reportRequest))
+    Future.Done
```
hawkw: Should this return a finished `Future`, or return a future that's satisfied when the report stream finishes?

istio/src/main/scala/io/buoyant/k8s/istio/IstioRequest.scala
```diff
@@ -0,0 +1,83 @@
...
+   in most cases this will be the same value, but eventually the correct solution
+   should be to get this directly from the label or the uid.
+   */
+  private def findTargetLabelApp: Option[String] = findTargetService.flatMap(_.split('.').headOption)
+}
+
+/**
+ * Resolves a valid Istio [[Path]], if any exists for the current [[DstBoundCtx]]
+ */
+object CurrentIstioPath {
```
adleong: I like the way this has been factored.  Feels much more explicit now :+1:

istio/src/main/scala/io/buoyant/k8s/istio/IstioRequestAuthorizerFilter.scala
```diff
@@ -0,0 +1,40 @@
+package io.buoyant.k8s.istio
+
+import com.twitter.finagle.{Filter, Service}
+import com.twitter.util.{Duration, Stopwatch, Try}
+import io.buoyant.k8s.istio.mixer.MixerClient
+
+trait IstioRequestAuthorizerFilter[Req, Resp] extends Filter[Req, Resp, Req, Resp] {
+
+  def mixerClient: MixerClient
```
adleong: TIOLI: this feels more like a dependency and less like something that needs to be implemented.  Could make this an abstract class instead of a trait and have MixerClient be a constructor argument.

istio/src/main/scala/io/buoyant/k8s/istio/IstioAttribute.scala
```diff
@@ -0,0 +1,153 @@
...
+  * > Istio uses attributes to control the runtime behavior of services running in the service mesh. Attributes are named and
+  * > typed pieces of metadata describing ingress and egress traffic and the environment this traffic occurs in. An Istio attribute
+  * > carries a specific piece of information such as the error code of an API request, the latency of an API request, or the original
+  * > IP address of a TCP connection
+  *
+  * @param name
+  * @param attributeType
+  * @param description
+  */
+sealed abstract class IstioAttribute[T](val name: String, val attributeType: MixerAttributeType, description: String) {
```
adleong: What about something like: 

```
sealed abstract class IstioAttribute[T](val name: String, val description: String, val value: T)

sealed abstract class StringAttribute(val name: String, val description: String, val value: String) extends IstioAttribute[String](name, description, value)
...

case class SourceIpIstioAttribute(val value: String) extends StringAttribute("source.ip", "blah blah", value)
...
```

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
adleong: With the change to IstioAttributes I suggest above, this could be something like:
```
val allStringAttributes = attributes.collect { case attr@StringAttribute(_) => attr }
```

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Checks this request agains  ixer' Pre-Condition system, using the [[https://istio.io/docs/reference/api/mixer/mixer-service.html#check Check API]]
```
adleong: typo: ixer'

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private def handleExceptionInPreconditionCheck(e: Throwable) = {
+    //TODO: Consider making this behaviour configurable, as per https://github.com/istio/mixerclient/blob/13460a96b4b3aa792ad47817d4fbf529b63c25e2/src/check_cache.cc#L163
+    log.error("Error checking Mixer pre-conditions: %s - %s", e.getClass, e.getMessage);
+    Future.value(MixerCheckStatus(codeUsedWhenUnknownStatus))
+  }
+
+  private[this] def mkStatus(stream: Stream.Releasable[CheckResponse]) = {
+    val result = stream.value.`result`.get
```
adleong: Instead of get and getOrElse, I think we can chain all these options together with flatMap and then convert to a Future with Future.const

adleong: we also have to call stream.release()

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  private def handleExceptionInPreconditionCheck(e: Throwable) = {
+    //TODO: Consider making this behaviour configurable, as per https://github.com/istio/mixerclient/blob/13460a96b4b3aa792ad47817d4fbf529b63c25e2/src/check_cache.cc#L163
+    log.error("Error checking Mixer pre-conditions: %s - %s", e.getClass, e.getMessage);
+    Future.value(MixerCheckStatus(codeUsedWhenUnknownStatus))
+  }
+
+  private[this] def mkStatus(stream: Stream.Releasable[CheckResponse]) = {
```
adleong: I think a `Stream.Releasable` is an item from the stream, not the stream itself.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/istio/HttpIstioRequest.scala
```diff
@@ -0,0 +1,11 @@
+package io.buoyant.linkerd.protocol.http.istio
+
+import com.twitter.finagle.http.Request
+import io.buoyant.k8s.istio.{CurrentIstioPath, IstioRequest}
+
+object HttpIstioRequest {
+  def apply(req: Request): IstioRequest[Request] =
```
adleong: I think we should pass `Option[IstioPath]` in here so that callers are explicit about when they are pulling the local context.

adleong: This means that code in the identifiers that construct an HttpIstioRequest can explicitly pass None for IstioPath

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
pcalcado: It's definitely fire-and-forget, and because nobody cares about when it's completed I decided to just return a constant. On the return type, I generally like methods from a single type to be symmetric, i.e. given a set `A` it always map from `A` to `S[A]`, but no strong preference.

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
pcalcado: Fixed, thanks!

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
pcalcado: Fixed, thanks!

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
pcalcado: On the chaining, I am not sure I understand: each option can be present or not, and if not we're using the default value. How would we deal with the default values on a `flatMap` chain?

istio/src/main/scala/io/buoyant/k8s/istio/IstioConfigurator.scala
```diff
@@ -0,0 +1,43 @@
+package io.buoyant.k8s.istio
+
+import com.twitter.finagle.buoyant.H2
+import com.twitter.finagle.{Address, Name}
+import com.twitter.logging.Logger
+import io.buoyant.config.types.Port
+import io.buoyant.k8s.istio.mixer.MixerClient
+
+object IstioServices {
```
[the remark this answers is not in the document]
pcalcado: Done!

istio/src/main/scala/io/buoyant/k8s/istio/IstioConfigurator.scala
```diff
@@ -0,0 +1,43 @@
...
+  protected val log = Logger.get(this.getClass.getSimpleName)
```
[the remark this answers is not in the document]
pcalcado: Aye

istio/src/main/scala/io/buoyant/k8s/istio/IstioRequest.scala
pcalcado: 😻

istio/src/main/scala/io/buoyant/k8s/istio/IstioRequestAuthorizerFilter.scala
pcalcado: Absolutely agree, this is a leftover from the previous shape of this type. Let me fix it.

istio/src/main/scala/io/buoyant/k8s/istio/IstioAttribute.scala
hawkw: I like that alignment style for parameters, but most of the rest of the Linkerd codebase puts parameter lists on the next line, one tab level in, like
```scala
class Foo(
  a: A,
  b: B
) extends Bar {
...
}
```
rather than 
```scala
class Foo(a: A,
          b: B) 
  extends Bar {
...
}
```

hawkw: I agree that we definitely ought to have a shared config file for formatting. Do note that, while we don't have one you can plug into IDEA at the moment, running `sbt test` will also run Scalariform on everything.

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
adleong: Hmmm... I don't understand your comment about symmetric return types.  What would be the difference between having the return type be `Unit` instead of `Future[Unit]`?

pcalcado: I mean that I would expect all functions in a given type/namespace to return values of the same container, or at least the same _kind_. In this particular case, it means that I would expect the published interface of `MixerClient` to be either this:

```scala
class MixerClient(client: Mixer) {
  def report(/* ... */): Future[Unit]
  def checkPreconditions(/* ... */): Future[MixerCheckStatus]
}
```

Or:
```scala
class MixerClient(client: Mixer) {
  def report(/* ... */): Unit
  def checkPreconditions(/* ... */): MixerCheckStatus
}
```

But **not**:
```scala
class MixerClient(client: Mixer) {
  def report(/* ... */): Unit
  def checkPreconditions(/* ... */): Future[MixerCheckStatus]
}
```

But, again, this is just a preference, I'm happy to make it `Unit` if it makes it more aligned with our internal philosophy.

istio/src/main/scala/io/buoyant/k8s/istio/IstioRequestAuthorizerFilter.scala
pcalcado: Done!

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
pcalcado: Oh, and on the `release()` comment, do we have to manually release each stream item in our gRPC implementation? [The current implementation seems to be a noop](https://github.com/linkerd/linkerd/blob/1a39277/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/Stream.scala#L18) and I couldn't find an example of something else consuming gRPC like this, so I'm not super sure what's the usual idiom.

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerApiRequests.scala
pcalcado: I've made the changes @adleong mentioned. @hawkw that's a valid concern and interesting solution, but I would argue that we should rather get through with istio 0.1.x features and address performance on our 0.2 work. Istio 0.1.x doesn't focus on performance as much and I find it unlikely that this would be an actual bottleneck as opposed to, for e.g. the lack of caching in this release.

istio/src/main/scala/io/buoyant/k8s/istio/IstioAttribute.scala
```diff
@@ -0,0 +1,247 @@
...
+  name: String,
+  description: String
+) extends Int64Attribute(name, description)
+
+sealed abstract class DurationAttribute(
+  name: String,
+  description: String
+) extends IstioAttribute[Duration](name, description)
+
+case class SourceIpIstioAttribute(val value: String)
```
adleong: no need for `val` on case class params

istio/src/main/scala/io/buoyant/k8s/istio/IstioRequestAuthorizerFilter.scala
```diff
@@ -0,0 +1,38 @@
+package io.buoyant.k8s.istio
+
+import com.twitter.finagle.{Filter, Service}
+import com.twitter.util.{Duration, Stopwatch, Try}
+import io.buoyant.k8s.istio.mixer.MixerClient
+
+abstract class IstioRequestAuthorizerFilter[Req, Resp](mixerClient: MixerClient) extends Filter[Req, Resp, Req, Resp] {
+
+  def report(request: IstioRequest[Req], response: IstioResponse[Resp], duration: Duration) = {
```
adleong: I like explicit return type on public methods

istio/src/main/scala/io/buoyant/k8s/istio/IstioServices.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.twitter.finagle.buoyant.H2
+import com.twitter.finagle.{Address, Name}
+import com.twitter.logging.Logger
+import io.buoyant.config.types.Port
+import io.buoyant.k8s.istio.mixer.MixerClient
+
+object IstioServices {
+  protected val log = Logger()
+
+  def mkMixerClient(mixerHost: Option[String], mixerPort: Option[Port]) = {
```
adleong: I like explicit return types on public methods

istio/src/main/scala/io/buoyant/k8s/istio/mixer/MixerClient.scala
adleong: Ah, I see.  I think `Unit` is more aligned with my personal philosophy anyway.  (since this method is fire-and-forget)

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/istio/IstioIdentifier.scala
```diff
@@ -0,0 +1,67 @@
...
+  baseDtab: () => Dtab,
+  routeCache: RouteCache,
+  clusterCache: ClusterCache,
+  mixerClient: MixerClient
+)
+  extends Identifier[Request] {
+  val istioIdentifier = new InternalTrafficIdentifier[Request](pfx, baseDtab, routeCache, clusterCache, mixerClient, new H2IstioRequestHandler)
+
+  override def apply(originalRequest: Request): Future[RequestIdentification[Request]] = {
+    val req = H2IstioRequest(originalRequest, None)
```
adleong: :heart_eyes: explicit None

istio/src/main/scala/io/buoyant/k8s/istio/IstioServices.scala
pcalcado: Me too, will fix it. I think Intellij has a setting for this somewhere?

pcalcado:
@adleong cool! In that case, I'd recommend we also review/merge and test #1606 , as the only difference between that PR and this one is this commit: https://github.com/linkerd/linkerd/pull/1606/commits/b9dd7dfd1f2cd4853c1ba53f39bb90437911ad4e

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/istio/IstioIdentifier.scala
hawkw: this phrase gives me unpleasant [FORTRAN flashbacks](http://www.personal.psu.edu/jhm/f90/statements/implicit.html)

PR #1651, opened 2017-09-20
Pass set differences to K8s EventLogger lazily

hawkw:
As @adleong mentioned in https://github.com/linkerd/linkerd/pull/1644#discussion_r139780705, this should reduce the performance hit from computing set differences for `trace` logging when the trace log level is disabled.


hawkw:
I'd like to refactor the logging code in K8s some more (it's a little inelegant) but this should improve performance in the short term.

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -7,7 +7,7 @@ private[k8s] trait EventLogging {
   @inline private[this] def logAction[A](verb: String, descriptor: => String, format: A => String)(value: A): Unit =
     log.ifTrace(s"k8s ns $ns service $srv $verb $descriptor ${format(value)}")
 
-  protected def logActions[A](verb: String, descriptor: => String, format: A => String)(values: Iterable[A]): Unit =
+  protected def logActions[A](verb: String, descriptor: => String, format: A => String)(values: => Iterable[A]): Unit =
     if (values.nonEmpty) {
```
adleong: values is immediately evaluated here regardless of log level so I don't think we benefit at all from passing by name.  Ideally we would avoid evaluating unless the log level was high enough.

adleong: To make matters worse, we actually evaluate `values` twice, incurring the cost of the set operations twice.

hawkw: Ah, you're right – I think I'll have to rewrite the whole event logger. Which is fine, this implementation is kind of ossified.

adleong: should we close this PR then?

hawkw: I'm working on _actually_ fixing this, but would be fine with closing this PR & opening a new one?

hawkw:
@adleong, WDYT now?

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+    }
 
-  protected def logModification[A](descriptor: String)(mods: Iterable[(A, A)]): Unit =
-    if (mods.nonEmpty) {
+  def deletion[A, B](deleted: Map[A, B]): Unit =
+    mkDeletion("port mapping", formatMapping)(deleted)
+
+  protected def mkNewState[A](
+    descriptor: String,
+    format: A => String = defaultFormat[A](_)
```
adleong: I think this is always `toString` (either explicitly with `_.toString` or by default with the default formatter.  Can we just remove it?

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+  )(
```
adleong: I don't think we ever partially apply this?  Combine into one argument list?

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+        deleted.foreach(logAction("deleted", descriptor, format))
+      }
+    }
 
-  protected def logModification[A](descriptor: String)(mods: Iterable[(A, A)]): Unit =
-    if (mods.nonEmpty) {
+  def deletion[A, B](deleted: Map[A, B]): Unit =
+    mkDeletion("port mapping", formatMapping)(deleted)
+
+  protected def mkNewState[A](
```
adleong: I don't really understand the naming convention here.  Why `mk`?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -239,13 +235,46 @@ object EndpointsNamer {
...
-  private case class Endpoint(ip: InetAddress, nodeName: Option[String])
+  private[EndpointsNamer] case class Endpoint(ip: InetAddress, nodeName: Option[String])
 
-  private object Endpoint {
+  private[EndpointsNamer] object Endpoint {
     def apply(addr: v1.EndpointAddress): Endpoint =
       Endpoint(InetAddress.getByName(addr.ip), addr.nodeName)
   }
 
+  private[EndpointsNamer] implicit class RichSubsetsSeq(
```
adleong: any change here or just a move?

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+    format: A => String = defaultFormat[A](_)
+  )(
+    deleted: Iterable[A]
+  ): Unit =
+    if (deleted.nonEmpty) {
+      log.debug(
+        "k8s ns %s service %s deleted %ss",
+        ns, srv, descriptor
+      )
+      if (log.isLoggable(Logger.TRACE)) {
```
adleong: in the case of a straight up deletion of the whole object, I don't think we need to list all elements of the object that was deleted.

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
-    logActions("deleted", "port mapping", formatMapping)(deleted)
+  @inline private[this] def defaultFormat[A](a: A): String = a.toString
 
-  @inline def addition[A, B](added: Map[A, B]): Unit =
-    logActions("added", "port mapping", formatMapping)(added)
+  protected def mkDeletion[A](
+    descriptor: String,
+    format: A => String = defaultFormat[A](_)
+  )(
+    deleted: Iterable[A]
```
adleong: we may not need this?

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+    mkDeletion("port mapping", formatMapping)(deleted)
+
+  protected def mkNewState[A](
+    descriptor: String,
+    format: A => String = defaultFormat[A](_)
+  )(
+    old: Set[A],
+    newState: Set[A]
+  ): Unit =
+    if (old.nonEmpty && old != newState) {
```
adleong: what about on object creation?  (ie where `old.nonEmpty`)  shouldn't we log then?

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+          logAction("deleted", descriptor, format)
+        }
+
+      }
     }
 
-  def modification[A, B](old: Map[A, B], replacement: Map[A, B]): Unit =
-    if (old.nonEmpty && replacement.nonEmpty) {
+  def newState[A, B](old: Map[A, B], newState: Map[A, B]): Unit =
+    if (old.nonEmpty && old != newState) {
```
adleong: ditto

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
       )
-      for {
-        key <- old.keySet.intersect(replacement.keySet)
-        oldValue <- old.get(key)
-        newValue <- replacement.get(key)
-        if oldValue != newValue
-      } log.ifTrace(s"k8s ns $ns service $srv remapped port '$key' from $oldValue to $newValue")
+      if (log.isLoggable(Logger.TRACE)) {
+
+        val (remapped: Set[Option[String]], removed: Set[Option[String]]) = (for {
```
adleong: unzip is cute but this is probably more readable by building up :scream: mutable sets.

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
hawkw: ah, you're right, whoops.

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
hawkw: Oh, okay, that'll make this a lot simpler, then.

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
hawkw: noooooo, the horror

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
hawkw: an (unnecessary) move for changes I didn't end up making. Didn't mean to commit this.

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,91 @@
...
       log.debug(
         "k8s ns %s service %s modified %ss",
-        ns, srv, descriptor
+        ns, srv, noun
       )
-      for {
-        (old, replacement) <- mods
-      } logAction[A](s"replaced $old with", descriptor, _.toString)(replacement)
+      if (log.isLoggable(Logger.TRACE)) {
+        val was: String => A => Unit = logAction(noun)
```
adleong: We're paying a price here for being more general than is necessary.  There are a very small number of nouns (service, endpoint, portMapping?).  Instead, you could have functions like `val logEndpointAdded: Endpoint => Unit` defined on the trait instead of being allocated on each `newState` call.  I think the goal should be to not allocate any functions in the logging path; not because allocating functions is necessarily that expensive, but because the result will be more straightforward code.

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,91 @@
...
-        newValue <- replacement.get(key)
-        if oldValue != newValue
-      } log.ifTrace(s"k8s ns $ns service $srv remapped port '$key' from $oldValue to $newValue")
+      if (log.isLoggable(Logger.TRACE)) {
+        val was: String => ((Any, Any)) => Unit =
+          logAction("port mapping", formatMapping)
+        val wasRemapped =
+          logAction("port", formatMapping)("remapped")(_)
+
+        val (remapped, removed) = (for {
```
adleong: I still think building these sets mutably will help readability and performance

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
hawkw: fair enough

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+        (oldState -- newState).foreach { wasDeleted }
+      }
     }
 
-  def modification[A, B](old: Map[A, B], replacement: Map[A, B]): Unit =
-    if (old.nonEmpty && replacement.nonEmpty) {
+  def deletion(noun: String = ""): Unit =
+    log.debug("k8s ns %s service %s deleted %s", ns, srv, noun)
+
+  private val mappingWas = logAction("port mapping", formatMapping)(_)
```
adleong: it looks like this class is supposed to be type agnostic but this assumes that all mappings are port mappings?

k8s/src/main/scala/io/buoyant/k8s/EventLogging.scala
```diff
@@ -1,55 +1,95 @@
...
+import com.twitter.logging.Logger
+
 private[k8s] trait EventLogging {
   def ns: String
   def srv: String
+  
+  @inline 
+  protected def logAction[A](
+    noun: String,
+    format: A => String = (a: A) => a.toString
```
adleong: is this ever anything other than toString?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -344,48 +372,13 @@ object EndpointsNamer {
...
-    def addition(endpoints: Iterable[Endpoint]): Unit =
-      logActions[Endpoint]("added", "endpoint", _.toString)(endpoints)
-
-    def deletion(endpoints: Iterable[Endpoint]): Unit =
-      logActions[Endpoint]("deleted", "endpoint", _.toString)(endpoints)
-
-    def modification(endpoints: Iterable[(Endpoint, Endpoint)]): Unit =
-      logModification("endpoint")(endpoints)
+    private val endpointWas = logAction[Endpoint]("endpoint", _.toString)(_)
+    private val endpointWasAdded = endpointWas("added")(_)
```
adleong: could combine the first two argument lists of logAction if you write this as

```
private val endpointWasAdded = logAction[Endpoint]("endpoint", "added")(_)
```

hawkw:
Okay @adleong, you were right, this could have been much simpler (though I was having fun over-engineering it...). It turns out the code for handling port remappings was the only thing that was really complex enough to be factored out – for everything else, it turns out that just logging directly from the `update` function took about as many lines of code as having `EventLogging` do it...

Let me know what you think of fa68e421c9326022e9fed839f52b411d675e282f? I also changed namer event handling slightly to try and make it more consistent between different namers.

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -140,33 +140,35 @@ abstract class EndpointsNamer(
...
+                log.debug(
+                  "k8s ns %s service %s deleted",
+                  ns, srv
+                )
                 Map.empty
               case (oldMap, v1.ServiceError(error)) =>
                 log.warning(
                   "k8s ns %s service %s watch error %s",
-                  nsName, serviceName, error
+                  ns, srv
```
adleong: we've got 3 format thingies and 2 arguments

k8s/src/main/scala/io/buoyant/k8s/PortMapLogging.scala
```diff
@@ -0,0 +1,35 @@
...
+            "k8s ns %s service %s remapped port %s from %s to %s",
+            nsName, serviceName, port, oldValue, newValue
+          )
+        case None =>
+          log.trace(
+            "k8s ns %s service %s removed port mapping from %s to %s",
+            nsName, serviceName, port, oldValue
+          )
+      }
+      (newPorts -- oldPorts.keys).foreach { case (to, from) =>
```
adleong: the naming of `to` and `from` seems backwards

k8s/src/main/scala/io/buoyant/k8s/PortMapLogging.scala
```diff
@@ -0,0 +1,35 @@
...
+private[k8s] trait PortMapLogging {
+  def nsName: String
+  def serviceName: String
+  def logPortDiff[A,B](oldPorts: Map[A, B], newPorts: Map[A, B]): Unit =
+    if (log.isLoggable(Logger.TRACE)) {
+      for {
+        port <- oldPorts.keySet
+        oldValue <- oldPorts.get(port)
+      } newPorts.get(port) match {
+        case Some(`oldValue`) =>
```
adleong: any reason for the backtick escaped name?

k8s/src/main/scala/io/buoyant/k8s/PortMapLogging.scala
```diff
@@ -0,0 +1,35 @@
+package io.buoyant.k8s
+
+import com.twitter.logging.Logger
+
+private[k8s] trait PortMapLogging {
```
adleong: Would prefer to make this a class with `nsName` and `serviceName` as constructor arguments instead of a trait

k8s/src/main/scala/io/buoyant/k8s/ServiceNamer.scala
```diff
@@ -23,17 +23,107 @@ class ServiceNamer(
...
    * @param portMappings a map of port numbers to port names.
    */
-  private[this] case class Svc(
+  case class Svc(
+    nsName: String,
+    serviceName: String,
     ports: Map[String, Address],
     portMappings: Map[Int, String]
-  ) {
+  ) extends PortMapLogging {
```
adleong: would prefer to have the PortMapLogging as a private member rather than a parent.

k8s/src/main/scala/io/buoyant/k8s/PortMapLogging.scala
hawkw: yeah, this is matching the case where the value contained in the `Some` is _equal to_ the value of `oldValue`, not shadowing the binding of `oldValue` to a new value.

hawkw: (how many times can I say "value" in one sentence?)

adleong: Oh!  I definitely missed that.  Would `case Some(value) if value == oldValue =>` be more clear perhaps?

k8s/src/main/scala/io/buoyant/k8s/PortMapLogger.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.twitter.logging.Logger
+
+private[k8s] case class PortMapLogger(nsName: String, serviceName: String) {
+  def logDiff[A,B](oldPorts: Map[A, B], newPorts: Map[A, B]): Unit =
+    if (log.isLoggable(Logger.TRACE)) {
+      for {
+        port <- oldPorts.keySet
+        oldValue <- oldPorts.get(port)
+      } newPorts.get(port) match {
+        case Some(`oldValue`) =>
```
adleong: this is pretty confusing.  would prefer `case Some(value) if value == oldValue =>`

k8s/src/main/scala/io/buoyant/k8s/PortMapLogger.scala
```diff
@@ -0,0 +1,33 @@
...
+        case None =>
+          log.trace(
+            "k8s ns %s service %s removed port mapping from %s to %s",
+            nsName, serviceName, port, oldValue
+          )
+      }
+      (newPorts -- oldPorts.keys).foreach { case (from, to) =>
+        log.trace(
+          "k8s ns %s service %s mapped port %s to %s",
+          nsName, serviceName, to, from
```
adleong: `from, to`

hawkw: argh. i thought i fixed that.

k8s/src/main/scala/io/buoyant/k8s/PortMapLogger.scala
hawkw: yeah, was going to mention on your earlier comment that if this wasn't immediately clear, it probably ought to be changed.

Issue #1655, opened 2017-09-25
prometheus: Add the ability to configure a prefix


amitsaha:
Issue Type:

- [ ] Bug report
- [x] Feature request

Currently, the prometheus metrics are of the form `namer:xx` and `rt:xx` and so on. Based of this [discourse discussion](https://discourse.linkerd.io/t/prometheus-metrics-namespacing/297/5), it would be useful to have a configurable prometheus metrics prefix. The default will still be no prefix.

wmorgan:
@amitsaha thanks for filling this!

amitsaha:
@wmorgan happy to have been able to get my colleague to work on a PR as well :)

hawkw:
Closed by #1657

PR #1657, opened 2017-09-28
Add Prefix Option to Prometheus Telemetry Config

joshleeb:
prometheus: Add the ability to configure a prefix #1655.

The new config key is optional and can be used like this:

```yaml
telemetry:
- kind: io.l5d.prometheus
  path: /admin/metrics/prometheus
  prefix: linkerd_
```

By default, the `prefix` value is an empty string and so no prefix will be attached to the exported Prometheus metrics. 

Also worth noting, currently the prefix will be inserted as is, with no delimeter. So if the prefix value is `linkerd_`, as above, then the exported metrics will look like: `linkerd_foo:bar...`.

### DIFF
--- a/linkerd/docs/telemetry.md
+++ b/linkerd/docs/telemetry.md
@@ -22,6 +22,7 @@ experimental | `false` | Set this to `true` to enable the telemeter if it is exp
 telemetry:
 - kind: io.l5d.prometheus
   path: /admin/metrics/prometheus
+  prefix: linkerd_
 ```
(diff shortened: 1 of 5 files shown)


wmorgan:
@joshleeb thanks for submitting this! We will take a look!

wmorgan:
@joshleeb thanks again! email me your address and preferred size and I will send a t-shirt your way! My first name at buoyant.io

PR #1664, opened 2017-10-06
Add New Relic telemeter

hawkw:
This PR adds an experimental telemeter for reporting metrics to New Relic. 

A New Relic plugin which ingests these metrics and displays them on the New Relic dashboard is in the works.

### DIFF
--- a/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
+++ b/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
@@ -6,6 +6,7 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicReference
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
+import scala.collection.mutable
 
 trait MetricsTree {
   def children: Map[String, MetricsTree]
@@ -119,5 +120,20 @@ object MetricsTree {
       trees.clear()
       metricRef.set(Metric.None)
     }
+  }
+
+  def flatten(
+    tree: MetricsTree,
+    prefix: String = "",
+    acc: mutable.Buffer[(String, Metric)] = mutable.Buffer()
+  ): Seq[(String, Metric)] = {
+    acc += (prefix -> tree.metric)
+    for ((name, child) <- tree.children) {
+      if (prefix.isEmpty)
+        flatten(child, name, acc)
+      else
+        flatten(child, s"$prefix/$name", acc)
+    }
+    acc
   }
 }
--- a/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
+++ b/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
@@ -152,7 +152,7 @@ object MetricsTree {
       case metric => acc += prefix -> tree.metric
     }
 
-    for ((name, child) <- tree.children ) {
+    for ((name, child) <- tree.children) {
       if (prefix.isEmpty)
         flattenNones(child, name, acc)
       else
--- a/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
+++ b/telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
@@ -122,25 +122,10 @@ object MetricsTree {
     }
   }
 
-  def flatten(
-    tree: MetricsTree,
-    prefix: String = "",
-    acc: mutable.Buffer[(String, Metric)] = mutable.Buffer()
-  ): Seq[(String, Metric)] = {
-    acc += (prefix -> tree.metric)
-    for ((name, child) <- tree.children) {
-      if (prefix.isEmpty)
-        flatten(child, name, acc)
-      else
-        flatten(child, s"$prefix/$name", acc)
-    }
-    acc
-  }
-
   /**
    * Flatten `tree`, skipping empty scopes.
    */
-  def flattenNones(
+  def flatten(
     tree: MetricsTree,
     prefix: String = "",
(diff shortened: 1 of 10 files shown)


telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
```diff
@@ -120,4 +121,43 @@ object MetricsTree {
...
+      else
+        flatten(child, s"$prefix/$name", acc)
+    }
+    acc
+  }
+
+  /**
+   * Flatten `tree`, skipping empty scopes.
+   */
+  def flattenNones(
```
adleong: I think it's safe to just skip Metric.None in flatten.  Can't think of a reason why we'd want both behaviors.

telemetry/newrelic/src/main/scala/io/buoyant/telemetry/newrelic/NewRelicTelemeter.scala
```diff
@@ -0,0 +1,107 @@
...
+  // only run at most once
+  def run(): Closable with Awaitable[Unit] =
+    if (started.compareAndSet(false, true)) run0()
+    else Telemeter.nopRun
+
+  private[this] def run0() = {
+    val task = timer.schedule(NewRelicTelemeter.Interval) {
+      val _ = sendMetrics().respond {
+        case Return(r) if r.statusCode == 200 =>
+          log.trace("successfully published metrics to New Relic")
```
adleong: consider a counter instead of a trace

telemetry/newrelic/src/main/scala/io/buoyant/telemetry/newrelic/NewRelicTelemeter.scala
```diff
@@ -0,0 +1,107 @@
...
+        case Return(r) =>
+          log.warning("New Relic API returned error %s, %s", r.status, r.contentString)
```
adleong: consider counters for the error cases too (probably in addition to logging)

telemetry/newrelic/src/main/scala/io/buoyant/telemetry/newrelic/NewRelicTelemeter.scala
```diff
@@ -0,0 +1,107 @@
...
+import com.twitter.finagle.stats.NullStatsReceiver
+import com.twitter.finagle.tracing.NullTracer
+import com.twitter.util._
+import io.buoyant.config.Parser
+import io.buoyant.telemetry.Metric.{Counter, Gauge, Stat, None => MetricNone}
+import io.buoyant.telemetry.{MetricsTree, Telemeter}
+import java.util.concurrent.atomic.AtomicBoolean
+import com.twitter.logging.Logger
+
+class NewRelicTelemeter(
```
pcalcado: How are we testing this class?

telemetry/newrelic/src/main/scala/io/buoyant/telemetry/newrelic/NewRelicTelemeter.scala
hawkw: good call!

telemetry/core/src/main/scala/io/buoyant/telemetry/MetricsTree.scala
hawkw: Yeah, I just wasn't sure, since you'd originally written it that way. Happy to make this the default behaviour

telemetry/newrelic/src/main/scala/io/buoyant/telemetry/newrelic/NewRelicTelemeter.scala
hawkw: Added tests for initializers in 7fdfa1d, will add tests for telemeter reporting as well

jippi:
Hi @hawkw 

I'm probably the reason why this telmeter was implemented, as I requested NewRelic integration from @wmorgan 

Looking at the code, it looks like something different from what I asked was possible.

My request was to support [request queue](https://docs.newrelic.com/docs/apm/applications-menu/features/request-queue-server-configuration-examples) in linkerd, which probably is a way simpler thing to implement - or maybe an additional thing on top of this.

What we would like to see Linkerd do, is simply adding a `X-Request-Start` header to outgoing requests, with the value (in msec) of when linkerd received the request. There is examples on this on their [configuration page](https://docs.newrelic.com/docs/apm/applications-menu/features/request-queue-server-configuration-examples).

This setup do not require any kind of newrelic license key or anything else, as the NewRelic SDK's automatically process the header if present. 

The benefit of the `X-Request-Start` header is that the latency it measures, will show up per-app next to $LANGUAGE, Redis, Database etc in the performance charts, rather than being a distinct metric like this PR have.

hawkw:
Hi @jippi, thanks for letting me know! I think must have been some miscommunication about what kind of New Relic integration you were interested in. I think supporting the New Relic request queue shouldn't take too long, so expect to see that soon.

I do think there's still a potential use-case for a New Relic plugin like in this PR as well, though that might be better off focusing on metrics related to Linkerd performance rather than on the application itself?

jippi:
@hawkw really sorry I haven't been clear in what i wanted to see. What's done here is excellent too, and something that can be super useful as well. My request was a simple header though :) I hope to see that being added as a config (that doesn't require NR license key)

hawkw:
@jippi no worries, and yeah, we can definitely do that.

hawkw:
Hi again @jippi, just wanted to follow up and let you know that we've just merged #1672, which adds an optional `timestampHeader` configuration field to HTTP servers, which will allow you to add the `X-Request-Start` header. Sorry it took a little while – I actually spent most of last week moving – but it should be in the next Linkerd release!

jippi:
amazing @hawkw ! thank you so much!

PR #1665, opened 2017-10-09
Correct ip/port confusion in namers documentation

christophetd:
It seems that there was a typo in the docs of the rewriting namers `hostportPfx` and `porthostPfx`

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -800,7 +800,7 @@ would be rewritten to `/pfx/foo/resource/name`
 /$/io.buoyant.hostportPfx/<prefix>/<host>:<port>/etc
 ```
 
-Rewrites a name of the form "host:ip" as a path with host followed by ip. Does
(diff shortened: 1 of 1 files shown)


linkerd/docs/namer.md
```diff
@@ -821,7 +821,7 @@ would be rewritten to `/pfx/host/port/etc`.
 /$/io.buoyant.porthostPfx/<prefix>/<host>:<port>/etc
 ```
 
-Rewrites a name of the form "host:ip" as a path with ip followed by host. Does
+Rewrites a name of the form "host:port" as a path with ip followed by the port. Does
```
esbie: I think we want "as a path with port followed by host"

christophetd: Yes, definitely. I'll update that tomorrow.

christophetd:
Should be good now

christophetd:
Done

klingerf:
Great, thank you!

PR #1674, opened 2017-10-19
Call non-watch endpoint first when restarting watches

### LINKED ISSUE #1668: Linkerd fails to route to kubernetes service seemingly after persistent k8s watch connection dies
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:

After upgrading to 1.3.0, occasionally linkerd will fail to route to a k8s service after a rolling update. The k8s deployments all specify the following:

```
  replicas: <n replicas>
  strategy:
    rollingUpdate:
      maxUnavailable: 0
      maxSurge: 2
```

to guarantee that there are always `n` healthy pods running during the rolling update. When the issue occurs, `kubectl -n <namespace> get pods` shows that the pods are ready. Looking at the logs, it seems to be marking healthy services as being dead. It seems to only happen after the persistent watch connection w/`kubectl proxy` dies.

The issue resolves itself after a while (usually minutes?), but causes downtime until it does.

kubectl and`TRACE`-level linkerd logs: https://gist.github.com/agunnerson-ibm/6384e35b9e1239ae1524e5894b1c8e47

Specifically, the issue was seen at 20:26:23 UTC for `redacted-service-1` in `redacted-namespace-1`.

linkerd should not fail to route to a healthy service.

It seems to happen when several people are deploying and hitting services, but I've been unable to reproduce it at will.

It's possible this might have also affected the 1.2.x releases. If it did, it would've been masked by other issues we were encountering.

**Environment**:
- linkerd 1.3.0 in k8s daemonset configuration.
- dtab:
  ```
  /k8s => /#/io.l5d.k8s;
  /ns  => /k8s/master;
  /srv => /ns/http;
  /svc => /srv;
  ```
  Requests to services in other namespaces are done by passing the `l5d-dtab: /ns => /k8s/some-namespace` header.
- kubernetes 1.6.7 in HA setup with 3 masters. Haproxy sits in front of the masters and will terminate connections after half an hour.

### LINKED ISSUE #1669: Namerd not updated automatically
Hello,

We are using namerd for dynamic dtab, but when we want to update a route, change are not take into account without restart of namerd pod.

Kubernetes version:

```
NAME                                           STATUS    AGE       VERSION   EXTERNAL-IP   OS-IMAGE                      KERNEL-VERSION
ip-xxx.eu-central-1.compute.internal   Ready     7d        v1.7.6    <none>        Debian GNU/Linux 8 (jessie)   4.4.78-k8s
ip-xxx.eu-central-1.compute.internal   Ready     15d       v1.7.6    <none>        Debian GNU/Linux 8 (jessie)   4.4.78-k8s
```

Deployed over AWS.

Linkerd Configuration:

```yaml
    # Custom HTTP Ingress Controller listening on port 80
    # Accessible Externaly
    - protocol: http
      label: http-ingress
      servers:
        - port: 80
          ip: 0.0.0.0
          clearContext: true
      interpreter:
        kind: io.l5d.namerd
        dst: /#/io.l5d.k8s/linkerd/4100/namerd
        namespace: external
      identifier:
        kind: io.l5d.path
        segments: 1
        consume: true
```

Namerd Configuration:

```yaml
    admin:
      ip: 0.0.0.0
      port: 9991

    namers:
    - kind: io.l5d.k8s
      experimental: true
      host: localhost
      port: 8001

    storage:
      kind: io.l5d.k8s
      host: localhost
      port: 8001
      namespace: linkerd

    interfaces:
      ip: 0.0.0.0
      port: 4100
    - kind: io.l5d.httpController
      ip: 0.0.0.0
      port: 4180
```

Namerd Service:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: namerd
  namespace: linkerd
spec:
  selector:
    app: namerd
  type: LoadBalancer
  ports:
  - name: thrift
    port: 4100
  - name: http
    port: 4180
  - name: admin
    port: 9991
```

Linkerd Metrics:

[metrics.txt](https://github.com/linkerd/linkerd/files/1376844/metrics.txt)

klingerf:
## Problem

When restarting Kubernetes watches after the previous watch has died, it's possible that the newly established watch creation is delayed, causing Linkerd's representation of the watch data to become stale. This is caused by our watch code seeding the new watch by calling the Kubernetes watch endpoint for the resource, which might not return immediately in some Kubernetes environments.

## Solution

When restarting Kubernetes watches, seed the watch by first calling the non-watch endoint, which returns immediately.

Fixes #1668. Fixes #1669.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/resources.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/resources.scala
@@ -150,11 +150,9 @@ private[k8s] class ListResource[O <: KubeObject: TypeReference, W <: Watch[O]: T
     fieldSelector: Option[String]
   ): Future[(Seq[W], Option[String])] =
     get(
-      labelSelector,
-      fieldSelector,
-      None,
-      retryIndefinitely = true,
-      watch = true
+      labelSelector = labelSelector,
+      fieldSelector = fieldSelector,
+      retryIndefinitely = true
     ).map { maybeList =>
       val list = maybeList.get // list resources should always exist
       (list.items.map(od.toWatch), list.metadata.flatMap(_.resourceVersion))
@@ -226,19 +224,14 @@ private[k8s] class NsObjectResource[O <: KubeObject: TypeReference, W <: Watch[O
     fieldSelector: Option[String] = None
   ): Future[(Seq[W], Option[String])] =
     get(
-      labelSelector,
-      fieldSelector,
-      None,
-      retryIndefinitely = true,
-      watch = true
+      labelSelector = labelSelector,
+      fieldSelector = fieldSelector,
+      retryIndefinitely = true
     ).map { maybeObj =>
       val watch = maybeObj.toSeq.flatMap { obj => Seq(od.toWatch(obj)) }
-      val version =
-        for {
-          obj <- maybeObj
-          meta <- obj.metadata
-          version <- meta.resourceVersion
-        } yield version
-      (watch, version)
+      // object watches don't supply a resource version, since the version associated
+      // with the object is tied to when the object was last modified, and therefore
+      // it may be too old to successfully establish a watch
+      (watch, None)
     }
 }
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -45,12 +45,11 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
     labelSelector: Option[String] = None,
     fieldSelector: Option[String] = None,
     resourceVersion: Option[String] = None,
-    retryIndefinitely: Boolean = false,
-    watch: Boolean = false
+    retryIndefinitely: Boolean = false
   ): Future[Option[G]] = {
     val req = Api.mkreq(
       http.Method.Get,
-      if (watch) watchPath else path,
+      path,
       None,
       LabelSelectorKey -> labelSelector,
       FieldSelectorKey -> fieldSelector,
@@ -168,8 +167,7 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
                   case None =>
                     // In this case, we want to try loading the initial information instead before watching again.
                     restartWatches(labelSelector, fieldSelector).map {
-                      case (ws, ver) =>
-                        AsyncStream.fromSeq(ws) ++ _watch(ver)
+                      case (ws, ver) => AsyncStream.fromSeq(ws) ++ _watch(ver)
                     }
                 }
               }.flatten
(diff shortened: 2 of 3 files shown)


k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -226,19 +224,14 @@ private[k8s] class NsObjectResource[O <: KubeObject: TypeReference, W <: Watch[O
...
     ).map { maybeObj =>
       val watch = maybeObj.toSeq.flatMap { obj => Seq(od.toWatch(obj)) }
-      val version =
-        for {
-          obj <- maybeObj
-          meta <- obj.metadata
-          version <- meta.resourceVersion
-        } yield version
-      (watch, version)
+      // object watches don't supply a resource version, since the version associated
```
adleong: I don't really understand.

klingerf: I can't seem to find this documented anywhere, but as far as I can tell, Kubernetes list resources (e.g. `/api/v1/namespaces/ns/services`) return the global resource version as part of the metadata response. This means that you can always use the version returned to establish a new watch.

On the other hand, Kubernetes object resources (e.g. `/api/v1/namespaces/ns/services/mysvc`) do not return the global resource version. They instead return the resource version from when the object was last modified, and that version might be too old to establish a new watch.

For instance, testing in one of our clusters, I make the following request for an endpoints object:

```
# curl localhost:8001/api/v1/namespaces/test/endpoints/namerd
{
  "kind": "Endpoints",
  "apiVersion": "v1",
  "metadata": {
    "name": "namerd",
    "namespace": "test",
    "selfLink": "/api/v1/namespaces/test/endpoints/namerd",
    "uid": "28d76695-af7d-11e7-a88a-42010af00004",
    "resourceVersion": "65078108",
    "creationTimestamp": "2017-10-12T18:42:54Z"
  },
```

That "resourceVersion" reflects when the endpoints were last modified, rather than the global resource version. If I then immediately make a watch request with the returned resource version, I get:

```
# curl localhost:8001/api/v1/watch/namespaces/test/endpoints/namerd?resourceVersion=65078108
{"type":"ERROR","object":{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"too old resource version: 65078108 (65950030)","reason":"Gone","code":410}}
```

So when watching objects, we need to omit the version, rather than risk having the watch fail because the resource version is too old.

Would love to get a bit more clarity on this though from folks with more familiarity of Kubernetes internals.

klingerf: Have confirmed that this is the same behavior in Kubernetes 1.7 as well.

thedebugger: @klingerf if the resource version is actually "65078108" and we are seeing above error, that to me sounds like a k8s bug. Based on my experience working with zk/consul, the client should always send a last known version because the client could miss the events when the client is not watching. So if the client has an old version, k8s should return what has changed and latest resource version. I hope we tested that linkerd doesn't miss a watch change b/w the retries. But again, i'm speaking based on my past experience and may not be true for k8s.

PR #1681, opened 2017-10-21
Stop k8s watches from time travelling

### LINKED ISSUE #1680: Kubernetes API clients should filter watch events with a resource version that’s less than the object’s resource version
As discussed previously w @adleong and @klingerf, in order to prevent temporary jumps "back in time" to older states, we should ensure that, before updating an activity from a new watch event, that the received event's resource version is later than the current object's resource version.

hawkw:
I've modified our code for watching Kubernetes resources so that it will ignore events whose resource versions are older than the resource version of the previous event. We do this by creating a two-item sliding window over the stream of watch events, and filtering out those tuples where the first item (the previous event) is older than the second item (the received event). This should ensure that we never travel back in time when the Kubernetes API is out of sync.

I've added a test for `EndpointsNamer` and `ServiceNamer` that assert that the namer's state does not update when receiving events in reverse order. Furthermore, a number of existing tests had to be modified to ensure all resource versions were sequential – we probably had been lazy in the past about copying and pasting JSON strings in tests without updating their resource versions.

Closes #1680


k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -182,6 +184,24 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
             AsyncStream.fromFuture(f)
         }
       }
+      // Ignore any cases where we receive a resource version lower
+      // than the last received resource version, by creating a sliding
+      // two-item window over the stream of watch events.
+      stream.paired
```
adleong: I don't think this works.  Consider:

stream = [3, 1, 2]
paired = [(x,3), (3,1), (1,2)]
filtered = [(x,3), (1,2)]
mapped = [3,2]

I think you could probably accomplish this with a scanLeft...

```
stream.scanLeft[(W, Boolean)]((Zero, false)) { case ((largest, incr), w) =>
  if (w > largest) (w, true)
  else (largest, false)
}.filter(_._2).map(_._1)
```

or something like that

hawkw: Yeah, you're right; 0208ec44f22688d16fd14786916f25736b78daf4 should be more correct.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -182,6 +184,15 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
             AsyncStream.fromFuture(f)
         }
       }
+      // Ignore any cases where we receive a resource version lower
```
adleong: defining this as a method `def increasingSubsequence[T <: Ordered[T]](stream: AsyncStream[T]): AsyncStream[T]` would make the logic of this method easy to test (since you can just test it on an AsyncStream of numbers).

k8s/src/main/scala/io/buoyant/k8s/Api.scala
```diff
@@ -161,8 +161,26 @@ case class ObjectReference(
  * create actual subclasses (i.e. FooWatch extends Watch[Foo]) rather than using Watch[Foo]
  * directly, and set the correct Jackson annotations on those, to ensure correct parsing.
  */
-trait Watch[O <: KubeObject] {
+trait Watch[O <: KubeObject] extends Ordered[Watch[O]] {
   def resourceVersion: Option[String]
+
+  private[Watch] def versionNum: Option[Long] = for {
```
adleong: what is the visibility scope here?  Is this just equivalent to `private`?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: Smart!

k8s/src/main/scala/io/buoyant/k8s/Api.scala
hawkw: yeah, i just thought that – since Scala's visibility modifiers can be confusing – it was clearer to make the scope explicit?

PR #1682, opened 2017-10-23
Denoise Consul namer logging and possibly fix a resource leak

### LINKED ISSUE #1670: Undesired Warning in namerd output when consul observation is released
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
After upgrade to namerd 1.3.0 there are warnings in namerd output
```
W 1011 15:35:08.872 UTC THREAD41 TraceId:5fbf736f9b6f2da2: consul datacenter 'bethesda-dev' service 'namerd-grpc' observation error Failure(service observation released, flags=0x02) with RemoteInfo -> Upstream Address: Not Available, Upstream id: Not Available, Downstream Address: /127.0.0.1:8500, Downstream label: client, Trace Id: 4b87215a3717cc59.4b87215a3717cc59<:4b87215a3717cc59, falling back to last good state
```

Resolving any service via namerd with consul namer
```
$ curl 'http://namerd-instance:4180/api/1/resolve/default?path=/consul/someservice
```
```
admin:
  ip: 127.0.0.1
  port: 9001
namers:
- kind: io.l5d.consul
  experimental: true
  host: 127.0.0.1
  port: 8500
  setHost: true
  useHealthCheck: true
  consistencyMode: stale
  failFast: false
  prefix: /consul
interfaces:
- kind: io.l5d.httpController
  ip: 127.0.0.1
  port: 4180
```

**Environment**:
namerd v1.3.0
consul v0.9.3

hawkw:
As described in issue #1670, the Consul namer currently logs a scary-looking error message every time we close a watch on a Consul service. This is because the namer's activity handles the `ServiceRelease` exception that's thrown to close a watch that's no longer needed as though it were an error. 

In addition to logging a spurious error message, if the namer has previously seen a good state from Consul, it handles this "error" by falling back to the last observed good state and restarting the watch, which is, of course, not what we want to do when we've intentionally closed a watch activity that's no longer needed. It's possible that these zombie activities may also cause a resource leak, although I'm not 100% certain this is the case – something else may be closing these activities regardless?

I've modified `SvcAddr` to handle the `ServiceRelease` exception separately from other `Throw`s, which should stop the spurious error message from being logged and closed watches from being restarted. 

Fixes #1670

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -67,7 +67,20 @@ private[consul] object SvcAddr {
             )
             // Drop the index, in case it's been reset by a consul restart
             loop(None)
+          case Throw(ServiceRelease) =>
+            // this exception is raised when we close a watch - thus, it needs
+            // to be special-cased so that we don't continue observing that
+            // service.
+            log.trace(
+              "consul datacenter '%s' service '%s' observation closed",
+              datacenter,
+              key.name
+            )
+            stopped = true
+            Future.Unit
           case Throw(e) =>
+            // an error occurred. if we've previously seen a good state, fall
+            // back to that state; otherwise, fail.
             stats.errors.incr()
             lastGood match {
               case Some(addr) =>
--- a/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
+++ b/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
@@ -3,11 +3,12 @@ package io.buoyant.namer.consul
 import com.twitter.finagle._
 import com.twitter.finagle.stats.InMemoryStatsReceiver
 import com.twitter.io.Buf
-import com.twitter.util.{Activity, Future, Promise}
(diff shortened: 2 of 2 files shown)


hawkw:
Further investigation will be necessary in order to determine whether this also resolves any of the resource leaks that have been plaguing the Consul namer for a while.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -67,7 +67,20 @@ private[consul] object SvcAddr {
...
+          case Throw(ServiceRelease) =>
+            // this exception is raised when we close a watch - thus, it needs
+            // to be special-cased so that we don't continue observing that
+            // service.
+            log.trace(
+              "consul datacenter '%s' service '%s' observation closed",
+              datacenter,
+              key.name
+            )
+            stopped = true
```
adleong: stopped should already be true at this point.

hawkw:
@adleong i'm working on that test as we speak.

hawkw:
Okay, update on this: it looks like there's no resource leak here, since I've added a test in ae0662ed3e4973850ac6ef2a797a472b6cef25d8 to assert that Consul isn't polled again after releasing a service.

PR #1683, opened 2017-10-24
Add test for closing activity twice

hawkw:
As @adleong requested in https://github.com/linkerd/linkerd/pull/1682#pullrequestreview-71344796, the `eliza/consul-leak-test` branch has a test that we don't observe Consul twice after closing an activity. It fails against master. Merging it in to my branch to see if my changes fix it


hawkw:
Okay, I don't know if this test is written correctly – @adleong, I might need your advice on this tomorrow.

PR #1688, opened 2017-10-30
add ability to update newly added CRDs in Kubernetes 1.7 and 1.8

### LINKED ISSUE #1661: namerd needs an alternative to the deprecated k8s ThirdPartyResource
Kubernetes 1.7 has deprecated ThirdPartyResources in favor of CustomResourceDefinitions, and TPRs are no longer supported in Kubernetes 1.8. This will break namerd on k8s 1.8.

deebo91:
Kubernetes deprecated support for ThirdPartyResources in version 1.8 in favor of CustomResourceDefinitions. Migration of the ThirdPartyResources is relatively simple and only requires that you create a CustomResourceDefinition that has the same properties as the older ThirdPartyResource. The process is documented [here](https://kubernetes.io/docs/tasks/access-kubernetes-api/migrate-third-party-resource/).  The namerd k8s storage plugin needs to be compatible with the new CustomResourceDefinition and this pull request addresses this need and hopefully should fix #1661


namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStoreInitializer.scala
```diff
@@ -26,3 +26,5 @@ class K8sDtabStoreInitializer extends DtabStoreInitializer {
   override def configClass = classOf[K8sConfig]
   override def configId = "io.l5d.k8s"
 }
+
+object K8sDtabStoreInitializer extends K8sDtabStoreInitializer
```
hawkw: What exactly is the purpose of the empty companion object here?

k8s/src/test/scala/io/buoyant/k8s/CustomResourceTest.scala
```diff
@@ -14,8 +14,8 @@ import org.scalatest.{FunSuite, OptionValues}
  * This test both exercises the versioning functionality in `io.buoyant.k8s`, and provides
  * an example implementation. (See `namerd` for the current usage in the Buoyant suite of projects).
  */
-class ThirdPartyResourceTest extends FunSuite with Awaits with OptionValues {
-  import ThirdPartyResourceTest._
+class CustomResourceTest extends FunSuite with Awaits with OptionValues {
+  import CustomResourceTest._
```
hawkw: I don't think it's correct to replace the existing `ThirdPartyResourceTest` with a new `CustomResourceTest`, since we will need to continue supporting third-party resources as well in order to remain backwards-compatible. I think that rather than replacing the test, we ought to have separate tests for TPR and CRD.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sV2DtabStoreInitializer.scala
```diff
@@ -0,0 +1,28 @@
+package io.buoyant.namerd.storage
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import io.buoyant.config.types.Port
+import io.buoyant.k8s.ClientConfig
+import io.buoyant.namerd.{DtabStore, DtabStoreConfig, DtabStoreInitializer}
+
+case class K8sConfigV2(
```
hawkw: I'm not a fan of this naming. I don't think `V2` is very descriptive – I think the class name should make it clear that the `DtabStore` created by this config is the CRD-based dtab store. Consider `K8sCrdConfig`?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sV2DtabStoreInitializer.scala
```diff
@@ -0,0 +1,28 @@
...
+  @JsonIgnore
+  override def mkDtabStore: DtabStore = {
+    val client = mkClient()
+    new CrdK8sDTabStore(client, dst, namespace.getOrElse(DefaultNamespace))
+  }
+}
+
+class K8sV2DtabStoreInitializer extends DtabStoreInitializer {
+  override def configClass = classOf[K8sConfigV2]
+  override def configId = "io.l5d.v2.k8s"
```
hawkw: Similarly to my above comment, I think this should be `io.l5d.k8s.crd`. We want the config id to make it clear that this is the CRD dtab store. Also, I think "k8s" should come _before_ the part of the config id that designates what _type_ of dtab store this is, since it's less specific.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/kubernetes/resources.scala
```diff
@@ -1,9 +1,9 @@
 package io.buoyant.namerd.storage.kubernetes
 
 import com.fasterxml.jackson.core.`type`.TypeReference
-import io.buoyant.k8s.{Client, NsThirdPartyVersion, ThirdPartyVersion}
+import io.buoyant.k8s.{Client, NsCustomResourceVersion, CustomResourceVersion}
```
hawkw: Once again, since not all users will be immediately upgrading to a version of Kubernetes that supports CRDs, we need to support both TPRs and CRDs. Thus we can't just replace the third party resource code, we need _both_.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/CrdK8sDTabStore.scala
```diff
@@ -0,0 +1,24 @@
...
+          r <- m.resourceVersion
+        } yield r
+
+        resourceVersion match {
+          case Some(r) => update(ns, dtab, Buf.Utf8(r))
+          case _ => Future.exception(new Exception())
+        }
+      case _ => Future.Unit
+    }
+    Future.Unit
```
hawkw: I think this `Future.Unit` should be removed – shouldn't the value returned from this function just be the result of the `flatMap` expression above?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/CrdK8sDTabStore.scala
```diff
@@ -0,0 +1,24 @@
...
+        val resourceVersion = for {
+          m <- v.metadata
+          r <- m.resourceVersion
+        } yield r
+
+        resourceVersion match {
+          case Some(r) => update(ns, dtab, Buf.Utf8(r))
+          case _ => Future.exception(new Exception())
+        }
+      case _ => Future.Unit
```
hawkw: Shouldn't the case where the dtab doesn't already exist be handled by creating it, rather than doing nothing?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sV2DtabStoreInitializer.scala
deebo91: Thats a good point. I was thinking that we calling it `K8sCrdConfig` would give details about the underlying implementation of the k8s storage plugin but now that I think of it. This is an open source project. :facepalm:

hawkw: I think whether CRDs or TPRs are used reflects both the implementation of the object and its _purpose_.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/kubernetes/resources.scala
esbie: Given the only differences between the TPRs and CRDs in `k8s/src/main/scala/io/buoyant/k8s/resources.scala` is the name of the classes/objects, I'd argue that the upgraded version is backwards compatible. (lmk if I'm missing something)

hawkw: @esbie yeah, @deebo91 explained to me that the APIs are equivalent, so I retract that comment. I do think it ought to be documented in that these are equivalent, though.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStoreInitializer.scala
adleong: so you don't have to type `new` to use it :)

deebo91:
@adleong We would need to do an integration test to make sure things don't break with namerd especially in Kubernetes 1.7 and 1.8. I definitely like the idea since its a "small" code change in the existing functionality. Smaller than this PR

namerd/storage/k8s/src/main/resources/META-INF/services/io.buoyant.namerd.DtabStoreInitializer
```diff
@@ -1 +1 @@
-io.buoyant.namerd.storage.K8sDtabStoreInitializer
```
adleong: looks like this removed the newline from the end of the file.  please revert this.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
```diff
@@ -112,16 +112,25 @@ class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
...
+        val resourceVersion = for {
+          meta <- value.metadata
+          version <- meta.resourceVersion
+        } yield version
+        nsDTab.put(namedDtab(ns, dtab, resourceVersion)).unit.rescue {
+          // In the event of a 404 on a PUT, we should POST instead, as the k8s API doesn't appear to
+          // support creating on PUT. Note that this is potentially racy if the namespace is created
+          // between the PUT and the POST, but it's likely safer to surface that error all the way to
+          // the user in that case than to attempt to recover beyond this point.
+          case io.buoyant.k8s.Api.NotFound(_) => create(ns, dtab)
```
adleong: Since we're now specifying a resource version, I think we now need to handle:

```
case io.buoyant.k8s.Api.Conflict(_) => Future.exception(new DtabVersionMismatchException())
```

We may also want to update the documentation or add a comment indicating that this method can return a DtabVersionMismatchException if the dtab was changed during the update and that the put should be simply retried.

namerd/storage/k8s/src/test/scala/io/buoyant/namerd/storage/K8sConfigTest.scala
```diff
@@ -0,0 +1,39 @@
...
+    """.stripMargin
+
+    val mapper = Parser.objectMapper(yaml, Iterable(Seq(K8sDtabStoreInitializer)))
+    val k8s = mapper.readValue[DtabStoreConfig](yaml).asInstanceOf[K8sConfig]
+    assert(k8s.host.value == "localhost")
+    assert(k8s.port.value == Port(8001))
+    assert(k8s.namespace.value == "default")
+  }
+
+  test("parse v2 config") {
```
adleong: rm

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
deebo91: Good point. I will be sure to add that.

deebo91: @adleong you mentioned documentation, Is that in the Readme?

namerd/storage/k8s/src/test/scala/io/buoyant/namerd/storage/K8sConfigTest.scala
deebo91: In terms of manual testing, I was using namerctl and minikube using both TPRs and CRDs. I mainly tested whether an update and create was successful. I am going to test delete to see if it works as well but I am confident that it will.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
adleong: https://github.com/linkerd/linkerd/blob/master/namerd/docs/storage.md

deebo91:
I have tested this manually on a minikube on my local machine and all CRUD operations for the DTabs seem to work in using namerctl. Next step would be to run integration tests.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStoreInitializer.scala
olix0r: rather, so you have a static instance on which you can reference its members without allocating a new initializer instance

shinofara:
+1

namerd/docs/storage.md
```diff
@@ -40,6 +40,13 @@ The Kubernetes storage plugin does not support TLS.  Instead, you should run `ku
...
 </aside>
 
+
+**Deprecation Notice**:
+As of Kubernetes 1.8+, The ThirdPartyResource API has been [deprecated](https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-third-party-resource/) in favor of CustomResourceDefinitions
+API. Kubernetes 1.7 supports both API to help migrate ThirdPartyResource objects to CustomResourceDefinition objects. More details on how to migrate to the new API in Kubernetes 1.8+ can be found
+[here](https://kubernetes.io/docs/tasks/access-kubernetes-api/migrate-third-party-resource/). 
+
+
 **How to check ThirdPartyResource is enabled**
```
siggy: this section below should be updated/replaced

deebo91: I added the Deprecation Notice as a way to notify people of the change so that they could migrate their DTabs if they were using K8s 1.7. I can remove this if it is no longer necessary.

siggy: ah, the Deprecation Notice is good. i was referring to the section titled `How to check ThirdPartyResource is enabled`. if we're still supporting `ThirdPartyResource`'s for k8s <= 1.7, i suppose this gets left as is?

deebo91: We may have to make the distinction on that line, something like "for Kubernetes <= v1.7". Hopefully it shouldn't be too confusing.

namerd/docs/storage.md
```diff
@@ -83,6 +107,25 @@ versions:
   - name: v1alpha1 # Do not change this value as it hardcoded in Namerd and doesn't work with other value.
 ```
 
+**Example of configuration for ThirdPartyResource in Kubernetes (for Kubernetes >= 1.8)**
```
siggy: `CustomResourceDefinition`

namerd/examples/k8s/3rdparty.yaml
```diff
@@ -1,7 +1,12 @@
...
-  - name: v1alpha1
+  name: dtabs.l5d.io
+spec:
+  scope: Namespaced
+  group: l5d.io
+  version: v1alpha1
+  names:
+    kind: DTab
+    plural: dtabs
+    singular: dtab
```
siggy: set your editor to add newlines to the end of files.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -91,39 +91,43 @@ private[k8s] class NsVersion[O <: KubeObject](
...
- * design model.
- *
- * Implementors are likely to want to extend this to provide convenience methods on top of the `list` method.
- *
- * @tparam O The parent type for all [[KubeObject]]s served by this Version.
- */
-trait ThirdPartyVersion[O <: KubeObject] extends Version[O] {
+  * A custom resource version contains CustomResource-typed objects. The API was called ThirdPartyResources in Kubernetes
+  * < 1.7 and has been deprecated since version 1.8. The APIs for both resource versions are identical in implementation. However,
+  * various naming standards have been changed to make the apis more meaningful. For me details
```
siggy: For *more

namerd/docs/storage.md
```diff
@@ -76,6 +81,48 @@ versions:
...
 ```
 
+### How to check if CustomResourceDefinition is enabled (for Kubernetes v1.8+)
+
+1. Open `apiextensions.k8s.io/v1beta1` api - `https://<k8s-cluster-host>/apis/apiextensions.k8s.io/v1beta1`.
+2. Check that kind `CustomResourceDefinition` exists in the response.
+
+> Example CustomResourceDefinition response
+
+```yaml
```
siggy: the block below is json ;)

deebo91:
Should I also add other changes that are going to be in this release as well? @siggy

PR #1692, opened 2017-11-06
Add test for TLS certificate reloading

hawkw:
I've added a test class to `linkerd/protocol/http/integration` which exercises the reloading of TLS certificates. The test prepares 2 different cert-key pairs: one with `commonName = foo `and one with `commonName = bar`. It configures an incoming Linkerd router with server TLS and points it at the "foo" cert and key, and an outgoing Linkerd router with client TLS and set it to expect `commonName = bar`.

The test sends a request through and asserts that it should fail validation because the cert has commonName foo instead of bar. Then, it replaces the bar cert and key with the foo cert and key, and sends another request, asserting that this request succeeds. If the incoming Linkerd reloads the cert and key, the second request should succeed.

### DIFF
--- a/linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsCertReloadingTest.scala
+++ b/linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsCertReloadingTest.scala
@@ -1,17 +1,15 @@
 package io.buoyant.linkerd.protocol
 
-import java.net.InetSocketAddress
-
 import com.twitter.finagle.http.{Request, Status}
 import io.buoyant.linkerd.Linker
-import io.buoyant.linkerd.protocol.TlsUtils.{Certs, Downstream}
 import io.buoyant.test.{Awaits, FunSuite}
+import java.net.InetSocketAddress
 
 class TlsCertReloadingTest extends FunSuite with Awaits {
 
   test("cert reloading") {
     TlsUtils.withCerts("foo", "bar") { certs =>
-      val downstream = Downstream.const("foo", "foo")
+      val downstream = TlsUtils.Downstream.const("foo", "foo")
 
       val fooCerts = certs.serviceCerts("foo")
       val barCerts = certs.serviceCerts("bar")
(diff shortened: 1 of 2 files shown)


linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsCertReloadingTest.scala
```diff
@@ -0,0 +1,94 @@
+package io.buoyant.linkerd.protocol
+
+import java.net.InetSocketAddress
```
adleong: group and alphasort imports please

linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsUtils.scala
```diff
@@ -1,17 +1,18 @@
...
-import com.twitter.util.{Future, Var}
-import java.io.{FileInputStream, File}
-import java.net.{SocketAddress, InetSocketAddress}
+import com.twitter.util.{Future, Try, Var}
+import java.io.{File, FileInputStream}
+import java.net.{InetSocketAddress, SocketAddress}
 import java.security.KeyStore
 import java.security.cert.CertificateFactory
 import javax.net.ssl.{SSLContext, TrustManagerFactory}
+
```
adleong: rm

linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsUtils.scala
```diff
@@ -23,7 +24,15 @@ object TlsUtils {
 
   def run(p: ProcessBuilder): Int = p ! DevNull
 
-  case class ServiceCert(cert: File, key: File)
+  case class ServiceCert(cert: File, key: File) {
+    def delete(): Try[Unit] =
+      Try(assert(cert.delete()))
```
adleong: what about 

``` if (cert.delete() && key.delete()) Return(()) else Throw(new SomethingSomething)```

File: linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsUtils.scala [hunk 4]
```diff
@@ -23,7 +24,15 @@ object TlsUtils {
...
+        .flatMap(_ => Try(assert(key.delete())))
+
+    def renameTo(name: String): Try[Unit] =
+      Try(assert(cert.renameTo(new File(cert.getParentFile, s"${name}_cert.pem"))))
```
adleong: ditto-ish

PR #1698, opened 2017-11-14
Properly handle cancellation in ThirftNamerInterface

adleong:
# Problem

When a client establishes a pending request on Namerd's ThirftNamerInterface and then disconnects, the request stays pending rather than being properly cancelled.  If this occurs repeatedly, it can lead to a memory leak in Namerd as the number of pending requests grows without bound.

# Solution

Add an interrupt handler to the Promise returned by Observer so that the Future fails when it is cancelled.

# Testing done

Manually tested that the pending requests stat properly resets when the client disconnects.


adleong:
All relevant changes are in https://github.com/linkerd/linkerd/pull/1698/files#diff-82c97ffce6d2e44e82d279b37b62865c

The rest is just scalariform

PR #1703, opened 2017-11-16
Move Namerd retries to the application layer

adleong:
Namerd retries are difficult to reason about because they happen inside the stack of the namerd thriftmux client.

Since we have a long-polling loop anyway, move all retry logic to that long-polling loop.  We do this by turning off retries in the thriftmux client and modifying the long-polling loop to continue looping in the event of an error.


adleong:
Poked this a bunch manually and everything looked fine.

PR #1705, opened 2017-11-17
1.3.2

deebo91:
## 1.3.2 2017-11-17
* Kubernetes
  * Deprecate ThirdPartyResources in favor of CustomResourceDefinitions for namerd dtab storage (#1688)
* Thrift
    * Fix a bug with the Thrift Identifier not properly identifying TTwitterThrift requests
* Consul 
    * Add tag weight configuration to Consul node addresses
* Namerd
    * Fix a memory leak involving the `io.l5d.thriftNameInterpreter` interface
    * Fix a bug where a Namerd client could sometimes becomes unresponsive
* gRPC
    * Fix a bug where large message bodies in unary gRPC requests could cause Linkerd to hang
* TLS
    * Add `forwardClientCert` to HTTP and HTTP/2 server configurations to forward client TLS certificates to services behind Linkerd

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,17 @@
 ## 1.3.2 2017-11-17
 * Kubernetes
-  * Deprecated ThirdPartyResources in favor of CustomResourceDefinitions for namerd dtab storage (#1688)
+  * Deprecate ThirdPartyResources in favor of CustomResourceDefinitions for namerd dtab storage (#1688)
+* Thrift
+    * Fix a bug with the Thrift Identifier not properly identifying TTwitterThrift requests
+* Consul 
+    * Add tag weight configuration to Consul node addresses
+* Namerd
+    * Fix a memory leak involving the `io.l5d.thriftNameInterpreter` interface
+    * Fix a bug where a Namerd client could sometimes becomes unresponsive
+* gRPC
+    * Fix a bug where large message bodies in unary gRPC requests could cause Linkerd to hang
+* TLS
+    * Add `forwardClientCert` to HTTP and HTTP/2 server configurations to forward client TLS certificates to services behind Linkerd
 
 ## 1.3.1 2017-10-24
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,6 +1,17 @@
 ## 1.3.2 2017-11-17
 * Kubernetes
-  * Deprecated ThirdPartyResources in favor of CustomResourceDefinitions for namerd dtab storage (#1688)
+  * Deprecate ThirdPartyResources in favor of CustomResourceDefinitions for namerd dtab storage (#1688)
+* Thrift
+    * Fix a bug with the Thrift Identifier not properly identifying TTwitterThrift requests
+* Consul 
+    * Add tag weight configuration to Consul node addresses
```
adleong: Perhaps: "Added the ability to weight addresses based on Consul tags."

CHANGES.md
```diff
@@ -1,6 +1,17 @@
...
+* Namerd
+    * Fix a memory leak involving the `io.l5d.thriftNameInterpreter` interface
+    * Fix a bug where a Namerd client could sometimes becomes unresponsive
+* gRPC
+    * Fix a bug where large message bodies in unary gRPC requests could cause Linkerd to hang
+* TLS
+    * Add `forwardClientCert` to HTTP and HTTP/2 server configurations to forward client TLS certificates to services behind Linkerd
```
adleong: I've just discovered some issues with the `forwardedClientCert`.  We should fix them, but I don't think we need to block the release.  This property is missing documentation, has an invalid test (that isn't run in CI), and is extraneously settable on H2 servers.

As for this line item:
* The property applies to client configurations, not server.
* "... configurations which causes Linkerd to forward client TLS certificates in the `x-forwarded-client-cert` header.

CHANGES.md
```diff
@@ -1,6 +1,17 @@
 ## 1.3.2 2017-11-17
```
adleong: 11-17?

PR #1706, opened 2017-11-17
Remove broken links

deebo91:
In `CurrentIstioPath.scala` there are broken links that cause issues during the release process. This PR removes these issues.


istio/src/main/scala/io/buoyant/k8s/istio/CurrentIstioPath.scala
```diff
@@ -20,7 +20,7 @@ object CurrentIstioPath {
   }
 
   /**
-   * Returns the target service requested based on the supplied [[Path]].
```
hawkw: Maybe add backtick (`) characters around the former links so they'll be typeset as code?

hawkw:
@adleong beat me to it :P

PR #1709, opened 2017-11-21
Reset client streams on errors when the server stream has already closed

### LINKED ISSUE #1696: Linkerd fails to reset its client streams in Half Closed (remote) state when the upstream client disconnects
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

A gRPC client initiates a bidirectional streaming GET request to Linkerd, which initiates the same request to a server. The server responds with a headers frame with the EOS bit set (in this case, a gRPC error), placing the stream in the "Half Closed (remote)" state, while Linkerd is still sending the request body. Then, the upstream client disconnects. This results in a synthetic remote reset. The reset cancels all of Linkerd's server streams to the client, but fails to cancel the client streams to the downstream server which are in the Half Closed (remote) state (client streams not in this state appear to be properly reset). 

This results in a memory leak, as these streams continue to be tracked after the connection is terminated. This can be observed using linkerd's `rt/client/.../open_streams` metric, which should remain the same after the connection is closed.

The synthetic remote reset on the server dispatcher should be propagated to the client dispatcher. All client streams should be reset.

**Environment**:
- linkerd/namerd version, config files: linkerd 1.3.2
- Platform, version, and config files (Kubernetes, DC/OS, etc): running locally with strest-grpc
- Cloud provider or hardware configuration: running locally (MacBook)

hawkw:
Currently, an error (#1696) exists where stream transports on the H2 `ClientDispatcher` can be leaked when a global reset occurs when the request stream has not been fully sent, but the response stream has finished. The sequence of events is as follows:

1. Request headers are sent from the client to Linkerd
2. Request headers are sent from Linkerd to the server.
3. Response headers with the EOS flag set are sent from the server to Linkerd.
4. Response headers are sent from Linkerd to the client, and the stream is closed by the EOS. 
5. A request data Aframe with an EOS is sent from the client to Linkerd. At this point, the `StreamTransport` in Linkerd's `ServerDispatcher` streams hash map is removed.
Meanwhile, the request body is still being sent from Linkerd to the server.
6. The client disconnects abruptly, causing Linkerd to reset all streams.
7. When the connection terminates, the stream in Linkerd's `ClientDispatcher` is left in the `StreamOpen(RemoteClosed)` state, rather than the `StreamRemoteReset` state, and is never removed.

Typically, we propagate this error by failing the client transport's send queue, but when the client has no more frames to send, it never offers frames into the excepting queue, and thus never sees the error. 

This can happen when the response body is much shorter than the request body (or the response consists only of a headers frame), and the response is started before the request finishes. Although this is a fairly uncommon case, it does seem to actually happen, such as when proxying bidirectional gRPC streaming requests which result in gRPC errors (represented as a HEADERS-only response).

I've added a `Future` to H2 `Request`s that is failed when the server dispatcher resets, and modified the way streams transition to the closed state in `Netty4StreamTransport`. Rather than just immediately removing closed streams from the `streams` hash map, we replace these streams with a `StreamClosed` object that holds a reference to a future that can be used to propagate errors to the client dispatcher. This allows streams on the server dispatcher which have been closed but still have streams open on the client dispatcher to close their corresponding client streams when there's a global reset.

Closes #1696.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
@@ -1,5 +1,6 @@
 package com.twitter.finagle.buoyant.h2
 
+import com.twitter.logging.Logger
 import com.twitter.util.{Future, Promise}
 
 /**
@@ -22,6 +23,7 @@ sealed trait Message {
 
   /** Create a deep copy of the Message with a new copy of headers (but the same stream). */
   def dup(): Message
+
 }
 
 trait Headers {
@@ -92,6 +94,8 @@ trait Request extends Message {
   def authority: String
   def path: String
   override def dup(): Request
+
+  def onFail: Future[Unit]
 }
 
 object Request {
@@ -108,13 +112,14 @@ object Request {
       Headers.Authority -> authority,
       Headers.Path -> path
     ),
-    stream
+    stream,
+    Future.never
   )
 
-  def apply(headers: Headers, stream: Stream): Request =
-    Impl(headers, stream)
+  def apply(headers: Headers, stream: Stream, fail: Future[Unit] = Future.never): Request =
+    Impl(headers, stream, fail)
 
-  private case class Impl(headers: Headers, stream: Stream) extends Request {
+  private case class Impl(headers: Headers, stream: Stream, onFail: Future[Unit]) extends Request {
     override def toString = s"Request($scheme, $method, $authority, $path, $stream)"
     override def dup() = copy(headers = headers.dup())
     override def scheme = headers.get(Headers.Scheme).getOrElse("")
@@ -124,6 +129,7 @@ object Request {
     }
     override def authority = headers.get(Headers.Authority).getOrElse("")
     override def path = headers.get(Headers.Path).getOrElse("/")
+    private[this] val log = Logger.get("h2")
   }
 }
 
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
@@ -97,7 +97,6 @@ trait Request extends Message {
   def headers: Headers
   override def dup(): Request
   def onFail: Future[Unit]
-  onFail.onFailure { e => log.debug("Request trait (not Impl) observed onFail: %s; onFail=%s; ", e, onFail.hashCode()) }
 
   def copy(
     headers: Headers = this.headers.dup(),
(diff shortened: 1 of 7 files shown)


hawkw:
TODO:
 - [ ] probably need to also remove `StreamClosed` streams from the `streams` hashmap when the corresponding client stream finishes and/or not put them in the map unless we know the client stream is still open
 - [ ] reap reset/errored streams after a TTL

adleong:
Because of the complexity of this issue, I think the description could benefit from being a lot more precise.  What are the exact sequence of events that leads to the issue?  I think it is:

1. Request headers are sent from the client to linkerd and then sent from linker to the server.
2. Response headers with EOS are sent from the server to linkerd and then sent from linkerd to the client.
3. The client disconnects.

At this point, what are the symptoms?  Does linkerd fail to send a Reset to the server (correctness issue)?  Are items permanently left in one of the dispatcher stream maps (memory leak)?  Which one?  What state are they left in?

Based on reading the code, here is my interpretation of what I would expect to happen in this scenario:

1. Before the client disconnects, the Linkerd server dispatcher should have a StreamTransport in the Half-Closed (local) ie `LocalClosed(RemoteStreaming)` state.  It should be in this state because it has fully sent the response but is still receiving the request.  
2. Meanwhile, the Linkerd client dispatcher should have a StreamTransport in Half-Closed (remote) because it has not fully sent the request but has finished receiving the response.
3. When the client disconnects, the `ClientDispatcher.demux` method should see either a GoAway frame or a ChannelClosedException depending on how gracefully the client disconnected.  Either one should trigger a call to `resetStreams` which should iterate through the streams and call `remoteReset` on each one.
4. When the Server dispatcher's StreamTransport gets `remoteReset`, it causes the LocalClosed(RemoteStreaming) to fail it's AsyncQueue with discard=true.
5. Meanwhile, the client dispatcher's StreamTransport is attempting to write the request stream by calling `stream.read` in a loop in the `writeStream` method.  When the AsyncQueue is failed, this `stream.read` should result in a `Reset.Cancel` exception which is then wrapped in a `StreamError.Local`
6. This exception should cause the `writeF` future to become an exception and hit the `Throw(StreamError.Local)` case of respond in the `send` method.  This results in a call to `localReset`.
7. This causes `resetP` to become a `StreamError.Local(Reset.Cancel)`.  In the client dispatcher, an onReset handler was added to this stream transport when it was added to the stream map.  This onReset handler causes the StreamTransport to be replaced by a StreamLocalReset in the streams map and causes a Reset to be sent to the server.

Does this description match what is actually happening?

One thing that stands out to me is that in step 7 when the stream is reset from local, the client dispatcher could theoretically look at the stream transport's state and see that it is already in RemoteClosed.  Therefore, a local reset is enough to fully close the stream and remove it from the streams map entirely rather than replacing it with StreamLocalReset.

hawkw:
@adleong thanks for the review, I've edited this PR to describe the issue more accurately. 

Your description of what should be happening in this scenario is correct. However, when we reach step 5, the observed behaviour no longer lines up with your description:
> 5. Meanwhile, the client dispatcher's StreamTransport is attempting to write the request stream by calling `stream.read` in a loop in the `writeStream` method.  When the AsyncQueue is failed, this `stream.read` should result in a `Reset.Cancel` exception which is then wrapped in a `StreamError.Local`

For streams which are leaked due to this bug, the closure `flatMap`ped over `stream.read()` in `writeStream` is never called again after the client disconnects. When I was investigating this issue, I added logging to `AsyncQueue`, and I believe that the future returned by `read()` never finishes, because the stream's `AsyncQueue` has 0 offers.

hawkw:
Closing this for now as we are going to try and find a better solution.

PR #1710, opened 2017-11-21
Replace total setInterruptHandlers with Promise.InterruptHandler

hawkw:
I've replaced calls to `Promise.setInterruptHandler` where the partial function passed as the interrupt handler is total with `new Promise with Promise.InterruptHandler`. According to the [twitter.util docs](https://twitter.github.io/util/docs/com/twitter/util/Promise$$InterruptHandler.html), this is equivalent to `Promise.setInterruptHandler` but reduces the number of allocations, since it doesn't allocate an additional partial function.

This might have a meaningful impact on performance if any of these calls to `setInterruptHandler` happen in a hot path.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
@@ -105,11 +105,11 @@ object Stream {
   }
 
   private[this] def failOnInterrupt[T, Q](f: Future[T], q: AsyncQueue[Q]): Future[T] = {
-    val p = new Promise[T]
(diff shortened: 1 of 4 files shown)


hawkw:
@adleong agreed, any suggestions?

hawkw:
Some improvements in allocations vs 1.3.2 in some test runs:

# test 7
test7 mem usage

# test 2
screen shot 2017-11-24 at 1 54 46 pm

(my branch is on the left and 1.3.2 is on the right)

hawkw:
Some of the other test runs appear to have broken after I redeployed 1.3.2, or to have not been reporting metrics for my branch, so I'm still working on getting a trustworthy validation of this.

Issue #1504, opened 2017-07-14
End to end tests intermittently fail due to timeouts


hawkw:
I'm seeing a weird issue where the test `UsageDataTelemeterEndToEndTest ` seems to fail about 1/2 to 3/4ths of the time with the message:

```scala
operation timed out after 2.seconds (UsageDataTelemeterEndToEndTest.scala:23)
```

This is happening both on my dev machine and on CI – I've had to re-run several [spuriously-failed](https://circleci.com/gh/linkerd/linkerd/5301) [CI builds](https://circleci.com/gh/linkerd/linkerd/5302). Issue appears to persist both on my dev branch and on `master`.

hawkw:
I've been rerunning this test against `master` over and over on my laptop and I don't think it's passed a single time, so far?

esbie:
I ran `./sbt e2e:test` on master on my laptop a number of times and `UsageDataTelemeterEndToEndTest` passed every time (although other* tests intermittently failed)

* MuxEndToEndTest, HttpEndToEndTest, TlsEndToEndTest, RouterEndToEndTest

hawkw:
@esbie issue renamed :)

hawkw:
This seems like it's probably the same issue as/linked to #1225?

PR #1714, opened 2017-11-27
Fix AsyncStream leak in kubernetes watch (#1694)

### LINKED ISSUE #1694: Kubernetes ingress memory leak
Issue Type:  Bug report

**What happened**:
Slow memory leak while using linkerd as a kubernetes ingress.

No memory leak :)

Appears to consistently reproduce in our configuration:
- Linkerd 1.3.1 docker image (original & modified with updated java8 jdk)
- Configuration as in https://gist.github.com/sgrankin/360471146e7c273fecce48e859d498e5 and with several 'default' and non-'default' namespace ingress  configurations inside kubernetes. 
- Not sure if there's a particular trigger since the start of the leak appears slow.
- Typical heap size (note the Y axis is logarithmic): 
screen shot 2017-11-07 at 17 03 05

- I have heap dumps of a few hours after process start and then 15 hours later that I can provide (directly—I don't think there's anything confidential in there but would prefer to avoid posting publicly).
- Via the heap dumps, I believe a (the?) potential leak may be  here: https://github.com/linkerd/linkerd/blob/35db8060ae4d3398381b5bed285f1c42b3956cb6/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala#L294 I think the closure is holding a reference to `this`, which is holding a reference to `as: AsyncStream[T]`, keeping the full stream in memory.  (Not sure why the growth is non-linear though...)

**Environment**:
1.3.1, config attached above.
v1.7.6+coreos.0
AWS

sgrankin:
k8s watch returned an AsyncStream, the head of which was leaking (as
determined via heap dumps), leading to significant leaks over 24-48hrs.

Since the AsyncStream was not strictly necessary, rewrite `watch` as
an update function suitable for Var.async (as used by `activity`).
- The change attempts to preserve previous bug fixes and comments
  documenting them.
- The AsyncStream enrichments and their tests have been removed as they
  are no longer used.
- A `watch` helper was added to ApiTest to hide the more complicated
  usage pattern around .watch.
- Fix orphaned infinite-retry watches in ingress tests by adding
  no-data /watch/ handlers.

Fixes #1694


siggy:
Ran this through our integration test suite. Results look good, memory is stable:
screen shot 2017-11-27 at 12 11 49 pm

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -83,19 +83,19 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
    *
    * TODO: investigate k8s websockets support as an alternative HTTP approach
    *
-   * @return An asynchronous stream of [[Watch]] objects containing additions, modifications, and
-   *         deletions to the items in the list being watched, and a Closable handle allowing for
-   *         the watch to be terminated.
+   * @return An update function that may be passed to [[Var.async]] to create stream of [[Watch]] objects containing
+   *         additions, modifications, and deletions to the items in the list being watched, and a Closable handle
+   *         allowing for the watch to be terminated.
```
olix0r: nit: i suspect these comment lines are a big longer than the usual 100 char limit we like to apply.

Would you mind wrapping comments at 100 chars here & throughout?  Thanks!

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -83,19 +83,19 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
    */
   def watch(
     labelSelector: Option[String] = None,
     fieldSelector: Option[String] = None,
     resourceVersion: Option[String] = None
-  ): (AsyncStream[W], Closable) = {
+  ): (Activity.State[W] => Unit) => Closable = { (output: Activity.State[W] => Unit) =>
```
olix0r: nit: probably most idiomatic to call _output_ `state`

siggy:
memory usage flat after 24 hours. that drop at the end is me forcing a gc with `jmap`:
screen shot 2017-11-28 at 10 46 03 am

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -139,71 +121,108 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+            state(Activity.Failed(exc))
+            Future.exception(exc)
         }
       }
-        // Ignore any cases where we receive a resource version lower
-        // than the last received resource version.
-        .increasingOnly
+    }
+
+    def _processEventStream(
```
hawkw: A comment summarizing this function's behaviour might be nice.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -139,71 +121,108 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+      stream: () => AsyncStream[W],
+      resourceVersion: Option[String],
+      previousEvent: Option[W] = None
+    ): Future[Unit] = {
+      stream().uncons.flatMap {
+        case Some((w, ws)) =>
```
hawkw: Take it or leave it, but I think this _might_ be a little clearer if we collapsed the nested `match` expression into one `match` with guards, like:
```scala
// special case to handle Kubernetes bug where "too old
// resource version" errors are returned with status code 200
// rather than status code 410.
// see https://github.com/kubernetes/kubernetes/issues/35068
// for details
case Some((e: Watch.Error[0], ws))  if e.status.code.contains(410) =>
  log.debug(
   "k8s returned 'too old resource version' error with " +
     "incorrect HTTP status code, restarting watch"
  )
  _resourceVersionTooOld()
  
case Some((event, ws)) =>
  import Ordering.Implicits._
  if (previousEvent.forall((_: W) < event)) {
    state(Activity.Ok(event))
  } else {
    // Ignore any events where we receive a resource version lower
    // than the last received resource version.
    _processEventStream(ws, resourceVersion, previousEvent)
  }

// if the stream ends (k8s will kill connections after ~30m), restart it.
case None if resourceVersion.isDefined => _watch(resourceVersion)

// In this case, we want to try loading the initial information instead before
// watching again.
case None => _resourceVersionTooOld()
```

k8s/src/test/scala/io/buoyant/k8s/WatchableTest.scala
```diff
@@ -1,24 +0,0 @@
-package io.buoyant.k8s
-
-import com.twitter.concurrent.AsyncStream
-import io.buoyant.test.FunSuite
-import org.scalatest.prop.GeneratorDrivenPropertyChecks
-
-class WatchableTest extends FunSuite
```
hawkw: Since `IncreasingOnlyAsyncStream` was removed, I'd like to see another test somewhere that asserts that the events on `Watchable.Activity` have monotonically-increasing resource versions even if the Kubernetes API sends events out of order...

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
sgrankin: (Also rephrased the comments slightly for clarity)

k8s/src/test/scala/io/buoyant/k8s/WatchableTest.scala
sgrankin: Good catch!

sgrankin:
Updated with CR fixes and rebased on latest master.  
Note that io.buoyant.grpc.interop.NetworkedInteropTest seems to be flaky (failed, then passed on re-run).

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -83,19 +84,19 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+   * @return An update function that may be passed to [[Var.async]] to create stream of [[Watch]]
+   *         objects containing additions, modifications, and deletions to the items in the list
+   *         being watched, and a Closable handle allowing for the watch to be terminated.
    */
   def watch(
     labelSelector: Option[String] = None,
     fieldSelector: Option[String] = None,
     resourceVersion: Option[String] = None
-  ): (AsyncStream[W], Closable) = {
+  ): (Activity.State[W] => Unit) => Closable = { (state: Activity.State[W] => Unit) =>
```
adleong: Could this return type just be `Activity[W]` instead?  I think this should be as simple as moving `Activity(Var.async(...) { ... })` from the caller into this method.  That would make this method signature a little easier to understand.

hawkw:
@sgrankin yeah, that's one of a handful of integration tests that are known to be somewhat flaky on CI (https://github.com/linkerd/linkerd/issues/1504), please feel free to restart any builds failed by that test!

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -139,71 +121,106 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+    }
+
+    /**
+     * Recursively consume an AsyncStream of watch events from an established connection,
+     * updating [[state]] when appropriate.  Restart the watch on various error conditions.
+     */
+    def _processEventStream(
+      stream: () => AsyncStream[W],
+      resourceVersion: Option[String],
+      previousEvent: Option[W] = None
```
adleong: consider calling this "largestVersion" or something since it may not actually be the previous event

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -139,71 +121,106 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+            "k8s returned 'too old resource version' error with incorrect HTTP status code, " +
+              "restarting watch"
+          )
+          _resourceVersionTooOld()
+
+        case Some((event, ws)) =>
+          import Ordering.Implicits._
+          // Register the update only if its resource version is larger than the largest version
+          // seen so far.
+          if (previousEvent.forall((_: W) < event)) {
```
adleong: is the `_: W` type annotation necessary?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -139,71 +121,106 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+    def _processEventStream(
+      stream: () => AsyncStream[W],
+      resourceVersion: Option[String],
+      previousEvent: Option[W] = None
+    ): Future[Unit] = {
+      stream().uncons.flatMap {
+        // Special case to handle Kubernetes bug where "too old resource version" errors are
+        // returned with status code 200 rather than status code 410.
+        // see https://github.com/kubernetes/kubernetes/issues/35068 for details
+        case Some((e: Watch.Error[O], _)) if e.status.code.contains(410) =>
```
hawkw: Thanks for flattening the match arms like I asked, this looks great. The rephrased comments are much clearer too!

k8s/src/test/scala/io/buoyant/k8s/v1/ApiTest.scala
```diff
@@ -432,30 +431,77 @@ class ApiTest extends FunSuite
...
+            Seq("10.248.2.8", "10.248.7.10", "10.248.8.8"))
+      }
+
+      poll { case None => }
+
+      // repeat the event: no update since resource version is the same
+      await(w.write(modified0))
+      poll { case None => }
+
+      // write an earlier event: no update since resource version is too low
```
hawkw: Thanks for adding this test!

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -139,71 +121,106 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
-        .increasingOnly
+    }
+
+    /**
+     * Recursively consume an AsyncStream of watch events from an established connection,
+     * updating [[state]] when appropriate.  Restart the watch on various error conditions.
+     */
+    def _processEventStream(
+      stream: () => AsyncStream[W],
+      resourceVersion: Option[String],
```
adleong: I think this variable shadows the resourceVersion passed to `watch` which is a bit confusing.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
sgrankin: IIRC I actually started this change with watch returning a Event[].  It was indeed cleaner, and worked in deployment, but the tests failed:
- some of the existing tests write multiple events into the response to be processed before actually testing what came out on the other end.
- the only way to consume a stream of events from a Var or Activity seems to be Var.changes or Activity.states (which then calls Var.changes)
- in the observe method of the Var created in Var.async, I *think* multiple updates were managing to sneak in between the synchronized creation of the nested Var and the call to .observe which would register the Observer, so the first update would not necessarily be seen
- and so the tests would not necessarily see the first event and would fail.

I thought this behavior of missing some initial events should be ok in the wild (we'd get the latest event eventually), but wanted to make sure the tests (that test the `watch` function directly) saw the complete stream of events, as they did previously with AsyncStream.  I was also trying to avoid changing the logic of the tests.  So I ended up with this function  thus this function and the concurrent-queue machinery in the tests. 

I don't see a way of extracting the complete stream of events out of a Var (otherwise we'd likely run into the same issue as with AsyncStream, right?).  Anything I may be missing about Vars/Activities?

Actually... another option to improve the readability without changing the signature much (obvious in retrospect):
```
def watch(
    labelSelector: Option[String] = None,
    fieldSelector: Option[String] = None,
    resourceVersion: Option[String] = None,
    state: Activity.State[W] => Unit
  ) 
```
The watchable call-site doesn't actually benefit from being able to partially watch, and the ApiTest callsite is easily changed to work with this signature as well.

Thoughts?   It'd still keep the weird `state` type but make the `watch` signature plainer, and the tests can still avoid dealing with Var quirks.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
sgrankin: It was in some other incarnation of this lambda, but it seems no longer (or was possibly an IntelliJ-only error).  Will remove.

adleong:
Ah, that's super interesting!  Yes, I think you are correct that it is a fundamental property of `Var` (and `Event`) that they do not keep history.  With that in mind, your approach here makes sense.  I do like the updated signature you propose (moving state into the param list).

siggy:
Hi @sgrankin, just checking in to see how things are going. We'd love to get this merged for our 1.3.3 release this week.

sgrankin:
I just got home :) 
I’ll do the last round of changes sometime in the next few hours — since these are all effectively cosmetic change stability should be unaffected (in case you are concerned about re-running the stress tests).

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
sgrankin: updated to 'largestVersion' and 'largestEvent'.  

Aside: a good cleanup to do afterwards may be to get rid of the event argument here.  I think the resource version comparison from ResourceVersionOrdering can be factored out and used directly with the resource version here, avoiding the need to keep a whole event around.  (I think this may be the only use case of that Ordering type class ...  I didn't do the change wanting to limit the diff size.)

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
sgrankin: per discussion, went with the l updated signature

sgrankin:
@siggy: I'm done with all the changes. Thanks!

Issue #1718, opened 2017-11-29
dnssrv namer does not update the DNS records.


carloszuluaga:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**: After first DNS query response the poll mechanism does not update the DNS records.

**What you expected to happen**: Updated records get available for load balancing.

**How to reproduce it (as minimally and precisely as possible)**: 
1. Start Linkerd with a DNS Record
2. Change DNS Record and wait for DNS refresh interval.
3. Check that load balancing nodes are unchanged.

**Environment**: Official docker image
- linkerd/namerd version, config files: 1.3.2

PR #1719, opened 2017-11-29
Ensure dnssrv namer update DNS updated records

carloszuluaga:
* Refactor the DNS lookup function using a Finagle Timer for DNS refresh interval and wrap it with an Activity.
 * Refactor the lookup function to map all in terms of Activity.
Fixes: #1718

### DIFF
--- a/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
+++ b/namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
@@ -1,15 +1,14 @@
 package io.buoyant.namer.dnssrv
 
-import java.io.IOException
 import java.net.{InetAddress, InetSocketAddress, UnknownHostException}
-import java.util.concurrent.atomic.AtomicBoolean
 
 import com.twitter.finagle._
 import com.twitter.finagle.stats.{Stat, StatsReceiver}
 import com.twitter.logging.Logger
-import com.twitter.util.Activity.State
 import com.twitter.util._
 import org.xbill.DNS
+
+import scala.util.control.NoStackTrace
 
 class DnsSrvNamer(
   prefix: Path,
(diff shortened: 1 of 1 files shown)


namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -32,80 +30,84 @@ class DnsSrvNamer(
...
     path.take(1) match {
       case id@Path.Utf8(address) =>
-        Activity(Var.async[State[NameTree[Name]]](Activity.Pending) { state =>
+        val vaddr = watchDns(address).run.map {
+          case Activity.Ok(rsp) =>
+            Addr.Bound(rsp: _*)
+          case Activity.Pending =>
+            Addr.Pending
+          case Activity.Failed(e) =>
+            log.debug("SDS lookup failure: %s", e.getMessage)
```
ccmtaylor: maybe replace `SDS` with `DNS-SRV` for consistency? I've only run across the term SDS in Envoy docs, but I don't think it applies here.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -32,80 +30,84 @@ class DnsSrvNamer(
...
-          val srvRecords = answers.flatMap {
-            case srv: DNS.SRVRecord => try {
-              val inetAddress = InetAddress.getByName(srv.getTarget.toString())
-              Some(Address(new InetSocketAddress(inetAddress, srv.getPort)))
-            } catch {
-              case _: UnknownHostException =>
-                log.warning(s"srv lookup of $address returned unknown host ${srv.getTarget}")
-                badHosts.incr()
-                None
+  private def watchDns(dsnsrvRecord: String)(implicit timer: Timer = DefaultTimer): Activity[List[Address]] = {
```
ccmtaylor: I don't think you need to provide `DefaultTimer` here; there should be an implicit timer in scope for the whole class -- see line 20

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -32,80 +30,84 @@ class DnsSrvNamer(
...
+          case code =>
+            val msg = s"unexpected result: $code for $dsnsrvRecord: ${lookup.getErrorString}"
+            log.error(msg)
+            update.update(Activity.Failed(new Exception(msg)))
+        }
+      }
+
+      val _ = doUpdate()
+
+      timer.schedule(refreshInterval) {
```
ccmtaylor: related: if you use the timer explicitly, maybe pass it as a regular parameter?

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -32,80 +30,84 @@ class DnsSrvNamer(
...
-          log.error(msg)
-          throw new IOException(msg)
+          case code =>
+            val msg = s"unexpected result: $code for $dsnsrvRecord: ${lookup.getErrorString}"
+            log.error(msg)
+            update.update(Activity.Failed(new Exception(msg)))
+        }
+      }
+
+      val _ = doUpdate()
```
ccmtaylor: if you don't use the return value, maybe give the method a type of `doUnit(): Unit`?

siggy:
Thanks for the PR @carloszuluaga, and also thanks @ccmtaylor for the reviews! Heads up we're targeting a 1.3.3 release by the end of this week. If this branch is ready by then we'd love to include it.

carloszuluaga:
@ccmtaylor fixed according to your code review.

carloszuluaga:
I think the CI failure is not related to this changes, @siggy can you please take a look?

siggy:
@carloszuluaga I agree the ci failure looks unrelated to your change. Can you hit rebuild on the circle ci dashboard? I don't have access because it's running under your user.

carloszuluaga:
@siggy I don't have access either, the button is disabled with this message "You need write permissions to trigger builds."

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -32,80 +30,84 @@ class DnsSrvNamer(
...
-            NameTree.Leaf(Name.Bound(Var.value(Addr.Bound(srvRecords: _*)), id, residual))
-          }
-        case code =>
-          val msg = s"unexpected result: $code for $address: ${lookup.getErrorString}"
-          log.error(msg)
-          throw new IOException(msg)
+          case code =>
+            val msg = s"unexpected result: $code for $dsnsrvRecord: ${lookup.getErrorString}"
+            log.error(msg)
+            update.update(Activity.Failed(new Exception(msg)))
```
hawkw: Very minor, but I notice the type of exception here was changed from `IOException` to generic `Exception`, which is a loss of specificity --- is there a more specific exception that can be set here, or do you think `new Exception` is the most correct?

carloszuluaga: @hawkw I think Excepcion is the most correct because in this case the error is not produced by failed or interrupted I/O operations.

hawkw: @carloszuluaga I agree that `IOException` is not the correct type here, I'm just wondering if there's something that conveys more meaning than `Exception`.

carloszuluaga:
@hawkw thanks for the review. I agree. Defined a custom exception with the right meaning of the error. Let me know if this is ok.

namer/dnssrv/src/main/scala/io/buoyant/namer/dnssrv/DnsSrvNamer.scala
```diff
@@ -32,80 +31,90 @@ class DnsSrvNamer(
...
+
+      timer.schedule(refreshInterval) {
+        doUnit()
       }
     }
+    Activity(state)
   }
 }
+
+case class DNSLookupException(msg: String)
```
hawkw: This looks good to me – I might consider having separate subtypes for the "lookup errored" and "lookup contained no SRV records" cases, but that's bikeshedding at this point... :)

siggy:
@carloszuluaga please send us your t-shirt size so we can thank you for this work!

carloszuluaga:
@siggy wow! thank you so much :smiley:. my t-shirt size is M

PR #1734, opened 2017-12-06
Add request parameters to identified request path in the `io.l5d.path` identifier

### LINKED ISSUE #1725: PathIdentifier consumes a segment of the path with query parameters
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
`PathIdentifier` consumes a segment of the path with query parameters.

Expect query parameters to remain in the query

I have:
1. dtab record `/svc/play => /#/io.l5d.k8s/namespace/http/play-service`. 
1. Service in Kubernetes that handles requests under the `/`.
1. Linkerd with `PathIdentifier`
    ```
        identifier:
          - kind: io.l5d.path
            segments: 1
            consume: true
    ```

So when I query service with url `/play?param=value` query parameters don't come to the service. But when I query `/play/?param=value` everything works well.

deebo91:
When using the `io.l5d.path` identifier this these config values
```yaml
kind: io.l5d.path
segments: 1
consume: true
```
and a request url like `GET /mysvc?param=1&param=2` Linkerd would drop the query parameters when sending the request for dtab delegation. 

This PR makes sure that query parameters are added to the dtab request url if the query parameters are available. This was tested with a local instance of Linkerd talking to a simple service that accepted requests with params. 

fixes #1725

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
@@ -20,13 +20,23 @@ case class PathIdentifier(
   def apply(req: Request): Future[RequestIdentification[Request]] =
     req.path.split("/").drop(1) match {
       case path if path.size >= segments =>
+        val params = req.getParams()
+
         if (consume) {
-          req.uri = req.uri.split("/").drop(segments + 1) match {
+          req.uri = req.path.split("/").drop(segments + 1) match {
             case Array() => "/"
             case x =>
-              val trailingSlash = if (req.uri.endsWith("/")) "/" else ""
+              val trailingSlash = if (req.path.endsWith("/")) "/" else ""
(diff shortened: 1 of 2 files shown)


router/http/src/main/scala/io/buoyant/router/http/PathIdentifier.scala
```diff
@@ -20,13 +20,23 @@ case class PathIdentifier(
   def apply(req: Request): Future[RequestIdentification[Request]] =
     req.path.split("/").drop(1) match {
       case path if path.size >= segments =>
+        val params = req.getParams()
```
adleong: rather than getting and re-encoding the query string, I would just do something like:

```
val query = {
  val u = req.uri
  u.indexOf('?') match {
    case -1 => ""
    case n => u.substring(n)
  }
}
```

PR #1738, opened 2017-12-08
Fix incorrect handling of Consul ServiceRelease exception

### LINKED ISSUE #1670: Undesired Warning in namerd output when consul observation is released

hawkw:
In the Consul namer `SvcAddr` class, a singleton exception called `ServiceRelease` is raised when an `Activity` watching Consul is closed. Previously, this exception was a `val`. However, due to how equality is implemented for `Failure`s, the by the time exception raised reached the `match` in `loop()`, the `Failure` did not always equal the value of `ServiceRelease`, seemingly because of the `ProcessFailure` filter masking some of the flags on the `Failure`. This led to service releases being handled as errors, rather than gracefully cancelling the observation, leading to a _lot_ of unnecessary exceptions in the logs.

I've fixed this issue by setting the cause of the raised exception to a singleton object rather than a `val`, so that the match arm can test type rather than equality. The exception raised to cancel an observation should now always match the correct pattern.

I tried to write a unit test for this that makes assertions about the contents of the logs, but I couldn't make it fail against the current master. It appeared that the failure was being munged only when the namer was fully integrated with Linkerd. Instead, I tested this manually using Consul running locally and setting the number of bound names in the binding cache to 1. I would request one service in Consul, and then a second one – thus forcing the namer to evict the first bound name from the bindcache, stopping the observation.

Fixes #1670, for real this time.

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -64,15 +64,7 @@ private[consul] object SvcAddr {
 
         if (stopped) Future.Unit
         else getAddresses(index0).transform {
-          case Throw(Failure(Some(err: ConnectionFailedException))) =>
-            log.warning(
-              "consul datacenter '%s' service '%s' retrying on connection" +
-                " failure: %s",
-              datacenter, key.name, err
-            )
-            // Drop the index, in case it's been reset by a consul restart
-            loop(None)
-          case Throw(f: Failure) if f.cause.contains(ServiceRelease) =>
+          case Throw(Failure(Some(ServiceRelease))) =>
             // this exception is raised when we close a watch - thus, it needs
             // to be special-cased so that we don't continue observing that
             // service.
@@ -83,6 +75,14 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
+          case Throw(Failure(Some(err: ConnectionFailedException))) =>
+            log.warning(
+              "consul datacenter '%s' service '%s' retrying on connection" +
+                " failure: %s",
+              datacenter, key.name, err
+            )
+            // Drop the index, in case it's been reset by a consul restart
+            loop(None)
           case Throw(e) =>
             // an error occurred. if we've previously seen a good state, fall
             // back to that state; otherwise, fail.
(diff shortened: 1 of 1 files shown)


namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -18,6 +18,8 @@ private[consul] case class SvcKey(name: String, tag: Option[String]) {
 
 private[consul] object SvcAddr {
 
+  private[this] object ServiceRelease extends Exception with NoStackTrace
```
adleong: what do you think about making this a `private[this] val ServiceRelease = new Exception("fjdklasf") with NoStackTrace` since SvcAddr is already a singleton anyway

hawkw: I made it an object because I was concerned that if it was a singleton `val` the equality might still get munged, but the type test would still pass?

adleong: I think what you have right now is an equality check, not a type check.  (since ServceRelease is an object and not a type)

hawkw:
@adleong added to the description & made the changes you requested!

yangzhares:
I also want to file an issue for this, good to see guys have fixed it, we also hit it under namerd 1.3.0, thanks guys.

PR #1739, opened 2017-12-08
Add a check to disallow invalid namespaces

### LINKED ISSUE #1733: infinite consul log messages when invalid dtab namespace given
Using this modified consul example setup:
https://github.com/linkerd/linkerd-examples/tree/siggy/consul-namerd/consul

Providing an invalid namespace:
```bash
curl -v "192.168.99.100:4180/api/1/dtabs/%2e%2e/foo"
```

Namerd logs these forever:
```
E 1206 19:06:58.113 UTC THREAD31 TraceId:c7f4d1ae66700e4b: consul ns %2e%2e/foo dtab observation error UnexpectedResponse(301: <a href="/v1/kv/namerd/foo?raw=true">Moved Permanently</a>.

)
```

deebo91:
Problem
When namerd uses consul for dtab storage, invalid characters in a namespace e.g. `%2e%2e/foo` causes namerd to infinitely log a `dtab observation error`

Solution
Add a check to ensure that only alphanumeric characters can be used as keys in consul to store dtabs. 

Validation
Created a varying number of valid and invalid strings with a regex pattern and curled namerd with both types of strings. The tests worked successfully. However, curling a large number of requests to namerd caused consul to remove namerd as a service and subsequently caused `ChannelClosedExceptions`. I believe that is probably caused by DoSing consul with a bash script and not because of the proposed change.

Fixes #1733

### DIFF
--- a/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
+++ b/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
@@ -52,6 +52,9 @@ object DtabStore {
   class DtabNamespaceDoesNotExistException(ns: Ns)
     extends Exception(s"The dtab namespace $ns does not exist")
 
+  class DtabNamespaceInvalidException(ns: Ns)
+    extends Exception(s"The dtab namespace $ns contains invalid characters")
+
   object Forbidden extends Exception("You do not have sufficient permissions")
 
   class Proxy(underlying: DtabStore) extends DtabStore {
--- a/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
+++ b/namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
@@ -53,7 +53,7 @@ object DtabStore {
     extends Exception(s"The dtab namespace $ns does not exist")
 
   class DtabNamespaceInvalidException(ns: Ns)
-    extends Exception(s"The dtab namespace $ns contains invalid characters")
+    extends Exception(s"invalid dtab namespace $ns: namespace must contain only letter, number or '-' characters")
 
   object Forbidden extends Exception("You do not have sufficient permissions")
 
--- a/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
+++ b/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
@@ -57,6 +57,7 @@ class ConsulDtabStore(
             }
         else
           Future.Unit
+
       val pending = cycle(None)
 
       Closable.make { _ =>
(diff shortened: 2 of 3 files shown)


namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -1,13 +1,15 @@
 package io.buoyant.namerd
 package storage.consul
 
+import java.util.regex.Pattern
```
hawkw: tioli, but can we use the scala stdlib's regex stuff instead?

namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
```diff
@@ -52,6 +52,9 @@ object DtabStore {
   class DtabNamespaceDoesNotExistException(ns: Ns)
     extends Exception(s"The dtab namespace $ns does not exist")
 
+  class DtabNamespaceInvalidException(ns: Ns)
+    extends Exception(s"The dtab namespace $ns contains invalid characters")
```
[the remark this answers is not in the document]
olix0r: i don't think we need to be that fancy, however, it's probably better to say something like _"Invalid dtab namespaces $ns: must contain only DNS-safe characters"_ -- or something.  We should ideally indicate what set of characters are _allowed._

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -19,6 +21,8 @@ class ConsulDtabStore(
 
   private[this] val log = Logger.get("consul")
 
+  def namespaceIsValid(ns: Ns): Boolean = Pattern.matches("[\\w+\\-?/?]+", ns)
```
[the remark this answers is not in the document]
olix0r: We should definitely build the regex statically so that it doesn't need to be compiled on every invocation of this method.  Regex invocation is slow.

```scala
val validNs = "[\\w+\\-?/?]+".r

def namespaceIsValid(ns: Ns): Boolean = validNs.matches(ns)
```

Or something like that

namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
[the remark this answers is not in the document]
hawkw: Yeah, that's fine too – I think if we're going to have restrictions on what characters are allowed, they should be in the namerd docs as well.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
olix0r: Also, this regex doesn't make it very clear what character set we're supporting.

Can this just be `^[A-Za-z0-9_-]+$`, for instance?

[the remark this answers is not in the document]
hawkw: Also, if we want a valid name to consist only of canonical DNS label characters, we already have a regex for matching them: 
https://github.com/linkerd/linkerd/blob/642e14fd7e7f0d961a023cbf55ecaa23b070f649/namer/core/src/main/scala/io/buoyant/hostport.scala#L12-L20
We could put this regex someplace where it's publicly visible and just reuse it...

hawkw: Oh, also, TIOLI, but if you use a [raw string](https://docs.scala-lang.org/overviews/core/string-interpolation.html#the-raw-interpolator) you won't have to escape the `\` characters in your regex, which might make it a little more readable.

namerd/core/src/main/scala/io/buoyant/namerd/DtabStore.scala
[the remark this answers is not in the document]
deebo91: I can definitely include what characters are allowed in the docs and possibly in the log message.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
deebo91: I can do that, wasn't aware scala had its own regex lib

deebo91:
All great points @olix0r and @hawkw! thanks for the review. Should have another commit by end of day.

olix0r:
@adleong i assumed that since consul exposes things via DNS, that there is _some_ DNS restriction somewhere. I don't actually know if that's appropriate here...

deebo91:
@hawkw I like the idea of moving the validation to the a new `Ns` class, but I am a little apprehensive about changing Ns from a type to a class. Would changing that introduce breaking changes to other classes that use `Ns`?

hawkw:
@deebo91 If you were to just copy and paste the code example I gave above, yes. But if you were to also add an implicit conversion from `Ns` back to `String`, consumers of `Ns` would still be able to call all the `String` methods on it. Something like
```scala
implicit def nsToString(ns: Ns): String = ns.ns
```
and then just make sure the implicit is in scope at the call site, and all consumers of `Ns` that call `String` methods on it will still work.

hawkw:
Of course, that does break some of the _advantages_ of type-level validation; if I can implicitly convert an `Ns` back into a string then I can edit the string to include invalid characters again. You might prefer to have your `Ns` class only implement the methods on `String` that consumers of `Ns` will actually be able to use safely.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -62,7 +69,8 @@ class ConsulDtabStore(
     Activity(run).stabilize
   }
 
-  def create(ns: Ns, dtab: Dtab): Future[Unit] =
+  def create(ns: Ns, dtab: Dtab): Future[Unit] = {
+    if (!namespaceIsValid(ns)) Future.exception(new DtabNamespaceInvalidException(ns))
```
adleong: I don't think this does anything.  I think you either need to `return Future.exception(...)` here or (my preference) put the rest of this method in an `else` block.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -162,4 +177,6 @@ class ConsulDtabStore(
     }
     Activity(run).stabilize
   }
+
```
adleong: remove newlines

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -19,6 +19,13 @@ class ConsulDtabStore(
 
   private[this] val log = Logger.get("consul")
 
+  val validNs = raw"^[A-Za-z0-9_-]+".r
```
olix0r: probably should be `private[this]` and/or moved into companion `object`

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -62,54 +70,70 @@ class ConsulDtabStore(
...
-
-  def delete(ns: Ns): Future[Unit] = {
-    val key = s"${root.show}/$ns"
-    api.get(key, datacenter = datacenter, consistency = writeConsistency).transform {
-      case Return(_) => api.delete(
-        key,
+  def create(ns: Ns, dtab: Dtab): Future[Unit] = {
+    if (!namespaceIsValid(ns))
+      Future.exception(new DtabNamespaceInvalidException(ns))
+    else
```
olix0r: nit: prefer braces on multi-line conditional statements.

i.e.
```
// ok
if (true) foo else bar

// ok
if (true) foo.withSomethingLongish()
else bar

// ok
if (true) {
  foo
} else {
  bar.spanning(
    mutliple,
    lines
  )
}

```

olix0r: here & throughout

namerd/storage/consul/src/test/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreTest.scala
```diff
@@ -0,0 +1,70 @@
...
+    )
+
+    @volatile var state: Activity.State[Set[Ns]] = Activity.Pending
+    store.list.states respond {
+      state = _
+    }
+    assert(state == Activity.Ok(Set("foo/bar", "foo/baz")))
+  }
+
+  test("Block when namespaces are not found") {
```
adleong: I think this test is mistitled.  it looks like it tests that when the dtab directory is absent, an empty list of dtabs is returned

namerd/storage/consul/src/test/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreTest.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import com.twitter.finagle.http.{Request, Response, Status}
+import com.twitter.finagle.{Path, Service}
+import com.twitter.io.Buf
+import com.twitter.util.{Activity, Future}
+import io.buoyant.consul.v1.KvApi
+import io.buoyant.namerd.Ns
+import io.buoyant.test.{Awaits, Exceptions, FunSuite}
+
+class ConsulDtabStoreTest extends FunSuite with Awaits with Exceptions {
```
adleong: it doesn't look like this tests the invalid namespace check at all.  that check applies to the create, delete, put, and get methods.

deebo91: I can tests for that functionality. I was under the impression that we were trying to get this out the door by today so I just wrote basic tests for the ConsulDTabStore

chrismikehogan:
Were names like `foo.bar` actually causing problems? We use namespace names following this pattern, and have not seen them cause any identifiable problems.

But since 1.3.4 these namespace names are invalid and 500 when queried via namerd, which is a breaking change from 1.3.3.

PR #1740, opened 2017-12-10
Added a namer for Rancher

### LINKED ISSUE #1146: Linkerd Namer for Rancher Cattle
**_Enhancement request_**

I'd like to be able to have a namer consume the Rancher Cattle Metadata service that would allow linkerd within the cattle CNI managed network (ideally also in the docker and host networks too).

The metadata service appears to support long-blocking connections which unblock when the metadata is refreshed. This could be used to build a watcher. 

The following scenarios should all be considered (as well as `linkerd-->linkerd` paths):
``` Public --> Rancher LB --> linkerd --> service/container```
``` Public --> linkerd --> service/container```
``` Public --> linkerd --> Rancher LB --> service/container```

See Example blocking/wait call at: 
https://github.com/rancher/go-rancher-metadata/blob/master/metadata/change.go#L51

More info on Rancher Cattle Metadata Service:
https://docs.rancher.com/rancher/v1.5/en/rancher-services/metadata-service/

### PR DESCRIPTION
**Pull request by fangel**
Linkerd currently does not support service-discovery in environments that uses Rancher for orchestration. This makes Linkerd less attractive to people using Rancher for the Docker environments, as they would have to implement additional infrastructure to get service-discovery that way. Rancher has a JSON-based metadata-API that contains information about all of the containers that are running in Rancher, so it would be possible to perform service-discovery that way.

This commit introduces a new namer that utilizes the long-polling functionality of the Rancher metadata-API to achieve near-instant updates to the discovered services.

I have run some small-scale validation tests on a Rancher platform to ensure that the client accuratly detects the running services and can route requests to them.

Fixes #1146

### DIFF
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -15,7 +15,7 @@ These parameters are available to the namer regardless of kind. Namers may also 
 
 Key | Default Value | Description
 --- | ------------- | -----------
-kind | _required_ | Either [`io.l5d.fs`](#file-based-service-discovery), [`io.l5d.serversets`](#zookeeper-serversets-service-discovery), [`io.l5d.consul`](#consul-service-discovery), [`io.l5d.k8s`](#kubernetes-service-discovery), [`io.l5d.marathon`](#marathon-service-discovery), [`io.l5d.zkLeader`](#zookeeper-leader), [`io.l5d.curator`](#curator), or [`io.l5d.rewrite`](#rewrite).
+kind | _required_ | Either [`io.l5d.fs`](#file-based-service-discovery), [`io.l5d.serversets`](#zookeeper-serversets-service-discovery), [`io.l5d.consul`](#consul-service-discovery), [`io.l5d.k8s`](#kubernetes-service-discovery), [`io.l5d.marathon`](#marathon-service-discovery), [`io.l5d.zkLeader`](#zookeeper-leader), [`io.l5d.curator`](#curator), [`io.l5d.rancher`](#rancher), or [`io.l5d.rewrite`](#rewrite).
 prefix | namer dependent | Resolves names with `/#/<prefix>`.
 experimental | `false` | Set this to `true` to enable the namer if it is experimental.
 transformers | No transformers | A list of [transformers](#transformer) to apply to the resolved addresses.
@@ -692,6 +692,54 @@ Key | Required | Description
 --- | -------- | -----------
 prefix | yes | Tells linkerd to resolve the request path using the curator namer.
 serviceName | yes | The name of the Curator service to lookup in ZooKeeper.
+
+<a name="rancher"></a>
+## Rancher
+
+kind: `io.l5d.rancher`
+
+### Rancher configuration
+
+> Configure a Rancher namer:
+
+```yaml
+namers:
+- kind: io.l5d.rancher
+  experimental: true
+  portMappings:
+    proxy: 8080
+```
+
+> Then reference the namer in the dtab to use it:
+
+```yaml
+dtab: |
+  /rancher => /#/io.l5d.rancher
+  /svc => /$/io.buoyant.http.domainToPathPfx/rancher/http;
+```
+
+linkerd provides support for service discovery via Rancher's [Metadata-API](https://rancher.com/docs/rancher/v1.6/en/rancher-services/metadata-service/).
+
+Key | Default Value | Description
+--- | ------------- | -----------
+prefix | `io.l5d.rancher` | Resolves names with `/#/<prefix>`.
+experimental | `false` | Since the Rancher namer is still considered experimental, this must be set to `true`.
+portMappings | `<empty map>` | If specified, you can use the names of these port-mappings for the `<port>` path parameter. By default, the namer knows `http` and `https`.
+
+### Rancher Path Parameters
+
+> Dtab Path Format
+
+```yaml
+/#/<prefix>/<port>/<stack>/<service>
+```
+
+Key | Required | Description
+--- | -------- | -----------
+prefix | yes | Tells linkerd to resolve the request path using the Rancher namer.
+port | yes | The port name or number to route the request to
+stack | yes | Name of the stack of the service
+service | yes | The name of the service
 
 <a name="rewrite"></a>
 ## Rewrite
--- a/linkerd/docs/namer.md
+++ b/linkerd/docs/namer.md
@@ -714,8 +714,9 @@ namers:
 
 ```yaml
 dtab: |
-  /rancher => /#/io.l5d.rancher
-  /svc => /$/io.buoyant.http.domainToPathPfx/rancher/http;
+  /rancher => /#/io.l5d.rancher;
+  /s => /rancher/http;
+  /svc => /$/io.buoyant.http.domainToPathPfx/s;
 ```
 
 linkerd provides support for service discovery via Rancher's [Metadata-API](https://rancher.com/docs/rancher/v1.6/en/rancher-services/metadata-service/).
@@ -725,6 +726,7 @@ Key | Default Value | Description
 prefix | `io.l5d.rancher` | Resolves names with `/#/<prefix>`.
 experimental | `false` | Since the Rancher namer is still considered experimental, this must be set to `true`.
 portMappings | `<empty map>` | If specified, you can use the names of these port-mappings for the `<port>` path parameter. By default, the namer knows `http` and `https`.
+maxWait | 30 | The max number of seconds to wait for changes in the Rancher Metadata-API before starting a new request.
 
 ### Rancher Path Parameters
 
--- a/namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
+++ b/namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
@@ -1,45 +1,47 @@
 package io.buoyant.namer.rancher
 
-import com.fasterxml.jackson.databind.{ObjectMapper, PropertyNamingStrategy}
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.{DeserializationContext, ObjectMapper, PropertyNamingStrategy}
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer
+import com.fasterxml.jackson.databind.module.SimpleModule
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties
-
 import com.twitter.finagle.{Service, Http, Address, Stack}
 import com.twitter.finagle.http
 import com.twitter.finagle.service.Retries
-import com.twitter.finagle.tracing.NullTracer
 import com.twitter.finagle.stats.{Stat, StatsReceiver}
-
+import com.twitter.finagle.tracing.NullTracer
 import com.twitter.logging.Logger
 import com.twitter.util.Activity.State
 import com.twitter.util.{Timer, Var, Activity, Duration, Future, Closable, Return, Throw}
-
 import java.util.concurrent.atomic.AtomicBoolean
+import scala.reflect.{ClassTag, classTag}
 
 case class PortMapping(
-  val ip: String,
-  val publicPort: Int,
-  val privatePort: Int,
-  val protocol: String
+  ip: String,
+  publicPort: Int,
+  privatePort: Int,
+  protocol: String
 ) {
-  // This is a bit of a hack - and should have more error-checking. But it
-  // makes the JSON parsing work.
-  // If anyone knows how to do a better constructor - or JSON Deserializer,
-  // that would be nice
-  def this(mapping: String) = this(
-    mapping.split(":")(0),
-    mapping.split(":")(1).toInt,
-    mapping.split(":")(2).split("/")(0).toInt,
-    mapping.split(":")(2).split("/")(1)
-  )
   override def toString(): String = s"$ip:$privatePort:$publicPort/$protocol"
+}
+class PortMappingDeserializer extends StdDeserializer[PortMapping](RancherParser.jClass[PortMapping]) {
+  private[this] val Mapping = raw"([\.0-9]+):([0-9]+):([0-9]+)/(tcp|udp)".r
+  override def deserialize(jp: JsonParser, ctx: DeserializationContext): PortMapping = {
+    val mapping = _parseString(jp, ctx)
+    mapping match {
+      case Mapping(ip, privatePort, publicPort, protocol) =>
+        PortMapping(ip, privatePort.toInt, publicPort.toInt, protocol)
+      case _ =>
+        throw ctx.mappingException(s"Failed to understand ip-binding: $mapping")
+    }
+  }
 }
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 case class RancherContainer(
   createIndex: Int,
-  dns: List[String],
   ips: List[String],
   primaryIp: String,
   labels: Map[String, String],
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by fangel** [comment]
Please note that this is the first time I've ever touched Scala, so there are likely tons of non-idiomatic Scala-code in there. But I welcome any hints on what is considered better Scala-code!

**Comment by fangel** [comment]
So a little point for discussion: Right now the namer will route to the Docker-local ip's of the services (and the internal port of the service, e.g. `80`) – so each instance has it's own ip. This enables services to have a higher scale than the number of Rancher-hosts in your cluster.

The other alternative is to route to the ip of the Rancher-host, and the port that the internal port got mapped to (e.g. `80` in the container maps to `60001` externally on the host).
The reason I haven't done it this way, is because Rancher has a (weird?) idea of always wanting the external port of all instances to be the same across all insatnces- and since two instances on the same host can't share an external port, this limits the number of instances of your service to the number of physical hosts in your cluster.

However, I was just looking through some of the examples of using Linkerd-to-Linkerd, and they rely on a localhost-transformation to only send traffic to the instances running on the same machine. My understanding would be that if the routes are to the virtual Docker-local ip's, then the localhost transformation wouldn't be able to find the services, since their ip's aren't the same as the linkerd-instance, even if run on the same host. So for a similar setup to work in Rancher, I think the routing should go to the external port on the Rancher-hosts - no?

The question is, which mode is the more desirable. An alternative could be to offer a switch in the configuration between the two modes. Then if you need to scale to more instances than hosts (if your application can't utilise more than one core, and your hosts are multi-cored), then you just can't use the same setup for doing linkerd-to-linkerd.

**Comment by wmorgan** [comment]
Awesome. Thanks for the PR, @fangel! We'll take a look.

File: linkerd/docs/namer.md [hunk 1]
```diff
@@ -693,6 +693,54 @@ Key | Required | Description
 prefix | yes | Tells linkerd to resolve the request path using the curator namer.
 serviceName | yes | The name of the Curator service to lookup in ZooKeeper.
 
+<a name="rancher"></a>
+## Rancher
```
hawkw: Thanks for adding a section to the docs!

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+
+      Future.whileDo(!done.get) {
+        log.trace("starting version request")
+
+        val request = http.Request(http.Method.Get, s"/2015-12-19/version?wait=true&maxWait=${maxWait}&value=${v}")
+        request.host = "rancher-metadata"
+
+        service(request)
+          .map { _.getContentString().trim }
+          .liftToTry
```
adleong: you should be able to use `.transform` instead of `.liftToTry.flatMap`

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+          .flatMap {
+            case Return(newVersion) => {
```
adleong: style nit: no need for braces on case bodies

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+  private[this] val version: Activity[String] =
+    Activity(Var.async[State[String]](Activity.Pending) { state =>
+      val done = new AtomicBoolean(false)
+      val initialized = new AtomicBoolean(false)
+      var v = "initial"
+
+      Future.whileDo(!done.get) {
+        log.trace("starting version request")
+
+        val request = http.Request(http.Method.Get, s"/2015-12-19/version?wait=true&maxWait=${maxWait}&value=${v}")
```
adleong: I'm not familiar with the rancher API but let me just say, that is a wacky path.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+    .withParams(Http.client.params ++ params)
+    .withLabel("io.buoyant.namer.rancher.client")
+    .withTracer(NullTracer)
+    .newService(s"/$$/inet/rancher-metadata/80")
+
+  private[this] val version: Activity[String] =
+    Activity(Var.async[State[String]](Activity.Pending) { state =>
+      val done = new AtomicBoolean(false)
+      val initialized = new AtomicBoolean(false)
+      var v = "initial"
```
adleong: I think if you make this a `@volatile var v: Option[String] = None` you can get rid of the initialized boolean.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+                log.trace("metadata-version didn't change. Old: %s, received: %s", v, newVersion)
+              }
+              Future.Unit
+            }
+            case Throw(e) => {
+              // TODO: This should probably be handled by a exponentail back-off
+              // in the Finagle client - but I'm not sure how to achieve that
+              log.error("failed to fetch metadata-version. %s", e)
+              if (!initialized.get) {
+                state.update(Activity.Failed(e))
```
adleong: Depending on how this Var is intended to be used, you may not want to set it into an error state here.  It might be preferable to keep it in pending and keep retrying until you get a successful response.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherNamer.scala
```diff
@@ -0,0 +1,56 @@
...
+      case _ => None
+    }
+
+  def lookup(path: Path): Activity[NameTree[Name]] = path.take(3) match {
+    case phd@Path.Utf8(port, stack, service) => lookupPort(port) match {
+      case Some(portNum) =>
+        val containers: Activity[Option[Addr]] = client.activity.map { allContainers =>
+          val eligable: Set[Address] = allContainers
+            .filter(c => stack.equalsIgnoreCase(c.stackName) && service.equalsIgnoreCase(c.serviceName))
+            .collect(scala.Function.unlift(_.toAddr(portNum)))
```
adleong: I think you should be able to use flatMap here

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
+package io.buoyant.namer.rancher
+
+import com.fasterxml.jackson.databind.{ObjectMapper, PropertyNamingStrategy}
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+
```
adleong: linkerd import style is to group all imports together (no blank lines between imports) and sort them alphabetically.

fangel:
Thanks @adleong - I'll take a look at it when I get a bit of free time.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+  private[this] val numVersions = stats.counter("metadata_versions")
+
+  private[this] val maxWait = 30
+
+  private[this] val service = Http.client
+    .withStack(Http.client.stack.remove(Retries.Role))
+    .withParams(Http.client.params ++ params)
+    .withLabel("io.buoyant.namer.rancher.client")
+    .withTracer(NullTracer)
+    .newService(s"/$$/inet/rancher-metadata/80")
```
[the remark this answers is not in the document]
fangel: I can totally make it configurable if you would prefer that - but according to the documentation for the API, it will always be available at the magical host of "rancher-metadata" on port 80: http://rancher.com/docs/rancher/latest/en/rancher-services/metadata-service/

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
fangel: Agreed. I wouldn't wish it upon anyone to work on the Rancher Metadata-API. It's wrong in so many ways, and this long-polling API isn't even mentioned in their documentation.

But yes, they use dates as their API-versions, and the current API actually also mentions a version (eh, date) that isn't mentioned in the documentation so there is no way of knowing what you would get if you were to use that version)

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+import com.twitter.util.Activity.State
+import com.twitter.util.{Timer, Var, Activity, Duration, Future, Closable, Return, Throw}
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+case class PortMapping(
+  val ip: String,
+  val publicPort: Int,
+  val privatePort: Int,
+  val protocol: String
```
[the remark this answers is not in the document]
fangel: Ah, thanks for the hint about `val` being unnecessary. I'll definitely clean that up, then.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+) {
+  // This is a bit of a hack - and should have more error-checking. But it
+  // makes the JSON parsing work.
+  // If anyone knows how to do a better constructor - or JSON Deserializer,
+  // that would be nice
```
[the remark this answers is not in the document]
fangel: I'll look into doing a custom deserialiser first, but if that doesn't work out I'll definitely look into improving the constructor this way.

I had a lot of issues trying to overload the constructor of a case-class so I could use a regex, so to make progress I added the very ugly kludge to make some progress. My recollection was that the constructor for case-classes had to start with a call to `this(...)`, but I might be mistaken.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,140 @@
...
+  params: Stack.Params,
+  stats: StatsReceiver
+)(implicit val timer: Timer) {
+  private[this] val objectMapper = new ObjectMapper() with ScalaObjectMapper
+  objectMapper.registerModule(DefaultScalaModule)
+  objectMapper.setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE)
+
+  private[this] val numVersions = stats.counter("metadata_versions")
+
+  private[this] val maxWait = 30
```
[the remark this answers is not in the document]
fangel: Probably a good idea. Since the long-polling API isn't even documented from Rancher's side, I have no idea how long of a wait it actually supports - so making it configurable makes a lot of sense. I'll do that.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
fangel: Are there any rules for how many imports you can do from each package before switching to `xyz._`?

Right now I explicitly name 8 different classes from `com.twitter.utils` because I wanted to actually know where each class came from as I was figuring out how stuff worked. But I guess I could just as well import `com.twitter.utils._`

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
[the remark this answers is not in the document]
hawkw: I think that for a `case class`, rather than providing a constructor, you probably want a [companion object](https://twitter.github.io/scala_school/basics2.html#object) with an `apply()` method.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
fangel: Yes - that `.transform` seems to do exactly the same as `.liftToTry.flatMap`, just in a way more concise way. Thanks!

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherNamer.scala
fangel: Clearly, I need to get better about fully understanding `flatMap`, because I think you are right.

Right now I use the `unlift`-method to turn `toAddr` into a partial function that only returns something when it returns `Some(address)`, and then `collect` to do a filter/map-combo which maps the partial function and filters out all the ones where it wasn't defined (i.e. the ones that would have had a `None` returned from `toAddr`.

I thought my way was fairly elegant – but I agree that `flatMap` aught to do the same and in a even shorter way. I just hadn't quite gotten that `Option` is traversable too, and returns either no or a single element, which can then be flattened into desired list too.

fangel:
Thank you so much for your suggestions – they've been great! I'm slowly working my way through all of them – right now I still have to address the input about the long-polling loop in the `RancherClient`-class (i.e., the stuff about modifying it into a `@volatile var v: Option[String]` etc).

But I've successfully added a custom deserialiser which can verify the format of the port-binding info before constructing an instance of the case-class. I've also added a unit-test to ensure that the parser actually understands the JSON returned from the API.

Once I've addressed the last few comments, and run the corrected code on our Rancher-cluster, I'll push all the changes up for you to see.

fangel:
Okay, I've pushed my changes. Please take a look - I've tried to address all of your comments (except making the api-endpoint configurable - see the line-comment above for more info).

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
adleong: Personally, I usually wildcard it after about 5 imports in the same package.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,149 @@
...
+      val request = http.Request(http.Method.Get, "/2015-12-19/containers")
+      request.host = "rancher-metadata"
+      request.accept = "application/json"
+
+      service(request)
+        .map { _.getContentString() }
+        .map { RancherParser().readValue[List[RancherContainer]] }
+        .onSuccess { containers =>
+          log.debug("received %s new containers", containers.size)
+        }: Future[List[RancherContainer]]
```
adleong: you shouldn't nee this type annotation

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherInitializer.scala
```diff
@@ -0,0 +1,35 @@
+package io.buoyant.namer.rancher
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.{Path, Stack, param}
+import io.buoyant.namer.{NamerConfig, NamerInitializer}
```
adleong: alphasort :)

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  private[this] val version: Activity[String] =
+    Activity(Var.async[State[String]](Activity.Pending) { state =>
+      val done = new AtomicBoolean(false)
+      @volatile var v: Option[String] = None
+
+      Future.whileDo(!done.get) {
+        log.trace("starting version request")
+
+        val request = http.Request(http.Method.Get, s"/2015-12-19/version?wait=true&maxWait=${maxWait}&value=${v}")
```
fangel: Just leaving this here until I have the time to fix it: `${v}` needs to be updated after `v` changed from `String` to a `Option[String]` - so probably something like `${v.getOrElse('initial')}`.

namer/rancher/src/main/scala/io/buoyant/namer/rancher/RancherClient.scala
fangel: You are very right. While trying to figure out how stuff worked, I found it useful to add in extra typo annotations to make sure my assumptions about types held up...

fangel:
A belated Merry Christmas / Happy Holidays to you guys and girls! I've addressed the latest comments now, so please take another look at this now.

I also found and removed a rather unfortunate equivalent-to-infinite-loop situation stemming from the `String` to `Option[String]` refactoring in the long-polling. Basically it sent the wrong version name (it sent `Some([version])` leading it to instantly return with the same version. This would just make it repeat itself over and over again. Our ops-team was a little unhappy with the amount of log-data that generated. Oops. But that's fixed now.

The CI tests failed on one of the two builds on an error in the _NetworkedInteropTest.cancel_after_begin_ which I don't quite understand, and I haven't been anywhere near that. So if anyone has any hints as to why that could be, that would be much appreciated.

hawkw:
@fangel `NetworkedInteropTest.cancel_after_begin` is one of a small handful of tests that sometimes fail on CI for reasons we haven't been able to pin down yet; so that's probably spurious. I'll restart the build, and we'll review your most recent changes. Thanks again!

PR #1742, opened 2017-12-11
Add a backoff retry timer to the ConsulDtabStore observer

deebo91:
The Consul namer storage plugin retries a request to a Consul API instance when it receives a response error response that is not 4XX. In this case the namer retries infinitely, which is intended functionality. However, each retry is done in rapid succession with a log message printed right before each retry. This causes the namerd logs to grow rapidly. This PR adds the ability to exponentially increase the amount of time between each successive retry so that log sizes are more manageable. 

One thing I am not sure about is the The `Backoff` parameter for the Consul API in [`KvApi.scala`](https://github.com/linkerd/linkerd/blob/a9c1627d389fab206faac44d1f674695207ff5cd/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala#L18). right now it is set to `Backoff.exponentialJittered(1.milliseconds, 5.seconds)`. The 5 seconds is the max duration the API will wait to retry. This still feels a little to short of a window so I am open to feedback on what duration would be appropriate in this scenario.


namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -14,19 +15,23 @@ class ConsulDtabStore(
...
 
   private[this] val log = Logger.get("consul")
+  private[this] implicit val timer = _timer
 
   override val list: Activity[Set[Ns]] = {
     def namespace(key: String): Ns = key.stripPrefix("/").stripSuffix("/").substring(root.show.length)
 
     val run = Var.async[Activity.State[Set[Ns]]](Activity.Pending) { updates =>
       @volatile var running = true
+      var retPending: Future[_] = Future.Unit
```
adleong: unused

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -38,19 +43,20 @@ class ConsulDtabStore(
...
               case Throw(e: NotFound) =>
                 updates() = Activity.Ok(Set.empty[Ns])
-                cycle(e.rsp.headerMap.get(Headers.Index))
+                cycle(e.rsp.headerMap.get(Headers.Index), backoffs)
               case Throw(e: Failure) if e.isFlagged(Failure.Interrupted) => Future.Done
               case Throw(e) =>
                 updates() = Activity.Failed(e)
                 log.error("consul ns list observation error %s", e)
-                cycle(None)
+                val sleep #:: backoffs0 = backoffs
```
adleong: The convention is usually that `backoffs0` represents the prior backoffs and `backoffs1` represent the next backoffs.  Alternatively, `rest` or `tail` is sometimes used to denote the tail of a stream.  So this could be
`val sleep #:: backoffs1 = backoffs0` or `val sleep #:: rest = backoffs`.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -38,19 +43,20 @@ class ConsulDtabStore(
...
+                Future.sleep(sleep).flatMap(_ => cycle(None, backoffs0))
```
adleong: TIOLI: flatMap is perfectly fine here but there is also `Future.sleep(sleep).before(cycle(None, backoffs0))` that is equivalent.  Take your pick.

deebo91:
@adleong after changing the ConsulConfig to add a configurable backoff setting, I realized that ConsulConfig is used for the consul Namer plugin in linkerd. This change is for the Namerd Dtabstore plugin. Do we need to also change the Namer plugin in Linkerd to reflect this behavior?

adleong:
It would probably be a good idea for the consul namer to backoff on errors too.  But that doesn't have to be part of this change.

deebo91:
I was able to successfully test the config value by providing a max backoff of 1ms. The config was able to override the default value. Would love feedback on the config field name. I tried to make sure it made sense but also didn't to have a really long field name.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreInitializer.scala
```diff
@@ -16,14 +16,16 @@ case class ConsulConfig(
   datacenter: Option[String] = None,
   readConsistencyMode: Option[ConsistencyMode] = None,
   writeConsistencyMode: Option[ConsistencyMode] = None,
-  failFast: Option[Boolean] = None
+  failFast: Option[Boolean] = None,
+  maxBackoffDurationMs: Option[Int] = None
```
adleong: This should be an `Option[BackoffConfig]` instead of an `Option[Int]`.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -14,19 +15,22 @@ class ConsulDtabStore(
   root: Path,
   datacenter: Option[String] = None,
   readConsistency: Option[ConsistencyMode] = None,
-  writeConsistency: Option[ConsistencyMode] = None
+  writeConsistency: Option[ConsistencyMode] = None,
+  _timer: Timer = DefaultTimer
```
adleong: I think you can do `implicit val timer: Timer = DefaultTimer` here

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -9,13 +9,14 @@ import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
 import com.twitter.util._
 
 object KvApi {
-  def apply(c: Client): KvApi = new KvApi(c, s"/$versionString")
+  val DefaultMaxBackOffDuration = 10000
+  def apply(c: Client, backoffMs: Int = DefaultMaxBackOffDuration): KvApi = new KvApi(c, s"/$versionString", Backoff.exponentialJittered(1.milliseconds, backoffMs.milliseconds))
```
adleong: lets have this accept a `Stream[Duration]`

namerd/core/src/main/scala/io/buoyant/namerd/BackoffConfig.scala
```diff
@@ -0,0 +1,26 @@
+package io.buoyant.namerd
```
adleong: rm this file, I think

deebo91: oops that I deleted it. thanks

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
```diff
@@ -9,13 +9,13 @@ import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
 import com.twitter.util._
 
 object KvApi {
-  def apply(c: Client): KvApi = new KvApi(c, s"/$versionString")
+  def apply(c: Client, backoff: Stream[Duration]): KvApi = new KvApi(c, s"/$versionString", backoff)
```
hawkw: Should the default value removed from the constructor be added here? (i.e, 
```scala
def apply(
  c: Client,
  backoff: Stream[Duration] = Backoff.exponentialJittered(1.milliseconds, 5.seconds)
): KvApi = new KvApi(c, s"/$versionString", backoff)
```

hawkw: this would also make it no longer necessary to change all the tests in `KVApiTest`...

linkerd/core/src/main/scala/io/buoyant/linkerd/SvcConfig.scala
```diff
@@ -1,13 +1,12 @@
...
-import com.fasterxml.jackson.annotation.{JsonIgnore, JsonProperty, JsonSubTypes}
+import com.fasterxml.jackson.annotation.{JsonIgnore, JsonProperty}
 import com.twitter.conversions.time._
 import com.twitter.finagle.{param, Stack}
 import com.twitter.finagle.buoyant.TotalTimeout
 import com.twitter.finagle.service._
 import com.twitter.finagle.buoyant.ParamsMaybeWith
-import com.twitter.util.Duration
-import io.buoyant.config.PolymorphicConfig
+import io.buoyant.namer.BackoffConfig
```
hawkw: I'm not sure if I'm a fan of having this live in the namer package, since I'm pretty sure the backoff config is used in places other than in namers? I think maybe it should be moved someplace where it's visible to the namer module, but not necessarily _into_ that package?

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreInitializer.scala
```diff
@@ -45,6 +50,7 @@ case class ConsulConfig(
 object ConsulConfig {
   val DefaultHost = "localhost"
   val DefaultPort = Port(8500)
+  val DefaultBackoff = Backoff.decorrelatedJittered(1.millis, 1.minute)
```
hawkw: Was the default backoff (for `KvApi`) not previously `exponentialJittered` rather than `decorrelatedJittered`? Is there a motivation behind this change?

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
adleong: I like that the default is defined in one place only.

linkerd/core/src/main/scala/io/buoyant/linkerd/SvcConfig.scala
adleong: Not sure there exists a package that's a perfect fit.  It needs to be visible from both linkerd and namerd and it depends on finagle and config.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreInitializer.scala
adleong: BackoffConfig only produces `decorrelatedJittered` and `constant`.  Never `exponentialJittered`.  This is more consistent with elsewhere in linkerd.

consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
hawkw: Yeah, I saw that after scrolling through the rest of the PR; disregard that.

PR #1744, opened 2017-12-12
K8s namer should retry watches on 4XX responses.

### LINKED ISSUE #1741: If namerd receives a 4xx error from k8s, the request is never retried
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
If namerd/linkerd receives a `4xx` response from the k8s apiserver, the request is never retried and its cached set of endpoints can become stale.

The request to be retried, similarly to how a `5xx` would be.

We're still pinning down under what exact conditions the apiserver could return a `4xx` error, but it's likely that it's possible (for example, if the service account token used to authenticate expires.) In any case, it seems like these requests _should_ be retried similarly to a `5xx` would.

[obeattie/linkerd-1741-repro](https://github.com/obeattie/linkerd-1741-repro) contains a minimal setup to reproduce this:

1. Unpack the tar file
2. In a terminal, start the included Go program:

    ```
    $ go run main.go
    ```

    This program mocks out a (very dumb) k8s apiserver, and decides how to respond based on the contents of `responses.txt`. In the way it is configured, it will return with `200` to the initial endpoint list request for the `foo` service, but respond with a `400` to the watch request that namerd subsequently makes.
3. Start namerd (for obvious reasons I haven't included the namerd binary in the tar file):

    ```
    $ namerd-1.3.3-exec ./k8s-repro/namerd.yaml
    ```
4. Make a request to `namerd` via its HTTP interface to resolve `/api/1/resolve/ns?path=/http/foo` and observe that it returns the endpoint set successfully:

    ```
    $ curl -v "http://localhost:4102/api/1/resolve/ns?path=/http/foo"
    *   Trying ::1...
    * TCP_NODELAY set
    * Connection failed
    * connect to ::1 port 4102 failed: Connection refused
    *   Trying 127.0.0.1...
    * TCP_NODELAY set
    * Connected to localhost (127.0.0.1) port 4102 (#0)
    > GET /api/1/resolve/ns?path=/http/foo HTTP/1.1
    > Host: localhost:4102
    > User-Agent: curl/7.54.0
    > Accept: */*
    >
    < HTTP/1.1 200 OK
    < Content-Type: application/json
    < Content-Length: 92
    <
    {"type":"bound","addrs":[{"ip":"1.2.3.4","port":80,"meta":{"nodeName":"node1"}}],"meta":{}}
    ```
5. Observe from the logs of the stubbed apiserver thing that namerd's requests are never retried after the `400` response is returned:

    ```
    $ go run main.go
    2017/12/11 18:16:55 👂  Listening on :5000
    2017/12/11 18:16:59 ➡️  /api/v1/namespaces/ns/endpoints/foo
    2017/12/11 18:16:59     ⬅️  200
    2017/12/11 18:16:59 ➡️  /api/v1/watch/namespaces/ns/endpoints/foo
    2017/12/11 18:16:59     ⬅️  400
    ```

**Environment**:
- linkerd/namerd version, config files: namerd 1.3.3
- Platform, version, and config files (Kubernetes, DC/OS, etc): Kubernetes
- Cloud provider or hardware configuration: Not relevant

adleong:
Fixes #1741 

If the k8s namer encounters a non-200, non-5xx response from the k8s API on a watch call, it gives up and does not retry.

Modify the watch logic so that we retry with backoff on all unexpected status codes.  Also add http status code stats to the k8s api client such as `namer/#/io.l5d.k8s/client/http/status/4XX`.


k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
```diff
@@ -27,6 +27,7 @@ trait ClientConfig {
   ) = {
     val setHost = new SetHostFilter(getHost, getPort)
     Http.client.withParams(Http.client.params ++ params)
+      .withHttpStats
```
olix0r: are we providing the client with a properly-scoped stats receiver? i assume so?

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -21,6 +22,8 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
 
   protected def backoffs: Stream[Duration]
   protected def stats: StatsReceiver
+  protected implicit val timer = HighResTimer.Default
```
olix0r: is HighResTimer required? is this just the default recommended thing? (used to seeing DefaultTimer.twitter or whatever it was)

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -124,23 +129,18 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
-
           case status =>
-            close.set(Closable.nop)
+            rsp.reader.discard()
             log.debug("k8s failed to watch resource %s: %d %s", path, status.code, status.reason)
-            val exc = Api.UnexpectedResponse(rsp)
-            state.update(Activity.Failed(exc))
-            Future.exception(exc)
+            val sleep #:: backoffs1 = backoffs0
+            Future.sleep(sleep).before(_resourceVersionTooOld(backoffs1))
```
hawkw: Correct me if I'm wrong, but the `_resourceVersionTooOld` function is now being called due to errors other than just the resource version being too old? 🤷

adleong: you are correct

k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
adleong: yup.  final stat names look like `namer/#/io.l5d.k8s/client/http/status/4XX`

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
hawkw: maybe it needs a different name...

PR #1745, opened 2017-12-12
Add shell script for running h2spec against Linkerd

hawkw:
This PR adds a shell script for running [`h2spec`](https://github.com/summerwind/h2spec), a conformance testing tool for HTTP/2, against a Linkerd HTTP/2 proxy. The script starts an `nghttpd` server for testing, starts Linkerd configured to forward traffic to `nghttpd`, and then starts `h2spec` pointed at Linkerd's port. The script will automatically build Linkerd first if there is no executable in `linkerd/target`, and will download the correct h2spec binary for the host OS if it is not present.

Right now, I put the script in the `ci` directory, even though it does not actually run on CI yet.

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ npm-debug.log
 .DS_Store
 
 **/*.iml
+h2spec-*
+h2spec
+l5d-h2spec.log
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,3 @@ npm-debug.log
 .DS_Store
 
 **/*.iml
-h2spec-*
-h2spec
-l5d-h2spec.log
--- a/ci/h2spec.sh
+++ b/ci/h2spec.sh
@@ -47,13 +47,15 @@ else
               H2SPEC_TAR="h2spec_linux_amd64.tar.ghz"
               ;;
         esac
-
+        # cd into /usr/local/bin to install h2spec.
         DIR=$(pwd)
         cd /usr/local/bin
+        # download h2spec.
         H2SPEC_URL="https://github.com/summerwind/h2spec/releases/download/v2.1.0/${H2SPEC_TAR}"
         printf "${SEP} downloading h2spec..."
         wget -O "${H2SPEC_TAR}" "${H2SPEC_URL}" > /dev/null 2>&1 &
         spin
+        # untar.
         tar xf "${H2SPEC_TAR}" > /dev/null &
         spin
         printf " downloading h2spec...done!\n"
@@ -61,11 +63,22 @@ else
         H2SPEC_EXEC="/usr/local/bin/h2spec"
     fi
 fi
-set -e
 
+# test for presence of nghttpd
+set +e
+command -v nghttpd > /dev/null
+NGHTTPD_STATUS=$?
+set -e
+if [[ NGHTTPD_STATUS -ne 0 ]]; then
+    echo "${SEP} nghttpd not found; please install nghttpd before continuing."
+    exit "${NGHTTPD_STATUS}"
+fi
+
+# find linkerd executable.
 L5D_PATH=$(find . -name 'linkerd-*-exec')
 
 if ! [ -e "${L5D_PATH}" ]; then
+    # if we couldn't find a linkerd-exec, build it.
     echo "${SEP} linkerd executable not found!"
     printf "${SEP} building linkerd..."
     ./sbt linkerd/assembly > /dev/null 2>&1 &
@@ -76,7 +89,7 @@ fi
 
 echo "${SEP} found linkerd executable: $L5D_PATH"
 
-which nghttpd > /dev/null | sed "s/^/${SEP} /"
+# start nghttpd.
 nghttpd 8080 --no-tls 2>&1 &
 NGHTTPD_PID=$!
 echo "${SEP} started nghttpd"
--- a/ci/h2spec.sh
+++ b/ci/h2spec.sh
@@ -4,7 +4,7 @@ set -o pipefail
 shopt -s extglob
 
 SEP="$(tput bold)⣷$(tput sgr0)"
-LOGFILE="l5d-h2spec.log"
+LOGFILE=$(mktemp -t l5d-h2spec.log)
 SPIN='⣾⣽⣻⢿⡿⣟⣯⣷'
 
 spin() {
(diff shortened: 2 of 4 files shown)


ci/h2spec.sh
```diff
@@ -0,0 +1,95 @@
...
+          ;;
+        linux*)
+          echo "${SEP} detected OS as linux"
+          H2SPEC_FILE="h2spec_linux_amd64.tar.ghz"
+          ;;
+
+    esac
+    H2SPEC_URL="https://github.com/summerwind/h2spec/releases/download/v2.1.0/${H2SPEC_FILE}"
+    printf "${SEP} downloading h2spec..."
+    wget "${H2SPEC_URL}" > /dev/null 2>&1 &
```
siggy: on my system i needed to to `wget -O $H2SPEC_FILE...` otherwise i end up with crazy long filenames with `Amz-Credential`'s in it.

ci/h2spec.sh
```diff
@@ -0,0 +1,95 @@
...
+
+until [ -e "${L5D_PATH}" ]; do
+    echo "${SEP} linkerd executable not found!"
+    ./sbt linkerd:compile | sed "s/^/${SEP} sbt ${SEP} /"
+    L5D_PATH=$(find . -name 'linkerd-*-exec')
+done
+
+echo "${SEP} found linkerd executable: $L5D_PATH"
+
+nghttpd 8080 --no-tls 2>&1 &
```
siggy: check for `nghttpd` on the system?

ci/h2spec.sh
```diff
@@ -0,0 +1,95 @@
...
+          echo "${SEP} detected OS as linux"
+          H2SPEC_FILE="h2spec_linux_amd64.tar.ghz"
+          ;;
+
+    esac
+    H2SPEC_URL="https://github.com/summerwind/h2spec/releases/download/v2.1.0/${H2SPEC_FILE}"
+    printf "${SEP} downloading h2spec..."
+    wget "${H2SPEC_URL}" > /dev/null 2>&1 &
+    spin
+    tar xf "${H2SPEC_FILE}" > /dev/null 2>&1 &
```
siggy: this swallowed an error, as `$H2SPEC_FILE` was not present.

ci/h2spec.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+set -eu
+set -o pipefail
+shopt -s extglob
+
+SEP="$(tput bold)⣷$(tput sgr0)"
+LOGFILE="l5d-h2spec.log"
```
siggy: consider `LOGFILE =$(mktemp -t l5d-h2spec.log)`, to avoid gitignore and overwriting.

ci/h2spec.sh
```diff
@@ -0,0 +1,95 @@
...
+# run h2spec against linkerd, printing linkerd's logs if h2spec failed.
+set +e
+./h2spec -p 4140
+H2SPEC_STATUS=$?
+set -e
+if [ "${H2SPEC_STATUS}" -eq 0 ]; then
+    echo "${SEP} h2spec passed!"
+else
+    echo "${SEP} h2spec failed! linkerd logs:"
+    cat "${LOGFILE}"
```
siggy: I think I'd just prefer `echo "${LOGFILE}"`, as the logs are so noisy, and what I really want to see is the h2spec output.

ci/h2spec.sh
```diff
@@ -0,0 +1,95 @@
...
+    while [ "$(ps a | awk '{print $1}' | grep $pid)" ]; do
+        i=$(( (i+1) %8 ))
+        printf "\r$(tput bold)${SPIN:$i:1}$(tput sgr0)"
+        sleep .1
+    done
+}
+
+echo "${SEP} running h2spec against linkerd..."
+# if we can't find h2spec, download it
+if ! [ -e "h2spec" ] ; then
```
siggy: consider checking for `which h2spec`, and, if not found, put it into `~/bin`, and let the user know it's there. this also avoid gitignore foo.

hawkw: yeah, was considering that - will do.

ci/h2spec.sh
```diff
@@ -0,0 +1,95 @@
...
+    tar xf "${H2SPEC_FILE}" > /dev/null 2>&1 &
+    spin
+    printf "\n"
+fi
+
+L5D_PATH=$(find . -name 'linkerd-*-exec')
+
+until [ -e "${L5D_PATH}" ]; do
+    echo "${SEP} linkerd executable not found!"
+    ./sbt linkerd:compile | sed "s/^/${SEP} sbt ${SEP} /"
```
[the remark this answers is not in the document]
hawkw: WHOOPS my bad, I forgot the task name

hawkw:
@siggy I've made all the changes you've requested except for the docs change --- do you think that should go in `README.md` or `BUILD.md`?

ci/h2spec.sh
```diff
@@ -0,0 +1,129 @@
...
+set +e
+WHICH_H2SPEC=$(which h2spec)
+if [[ $? -eq 0 ]]; then
+    # if which found an h2spec executable, just return that path.
+    H2SPEC_EXEC="${WHICH_H2SPEC}"
+    set -e
+else
+    set -e
+    # which didn't find a h2spec on the path. check if it exists in
+    # the working dir, and if not, dow nload it.
```
siggy: `dow nload`

ci/h2spec.sh
```diff
@@ -0,0 +1,129 @@
...
+    if [ -e "h2spec" ] ; then
+        # h2spec was found in the working dir, use that.
```
siggy: i'd say forego this step, as i would not expect h2spec to be in the linkerd dir.

ci/h2spec.sh
```diff
@@ -0,0 +1,129 @@
...
+              H2SPEC_TAR="h2spec_linux_amd64.tar.ghz"
+              ;;
+        esac
+        # cd into /usr/local/bin to install h2spec.
+        DIR=$(pwd)
+        cd /usr/local/bin
+        # download h2spec.
+        H2SPEC_URL="https://github.com/summerwind/h2spec/releases/download/v2.1.0/${H2SPEC_TAR}"
+        printf "${SEP} downloading h2spec..."
+        wget -O "${H2SPEC_TAR}" "${H2SPEC_URL}" > /dev/null 2>&1 &
```
siggy: i'd put the tar into a temp dir so it gets cleaned up:
```bash
TMPDIR=$(mktemp -d -t h2spec)
wget -O "${TMPDIR}/${H2SPEC_TAR}" ...
```

ci/h2spec.sh
```diff
@@ -0,0 +1,129 @@
...
+              H2SPEC_TAR="h2spec_darwin_amd64.tar.gz"
+              ;;
+            linux*)
+              echo "${SEP} detected OS as linux"
+              H2SPEC_TAR="h2spec_linux_amd64.tar.ghz"
+              ;;
+        esac
+        # cd into /usr/local/bin to install h2spec.
+        DIR=$(pwd)
+        cd /usr/local/bin
```
siggy: recommend putting `h2spec` into `~/bin`, as it's less invasive, and we may not have perms to for `/usr/local/bin`.

.gitignore
```diff
@@ -13,3 +13,6 @@ npm-debug.log
 .DS_Store
 
 **/*.iml
+h2spec-*
+h2spec
+l5d-h2spec.log
```
siggy: should be able to remove all changes to .gitignore.

ci/h2spec.sh
hawkw: is `~/bin` always on the path?

siggy: it's not necessarily, but if you put it there, you can at least reference it directly.

hawkw:
that's fair

ci/h2spec.sh
hawkw: hmm – and forgive me if i'm entering into bikeshed territory here – it seems to me that if `~/bin` isn't necessarily on the path, we'll need to either add it to the path, or add a special case to check whether `h2spec` is installed in that location?

siggy: sure, i don't think we want to modify the path, but checking for that install location seems reasonable.

hawkw: yeah, i didn't think we'd want to have the script modify the path either, but seems like checking if it exists in `~/bin` complicates this logic a bit... happy to add that though

siggy: no more than checking for `./h2spec` ;)

siggy:
right, `BUILD.md` looks like an appropriate place.

Issue #509, opened 2016-06-28
Support LB'ing external IIS-hosted virtual hosts


garysweaver:
Our team has two external IIS-hosted virtual hosts that we wanted to EWMA load balance via Linkerd, but unfortunately that won't work currently for us without a lot of trouble.

While the mapping of logical name to concrete is working as we expected, the destination service (on the IIS nodes) is expecting a specific Host: header, e.g. "node1.domain.com" for node1.domain.com and "node2.domain.com" for node2.domain.com. However, Linkerd instead just passes the logical-name on as the Host header. We can't just specify the node name in the logical-name, because then a request with Host header "node1.domain.com" might get routed to node2.domain.com which is expecting Host header "node2.domain.com" (since Linkerd is trying to LB those nodes).

@olix0r said that getting this to work would be tricky to accomplish at the moment, but that Linkerd could perhaps find a way to add a filter into the client stack that modifies Host values based on the destination address similar to how TLS naming works in Linkerd to try to address this use case.

wmorgan:
Yeah, this is a bit outside the current use case, where we assume that load-balancing happens over a pool of identical replicas.

As a short-term hack, would it make sense to introduce something like nginx to rewrite those  headers? Not a great solution, but might be enough for testing purposes to see if EWMA buys you a performance gain.

garysweaver:
@wmorgan Thanks! We’re looking into alternate solutions for now until this can be fixed.

aclemmensen:
Couldn't you just add additional bindings to the site in IIS? Make both servers bind the site to the same name and it should work, right? Except if you can't control what `Host` header Linkerd will use when forwarding the requests as this will have to match the binding you are configuring.

wmorgan:
@garysweaver this ticket is a bit old... are you still interested in pursuing this? I would like to talk to you a bit offline if so.

olix0r:
Closing this for now.  Please feel free to reopen this if you think it's worth pursuing.

PR #1746, opened 2017-12-13
Set Host header to concrete resolved name

aidansteele:
This addresses the issue seen in #509. Right now when a request is proxied through linkerd, the `Host` header remains unchanged from what the client used. This can cause requests to fail when the request is forwarded to a backend that uses Host header-based virtual hosting.

The example in #509 is IIS, but this functionality is coming back into vogue with Amazon's Application Load Balancer service [now supporting host based routing](https://aws.amazon.com/blogs/aws/new-host-based-routing-support-for-aws-application-load-balancers/). 

As an aside, this is probably the worst code ever submitted to the linkerd project - it's my first Scala ever. I'm very happy to change it in response to any advice, also happy to have it completely rewritten by someone who actually knows what they're doing.

Thanks!

### DIFF
--- a/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
+++ b/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
@@ -6,7 +6,7 @@ import com.twitter.finagle.context.{Contexts, Deadline => FDeadline}
 import com.twitter.finagle.http._
 import com.twitter.finagle.tracing._
 import com.twitter.util.{Future, Return, Throw, Time, Try}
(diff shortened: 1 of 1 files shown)


adleong:
Hi @aidansteele, thanks for the pull request!  Unfortunately, I don't think we'll be able to merge this.  It's essential to Linkerd's routing model that the Host header not be changed.  Consider the following common scenario: in Linker-to-Linker mode, both the incoming and outgoing Linkerds must see the same Host header so that they can route consistently.  Having the first Linkerd mutate the Host header would violate this.

We're definitely interested in supporting ALB or IIS style host based routing though.  I think the best way to proceed would be to open an issue with as much detail as possible about the requirements in order to make this work and we can brainstorm together about a way to fit that into Linkerd's existing routing model.

Let me know what you think!

PR #1711, opened 2017-11-22
Fix netty ByteBuf reference counting

siggy:
Netty was reporting "LEAK: ByteBuf.release() was not called", direct
memory was growing unbounded.

Remove unnecessary call to retain in h2 code.

Verified netty leak messages go away with this fix.

Fixes #1678

Issue #1678, opened 2017-10-19
finagle/netty4-7: LEAK: ByteBuf.release() was not called before it's garbage-collected


klingerf:
I'm running the latest linkerd master (614811730d88f40c4ee8bc507b9f250789517035) in Kubernetes 1.7.8, and have started seeing warnings in the linkerd logs:

```
ERROR 1019 21:54:39.902 UTC finagle/netty4-7: LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.
```

In my environment, I'm running the [gob app](https://github.com/linkerd/linkerd-examples/tree/master/gob#the-application-gobs-microservice), with a sidecar h2 router configuration that looks like this:

```
    routers:
    - protocol: h2
      experimental: true
      label: outgoing
      servers:
      - port: 5252
        ip: 0.0.0.0
      client:
        tls:
          commonName: linkerd
          trustCerts:
          - /io.buoyant/linkerd/certs/cacertificate.pem
      dtab: |
        /host/proto.GenSvc => /srv/gen2;
        /host/proto.WordSvc => /srv/word2;
      identifier:
        kind: io.l5d.header.path
      interpreter:
        kind: io.l5d.namerd.http
        experimental: true
        dst: /$/inet/namerd.test.svc.cluster.local/4180
        namespace: default

    - protocol: h2
      experimental: true
      label: incoming
      servers:
      - port: 6262
        ip: 0.0.0.0
        tls:
          certPath: /io.buoyant/linkerd/certs/linkerdcertificate.pem
          keyPath: /io.buoyant/linkerd/certs/linkerdkey.pk8
      dtab: /svc => /$/inet/127.1/8282
```

The leaks are being reported for requests from the `web` service to the `gen` service, which uses gRPC streaming. Here's the full leak report from the log:

```
ERROR 1019 21:54:39.902 UTC finagle/netty4-7: LEAK: ByteBuf.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.
WARNING: 112 leak records were discarded because the leak record count is limited to 4. Use system property io.netty.leakDetection.maxRecords to increase the limit.
Recent access records: 4
#4:
	io.netty.buffer.AdvancedLeakAwareByteBuf.release(AdvancedLeakAwareByteBuf.java:936)
[...]

PR #1751, opened 2017-12-14
Data frames release the underlying ByteBuf, not the wrapped Buf

### LINKED ISSUE #1728: linkerd 1.3.3 issue with gRPC
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
I updated linkerd from `1.1.2` to `1.3.3` in DCOS 1.9 and started getting the following errors in gRPC clients:

from go client:
```
2017/12/05 17:30:05 rpc error: code = ResourceExhausted desc = grpc: received message larger than max (845559858 vs. 4194304)
```

from .net core client:
```
Unhandled Exception: Grpc.Core.RpcException: Status(StatusCode=Internal, Detail="Failed to deserialize response message.")
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Grpc.Core.Internal.AsyncCall`2.UnaryCall(TRequest msg)
   at Grpc.Core.DefaultCallInvoker.BlockingUnaryCall[TRequest,TResponse](Method`2 method, String host, CallOptions options, TRequest request)
   at Grpc.Core.Internal.InterceptingCallInvoker.BlockingUnaryCall[TRequest,TResponse](Method`2 method, String host, CallOptions options, TRequest request)
```

It looks like the message becomes invalid. The errors don't happen all the time, I need to call that endpoint multiple times. The message itself is not that big - around `40KB`.

I tried running linkerd locally, but couldn't reproduce this issue.

**Environment**:
- linkerd/namerd version, config files: linkerd 1.3.3, no namerd
- Platform, version, and config files (Kubernetes, DC/OS, etc): DCOS Enterprise 1.9
- Cloud provider or hardware configuration: AWS

linkerd configuration:
```
---
usage:
  enabled: false
admin:
  port: 9990
  ip: 0.0.0.0
namers:
- kind: io.l5d.marathon
  host: leader.mesos
  port: 443
  uriPrefix: "/marathon"
  prefix: "/io.l5d.marathon"
  tls:
    commonName: master.mesos
    trustCerts:
    - "/mnt/mesos/sandbox/.ssl/ca.crt"
routers:
- protocol: h2
  experimental: true
  identifier:
    kind: io.l5d.header.token
  dtab: "/ph=>/$/io.buoyant.rinet;/srv=>/$/io.buoyant.porthostPfx/ph;/svc=>/srv;/marathonId=>/#/io.l5d.marathon;/svc=>/$/io.buoyant.http.domainToPathPfx/marathonId"
  servers:
  - port: 4140
    ip: 0.0.0.0

```

hawkw:
The Finagle `ByteBufAsBuf` converter wraps a Netty 4 `ByteBuf`, but does not support reference counting (see [this comment](https://github.com/twitter/finagle/blob/82d00660373582d6bfe56df8155b52528df36691/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/ByteBufAsBuf.scala#L58-L61)). Therefore, when releasing a data frame whose content `ByteBuf` we `retain()` before wrapping it as a `Buf`, we must `release()` the underlying `ByteBuf`. 

This fixes the memory leak observed in #1678 without removing the call to `retain()` which was removed in #1711, which introduced a data corruption issue (#1728). If we `release()` the underlying `ByteBuf` after releasing the data frame, we can once again `retain()` it before converting to `Buf` without leaking, thus avoiding the data corruption caused by removing the `retain()`.

Thanks to @vadimi for helping with this fix!

Fixes #1728.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4Message.scala
@@ -102,10 +102,12 @@ private[h2] object Netty4Message {
 
     def apply(f: Http2DataFrame, updateWindow: Int => Future[Unit]): Frame.Data = {
       val sz = f.content.readableBytes + f.padding
-      val buf = ByteBufAsBuf(f.content)
-      val releaser: () => Future[Unit] =
+      val buf = ByteBufAsBuf(f.content.retain())
+      val releaser: () => Future[Unit] = {
+        f.content.release()
         if (sz > 0) () => updateWindow(sz)
         else () => Future.Unit
+      }
       Frame.Data(buf, f.isEndStream, releaser)
     }
   }
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4Message.scala
@@ -103,14 +103,18 @@ private[h2] object Netty4Message {
     def apply(f: Http2DataFrame, updateWindow: Int => Future[Unit]): Frame.Data = {
       val sz = f.content.readableBytes + f.padding
       val buf = ByteBufAsBuf(f.content.retain())
-      val releaser: () => Future[Unit] = {
+      val releaser: () => Future[Unit] =
         () =>
           {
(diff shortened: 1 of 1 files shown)


hawkw:
@adleong I completely agree; I have just enough understanding of this change to understand why it might fix some issues, but not enough to feel confident that it will fix _all_ ref-counting issues.

Issue #1730, opened 2017-12-06
Linkerd continues to talk to old endpoint after Kubernetes deployment


dpetersen:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

I upgraded our Linkerd from 0.8.6 (I know, ancient, but we had no complaints) to 1.3.3. We are running a Kubernetes cluster, and this Linkerd receives traffic from the Internet via an ELB and routes to the appropriate services in various staging environments. I updated our configuration, and everything appeared to be working fine.

Here is our configuration. This upgraded linkerd is named `beyond-thunderdome` because the old one was `thunderdome` and it all makes perfect sense if you've watched Mad Max:
```
    admin:
      port: 9990
      # because it's getting hit by k8s for health checks
      ip: 0.0.0.0
    usage:
      enabled: false
    routers:
      - label: in-beyond-thunderdome
        protocol: http
        #httpAccessLog: /dev/stdout # not in production, por favor
        servers:
          - port: 8080
            # on purpose, in-pod nginx sidecar will be able to proxy to
            # localhost
            ip: 127.0.0.1 
            # this clears out linkerd headers that let you manipulate cluster
            # routing. It is ABSOLUTELY CRITICAL that we don't accept these
            # headers from the internet.
            clearContext: true
        interpreter:
          kind: io.l5d.k8s.configMap
          experimental: true # sure, why not
          name: linkerd-dtabs
          filename: beyond-thunderdome
          namespace: default
    namers:
      - kind: io.l5d.k8s
        host: 127.0.0.1
        port: 8001
    telemetry:
      - kind: io.l5d.prometheus
        path: /admin/metrics/prometheus
        prefix: linkerd_
      - kind: io.l5d.recentRequests
        sampleRate: 0.0001
        capacity: 20
```

slimmed-down dtab ConfigMap looks like this:

```apiVersion: v1
kind: ConfigMap
metadata:
  name: linkerd-dtabs
data:
  beyond-thunderdome: |-
    /host/com/example/integrations-dashboard => /#/io.l5d.k8s/integrations-staging/http/integrations-dashboard;
    /svc => /$/io.buoyant.http.domainToPathPfx/host ;
```

[...]

PR #1755, opened 2017-12-18
Fix bug parsing large streaming JSON objects 🐽

obeattie:
Internally `JsonStreamParser` uses a buffered reader to read the streaming HTTP response. It is possible for a single chunk to exceed the size of this buffer. Previously, if one of these reads resulted in an empty chunk (specifically in the case we saw, the read resulted in just `"\n"`) the stream could get “stuck,” perpetually trying to parse an incomplete object.

In this condition the stream never terminates, causing the k8s namer to stop processing updates to an endpoint until the process is restarted. This was turned up by some internal chaos 🐒 testing, and I think it is quite likely to be related to #1730.

In doing this I have refactored some of the code of `JsonStreamParser`, hopefully making it simpler by removing code that, for example, compensated for [now-fixed bugs in `com.twitter.util`](https://github.com/twitter/util/issues/195).

Tests are included for this and some other previously-uncovered cases. 🔋

### DIFF
--- a/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
+++ b/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
@@ -75,7 +75,9 @@ class JsonStreamParser(mapper: ObjectMapper with ScalaObjectMapper) {
     // if reads fail
     val terminatingReader = new Reader {
       def read(n: Int): Future[Option[Buf]] = reader.read(n).handle {
-        case NonFatal(_) => None
+        case NonFatal(e) =>
+          log.debug(e, "Error reading JSON stream")
+          None
       }
       def discard(): Unit = reader.discard()
     }
--- a/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
+++ b/config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
@@ -1,15 +1,14 @@
 package io.buoyant.config
 
+import com.fasterxml.jackson.core.JsonProcessingException
 import com.fasterxml.jackson.core.`type`.TypeReference
-import com.fasterxml.jackson.core.{JsonParser, JsonProcessingException}
 import com.fasterxml.jackson.databind._
 import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
 import com.twitter.concurrent.AsyncStream
 import com.twitter.io.{Buf, Reader}
 import com.twitter.logging.Logger
-import com.twitter.util.{Return, Throw, Try}
+import com.twitter.util.{Future, Try}
 
-import scala.collection.mutable
 import scala.util.control.NonFatal
 
 class JsonStreamParser(mapper: ObjectMapper with ScalaObjectMapper) {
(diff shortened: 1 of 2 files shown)


wmorgan:
Amazing. Thanks for this, @obeattie. We'll take a look.

obeattie:
I think the rest failure is unrelated to this change 🤔

obeattie:
I think the test failure is unrelated to this change 🤔

deebo91:
@obeattie I can try rerunning the CI job to see if we run into the same thing.

hawkw:
@obeattie, this is great! I'll get back to you with more detailed review soon. 😄

config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
```diff
@@ -29,105 +28,66 @@ class JsonStreamParser(mapper: ObjectMapper with ScalaObjectMapper) {
...
-      case Throw(e) =>
-        log.warning(e, "json read error")
-    }.handle {
-      case NonFatal(e) => None
-    }
-
-    AsyncStream.fromFuture(read).flatMap {
-      //Fake None element to get around scanLeft being one behind bug
-      //Tracked in https://github.com/twitter/util/issues/195
-      //Can be removed once 195 is fixed
```
hawkw: Glad this could be removed! 🙌

k8s/src/test/scala/io/buoyant/k8s/JsonTest.scala
```diff
@@ -17,21 +18,22 @@ class JsonTest extends FunSuite {
 
   test("chunked: message") {
     val obj = TestType("foo", Some(TestType("bar")))
-    val (objs, rest) = Json.readChunked[TestType](Json.writeBuf(obj))
-    assert(objs == Seq(obj))
+    val (decoded, rest) = Json.readBuf[TestType](Json.writeBuf(obj))
+    assert(decoded == Seq(obj))
```
hawkw: I think this renaming makes the tests a bit clearer, thanks! 👍

config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
```diff
@@ -29,105 +28,66 @@ class JsonStreamParser(mapper: ObjectMapper with ScalaObjectMapper) {
...
-          }
-        }
-      )
-      .flatMap(s => AsyncStream.fromSeq(s._1))
-  }
+    // Wrap the reader so reads just terminate if a non-fatal exception occurs. This ensures the below AsyncStream ends
+    // if reads fail
+    val terminatingReader = new Reader {
+      def read(n: Int): Future[Option[Buf]] = reader.read(n).handle {
+        case NonFatal(_) => None
```
adleong: What do you think about logging the exception here?

config/src/main/scala/io/buoyant/config/JsonStreamParser.scala
```diff
@@ -29,105 +28,66 @@ class JsonStreamParser(mapper: ObjectMapper with ScalaObjectMapper) {
...
+      val o = parser.getCurrentLocation.getByteOffset.toInt
+      (v, o)
+    } catch {
+      case Incomplete() => (None, 0)
+    } finally {
+      parser.close()
     }
 
-    val rest =
-      if (offset >= chunk.length) Buf.Empty
```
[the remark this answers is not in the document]
obeattie: Given the offset is only ever updated to the parser's last location, I don't think this is possible. Even if it were, `Buf.slice`'s doc states that:
>Out of bounds indices are truncated.

adleong:
Hey @obeattie and happy new year!  I'm excited about this fix and I'd like to move forward but I could use some more help understanding the bug that's being fixed here (see my previous comment).

obeattie:
@adleong Sorry, I thought I had already responded to this. If you run the `stream: empty chunk reads` test that I added against the old version of the code, you will see that it fails – specifically, the stream stalls and the test hangs. The new version of the code fixes this.

The underlying reason for this is the old `Incomplete` extractor wouldn't match an exception that was thrown by Jackson: "No content to map due to end-of-input" because it expected such messages to _begin_ "Unexpected end-of-input". This may have worked better in an older version of Jackson but I think it's likely that this always happened, it was just hard to trigger.

We caught this "in the wild" where a HTTP chunk would be terminated with a newline (something that k8s does), with the newline positioned in such a way that it would be read in a separate read operation due to the [buffer size](https://github.com/monzo/linkerd/blob/8f6ae461480fcd5462e87b8fbb88199cbeef9251/k8s/src/main/scala/io/buoyant/k8s/Api.scala#L14). When this case was hit the stream would never recover and no further endpoint updates would be processed.

The rest of the refactor is, in a sense, window dressing, but when I was tracking this down I found the old code a little hard to follow; I think the refactor simplifies things.

adleong:
Ah, makes sense.  I do think your refactors make the code more clear and increase the code quality level.  However, since we currently have active investigations into potential bugs with the k8s namer, I think I'd prefer to make the smallest change possible that fixes the bug.  Even though the code looks solid to me, I'm afraid of changing too many things at a time.

PR #1756, opened 2017-12-19
Bump base Docker image to 8u151, add jdk configs

siggy:
Linkerd is currently built on openjdk:8u131-jre, which is deprecated,
and does not support jmap for heap dumps.

Update Linkerd to use openjdk:8u151-jre as a base image, and add a jdk
config for heap dumps. These new images will appear as
buoyantio/linkerd:1.3.4-jdk and buoyantio/namerd:1.3.4-jdk.

Signed-off-by: Andrew Seigner <andrew@sig.gy>

### DIFF
--- a/project/Base.scala
+++ b/project/Base.scala
@@ -153,7 +153,7 @@ class Base extends Build {
     assemblyJarName in assembly := s"${name.value}-${version.value}-${configuration.value}-exec",
     docker := (docker dependsOn (assembly in configuration)).value,
     dockerEnvPrefix := "",
-    dockerJavaImage := (dockerJavaImage in Global).?(_.getOrElse("openjdk:8u131-jre")).value,
+    dockerJavaImage := (dockerJavaImage in Global).?(_.getOrElse("openjdk:8u151-jre")).value,
     dockerfile in docker := new Dockerfile {
       val envPrefix = dockerEnvPrefix.value.toUpperCase
       val home = s"/${organization.value}/${name.value}/${version.value}"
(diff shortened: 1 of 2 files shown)


project/Base.scala
```diff
@@ -153,7 +153,7 @@ class Base extends Build {
     assemblyJarName in assembly := s"${name.value}-${version.value}-${configuration.value}-exec",
     docker := (docker dependsOn (assembly in configuration)).value,
     dockerEnvPrefix := "",
-    dockerJavaImage := (dockerJavaImage in Global).?(_.getOrElse("openjdk:8u131-jre")).value,
+    dockerJavaImage := (dockerJavaImage in Global).?(_.getOrElse("openjdk:8u151-jre")).value,
```
adleong: TIOLI: could factor the "8u151" version number into a constant to avoid repetition and ensure they stay in sync.

klingerf:
@siggy I noticed that this change bumps up the image size on the non-jdk images. Is that expected? From testing locally, it looks like it adds about 230mb to each image.

siggy:
@klingerf this is unfortunately due to the base image upgrade from `8u131-jre` to `8u151-jre`.

```
$ docker images|grep jre
openjdk                     8u151-jre            c49bf7000580        3 weeks ago         538 MB
openjdk                     8u131-jre            02d99451775e        5 months ago        275 MB
```

these are the recommended base images from https://hub.docker.com/r/library/openjdk/. however, we could evaluate the "slim" or "alpine" variants if needed.

klingerf:
@siggy Thanks for the explanation -- makes sense to me. I didn't realize that the base had bumped up so much.

Issue #1731, opened 2017-12-06
Exception when using admin UI to inspect dtab delegations


obeattie:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**: When using the admin UI to perform dtab delegations, this often seems to throw an error in our environment:

```W 1206 10:20:26.907 UTC THREAD27: Exception propagated to the default monitor (upstream address: /10.224.49.0:49572, downstream address: n/a, label: adminhttp).
DelegationFailure(The maximum number of concurrent observations has been reached (10))
	at io.buoyant.namerd.iface.thriftscala.DelegationFailure$.decode(DelegationFailure.scala:122)
	at io.buoyant.namerd.iface.thriftscala.Namer$Delegate$Result$.readDfValue(Namer.scala:1397)
	at io.buoyant.namerd.iface.thriftscala.Namer$Delegate$Result$.decode(Namer.scala:1341)
	at io.buoyant.namerd.iface.thriftscala.Namer$Delegate$Result$.decode(Namer.scala:1245)
	at io.buoyant.namerd.iface.thriftscala.Namer$FinagleClient.decodeResponse(Namer$FinagleClient.scala:79)
	at io.buoyant.namerd.iface.thriftscala.Namer$FinagleClient.$anonfun$delegate$1(Namer$FinagleClient.scala:233)
	at com.twitter.finagle.thrift.DeserializeCtx.deserialize(DeserializeCtx.scala:38)
	at com.twitter.finagle.thriftmux.service.ThriftMuxResponseClassifier$$anon$2.deserializeIfPossible(ThriftMuxResponseClassifier.scala:167)
	at com.twitter.finagle.thriftmux.service.ThriftMuxResponseClassifier$$anon$2.isDefinedAt(ThriftMuxResponseClassifier.scala:177)
	at com.twitter.finagle.thriftmux.service.ThriftMuxResponseClassifier$$anon$2.isDefinedAt(ThriftMuxResponseClassifier.scala:155)
	at scala.PartialFunction.applyOrElse(PartialFunction.scala:123)
	at scala.PartialFunction.applyOrElse$(PartialFunction.scala:122)
	at com.twitter.finagle.thriftmux.service.ThriftMuxResponseClassifier$$anon$2.applyOrElse(ThriftMuxResponseClassifier.scala:155)
	at com.twitter.finagle.service.StatsFilter.$anonfun$apply$1(StatsFilter.scala:175)
[...]

PR #1757, opened 2017-12-19
Allow delegate call to continue as one-shot when delegationCache is full

jackkleeman:
This is a simple workaround for #1731; the delegationCache being full shouldn't cause delegate to fail. It might be preferable to simply not use an Observer for this request, as discussed on slack [here](https://linkerd.slack.com/archives/C0JV5E7BR/p1507222184000355) but I'm not experienced enough with the codebase to determine whether that's a better option.

@obeattie


deebo91:
@jackkleeman thanks for the PR, we will take a look and provide any feedback if necessary!

PR #1758, opened 2017-12-20
Add h2load to test script

hawkw:
Test script now runs `h2load` as well as `h2spec`, controllable with command-line args. Also, UI improvement; fixed a bug where the Linkerd executable was not found correctly when there are multiple linkerd executables in `linkerd/target`.


hawkw:
@siggy how do you feel about changing the name of the script to `h2.sh`? i thought it was reasonable because the subcommands are now `load` and `spec`; but wasn't sure if that seemed clear to everyone else?

siggy:
@hawkw `h2.sh` seems fine to me.

siggy:
@hawkw that reminds me, update `BUILD.md`

PR #1760, opened 2017-12-20
Improve handling of dtab parse errors in ConsulDtabStore

### LINKED ISSUE #1759: Namerd request to Consul "times out" if DTab stored in Consul is malformed
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:

When requesting DTabs from Consul, if Namerd has not already cached a valid DTab configuration, Namerd will stall (as if it's waiting for data from Consul) if the DTab is malformed.  The logs do not display a warning that the DTab is malformed and the Namerd Admin console hangs waiting for a response from the Namerd node.

```
E 1220 00:26:45.879 UTC THREAD27: adminhttp
com.twitter.finagle.CancelledRequestException: request cancelled. Remote Info: Not Available
```

I expect an error in the Namerd logs saying it can't parse the DTab and any request via the Admin console to return that error.

1.  Create a Consul cluster.
2.  Create a Consul KV entry with an invalid DTab.
3.  Create a Namerd config that uses Consul Storage via that Consul keyspace
4.  Start Namerd.
5.  Go to Namerd Admin console.  Attempt to access the DTab.

**Environment**:

- This has occurred since we first started using Namerd (0.9.x) and is still present in the latest version (1.3.4)
- Running everything AWS ECS, with Consul versions 0.9.3 and 1.0.2 (tried both).  However, we've experienced the same problem via Docker Compose.

Here's the Consul info (just in case):

```
/ # consul info -http-addr 10.26.8.35:8500
agent:
    check_monitors = 0
    check_ttls = 0
    checks = 4
    services = 4
build:
    prerelease = 
    revision = 112c060
    version = 0.9.3
consul:
    bootstrap = true
    known_datacenters = 1
    leader = true
    leader_addr = 10.26.8.35:8300
    server = true
raft:
    applied_index = 1479
    commit_index = 1479
    fsm_pending = 0
    last_contact = 0
    last_log_index = 1479
    last_log_term = 2
    last_snapshot_index = 0
    last_snapshot_term = 0
    latest_configuration = [{Suffrage:Voter ID:10.26.8.35:8300 Address:10.26.8.35:8300}]
    latest_configuration_index = 1
    num_peers = 0
    protocol_version = 2
    protocol_version_max = 3
    protocol_version_min = 0
    snapshot_version_max = 1
    snapshot_version_min = 0
    state = Leader
    term = 2
runtime:
    arch = amd64
    cpu_count = 2
    goroutines = 85
    max_procs = 2
    os = linux
    version = go1.9
serf_lan:
    coordinate_resets = 0
    encrypted = false
    event_queue = 0
    event_time = 2
    failed = 0
    health_score = 0
    intent_queue = 0
    left = 0
    member_time = 2
    members = 2
    query_queue = 0
    query_time = 1
serf_wan:
    coordinate_resets = 0
    encrypted = false
    event_queue = 0
    event_time = 1
    failed = 0
    health_score = 0
    intent_queue = 0
    left = 0
    member_time = 1
    members = 1
    query_queue = 0
    query_time = 1

/ # consul members
Node                                                         Address          Status  Type    Build  Protocol  DC                              Segment
master-1                                                     10.26.8.35:8301  alive   server  0.9.3  2         qa-feature-something-something  <all>
qa-services-feature-something-something-i-0fb6030d99d904dc9  10.26.8.67:8301  alive   client  0.9.3  2         qa-feature-something-something  <default>
```

And our Namerd configuration (keep in mind the IP's are generated at deployment time):

```
admin:
  ip: 0.0.0.0
  port: 9991
interfaces:
    ip: 0.0.0.0
    port: 4100
  - kind: io.l5d.httpController
    ip: 0.0.0.0
    port: 4180
namers:
  - kind: io.l5d.consul
    host: 10.26.8.67
    prefix: /consul
    includeTag: true
storage:
  kind: io.l5d.consul
  host: 10.26.8.67
  pathPrefix: /namerd/dtabs
```

hawkw:
Fixes #1759.

Currently, when Namerd's `ConsulDtabStore` makes a request to Consul's KV API that returns a value, it passes that value to `Dtab.read()` to parse it as a dtab. Dtab parsing can fail if the string returned by the KV API is not properly formatted as a dtab; if this is the case, `Dtab.read()` will throw an `InvalidArgumentException`. However, this is not handled in `ConsulDtabStore`; instead, the exception will be thrown upwards. When the dtab observation was opened by the Admin UI's `DtabHandler`, no additional error handling is performed, causing the exception to cancel the request, printing a message like
```
E 1220 00:26:45.879 UTC THREAD27: adminhttp
com.twitter.finagle.CancelledRequestException: request cancelled. Remote Info: Not Available
```
and causing the admin UI to hang, as though it's still waiting to receive a response from Namerd.

I've updated the code handling Consul GET responses in `ConsulDtabStore` to `Try` the call to `Dtab.read()`, and update the activity's state to `Activity.Failed` when the dtab could not be parsed successfully. Namerd will now log messages which state that the dtab was malformed, and surface the parse error, including showing which character was invalid. The log message now looks like this:

```
E 1220 20:46:06.770 UTC THREAD31 TraceId:30cbad82deebb433: consul ns consul dtab parsing failed: java.lang.IllegalArgumentException: '/' expected but 'a' found at '[a]sjgfdskjhsdfg;'; dtab: 'asjgfdskjhsdfg;'
```

The admin UI web interface doesn't currently handle the error in this case, and will load a blank page rather than displaying a useful error message. However, it no longer appears to be awaiting a response, and the error messages in the logs are much more meaningful; and I think changes to the web UI should be in a separate PR.

I've written a new unit test to assert that an unparsable dtab fails the `Activity` returned by `ConsulDtabStore.observe()`. This unit test fails against the current master. I've also added some test utility code to make assertions about `Activity.State` values more ergonomic. I've also manually verified the changes by creating an invalid dtab value in Consul running locally and opening the admin UI for that namespace.

### DIFF
--- a/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
+++ b/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
@@ -188,8 +188,7 @@ class ConsulDtabStore(
             case Throw(e: NotFound) =>
               updates() = Activity.Ok(None)
               cycle(e.rsp.headerMap.get(Headers.Index), backoffs0)
-            case Throw(e: Failure) if e.isFlagged(Failure.Interrupted) =>
-              Future.Done
+            case Throw(e: Failure) if e.isFlagged(Failure.Interrupted) => Future.Done
             case Throw(e) =>
               updates() = Activity.Failed(e)
               log.error("consul ns %s dtab observation error %s", ns, e)
--- a/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
+++ b/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
@@ -168,13 +168,28 @@ class ConsulDtabStore(
           ).transform {
             case Return(result) =>
               val version = Buf.Utf8(result.index.get)
-              val dtab = Dtab.read(result.value)
-              updates() = Activity.Ok(Some(VersionedDtab(dtab, version)))
+              // the raw string, not yet parsed as a dtab.
+              val rawDtab = result.value
+              // attempt to parse the string as a dtab, and update the the
+              // Activity with  the new state - either Ok if the string was
+              // parsed successfully, or Failed if an error occurred.
+              val nextState = Try {
+                Dtab.read(rawDtab)
+              } match {
+                case Return(dtab) => // dtab parsing succeeded.
+                  Activity.Ok(Some(VersionedDtab(dtab, version)))
+                case Throw(e) => // dtab parsing failed!
+                  log.error("consul ns %s dtab parsing failed: %s; dtab: '%s'", ns, e, rawDtab)
+                  Activity.Failed(e)
+              }
+              updates() = nextState
               cycle(result.index, backoffs0)
+
             case Throw(e: NotFound) =>
               updates() = Activity.Ok(None)
               cycle(e.rsp.headerMap.get(Headers.Index), backoffs0)
-            case Throw(e: Failure) if e.isFlagged(Failure.Interrupted) => Future.Done
+            case Throw(e: Failure) if e.isFlagged(Failure.Interrupted) =>
+              Future.Done
             case Throw(e) =>
               updates() = Activity.Failed(e)
               log.error("consul ns %s dtab observation error %s", ns, e)
--- a/test-util/src/main/scala/io/buoyant/test/ActivityValues.scala
+++ b/test-util/src/main/scala/io/buoyant/test/ActivityValues.scala
@@ -4,7 +4,47 @@ import com.twitter.util.Activity
 import org.scalatest.exceptions.TestFailedException
 
 /**
- * Like TryValues but for Activity.
+ * Like ScalaTest's <code>TryValues</code> trait, but for <code>Activity.State</code>s.
+ *
+ * Trait that provides an implicit conversion that adds <code>ok</code>, <code>failed</code>, and <code>pending</code>
+ * methods to <code>com.twitter.util.Activity.State</code>, enabling you to make assertions about the value of an
+ * <code>Activity.Ok</code> state or the exception of an <code>Activity.Failed</code> state.
+ *
+ * <p>
+ * The <code>ok</code> method will return the value of the <code>Activity.State</code> on which it is invoked if the
+ * state is <code>Ok</code>, or throw <code>TestFailedException</code> if not.
+ * The <code>failed</code> method will return the exception value of an <code>Activity.State</code> on which it is
+ * invoked if the state is a <code>Activity.Failed</code>, or throw <code>TestFailedException</code> if not.
+ * Finally, the <code>pending</code> method, which is largely included for completeness, will return <code>()</code> if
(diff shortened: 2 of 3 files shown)


test-util/src/main/scala/io/buoyant/test/ActivityValues.scala
```diff
@@ -0,0 +1,37 @@
+package io.buoyant.test
+
+import com.twitter.util.Activity
+import org.scalatest.exceptions.TestFailedException
+
+/**
+ * Like TryValues but for Activity.
+ */
+trait ActivityValues {
```
deebo91: Would love a mini tutorial to see how this works with your unit tests!

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -168,13 +168,28 @@ class ConsulDtabStore(
...
+              }
+              updates() = nextState
               cycle(result.index, backoffs0)
+
             case Throw(e: NotFound) =>
               updates() = Activity.Ok(None)
               cycle(e.rsp.headerMap.get(Headers.Index), backoffs0)
-            case Throw(e: Failure) if e.isFlagged(Failure.Interrupted) => Future.Done
+            case Throw(e: Failure) if e.isFlagged(Failure.Interrupted) =>
+              Future.Done
```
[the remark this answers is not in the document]
hawkw: Ah, no, this is an unintentional change – I added a log line in this match arm while debugging, and then removed it, but forgot to remove the line break. I'll remove this change, thanks for catching it!

test-util/src/main/scala/io/buoyant/test/ActivityValues.scala
hawkw: Added some docs in b891c92813823fb988a3ce3100421df66d1449cf that should help explain how this is useful – let me know if you have any further questions!

hawkw:
merged from the Buoyant holiday party cable car 🙂

hawkw:
@adleong I'm not sure how the malformed dtabs were being written to Consul; while I was testing this fix I was using the Consul command line `consul kv put ...` to write them to a local Consul instance. We might want to ask @rclayton-the-terrible, who reported #1759, how unparsable dtabs ended up in Consul?

hawkw:
@adleong I'm not sure how the malformed dtabs were being written to Consul; while I was testing this fix I was using the Consul command line `consul kv put ...` to write them to a local Consul instance. We might want to ask @rclayton-the-terrible, who reported #1759, how unparsable dtabs ended up in Consul?

deebo91:
@adleong I think @rclayton-the-terrible mentioned that they manually wrote dtabs into consul. I think this is the slack message
> We create DTabs directly in Consul as a part of our deployments.  So namerd would not get a chance to validate.

hawkw:
@deebo91 ah, yeah, that's more or less what I suspected.

rclayton-the-terrible:
Yes, we don't use namerd's API to write DTabs.  We sidecar namerd on our ECS hosts, so it exists more as an ephemeral service to containers on the host than something we interact with as a dedicated Ops service.  Consul's role in our architecture is the exact opposite - it's an essential part of our control plane.  I guess we could theoretically write DTabs through the namerd API, but we never got the sense that that was mandatory.

PR #1768, opened 2018-01-03
Add a client_state.json admin endpoint to expose the client address set

adleong:
Add client_state.json admin endpoint which outputs the current address set for each client in the client registry.  

Sample output:
```
{
  "/$/inet/localhost/4100": {
    "state": "bound",
    "addresses": [
      "localhost:4100"
    ]
  },
  "/#/io.l5d.k8s/emojivoto/grpc/voting-svc": {
    "state": "bound",
    "addresses": [
      "172.17.0.8:8080",
      "172.17.0.7:8080",
      "172.17.0.5:8080"
    ]
  }
}
```

### DIFF
--- a/linkerd/admin/src/main/scala/com/twitter/finagle/client/buoyant/ClientStateHandler.scala
+++ b/linkerd/admin/src/main/scala/com/twitter/finagle/client/buoyant/ClientStateHandler.scala
@@ -1,0 +1,41 @@
+package com.twitter.finagle.client.buoyant
+
+import com.twitter.finagle.{Addr, Address, Service}
+import com.twitter.finagle.client.ClientRegistry
+import com.twitter.finagle.http.{MediaType, Request, Response}
+import com.twitter.finagle.loadbalancer.LoadBalancerFactory
+import com.twitter.util.Future
+import io.buoyant.config.Parser
+
+class ClientStateHandler extends Service[Request, Response] {
+
+  private[this] val mapper = Parser.jsonObjectMapper(Nil)
+
+  override def apply(request: Request): Future[Response] = {
+    val entries = ClientRegistry.registrants.map { entry =>
+      val LoadBalancerFactory.Dest(va) = entry.params[LoadBalancerFactory.Dest]
+      va.changes.toFuture().map(entry.addr -> _)
+    }
+    Future.collect(entries.toSeq).map { entries =>
+      val clientState = entries.map {
+        case (client, addr) =>
+          val state = addr match {
+            case Addr.Bound(addresses, meta) =>
+              addresses.map {
+                case Address.Inet(isa, _) => isa.getHostString
+                case a => a.toString
+              }.toSeq
+            case Addr.Failed(why) => s"Failed: ${why.getMessage}"
+            case a@Addr.Pending => a.toString
+            case a@Addr.Neg => a.toString
+          }
+          client -> state
+      }.toMap
+
+      val response = Response()
+      response.contentType = MediaType.Json + ";charset=UTF-8"
+      response.contentString = mapper.writeValueAsString(clientState)
+      response
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


adleong:
cc @obeattie

linkerd/admin/src/main/scala/com/twitter/finagle/client/buoyant/ClientStateHandler.scala
```diff
@@ -0,0 +1,41 @@
...
+import io.buoyant.config.Parser
+
+class ClientStateHandler extends Service[Request, Response] {
+
+  private[this] val mapper = Parser.jsonObjectMapper(Nil)
+
+  override def apply(request: Request): Future[Response] = {
+    val entries = ClientRegistry.registrants.map { entry =>
+      val LoadBalancerFactory.Dest(va) = entry.params[LoadBalancerFactory.Dest]
+      va.changes.toFuture().map(entry.addr -> _)
```
olix0r: can we limit this to entries where _entry.addr_ is a `Name` or `Path` so that we don't encode weird things for clients that don't do name resolution?

olix0r: Why `va.changes.toFuture()` and not `va.sample()`?

linkerd/admin/src/main/scala/com/twitter/finagle/client/buoyant/ClientStateHandler.scala
```diff
@@ -0,0 +1,41 @@
...
+          val state = addr match {
+            case Addr.Bound(addresses, meta) =>
+              addresses.map {
+                case Address.Inet(isa, _) => isa.getHostString
+                case a => a.toString
+              }.toSeq
+            case Addr.Failed(why) => s"Failed: ${why.getMessage}"
+            case a@Addr.Pending => a.toString
+            case a@Addr.Neg => a.toString
+          }
```
olix0r: wait what is the type of `state`?  This seems a little annoying to decode if the type is "a list or maybe a string".

Perhaps better to have a case class that encodes e.g. `"state": "bound", "addrs": []` etc (i'd bet that we even have this somewhere already for namerd's http api)

linkerd/admin/src/main/scala/com/twitter/finagle/client/buoyant/ClientStateHandler.scala
olix0r: Are we confident that the Var LoadBalancerFactory.Dest is reference counted appropriately?

Might we need to wrap it in `Var.async` when it's constructed?

linkerd/admin/src/main/scala/com/twitter/finagle/client/buoyant/ClientStateHandler.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  private[this] val mapper = Parser.jsonObjectMapper(Nil)
+
+  override def apply(request: Request): Future[Response] = {
+    val entries = ClientRegistry.registrants
+      .filter { entry =>
+        // only display clients with a Path name
+        Try(Path.read(entry.addr)).isReturn
+      }.map { entry =>
+        val LoadBalancerFactory.Dest(va) = entry.params[LoadBalancerFactory.Dest]
```
olix0r: Are we confident that `LoadBalancerFactory.Dest`'s Var is reference-counted?  Do we need to ensure this higher up in the stack when we get the result from an interpreter?

nikolay-pshenichny:
nit: this PR adds '/client_state.json', but in the [1.3.5 release info](https://github.com/linkerd/linkerd/releases/tag/1.3.5) the endpoint is '/admin/client_state.json`

hawkw:
@nikolay-pshenichny whoops! Will fix that.

obeattie:
Thank you so much for this 🙌

PR #1769, opened 2018-01-04
Add configurable failureThreshold for namerd interpreter

### LINKED ISSUE #1767: Make mux client's failure detector configurable
Linkerd's mux client uses the default _failure detector_, which pings its server to determine whether the endpoint is available.

In some environments, the failure detector can go into _closed_ too aggressively, and so users would like to be able to work around this by, for instance, using a larger closeTimeout.

We should expose these [settings](
https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/liveness/FailureDetector.scala#L98
) in the interpreter configuration so that, when they are specified, the failure detector's config is overridden for the namerd client.

```scala
  case class ThresholdConfig(
    minPeriod: Duration = 5.seconds,
    threshold: Double = 2,
    windowSize: Int = 100,
    closeTimeout: Duration = 4.seconds
) extends Config
```

deebo91:
Linkerd interpreter for Namerd uses a default failure detector and does not allow for configuration changes for users. This PR adds in four new fields in the intepreter section of the config to modify Linkerd's failure detection module when resolving names through Namerd 

Tests are being run currently to see if changing the values affects the failure detection but I am putting the PR up for visibility.

fixes #1767

### DIFF
--- a/interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
+++ b/interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
@@ -1,12 +1,14 @@
 package io.buoyant.namerd.iface
 
 import com.fasterxml.jackson.annotation.JsonIgnore
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import com.twitter.conversions.time._
 import com.twitter.finagle._
 import com.twitter.finagle.buoyant.TlsClientConfig
+import com.twitter.finagle.liveness.FailureDetector
 import com.twitter.finagle.liveness.FailureDetector.ThresholdConfig
 import com.twitter.finagle.naming.NameInterpreter
-import com.twitter.finagle.param.{HighResTimer, Label}
+import com.twitter.finagle.param.Label
 import com.twitter.finagle.service._
 import com.twitter.logging.Logger
 import com.twitter.util.{NonFatal => _, _}
@@ -14,10 +16,6 @@ import io.buoyant.admin.Admin
 import io.buoyant.admin.Admin.{Handler, NavItem}
 import io.buoyant.namer.{InterpreterInitializer, NamespacedInterpreterConfig}
 import io.buoyant.namerd.iface.{thriftscala => thrift}
-import com.twitter.conversions.time
-import com.twitter.finagle.liveness.FailureDetector
-
-import scala.util.control.NonFatal
 
 /**
  * The namerd interpreter offloads the responsibilities of name resolution to
@@ -54,24 +52,21 @@ case class ClientTlsConfig(commonName: String, caCert: Option[String]) {
 
 case class FailureThresholdConfig(
   minPeriodMs: Option[Int],
-  threshold: Double,
+  @JsonDeserialize(contentAs = classOf[java.lang.Double]) threshold: Option[Double],
   windowSize: Option[Int],
   closeTimeoutMs: Option[Int]
 ) {
-  val defaultConfig = ThresholdConfig()
+  @JsonIgnore
   def params: Stack.Params = {
-    val thresholdCfg = for {
-      minPeriod <- minPeriodMs
-      window <- windowSize
-      timeout <- closeTimeoutMs
-    } yield ThresholdConfig(
-      minPeriod.milliseconds,
-      threshold,
-      window,
-      timeout.milliseconds
+    val thresholdConfig = ThresholdConfig(
+      minPeriodMs.map(_.milliseconds).getOrElse(FailureThresholdConfig.DefaultMinPeriod),
+      threshold.getOrElse(FailureThresholdConfig.DefaultThreshold),
+      windowSize.getOrElse(FailureThresholdConfig.DefaultWindowSize),
+      closeTimeoutMs.map(_.milliseconds).getOrElse(FailureThresholdConfig.DefaultCloseTimeout)
     )
-    StackParams.empty + FailureDetector.Param(thresholdCfg.getOrElse(defaultConfig))
+    StackParams.empty + FailureDetector.Param(thresholdConfig)
   }
+
 }
 
 case class NamerdInterpreterConfig(
@@ -110,7 +105,7 @@ case class NamerdInterpreterConfig(
     val stats = stats0.scope(label)
 
     val tlsParams = tls.map(_.params).getOrElse(Stack.Params.empty)
-    val failureThresholdParams = failureThreshold.map(_.params).getOrElse(StackParams.empty)
+    val failureThresholdParams = failureThreshold.map(_.params).getOrElse(FailureThresholdConfig.defaultStackParam)
 
     val client = ThriftMux.client
       .withParams(ThriftMux.client.params ++ tlsParams ++ failureThresholdParams ++ params)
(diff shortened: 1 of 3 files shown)


deebo91:
PR for this enhancement is ready for review, tested the change on both the branch and master. I was able to configure the closeTimeout  from 4 seconds to various other timeout values (1 minute, 30 seconds, 1 second, 40 seconds)

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -48,11 +52,34 @@ case class ClientTlsConfig(commonName: String, caCert: Option[String]) {
...
+case class FailureThresholdConfig(
+  minPeriodMs: Option[Int],
+  threshold: Double,
+  windowSize: Option[Int],
+  closeTimeoutMs: Option[Int]
+) {
+  val defaultConfig = ThresholdConfig()
+  def params: Stack.Params = {
+    val thresholdCfg = for {
+      minPeriod <- minPeriodMs
```
adleong: This will only build a ThresholdConfig if *all* properties are set.  Instead, I think we want to fall back to defaults for each property.

```
ThresholdConfig(
  minPeriod = minPeriodMs.map(_.millis).getOrElse(DefaultMinPeriod),
  ... etc
)
```

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -48,11 +52,34 @@ case class ClientTlsConfig(commonName: String, caCert: Option[String]) {
   }
 }
 
+case class FailureThresholdConfig(
+  minPeriodMs: Option[Int],
+  threshold: Double,
+  windowSize: Option[Int],
+  closeTimeoutMs: Option[Int]
+) {
+  val defaultConfig = ThresholdConfig()
```
adleong: Let's explicitly set our own default values instead of inheriting from Finagle.  It's okay if the default values are the same, but this way they won't unexpectedly change out from under us if do a Finagle upgrade.

adleong: Needs @JsonIgnore

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -48,11 +52,34 @@ case class ClientTlsConfig(commonName: String, caCert: Option[String]) {
...
+  def params: Stack.Params = {
```
adleong: needs @JsonIgnore

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -48,11 +52,34 @@ case class ClientTlsConfig(commonName: String, caCert: Option[String]) {
...
+    val thresholdCfg = for {
```
adleong: needs @JsonIgnore

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -83,9 +110,10 @@ case class NamerdInterpreterConfig(
     val stats = stats0.scope(label)
 
     val tlsParams = tls.map(_.params).getOrElse(Stack.Params.empty)
+    val failureThresholdParams = failureThreshold.map(_.params).getOrElse(StackParams.empty)
```
adleong: Let's fall back to our defaults if the failureTreshold is absent, instead of omitting the param entirely and letting Finagle pick the default.

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
chris-goffinet-ck: Don't we need to allow the user to set threshold? It looks to be missing

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -110,3 +133,18 @@ case class NamerdInterpreterConfig(
 object NamerdInterpreterConfig {
   def kind = classOf[NamerdInterpreterConfig].getCanonicalName
 }
+
+object FailureThresholdConfig {
+
+  val DefaultMinPeriod = 5000.milliseconds
```
adleong: tioli `5.seconds`

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -110,3 +133,18 @@ case class NamerdInterpreterConfig(
...
+  val DefaultThreshold = 2.0
+  val DefaultWindowSize = 100
+  val DefaultCloseTimeout = 4000.milliseconds
```
adleong: tioli `4.seconds`

linkerd/docs/interpreter.md
```diff
@@ -45,6 +45,19 @@ dst | _required_ | A Finagle path locating the Namerd service.
...
+failureThreshold | no failureThreshold | Sets the [failure threshold](#failure-threshold) used by Linkerd's threshold failure detector to gauge a Namerd instance's health
+
+### Failure Threshold
+
+Linkerd uses a [Threshold Failure Detector](https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/liveness/ThresholdFailureDetector.scala) 
+to determine the health of a Namerd instance. Linkerd sends pings to Namerd periodically and evaluates the health of Namerd based on a set number of ping latencies
+
+Key | Default Value | Description
+--- | ------------- | -----------
+minPeriodMs | 5000 | period between session pings to Namerd
```
adleong: Use consistent capitalization at the beginnings of the descriptions please

linkerd/docs/interpreter.md
```diff
@@ -45,6 +45,19 @@ dst | _required_ | A Finagle path locating the Namerd service.
 namespace | `default` | The name of the Namerd dtab to use.
 retry | see [Namerd retry](#namerd-retry) | An object configuring retry backoffs for requests to Namerd.
 tls | no tls | Requests to Namerd will be made using TLS if this parameter is provided.  It must be a [Namerd client TLS](#namerd-client-tls) object.
+failureThreshold | no failureThreshold | Sets the [failure threshold](#failure-threshold) used by Linkerd's threshold failure detector to gauge a Namerd instance's health
+
+### Failure Threshold
+
+Linkerd uses a [Threshold Failure Detector](https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/liveness/ThresholdFailureDetector.scala) 
+to determine the health of a Namerd instance. Linkerd sends pings to Namerd periodically and evaluates the health of Namerd based on a set number of ping latencies
```
adleong: tioli: "... the health of the connection to a Namerd instance."

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -138,3 +133,18 @@ case class NamerdInterpreterConfig(
...
```
chris-goffinet-ck: @olix0r way in here, but as discussed I think this value should be higher. closer to 60.seconds or 300.seconds

interpreter/namerd/src/main/scala/io/buoyant/namerd/iface/NamerdInterpreterInitializer.scala
```diff
@@ -110,3 +133,18 @@ case class NamerdInterpreterConfig(
 object NamerdInterpreterConfig {
   def kind = classOf[NamerdInterpreterConfig].getCanonicalName
 }
+
+object FailureThresholdConfig {
+
+  val DefaultMinPeriod = 5.seconds
+  val DefaultThreshold = 2.0
+  val DefaultWindowSize = 100
+  val DefaultCloseTimeout = 4.seconds
```
adleong: After discussing with @olix0r, I think we should make this default to Duration.Top

PR #1770, opened 2018-01-05
Render an error instead of a blank page for failures on namerd dtab page

### LINKED ISSUE #1761: Surface dtab parse errors in Namerd admin UI
Follow on from PR #1760.

I've improved handling of dtab parsing errors in `ConsulDtabStore` by failing the activity when an unparsable dtab was received from Consul. Namerd now logs a much more explanatory error message when such an error occurs. However, the admin web UI now simply loads a blank page, rather than displaying an error message. 

We should have the admin web UI display the error message, so that a user encountering a dtab parse error doesn't have to consult the Namerd logs to see what went wrong.

adleong:
When an exception is encountered fetching or parsing the dtab for the dtab page in the Namerd admin UI, an empty 500 is returned which renders as a blank page.

Instead, render an error page with the exception message.

Fixes #1761

screen shot 2018-01-05 at 10 51 21 am

### DIFF
--- a/namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
+++ b/namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
@@ -21,27 +21,16 @@ class DtabHandler(
       getDtab(namespace).transform {
         case Return(None) => Future.value(Response(Status.NotFound))
         case Return(Some(dtab)) =>
-          render(namespace, dtab.dtab)
+          renderDtab(namespace, dtab.dtab)
         case Throw(e) =>
-          renderError(namespace, e)
+          renderError(e)
       }
     case _ =>
       Future.value(Response(Status.NotFound))
   }
 
-  def render(name: String, dtab: Dtab): Future[Response] = {
-    val content = s"""
-      <!doctype html>
-      <html>
-        <head>
-          <title>namerd admin</title>
-          <link rel="shortcut icon" href="../files/images/favicon.png" />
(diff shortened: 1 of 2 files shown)


hawkw:
Yay, thank you!

namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
```diff
@@ -53,11 +52,50 @@ class DtabHandler(
...
+            <div class="row">
+              <div class="col-lg-6">
+                <h2 class="router-label-title">Dtab Error</h2>
+                <p>An error was encountered when fetching or parsing the dtab:</p>
+                <pre>$e</pre>
+              </div>
+            </div>
+          </div>
+        </body>
+      </html>
```
klingerf: TIOLI -- all of the HTML outside of the `<div class="container-fluid">` tag here is the same as the HTML that's constructed on line 33 of this file. You could move the wrapper HTML to a separate string template so that you only have to define it once. Something like:

```
def render(name: String, dtab: Dtab): Future[Response] = {
  val content = s"""
    <div class="row">...</div>
  """
  return renderHtml(content)
}

def renderError(name: String, e: Throwable): Future[Response] = {
  val content = s"""
    <div class="row">...</div>
  """
  return renderHtml(content)
}

def renderHtml(content: String): Future[Response] = {
  val wrappedContent = s"""
    <!doctype html>
      ...
      <div class="container-fluid">$content</div>
      ...
    </html>
    """
    val response = Response()
    response.contentType = MediaType.Html + ";charset=UTF-8"
    response.contentString = wrappedContent
    Future.value(response)
}
```

Issue #1079, opened 2017-02-16
OpenTracing support


olix0r:
At some point, we may want to have native support for opentracing in linkerd (i.e instead of / in addition to the l5d-ctx-trace header).  What would this look like?

zhangmuxi:
Have you guys implemented this or not?

klingerf:
Hey @zhangmuxi -- this is still unimplemented in Linkerd. You might be interested in following this opentracing issue, which has a bit more context around trace implementations in proxies: opentracing/specification#86.

mqliang:
@olix0r @klingerf IIUC, currently, we already have a [linkerd-zipkin](https://github.com/linkerd/linkerd-zipkin) plugin, which is a linkerd telemeter plugins for writing trace data to zipkin, but this is dedicated to zipkin. And, we want implement a linkerd-opentracing plugin, which writes trace data to all tracing vendor, as long as it support opentracing?

klingerf:
Hey @mqliang -- your understanding is correct. Right now Linkerd supports exporting tracing data in the Zipkin format, which can be collected by any Zipkin-compatible collector, such as Zipkin itself and Stackdriver. We'd need to add a separate telemeter plugin to export tracing data in the OpenTracing format, and that would enable OpenTracing-compatible collectors, such as Jaeger and LightStep.

malafeev:
I think plugin is not enough. 
We need a filter to have access to http headers of incoming and outgoing requests.
Filter should:
1. Extract span context from http headers of incoming requests (and later use this span context as parent for created span)
2. Create span and inject it into http headers of outgoing request.
3. Call Trace.recordBinary("opentracing-span", span). It will be intercepted by OpenTracing linkerd plugin (using TelemeterInitializer)

The problem that there isn’t a way to inject arbitrary filters via plugin (https://discourse.linkerd.io/t/is-it-possible-to-set-common-client-and-server-filter-via-plugin/493/2?u=malafeev)

SecretLiang:
@malafeev Then, how linkerd-zipkin plugin handle this? Hardcoding the HTTP header in the core codebase, instead of injecting via plugin? And you mentioned at https://discourse.linkerd.io/t/is-it-possible-to-set-common-client-and-server-filter-via-plugin/493/2?u=malafeev that you can create a PR, do you mean a PR that add filter to have access to http headers ?

malafeev:
[...]

PR #1773, opened 2018-01-10
Add OpenTracing support #1079

malafeev:
Initial proposal for #1079

Create OpenTracing filter to
- Extract span context from http headers of incoming requests (and later use this span context as parent for created span)
- Create span and inject it into http headers of outgoing request.
- Call Trace.recordBinary("opentracing-span", span). It will be intercepted by OpenTracing linkerd plugin (using TelemeterInitializer)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# William and Oliver should approve all changelog entries.
+CHANGES.md @wmorgan @olix0r
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,19 @@
+## 1.3.5 2018-01-17
+
+This release focuses on quality, and on improving the debugging process. It includes improvements and fixes for Linkerd's Kubernetes support, administrative UI, and Namerd control plane. It officially graduates HTTP/2 support out of experimental, and also features a number of community contributions!
+
+* 🎓 H2 router and `io.l5d.mesh` Namerd interface are no longer experimental ([#1782](https://github.com/linkerd/linkerd/pull/1782))! 🎓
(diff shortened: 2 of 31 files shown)


deebo91:
@malafeev thank you very much for starting a PR for this feature. We always love it when we get a contribution from the community. Let us not when you are ready for us to review this PR!

malafeev:
@deebo91 please review it.
Maybe you will propose another approach.

adleong:
Hi @malafeev!  Thank you for starting work on this!  We're really excited to get Open Tracing support into Linkerd.  Before we dig too much into the implementation details, could you tell us a little more about how this works?

* What are the steps for testing this?
* I don't see the OpenTracing telemeter in this branch, is that in progress?

malafeev:
@adleong I will publish telemeter plugin.

adleong:
Hi @malafeev, have you had a chance to publish the telemetry plugin?

malafeev:
@adleong I'm currently stuck because linkerd uses libthrift version 0.5
but Jaeger tracer uses libthrift version 0.9.2
Most of OpenTracing tracers come with dependencies on transport libraries (thrift, grpc etc)
and they sometimes conflict with application dependencies (jar hell).

adleong:
Yikes.  Unfortunately, I don't have much advice to offer.  Perhaps some kind of dependency shadowing might help?

malafeev:
yeah, trying to solve issue with shadowing (https://github.com/jaegertracing/jaeger-client-java/issues/332)

adleong:
Hi @malafeev, any luck getting the shadowing to work?

malafeev:
Hi @adleong, no luck.
Problem that only Jaeger tracer supports Tracer Resolver, but there is an (issue)[https://github.com/jaegertracing/jaeger-client-java/issues/332#issuecomment-365903790]
which probably will not be fixed nearest time.
I think we can close this PR for now.

Issue #83, opened 2016-02-09
Kubernetes namespace watches sometimes hang indefinitely


olix0r:
We've observed a bug in the kubernetes namer where name resolution (and requests) hang indefinitely when initializing a watch on a second namespace.

For example, in a test cluster, we launch an application into the `prod` namespace and do continuous deployment to a `pre-prod` namespace that may be accessed with an override.

The configuration contains:
```yaml
namers:
- kind: io.l5d.experimental.k8s
  prefix: /ns
  tlsWithoutValidation: true
  authTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
and a base dtab like:
```
  /iface => /io.l5d.k8s/prod;
  /method => /$/io.buoyant.http.anyMethodPfx/iface ;
  /http/1.1 => /method ;
  ...
```

Requests are routed to the pre-prod namespace with an override like:
```
/iface => /io.l5d.k8s/pre-prod | /$/fail;
```

Initially, we observe requests through prod operating as expected, resolving through the k8s namer:
```
D 0126 23:45:28.991 THREAD29: k8s initializing service0
...
D 0126 23:45:29.327 THREAD30: k8s ns prod initial state: service0
...
D 0126 23:45:42.955 THREAD30: k8s lookup: /pre-prod/http/service0 /pre-prod/http/service0
D 0126 23:45:42.956 THREAD30: k8s initializing pre-prod
```

And we see no further k8s logging after this.  Furthermore, because the namer never satisfies this  lookup, the request never succeeds and remains pending until the client disconnects

Something appears to be wrong in the kubernetes namer that prevents this second watch from succeeding.

siggy:
#### linkerd-0.2.1-SNAPSHOT confirmed working

testing with `linkerd-helium:sig3` (`linkerd-0.2.1-SNAPSHOT`), in namespace `sig`, works with `prod` and  `pre-prod`:

```
$ curl -H 'L5d-sample: 0' -H 'Dtab-local: /ns/sig=>/ns/prod;' http://23.251.154.236
$ curl -H 'L5d-sample: 0' -H 'Dtab-local: /ns/sig=>/ns/pre-prod;' http://23.251.154.236
```
```
$ kubectl --namespace=sig logs sproxy-ix56f -c l5d -f
...
D 0226 05:53:17.541 THREAD31: k8s lookup: /prod/router/web /prod/router/web
D 0226 05:53:17.546 THREAD31: k8s initializing prod
[...]

PR #1774, opened 2018-01-10
Use a single client for the entire k8s namer

adleong:
Each watch on the Kubernetes API made by the io.l5d.k8s namer uses a separate finagle client.  This causes stat name conflicts which results in incorrect reporting of client stats.  It may also have an adverse affect on memory usage.

We switch the io.l5d.k8s namer to use a single finagle client for all watches on the Kubernetes API.

This isolation of watches to separate finagle clients was originally done to work around https://github.com/linkerd/linkerd/issues/83.  However, since then, the [underlying issue](https://github.com/twitter/finagle/issues/471) has been [fixed in finagle](https://github.com/twitter/finagle/pull/481) making this no longer necessary.


adleong:
@obeattie by chance have you had an opportunity to test this in your staging cluster?  This should hopefully improve the quality of your metrics.

PR #1786, opened 2018-01-19
add access logging for http2 requests

franziskagoltz:
Access logs are stored in `./logs/access.log ` and can be accessed with `cat ./logs/access.log `

Our h2 Request and Response objects do not have all the attributes listed for logging requests in the [Apache Log best practices](https://httpd.apache.org/docs/1.3/logs.html). The information we don't have is represented with a `-` in the log, as we do in regular http logging.

The current Request object does not store a protocol version, we could potentially hard code this, as this code is only executed when http/2 are being made.

Current log output for h2 requests:
screen shot 2018-01-19 at 11 06 24 am

By comparison, the log for http/1 requests:
screen shot 2018-01-19 at 11 14 18 am

Signed-off-by: Franziska von der Goltz <franziska@vdgoltz.eu>

### DIFF
--- a/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
+++ b/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
@@ -5,12 +5,16 @@ import com.twitter.finagle.buoyant.h2.{Request, Response}
 import com.twitter.finagle.context.RemoteInfo
 import com.twitter.logging._
 import com.twitter.util.{Time, TimeFormat}
+import java.net.{InetSocketAddress}
 
 case class H2AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
 
   def apply(req: Request, svc: Service[Request, Response]) = {
     val reqHeaders = req.headers
-    val remoteHost = RemoteInfo.Upstream.addr.getOrElse("-")
+    val remoteHost = RemoteInfo.Upstream.addr match {
+      case Some(isa: InetSocketAddress) => isa.getHostString
+      case _ => "-"
+    }
     val identd = "-"
     val user = "-"
     val referer = reqHeaders.get("referer").getOrElse("-")
--- a/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
+++ b/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
@@ -12,10 +12,10 @@ case class H2AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
     val remoteHost = "-"
     val identd = "-"
     val user = "-"
-    val referer = "-"
+    val referer = reqHeaders.get("referer").getOrElse("-")
     val userAgent = reqHeaders.get("user-agent").getOrElse("-")
     var hostHeader = req.authority
-    val reqResource = s"${req.method.toString.toUpperCase} ${req.path} -"
+    val reqResource = s"${req.method.toString.toUpperCase} ${req.path} HTTP/2"
 
     svc(req).onSuccess { rsp =>
       val statusCode = rsp.status.code
--- a/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
+++ b/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
@@ -2,6 +2,7 @@ package io.buoyant.linkerd.protocol.h2
 
 import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack, Stackable}
 import com.twitter.finagle.buoyant.h2.{Request, Response}
+import com.twitter.finagle.context.RemoteInfo
 import com.twitter.logging._
 import com.twitter.util.{Time, TimeFormat}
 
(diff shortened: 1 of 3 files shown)


linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.twitter.util.{Time, TimeFormat}
+
+case class H2AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
+
+  def apply(req: Request, svc: Service[Request, Response]) = {
+    val reqHeaders = req.headers
+    val remoteHost = "-"
+    val identd = "-"
+    val user = "-"
+    val referer = "-"
```
hawkw: I believe the `referer` header field may be present on h2 requests, so this probably ought to be 
```scala
reqHeaders.get("referer").getOrElse("-")
```

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack, Stackable}
+import com.twitter.finagle.buoyant.h2.{Request, Response}
+import com.twitter.logging._
+import com.twitter.util.{Time, TimeFormat}
+
+case class H2AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
+
+  def apply(req: Request, svc: Service[Request, Response]) = {
+    val reqHeaders = req.headers
+    val remoteHost = "-"
```
hawkw: I think it would be good to log something for the `remoteHost` field, but as far as I can tell, this would probably require changes in `h2.Request` to store the remote address like Finagle's HTTP `Request` does, so it's reasonable to punt on that for now.

hawkw:
@franziskagoltz 
> The current Request object does not store a protocol version, we could potentially hard code this, as this code is only executed when http/2 are being made.

I think this seems reasonable for now.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
franziskagoltz: yes, after discussing this with kevin he said to not change the Request object for now, so i'll leave this as is.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
franziskagoltz: happy to change it to this. kevin and i  did loop over all headers and referrer was not a key in that sequence: here's the output for you review, in case you see the referer being associated to a different key?

```
:method
POST
:scheme
http
:path
/helloworld.Hello/Greeting
:authority
localhost:4142
content-type
application/grpc
user-agent
grpc-go/1.8.0
te
trailers
l5d-dst-service
/svc/localhost:4142
via
h2 linkerd
l5d-dst-client
/$/inet/127.1/8888
l5d-ctx-trace
BFa50jQ7Na/DuK+B+vMbkMO4r4H68xuQAAAAAAAAAAA=
l5d-reqid
c3b8af81faf31b90
```

the current output after changing the value of `referer` to `reqHeaders.get("referer").getOrElse("-")` is still `"-"`

hawkw: As we discussed, i think that this header is not always set, but can be present depending on the request --- I think we should attempt to log it if it is present.

hawkw:
🌐📲📝👍

hawkw:
CI failure for 6efeca5da632c7c1a9b938c7450b4f24142c973d appears to have been spurious, restarting.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
adleong: I think you can get the remote host with this:

```
import com.twitter.finagle.context.RemoteInfo
RemoteInfo.Upstream.addr
```

franziskagoltz: awesome! i'll give this a shot - thanks!

hawkw: nice, i didn't know that existed!

franziskagoltz:
current view of access log:
screen shot 2018-01-19 at 3 02 27 pm

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
```diff
@@ -0,0 +1,84 @@
...
+import com.twitter.finagle.buoyant.h2.{Request, Response}
+import com.twitter.finagle.context.RemoteInfo
+import com.twitter.logging._
+import com.twitter.util.{Time, TimeFormat}
+
+case class H2AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
+
+  def apply(req: Request, svc: Service[Request, Response]) = {
+    val reqHeaders = req.headers
+    val remoteHost = RemoteInfo.Upstream.addr.getOrElse("-")
```
adleong: To avoid printing the port here I think you want to attempt to cast to `InetSocketAddress` and use `getHostString`.  Something like

```
RemoteInfo.Upstream.addr match {
  case Some(isa: InetSocketAddress) => isa.getHostString
  case _ => "-"
}
```

franziskagoltz: awesome, thanks @adleong! i had to import `InetSocketAddress` and hope I choose the correct one?!

hawkw:
Yay!

PR #1790, opened 2018-01-24
Log unexpected responses from the k8s api

adleong:
When Linkerd receives a response from the k8s api with an unexpected status code, the request is retried indefinitely in case the unexpected response is transient.  Unfortunately, this makes non-transient errors (RBAC misconfigurations, for example) difficult to debug because no error is logged.

We log an error each time Linkerd receives an unexpected status code to make the problem more visible.  For example:

```
E 0124 19:48:03.863 UTC THREAD28: retrying k8s request to /apis/extensions/v1beta1/ingresses on unexpected response code 403 with message {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"ingresses.extensions is forbidden: User \"system:serviceaccount:default:default\" cannot list ingresses.extensions at the cluster scope","reason":"Forbidden","details":{"group":"extensions","kind":"ingresses"},"code":403}
```

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -42,7 +42,7 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
       // Don't retry on interruption
       case (_, Throw(e: Failure)) if e.isFlagged(Failure.Interrupted) => false
       case (_, Throw(NonFatal(ex))) =>
-        log.error("retrying k8s request to %s on error %s", path, ex)
+        log.warning("retrying k8s request to %s on error %s", path, ex)
         true
     },
     HighResTimer.Default,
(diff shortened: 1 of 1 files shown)


wmorgan:
More debugging power = good

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -36,7 +36,9 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
       case (_, Return(rep)) if rep.status.code >= 200 && rep.status.code < 300 => false
       case (_, Return(rep)) if rep.status == http.Status.NotFound || rep.status == http.Status.Gone => false
       // All other status codes are unexpected and should be retried
-      case (_, Return(rep)) => true
+      case (_, Return(rep)) =>
+        log.error("retrying k8s request to %s on unexpected response code %d with message %s", path, rep.statusCode, rep.contentString)
```
olix0r: maybe _warn_ instead of error? let's try to reserve error for things that we _know_ are terrible. this could be something transient if i'm not mistaken.

Issue #1800, opened 2018-02-02
Pass Stack.Params to DtabStore plugins


edio:
Issue Type:

- [ ] Bug report
- [x] Feature request

**What happened**:
Right now it is not possible to record any stats from `DtabStore` plugins, because namerd doesn't pass `Stack.Params` to `DtabStoreConfig.mkDtabStore`

**What you expected to happen**:
`Stack.Params` should be passed as with the rest of plugins

PR is to follow shortly

deebo91:
Thanks for creating the issue @edio. Looking forward to your PR

PR #1801, opened 2018-02-02
Allow passing Stack.Params to DtabStoreConfig

edio:
Implements #1800

### DIFF
--- a/namerd/core/src/main/scala/io/buoyant/namerd/DtabStoreInitializer.scala
+++ b/namerd/core/src/main/scala/io/buoyant/namerd/DtabStoreInitializer.scala
@@ -1,7 +1,8 @@
 package io.buoyant.namerd
 
 import com.fasterxml.jackson.annotation.{JsonIgnore, JsonProperty}
-import io.buoyant.config.{PolymorphicConfig, ConfigInitializer}
+import com.twitter.finagle.Stack
+import io.buoyant.config.{ConfigInitializer, PolymorphicConfig}
 
 abstract class DtabStoreInitializer extends ConfigInitializer
 
@@ -22,5 +23,5 @@ abstract class DtabStoreConfig extends PolymorphicConfig {
   def disabled = experimentalRequired && !_experimentalEnabled.contains(true)
 
   @JsonIgnore
-  def mkDtabStore: DtabStore
+  def mkDtabStore(params: Stack.Params): DtabStore
 }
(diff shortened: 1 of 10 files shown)


namerd/core/src/main/scala/io/buoyant/namerd/DtabStoreInitializer.scala
```diff
@@ -22,5 +23,9 @@ abstract class DtabStoreConfig extends PolymorphicConfig {
   def disabled = experimentalRequired && !_experimentalEnabled.contains(true)
 
   @JsonIgnore
+  def mkDtabStore(params: Stack.Params): DtabStore = mkDtabStore
+
+  @JsonIgnore
+  @Deprecated
```
adleong: Rather than keeping this deprecated method, how would you feel about just updating all of the existing DtabStore implementations?  There are only a small handful in the codebase and it should be a very mechanical change.

edio: Won't this break all custom plugins that linked users might have?

I prepared this in haste, and I do see now, that fix is not optimal, as new plugin would need to implement both methods.
I can fix that and update all plugins that come with linked, but I believe it's better to keep the old method anyway.

edio: I won't insist though if you decide to remove the old method :)

edio:
I updated plugins to use the new method.

Currently, if no method is implemented by a plugin, namerd will fail in runtime. I wasn't sure what is the decision on the old method, so made it as simple as possible for now.

I see 3 options:
1) going your way and removing old method completely; Pro - simple code, Con - breaks compatibility with plugins, that linkerd users may have
2) introducing more complicated class hierarchy (particularly, smthng like `BaseDtabStoreConfig` and `ParameterizedDtabStoreConfig`. `DtabStoreConfig` is marked as deprecated and implements `mkDtabStore(Stack.Params)`, `ParameterziedDtabStoreConfig` is used in all bundled plugins and has `mkDtabStore()` implemented. Pro - no plugins should be broken on linkerd upgrade, Con - complication.
3) Leaving things as they are (throwing `NotImplementedException` in Runtime). Pro - compatibility and simple code, Con - no compile-time checking and somewhat hacky.

Let me know, what do you think please.

adleong:
My preference is option 1.  Yes, this does break compatibility with existing custom plugins that may exist outside the repo.  I think this is okay because a) plugin maintainers need to recompile against new versions of Linkerd anyway and b) the change they need to make is very simple and easy.

edio:
@adleong I removed the old method.
CI now fails due to NetworkedInteropTest, which appears to be flaky (different cases have failed for me before).

Issue #1805, opened 2018-02-05
linkerd serves requests for deleted Kube Service for 15+ minutes


negz:
Issue Type:

- [x] Bug report

**What happened**:
I deleted a Kubernetes Service resource that was backing a Kubernetes Ingress resource. Requests to the Kubernetes ingress resource continued to work fine for approximately 15-20 minutes after the deletion.

**What you expected to happen**:
I expected linkerd to notice the Service resource had been deleted and be unable to route the Ingress traffic.

**How to reproduce it (as minimally and precisely as possible)**:
1. Deploy linkerd on Kubernetes (see below for configs)
2. Deploy nginx backends, service, and ingress (see below for configs)
3. `kubectl delete svc nginx`

**Anything else we need to know?:**
I suspect this is the same issue, or related to, #1730. I wanted to raise a new issue to avoid polluting that one with my config dumps.

**Environment**:
- linkerd/namerd version, config files: linkerd v1.3.5.
- Platform, version: Kubernetes 1.9.2
- Cloud provider or hardware configuration: GCE, running bespoke Kubernetes deployment (not GKE)

**Configurations**
linkerd:
```yaml
    admin:
      ip: 0.0.0.0
      port: 9990

    namers:
    - kind: io.l5d.k8s
    - kind: io.l5d.k8s
      prefix: /io.l5d.k8s.http
      transformers:
      - kind: io.l5d.k8s.daemonset
        namespace: kube-system
        port: http-incoming
        service: l5d
        hostNetwork: true
    - kind: io.l5d.k8s
      prefix: /io.l5d.k8s.h2
      transformers:
      - kind: io.l5d.k8s.daemonset
        namespace: kube-system
        port: h2-incoming
        service: l5d
        hostNetwork: true
    - kind: io.l5d.k8s
      prefix: /io.l5d.k8s.grpc
      transformers:
      - kind: io.l5d.k8s.daemonset
        namespace: kube-system
        port: grpc-incoming
        service: l5d
        hostNetwork: true
    - kind: io.l5d.rewrite
      prefix: /portNsSvcToK8s
      pattern: "/{port}/{ns}/{svc}"
      name: "/k8s/{ns}/{port}/{svc}"

    telemetry:
    - kind: io.l5d.prometheus
    - kind: io.l5d.recentRequests
      sampleRate: 0.05
    - kind: io.zipkin.http
      host: zipkin:9411
      initialSampleRate: 0.05

    usage:
      enabled: false

    routers:
[...]

PR #1810, opened 2018-02-09
Fix issue where k8s ingresses are sometimes not deleted 

### LINKED ISSUE #1791: Linkerd fails to recognize k8s DELETED events
Issue Type:

- [X] Bug report

**What happened**:

After ingress is deleted Linkerd still routes to the backend service.

Linkerd should display this:

```
curl -H "Host: aservice.p.m.in.in.alutech24.com" localhost:3142
Unknown destination: Request("GET /", from /127.0.0.1:46336) / no ingress rule matches...
```

Ingress resources used:

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: aservice-l5d-pm-test
  annotations:
    kubernetes.io/ingress.class: "linkerd-dev"
spec:
  rules:
  - host: aservice.p.m.in.in.alutech24.com
    http:
      paths:
      - path:
        backend:
          serviceName: aservice
          servicePort: http
```

1. Create ingress resource
2. Delete ingress resource

Sometimes it works instantly (as I'm expecting) and sometime it doesn't. Most times it doesn't. However after 40 minutes now Linkerd responds with "no ingress rule matches...".

Also yesterday when I deleted some of my ingress resources I've lost all of them in Linkerd. It started responding with "no ingress rule matches..." for all of my ingresses. I'm not 100% sure that it happened due to deletion of one of my ingresses but it happened. Linkerd restart helped and everything is back to normal. I'm currently investigating what happened and maybe come up with another issue about that. 

**Environment**:
- linkerd/namerd version, config files: Linkerd version - 1.3.5, config file - https://gist.github.com/pavel-mikhalchuk/5456e28e4dd7e958280e459b0ced0381
- Platform, version, and config files (Kubernetes, DC/OS, etc): Kubernetes 1.8 created with kubeadm (https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/); RBAC config - https://gist.github.com/pavel-mikhalchuk/1e52278f62ea76fcab423976e1524ed1
- Cloud provider or hardware configuration: Private DC; Ubuntu 16.04.3 xenial

**Linkerd logs**:

https://gist.github.com/pavel-mikhalchuk/3698ef093a67e41018fe1306fb31e604

Please let me know if you need more details.

Thanks!

deebo91:
Problem
Linkerd's `io.l5d.ingress` identifier watches ingress resources in order to function as an ingress controller in k8s. This watch receives events from k8s and relies on a `resourceVersion` to make sure the most up to date ingress is being used. The `resourceVersion` is supposedly a number that increases after a change event occurs on a resource. For a particular ingress resource, its `resourceVersion` increases after Add, Modify and Change events. Sometimes k8s returns a `resourceVersion` that is unchanged. Linkerd does not recognize this as a change since it expects a `resourceVersion` to always  be larger than the previously observed resource and therefore does not update its `IngressCache`.
 
Solution
The solution is rather simple. When Linkerd encounters a `resourceVersion` that is the same as the 
the previously observed we do not ignore the event and notify the ingress watch process to update the `IngressCache`

Validation
Tests were done on a locally running Linkerd instance connected to k8s running on docker. Linkerd was able to update its `IngressCache` when an ingress was deleted in k8s.

Fixes #1791

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -175,11 +175,21 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
           // Register the update only if its resource version is larger than or equal to the largest version
           // seen so far.
           if (largestEvent.forall(_ <= event)) {
-            log.debug(s"k8s returned a new resource: $event")
+            log.trace(
+              "k8s watch on '%s' registered event with same or newer resource version %s (older resource version was %s)",
+              watchPath,
+              event.resourceVersion,
+              largestEvent.flatMap(_.resourceVersion)
+            )
             state.update(Activity.Ok(event))
             _processEventStream(ws, event.resourceVersion, Some(event))
           } else {
-            log.debug(s"k8s returned an old resource. Using most up to date resource: $event")
+            log.trace(
+              "k8s watch on '%s' skipping event with resource version %s (older than the most recent resource version %s)",
+              watchPath,
+              event.resourceVersion,
+              largestEvent.flatMap(_.resourceVersion)
+            )
             _processEventStream(ws, largestVersion, largestEvent)
           }
 
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -172,12 +172,14 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
 
         case Some((event, ws)) =>
           import Ordering.Implicits._
-          // Register the update only if its resource version is larger than the largest version
+          // Register the update only if its resource version is larger than or equal to the largest version
           // seen so far.
-          if (largestEvent.forall(_ < event)) {
+          if (largestEvent.forall(_ <= event)) {
+            log.debug(s"k8s returned a new resource: $event")
             state.update(Activity.Ok(event))
             _processEventStream(ws, event.resourceVersion, Some(event))
           } else {
+            log.debug(s"k8s returned an old resource. Using most up to date resource: $event")
             _processEventStream(ws, largestVersion, largestEvent)
           }
 
--- a/k8s/src/test/scala/io/buoyant/k8s/v1/ApiTest.scala
+++ b/k8s/src/test/scala/io/buoyant/k8s/v1/ApiTest.scala
@@ -488,13 +488,13 @@ class ApiTest extends FunSuite
 
     assert(events.size == 0)
 
-    // repeat the event: no update since resource version is the same
+    // repeat the event: update since resource version is the same
     await(w.write(modified0))
-    assert(events.size == 0)
+    assert(events.size == 1)
 
     // write an earlier event: no update since resource version is too low
     await(w.write(modified2))
-    assert(events.size == 0)
+    assert(events.size == 1)
     closable.close()
     if (failure != null) throw failure
   }


k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -172,12 +172,14 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
-          // Register the update only if its resource version is larger than the largest version
+          // Register the update only if its resource version is larger than or equal to the largest version
           // seen so far.
-          if (largestEvent.forall(_ < event)) {
+          if (largestEvent.forall(_ <= event)) {
+            log.debug(s"k8s returned a new resource: $event")
             state.update(Activity.Ok(event))
             _processEventStream(ws, event.resourceVersion, Some(event))
           } else {
+            log.debug(s"k8s returned an old resource. Using most up to date resource: $event")
```
hawkw: TIOLI, but I think it would be nice if this log line included the following information:
+ the path of the watched resource
+ the received resource version
+ the previous resource version

If we log all this information, it might not be necessary to log the entire event as well (which can be somewhat wordy). Consider something like:

```scala
log.debug(
  "k8s watch on '%s' skipping event with resource version %s (older than most recent resource version %s)",
  watchPath,
  event.resourceVersion,
  largestEvent.map(_.resourceVersion)
)
```
or similar.

hawkw: Note also that the `sprintf`-style log formatting (`%s` and friends) is preferable over Scala string interpolation (`$name` etc) in twitter.util.logging, as it's only evaluated if that log level is enabled, which will have a reduced performance impact. See: https://github.com/twitter/util#using

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -172,12 +172,14 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
 
         case Some((event, ws)) =>
           import Ordering.Implicits._
-          // Register the update only if its resource version is larger than the largest version
+          // Register the update only if its resource version is larger than or equal to the largest version
           // seen so far.
-          if (largestEvent.forall(_ < event)) {
+          if (largestEvent.forall(_ <= event)) {
+            log.debug(s"k8s returned a new resource: $event")
```
hawkw: TIOLI, but I think this can probably be logged at the trace level. Also see my below comments on logging.

deebo91:
@hawkw I agree with you on testing. I would want to make sure we aren't affecting code that uses `Watchable.scala`. We probably have to test k8s service discovery extensively.

hawkw:
@deebo91 Yeah, `Watchable` is involved in pretty much all of our code that communicates with Kubernetes, not just ingress. Although that does mean that we'll have to do a lot of testing to be confident in shipping this change, it also means that there's a distinct possibility that this fixes other k8s watch issues such as #1730, as well as the ingress-specific issue #1791!

adleong:
Any progress on integration testing this?

negz:
We might be able to help testing this if there's anything in particular you're interested in. Anecdotally we're seeing issues that feel like this one *a lot* recently - i.e. linkerd not noticing updates to service and ingress resources.

deebo91:
@negz thanks for offering to help! It would be interesting to see your findings with this PR. There is a new docker image https://hub.docker.com/r/buoyantio/linkerd/tags/1.3.6-watch-change/ that includes this change. It would be helpful to see if this image fixes the ingress issue you have been seeing.

negz:
@deebo91 We rolled `buoyantio/linkerd:1.3.6-watch-change` today to test. I've found one bug that I can consistently reproduce, though it's unclear whether it was caused by this PR or already in linkerd. In short, deleting *any* Kubernetes Ingress will break all ingress rules. i.e. all other ingresses will report `Unknown destination: Request("GET /graph", from /x.x.x.x:64296) / no ingress rule matches`.

I can provide some more detail later - got to run for the moment.

negz:
> I've found one bug that I can consistently reproduce, though it's unclear whether it was caused by this PR or already in linkerd. In short, deleting any Kubernetes Ingress will break all ingress rules.

Looks like it was a pre-existing bug that we just happened to repro while running this build. I've raised https://github.com/linkerd/linkerd/pull/1817 to fix it.

negz:
> The resourceVersion is supposedly a number that increases after a change event occurs on a resource. For a particular ingress resource, its resourceVersion increases after Add, Modify and Change events. Sometimes k8s returns a resourceVersion that is unchanged. 

> I'd like to have a better understanding of the Kubernetes API's resource versioning and what the actual semantics are.

I've been trying to wrap my head around this today too. I'm probably just paraphrasing @deebo91 here,  but based on the `WatchEvent` [docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#watchevent-v1-meta) my understanding is that `MODIFIED` objects will have a higher `resourceVersion`, while `DELETED` objects will have an unchanged `resourceVersion` because they return "the state of the object immediately before deletion".

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -172,12 +172,24 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+            )
             state.update(Activity.Ok(event))
             _processEventStream(ws, event.resourceVersion, Some(event))
           } else {
+            log.trace(
+              "k8s watch on '%s' skipping event with resource version %s (older than the most recent resource version %s)",
+              watchPath,
+              event.resourceVersion,
+              largestEvent.flatMap(_.resourceVersion)
+            )
```
negz: Would it be worth making this a `log.warn()`? It seems like this should always be indicative of something weird/out of sync between the watch events and linkerd, and thus would provide a good signal to triage that kind of thing.

hawkw: @negz totally agree --- I thought I said something about that in my original review of this PR, but I think I forgot to mention it.

hawkw:
@negz 
> my understanding is that `MODIFIED` objects will have a higher `resourceVersion`, while `DELETED` objects will have an unchanged `resourceVersion` because they return "the state of the object immediately before deletion".

That seems right to me --- thanks for putting that together. I have some thoughts on how to fix Linkerd's handling of watch events based on your observation, so hopefully we'll be able to start testing some changes soon! Thanks again.

adleong:
@negz have you had a chance to test if #1817 resolves all the issues you were seeing or if you're still encountering this bug with DELETEs and resource versions?  Would it be helpful for us to cut an image that includes both #1817 and this change for you to test?

negz:
@adleong I don't expect #1817 to fix the issues described in #1805. Unfortunately we haven't been able to reliably reproduce those issues, so I don't have an actual test case I can use.

I should be able to bake an image for this and push it out to test today.

adleong:
@negz #1828 also may or may not be related to this.  @deebo91 can you rebase this branch onto latest master so that when @negz cuts a test build off this branch, #1828 will be included?

deebo91:
Sorry @adleong, only saw this now. I will rebase right now.

siggy:
@negz this branch has been rebased on master to include #1828. Would you mind testing this?

negz:
Thanks! We're now running a build of this PR on a lightly trafficked cluster. A quick spot check shows historically problematic things (deleting and recreating ingresses, deleting and recreating services) working as expected. I'll leave this running and report back with any issues we see.

siggy:
Thanks for testing @negz ! Any update after a night of burn in?

negz:
@siggy No issues thus far! That said, the class of issues we've seen are mostly caused by cluster customers "doing stuff" i.e. creating/deleting service and ingress resources, so soaking overnight doesn't build a _huge_ amount of confidence for us yet.

I *can* say that the main service currently taking traffic in the cluster autoscales fairly frequently and our dashboards show no errors due to the changing set of backends.

negz:
https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#metadata

> resourceVersion: a string that identifies the internal version of this object that can be used by clients to determine when objects have changed. This value MUST be treated as opaque by clients and passed unmodified back to the server. Clients should not assume that the resource version has meaning across namespaces, different kinds of resources, or different servers.

I'm working with the watch API this evening and came across the above. It reads to me a little like clients are not expected to inspect the resource version.

PR #1815, opened 2018-02-15
add @Ashald to MAINTAINERS.md

wmorgan:
cc @Ashald


Ashald:
Thanks! 😍 It's a pleasure to be part of such a great thing as Linkerd and work with such amazing people as you!

PR #1816, opened 2018-02-15
Filter out dirs when listing namespaces in Consul

### LINKED ISSUE #1812: namerd admin interface hangs
## Filing a Linkerd issue ##

Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

I’ve configured namerd to work with consul and expose a namer service as mesh and it works: the dtab I set in `namerd/dtabs/default` appears instantly in all linkerds.

I can also fetch `/admin/ping`:

```
$ curl https://10.6.180.250:9991/admin/ping
pong
```

And metrics (both `/admin/metrics.json` and my configured prometheus) work too.

However if I want to fetch the admin interface itself, it hangs forever:

```
curl -v https://10.6.180.250:9991/
*   Trying 10.6.180.250...
* Connected to 10.6.180.250 (10.6.180.250) port 9991 (#0)
* found 149 certificates in /etc/ssl/certs/ca-certificates.crt
* found 687 certificates in /etc/ssl/certs
* ALPN, offering http/1.1
* SSL connection using TLS1.2 / ECDHE_RSA_AES_128_GCM_SHA256
* 	 server certificate verification OK
* 	 server certificate status verification SKIPPED
* 	 common name: XXX (matched)
* 	 server certificate expiration date OK
* 	 server certificate activation date OK
* 	 certificate public key: RSA
* 	 certificate version: #3
* 	 subject: CN=XXX
* 	 start date: Mon, 12 Feb 2018 10:34:40 GMT
* 	 expire date: Sat, 11 Feb 2023 10:35:08 GMT
* 	 issuer: CN=XXX
* 	 compression: NULL
* ALPN, server did not agree to a protocol
> GET / HTTP/1.1
> Host: 10.6.180.250:9991
> User-Agent: curl/7.47.0
> Accept: */*
>
```

There are **no** log entries neither in the consul nor in the namerd log.

Not to hang? :)

Configure as I did, curl /.

Until I got the consul configuration right, the interface *did* load with an error telling me that no namespaces were found (as a side note, the documentation around namerd and its integration with linkerd and other services was not quite up to the level of what I got used to from linkerd's docs – I needed additional guidance otherwise I’d still be stumbling).

**Environment**:

namerd.yml:

```yaml
---
admin:
  port: 9991
  ip: 10.6.180.250
  tls:
    certPath: /etc/ssl/certs/XXX.crt
    keyPath: /etc/ssl/private/XXX.pkcs8


namers:
  - kind: io.l5d.consul
    includeTag: true
    useHealthCheck: true


storage:
  kind: io.l5d.consul
  token: XXX
  readConsistencyMode: stale
  pathPrefix: /namerd/dtabs


interfaces:
  - kind: io.l5d.mesh
    ip: 10.6.180.250
    port: 4321


telemetry:
  - kind: io.l5d.prometheus
    path: /admin/metrics/prometheus
    prefix: namerd_
```

namerd: 1.3.5
consul: 1.0.1

Querying consul works without any problems:

```
$ curl -v -H "x-consul-token: XXX" "http://localhost:8500/v1/kv/?recurse=true"
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8500 (#0)
> GET /v1/kv/?recurse=true HTTP/1.1
> Host: localhost:8500
> User-Agent: curl/7.47.0
> Accept: */*
> x-consul-token: XXX
>
< HTTP/1.1 200 OK
< Content-Type: application/json
< X-Consul-Index: 137959041
< X-Consul-Knownleader: true
< X-Consul-Lastcontact: 0
< Date: Tue, 13 Feb 2018 14:34:53 GMT
< Content-Length: 524
<
[{"LockIndex":0,"Key":"namerd/dtabs/","Flags":0,"Value":null,"CreateIndex":137788444,"ModifyIndex":137788444},{"LockIndex":0,"Key":"namerd/dtabs/default","Flags":0,"Value":"L3N2YyAgICAgICAgICAgICAgID0+IC8kL2lvLmJ1b3lhbnQuaHR0cC5kb21haW5Ub1BhdGg7CgovYWcvdm0vbDVkL3Byb2QgICAgPT4gLyMvaW8ubDVkLmNvbnN1bC8ubG9jYWwvZW52LXByb2Q7Ci9hZy92bS9sNWQvb3RlICAgICA9PiAvIy9pby5sNWQuY29uc3VsLy5sb2NhbC9lbnYtb3RlOwovYWcvdm0vbDVkL3N0YWdpbmcgPT4gLyMvaW8ubDVkLmNvbnN1bC8ubG9jYWwvZW52LXN0YWdpbmc7","CreateIndex":137788454,"ModifyIndex":137801542}]
```

Ubuntu Xenial, since this is an isolation namerd issue, the rest shouldn’t matter.

LXC running on metal. More than 1 GB of RAM free.

Ashald:
In order to list namespaces available in Consul KV ConsulDtabStore makes a call like `/kv/some/path/?keys=true` that may return not only keys inside `/some/path/` but also an entry for `/some/path/` and all dirs inside `/some/path/` such as `/some/path/foo/` (but not `/some/path/foo/bar`).
This leads to an exception in API response parsing and fails listing activity so that it's pending forever.

Fixed by omiting the entry for `/some/path/` as wells as all entries ending with / and applying a "validNs" regex.

Added a test imitating a problematic Consul KV structure.

Fixes #1812

Signed-off-by: Borys Pierov <pierovbg@ncbi.nlm.nih.gov>

### DIFF
--- a/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
+++ b/namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
@@ -29,7 +29,13 @@ class ConsulDtabStore(
   }
 
   override val list: Activity[Set[Ns]] = {
-    def namespace(key: String): Ns = key.stripPrefix("/").stripSuffix("/").substring(root.show.length)
+    def namespace(key: String): Option[Ns] =
+      Some(s"/$key".stripPrefix(s"${root.show}/"))
+        // "list()" might return "root" dir or nested dirs - ignore 'em all
+        .filterNot(_.isEmpty)
+        .filterNot(_.endsWith("/"))
+        // avoid awkward situations when we cannot observe listed namespace
+        .filter(namespaceIsValid)
 
     val run = Var.async[Activity.State[Set[Ns]]](Activity.Pending) { updates =>
       @volatile var running = true
@@ -46,7 +52,7 @@ class ConsulDtabStore(
           )
             .transform {
               case Return(result) =>
-                val namespaces = result.value.map(namespace).toSet
+                val namespaces = result.value.flatMap(namespace).toSet
                 updates() = Activity.Ok(namespaces)
                 cycle(result.index, backoffs0)
               case Throw(e: NotFound) =>
--- a/namerd/storage/consul/src/test/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreTest.scala
+++ b/namerd/storage/consul/src/test/scala/io/buoyant/namerd/storage/consul/ConsulDtabStoreTest.scala
@@ -12,7 +12,8 @@ import io.buoyant.test.{ActivityValues, Awaits, Exceptions, FunSuite}
 
 class ConsulDtabStoreTest extends FunSuite with Awaits with Exceptions with ActivityValues {
 
-  val namespacesJson = """["namerd/dtabs/foo/bar/", "namerd/dtabs/foo/baz/"]"""
+  val namespacesJson = """["namerd/dtabs/foo", "namerd/dtabs/bar"]"""
+  val namespacesWithDirsJson = """["namerd/dtabs/", "namerd/dtabs/foo", "namerd/dtabs/bar/"]"""
   val namerdPrefix = "/namerd/dtabs"
   val constBackoff = Backoff.const(Duration.Zero)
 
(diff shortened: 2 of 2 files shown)


Ashald:
Decided to celebrate becoming Linkerd maintainer by fixing my own bug from 2016. 🙂

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -29,7 +29,13 @@ class ConsulDtabStore(
   }
 
   override val list: Activity[Set[Ns]] = {
-    def namespace(key: String): Ns = key.stripPrefix("/").stripSuffix("/").substring(root.show.length)
+    def namespace(key: String): Option[Ns] =
+      Some(s"/$key".stripPrefix(s"${root.show}/"))
```
adleong: This might benefit from an example to help see what's going on here.

If `key` is `/namerd/dtabs/foo` and `root` is `/namerd/dtabs`, then doesn't `s"/$key"` start with a double slash?

Ashald: I tried to illustrate that by an example in tests but probably that's to subtle.

Consul KV always returns paths without a leading `/` so we never going to get a double slash here. Should I add a comment explaining that?

Another way to achieve same behavior would be:
```scala
def namespace(key: String): Option[Ns] =
  Some(s"/$key")  // $key never has a leading slash
    .filterNot(_.endsWith("/")) // filter out dirs
    .map(_.stripPrefix(s"${root.show}/")) // strip $root
    .filter(namespaceIsValid)
```

klingerf:
I can confirm that the fixes from this branch fix the issue that I was seeing locally. 🎉  Will defer to @adleong for code review.

PR #1818, opened 2018-02-15
Accept DELETED events with equal resource versions

hawkw:
As @negz noted in https://github.com/linkerd/linkerd/pull/1810#issuecomment-366022516:
> my understanding is that `MODIFIED` objects will have a higher `resourceVersion`, while `DELETED` objects will have an unchanged `resourceVersion` because they return "the state of the object immediately before deletion".

I've modified the ordering defined on watch events so that `DELETED` events are considered "greater" than other events with the same resource version. This should prevent Linkerd from skipping `DELETED` events, while otherwise preventing time travel for other event types.

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
@@ -175,11 +175,21 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
           // Register the update only if its resource version is larger than or equal to the largest version
           // seen so far.
           if (largestEvent.forall(_ <= event)) {
-            log.debug(s"k8s returned a new resource: $event")
+            log.trace(
+              "k8s watch on '%s' registered event with same or newer resource version %s (older resource version was %s)",
+              watchPath,
+              event.resourceVersion,
+              largestEvent.flatMap(_.resourceVersion)
+            )
             state.update(Activity.Ok(event))
             _processEventStream(ws, event.resourceVersion, Some(event))
           } else {
-            log.debug(s"k8s returned an old resource. Using most up to date resource: $event")
+            log.trace(
+              "k8s watch on '%s' skipping event with resource version %s (older than the most recent resource version %s)",
+              watchPath,
+              event.resourceVersion,
+              largestEvent.flatMap(_.resourceVersion)
+            )
             _processEventStream(ws, largestVersion, largestEvent)
           }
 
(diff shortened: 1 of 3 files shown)


k8s/src/test/scala/io/buoyant/k8s/v1/ApiTest.scala
```diff
@@ -488,13 +491,136 @@ class ApiTest extends FunSuite
...
+
+      val Activity.Ok(event: EndpointsWatch) = await(events.poll())
+      assert(event.versionNum.value == resourceVersion1)
+      val EndpointsDeleted(eps) = event
+      assert(eps.subsets.get.flatMap(_.addresses).flatten.map(_.ip) ==
+        Seq("10.248.2.8", "10.248.7.10", "10.248.8.8"))
+    }
+
+    // write an event with higher version than the delete event.
+    await(w.write(modified3))
```
negz: Shouldn't we normally see an `ADDED` after a `DELETED` before any `MODIFIED`, assuming we had an uninterrupted watch? I don't think it matters for the purposes of this test, just curious if I misunderstand how things work.

negz:
Thanks for working on this! FWIW I'm happy to switch our clusters over to running a nightly build once this is in to soak all the new changes before 1.3.6.

k8s/src/test/scala/io/buoyant/k8s/v1/ApiTest.scala
hawkw: Yeah, I believe that's true --- this should probably be an `ADDED` event. Since Linkerd handles those events in essentially the same way, I just wanted to reuse the existing test data, but it would probably be more accurate to send an `ADDED` event here.

k8s/src/main/scala/io/buoyant/k8s/Watchable.scala
```diff
@@ -173,11 +173,23 @@ private[k8s] abstract class Watchable[O <: KubeObject: TypeReference, W <: Watch
...
+            log.trace(
+              "k8s watch on '%s' registered event with newer resource version %s (older resource version was %s)",
+              watchPath,
+              event.resourceVersion,
+              largestEvent.flatMap(_.resourceVersion)
+            )
             state.update(Activity.Ok(event))
             _processEventStream(ws, event.resourceVersion, Some(event))
           } else {
+            log.trace(
```
negz: Friendly reminder about potentially making this a `log.warn`  (or `info`?), assuming this PR will supercede #1810. :)

hawkw: 👍

hawkw: 👍

adleong:
Given that redundant updates when the resource versions are equal are probably not a problem, how do we feel about closing this in favor of #1810?

hawkw:
@adleong sounds good to me!

PR #1822, opened 2018-02-16
Backoff before retrying on unexpected erros

### LINKED ISSUE #1821: no backoff in consul namer on unexpected exceptions if there's previous state and on ConnectionFailedException
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

Every name in consul namer is backed by an infinite loop in `SvcAddr`, that obtains connection to consul and long polls for updates.
Some of the erros are handled by `infiniteRetryFilter` inside of buoyant consul API, and in this case back-off is used.

Other errors are handled in infinite loop inside of `SvcAddr` and do NOT have any back-offs.

In production environment this means that in case of specific intermittent issues with host where namerd is running or with consul agent it talks to, namerd will retry without back-off, potentially causing more harm to the infrastructure.

Back-off are used for all type of retries

```
--- a/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
+++ b/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
@@ -686,4 +686,23 @@ class ConsulNamerTest extends FunSuite with Awaits {
     assert(stats.counters.get(Seq("service", "updates")).contains(1))
     assert(stats.counters.get(Seq("lookups")).contains(1))
   }
+
+  test("Namer doesn't use back-off") {
+    class TestApi extends CatalogApi(null, "/v1") {
+      override def serviceNodes(
+        serviceName: String,
+        datacenter: Option[String] = None,
+        tag: Option[String] = None,
+        blockingIndex: Option[String] = None,
+        consistency: Option[ConsistencyMode] = None,
+        retry: Boolean = false
+      ): Future[Indexed[Seq[ServiceNode]]] = Future.exception(Failure("failed", new ConnectionFailedException()))
+    }
+    val stats = new InMemoryStatsReceiver
+    val namer = ConsulNamer.untagged(testPath, new TestApi(), new TestAgentApi("acme.co"), stats = stats)
+    @volatile var state: Activity.State[NameTree[Name]] = Activity.Pending
+
+    namer.lookup(Path.read("/dc1/servicename/residual")).states respond { state = _ }
+  }
+
 }
```

This is of course synthetic. On practice `ConnectionFailedException` is thrown if new port can't be allocated to make a request to consul.

Also any exception thrown by `ConsulApi`/`HealthApi` will cause retry without back-off if there is previously known state in `SvcAddr`.

**Environment**:
1.3.4

edio:
Consul namer uses ConsulApi with retry=true to
retry on certain types of errors, and ConsulApi
uses backoffs while retrying. However if error
is not handled by ConsulApi, it is propagated as
exception to SvcAddr and retry is attemted there
without backoff. This is potentially dangerous.

This commit makes SvcAddr to use backoffs stream
provded by ConsulApi.

Fixes #1821

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -7,6 +7,8 @@ import io.buoyant.consul.v1
 import io.buoyant.namer.Metadata
 import java.net.InetSocketAddress
 
+import com.twitter.finagle.util.DefaultTimer
+
 import scala.util.control.NoStackTrace
 
 private[consul] case class SvcKey(name: String, tag: Option[String]) {
@@ -42,6 +44,7 @@ private[consul] object SvcAddr {
     tagWeights: Map[String, Double] = Map.empty,
     stats: Stats
   ): Var[Addr] = {
+    def initialBackoffs = consulApi.backoffs
     val meta = mkMeta(key, datacenter, domain)
     def getAddresses(index: Option[String]): Future[v1.Indexed[Set[Address]]] =
       consulApi.serviceNodes(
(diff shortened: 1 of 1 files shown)


edio:
I'm working in different PR, that will not rely on retry logic in consul api but will handle everything in SvcAddr. Closing this one.

PR #1826, opened 2018-02-20
Adding support for DNS SAN names in XFCC.

### LINKED ISSUE #1802: x-forwarded-client-cert doesn't forward SAN DNS names, only one URIName record is forwarded

## Getting Help ##

## Filing a Linkerd issue ##

Issue Type:

- [ ] Bug report
- [x] Feature request

**What happened**:
x-forwarded-client-cert containing SAN DNS names are not forwarded, only URIName records are forwarded. 

https://github.com/linkerd/linkerd/blob/b390b5b16549ccb2d34d00480390364517ab9246/router/base-http/src/main/scala/io/buoyant/router/http/ForwardClientCertFilter.scala#L30

The ForwardClientCertFilter explicitly filters out SAN name types other than URI. For mutual TLS authentication, we would also need the DNS name type related alt name to be forwarded as part of XFCC SAN string: http://www.docjar.com/html/api/sun/security/x509/GeneralNameInterface.java.html

SAN DNS name type values should also be passed on as part of the SAN string in the XFCC header.

Create a certificate which has DNS="something" entries in the SAN. Run the ForwardClientCert and you can notice.

**Environment**:
1.3.5

shakti-das:
Signed-off-by: Shakti Das shakti.das2002@gmail.com

Problem
The x-forwarded-client-cert header only forwards the SAN entry of type URI. SAN entries of type DNS are also used for TLS and mTLS verification. We need to support the same for XFCC header so that client implementations can use SAN DNS entries as well for certificate verification.

Solution
The solution as discussed in the issue conversation is to add DNS SAN entries directly to the XFCC header. So, after the change, the XFCC header would look like:
`Hash=468ed33be74eee6556d90c0149c1309e9ba61d6425303443c0748a02dd8de688;SAN=https://example.com;DNS=example.com;DNS=www.example.com`

Validation
The TLS related tests were run to verify it causes no regression. The tests were modified to verify propagation of DNS SAN entries in XFCC. 

Fixes #1802

### DIFF
--- a/linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsBoundPathTest.scala
+++ b/linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsBoundPathTest.scala
@@ -212,8 +212,8 @@ class TlsBoundPathTest extends FunSuite with Awaits {
   }
 
   test("wrong common name") {
-
-    withCerts("bill.buoyant.io", "ted.buoyant.io") { certs =>
+    // Create certificates without SAN DNS entries that override CN.
+    withCertsWithCustomDnsAltNames(Seq("bill.buoyant.io", "ted.buoyant.io"), null) { certs =>
       val bill = Downstream
         .constTls(
           "bill",
--- a/linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
+++ b/linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
@@ -1,7 +1,6 @@
 package io.buoyant.linkerd.tls
 
 import java.io.{File, FileWriter}
-
 import scala.collection.mutable
 import scala.sys.process._
 
@@ -95,8 +94,7 @@ object TlsUtils {
   }
 
   def dnsAltNames(names: Seq[String]): String = {
-    val altNames = new Array[String](names.length)
-    names.indices.foreach(i => altNames(i) = "DNS." + (i + 1) + " = " + names(i))
+    val altNames = names.zipWithIndex.map { case(name, i) => s"DNS.${i+1} = $name" }
     altNames.mkString("\n")
   }
 
--- a/linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
+++ b/linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
@@ -26,19 +26,21 @@ object TlsUtils {
   case class Certs(caCert: File, serviceCerts: Map[String, ServiceCert])
 
   def withCerts(names: String*)(f: Certs => Unit): Unit = {
-    withCerts(false, names)(f: Certs => Unit)
+    // By default add DNS SAN entries for upstream/downstream certificates.
+    // Important: DNS SAN entries overrides CN as per RFC 6125. So, if adding CN specific logic, use the method {@link #withCertsWithCustomDnsAltNames}
+    // Reference: https://stackoverflow.com/questions/5935369/ssl-how-do-common-names-cn-and-subject-alternative-names-san-work-together
+    withCertsWithCustomDnsAltNames(names, names)(f: Certs => Unit)
   }
 
-  def withCerts(addDnsAltNames: Boolean, names: Seq[String])(f: Certs => Unit): Unit = {
+  def withCertsWithCustomDnsAltNames(names: Seq[String], dnsAltNames: Seq[String])(f: Certs => Unit): Unit = {
     // First, we create a CA and get a cert/key for linker
     val tmpdir = new File("mktemp -d -t linkerd-tls.XXXXXX".!!.stripLineEnd)
     try {
       val configFile = mkCaDirs(tmpdir)
-      if (addDnsAltNames) {
-        // When we add DNS SAN (Subject Alternative Name) names, it overrides CN (Common Name).
-        // So, tests which modify CN to simulate failure scenarios will start failing if this is added for all.
-        // Reference: https://stackoverflow.com/questions/5935369/ssl-how-do-common-names-cn-and-subject-alternative-names-san-work-together
-        addDnsAltNamesInConfig(tmpdir, names)
+      if (dnsAltNames != null) {
+        // More granular way of doing this would be to use -addext option in openssl command once it's available.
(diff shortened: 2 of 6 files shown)


shakti-das:
@briansmith @wmorgan @adleong please review the changes.

wmorgan:
Thanks for the PR! We'll take a look!

linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
```diff
@@ -22,11 +24,22 @@ object TlsUtils {
 
   }
   case class Certs(caCert: File, serviceCerts: Map[String, ServiceCert])
+
   def withCerts(names: String*)(f: Certs => Unit): Unit = {
+    withCerts(false, names)(f: Certs => Unit)
```
briansmith: `true` should be the default value. If I understand your comment below, some tests will fail if `true` is used. I think that instead we should modify *those* tests to explicitly pass in `false`.

In particular, we should move away from using the subject CN for anything except as a fallback when the DNS names aren't provided, following RFC 6125.

linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
```diff
@@ -128,6 +148,12 @@ object TlsUtils {
     |URI.1 = https://buoyant.io
     |""".stripMargin
 
+  def dnsAltNames(names: Seq[String]): String = {
```
briansmith: NIT: I would put this function right under `addDnsAltNamesInConfig` so that all this directly-related code is together.

router/base-http/src/main/scala/io/buoyant/router/http/ForwardClientCertFilter.scala
```diff
@@ -24,11 +25,15 @@ class ForwardClientCertFilter[Req, H: HeadersLike, Rep](implicit requestLike: Re
...
-            val nameValue = altName.get(1)
-            if (nameType == GeneralNameTypeUri) {
-              clientCertHeader ++= s";SAN=$nameValue"
+            val it = altNames.iterator
+            while (it.hasNext) {
+              val altName = it.next
+              val nameType = altName.get(0)
+              val nameValue = altName.get(1)
+              nameType match {
+                case GeneralNameTypeUri => clientCertHeader ++= s";SAN=$nameValue"
```
briansmith: Please add a comment here that says something like `Use "SAN:" instead of "URI:" for backward compatibility with previous releases.`

linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
```diff
@@ -22,11 +24,22 @@ object TlsUtils {
...
+
+  def withCerts(addDnsAltNames: Boolean, names: Seq[String])(f: Certs => Unit): Unit = {
     // First, we create a CA and get a cert/key for linker
     val tmpdir = new File("mktemp -d -t linkerd-tls.XXXXXX".!!.stripLineEnd)
     try {
       val configFile = mkCaDirs(tmpdir)
+      if (addDnsAltNames) {
+        // When we add DNS SAN (Subject Alternative Name) names, it overrides CN (Common Name).
+        // So, tests which modify CN to simulate failure scenarios will start failing if this is added for all.
+        // Reference: https://stackoverflow.com/questions/5935369/ssl-how-do-common-names-cn-and-subject-alternative-names-san-work-together
```
briansmith: If you make `true` the default for `addDnsAltNames` then you can just remove this comment. Instead, add a comment at each `withCerts(false, ...)` call site explaining that the test is explicitly testing Subject CN behavior and not SAN behavior.

linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
shakti-das: Sounds good, will make changes accordingly.

shakti-das:
@briansmith updated the PR to add DNS SAN entries by default.

linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/TlsBoundPathTest.scala
```diff
@@ -212,8 +212,8 @@ class TlsBoundPathTest extends FunSuite with Awaits {
   }
 
   test("wrong common name") {
-
-    withCerts("bill.buoyant.io", "ted.buoyant.io") { certs =>
+    // Create certificates without SAN DNS entries that override CN.
+    withCertsWithCustomDnsAltNames(Seq("bill.buoyant.io", "ted.buoyant.io"), null) { certs =>
```
shakti-das: Not so happy with this method name, if you have a better name suggestion, plz recommend.

shakti-das:
@briansmith @wmorgan @adleong sorry to pester you guys, but am currently in the flow with this one so some new review comments would help me proceed quickly. Thanks.

linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
```diff
@@ -72,6 +87,19 @@ object TlsUtils {
...
 
+  def addDnsAltNamesInConfig(dir: File, names: Seq[String]): Unit = {
+    val configFile = new File(dir, "openssl.cfg")
+    val cw = new FileWriter(configFile, true)
+    cw.write(dnsAltNames(names))
+    cw.close()
+  }
+
+  def dnsAltNames(names: Seq[String]): String = {
+    val altNames = new Array[String](names.length)
```
adleong: I think this can be written more simply as:

```
val altNames = names.zipWithIndex.map { case(name, i) => s"DNS.${i+1} = $name" }
altNames.mkString("\n")
```

linkerd/tls/src/main/scala/io/buoyant/linkerd/tls/TlsUtils.scala
```diff
@@ -1,6 +1,8 @@
 package io.buoyant.linkerd.tls
 
-import java.io.File
+import java.io.{File, FileWriter}
+
```
adleong: Linkerd import style is all imports grouped together (no blank lines) in alphabetical order.

shakti-das:
Updated PR with the Scala code style suggestions.

adleong:
Thank *you* for this great contribution!

Issue #1820, opened 2018-02-16
unexpected message about "service observation released" in logs


edio:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

Quite often message in logs appears

```
E 0212 04:58:19.591 UTC THREAD15 TraceId:6ee3dc947b0726a6: consul datacenter '******' service '******' observation error Failure(service observation released, flags=0x02) with RemoteInfo -> Upstream Address: Not Available, Upstream id: Not Available, Downstream Address: localhost/127.0.0.1:8500, Downstream label: client, Trace Id: 2ab10a2a5b11c7ac.2ab10a2a5b11c7ac<:2ab10a2a5b11c7ac, no previous good state to fall back to!
```

**What you expected to happen**:
`service observation release` should be handled differently

**How to reproduce it (as minimally and precisely as possible)**:
Can't reproduce manually

**Anything else we need to know?**:

Failure should have been handled by a dedicated case of pattern matching

```
          case Throw(Failure(Some(cause))) if cause == ServiceRelease =>
            // this exception is raised when we close a watch - thus, it needs
            // to be special-cased so that we don't continue observing that
            // service.
            log.trace(
              "consul datacenter '%s' service '%s' observation closed",
              datacenter,
              key.name
            )
            stopped = true
            Future.Unit
```

But judging from log message, the Failure is handled by case that matches all `Throw` instances

```
          case Throw(e) =>
            // an error occurred. if we've previously seen a good state, fall
            // back to that state; otherwise, fail.
            stats.errors.incr()
            lastGood match {
              case Some(addr) =>
                // if we've already seen a good state, fall back to that and
                // try again.
                state() = addr
                log.warning(
                  "consul datacenter '%s' service '%s' observation error %s," +
                    " falling back to last good state",
                  datacenter, key.name, e
                )
                loop(index0)
              case None =>
                // if no previous good state was seen, treat the exception
                // as effectively fatal to the service observation.
                state() = Addr.Failed(e)
                log.error(
                  "consul datacenter '%s' service '%s' observation error %s," +
                    " no previous good state to fall back to!",
                  datacenter, key.name, e
                )
                Future.exception(e)
            }
```

- linkerd/namerd version, config files:
namerd 1.3.4

edio:
This is related to #1670, but is way more bizarre

edio:
[...]

PR #1827, opened 2018-02-20
Do not rely on ConsulApi retries in consul namer

### LINKED ISSUE #1821: no backoff in consul namer on unexpected exceptions if there's previous state and on ConnectionFailedException

edio:
Consul namer uses ConsulApi with retry=true to
retry on certain types of errors, backoffs is used
by ConsulApi while retrying. However if error is
not handled by ConsulApi, it is propagated as
exception to SvcAddr and retry is attemted there
without backoff. This is potentially dangerous.

This commit delegates retires fully to SvcAddr

Fixes #1821

Signed-off-by: Dmytro Kostiuchenko <kostiuchenkod2@ncbi.nlm.nih.gov>

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
@@ -1,12 +1,16 @@
 package io.buoyant.namer.consul
 
-import com.twitter.finagle.Address.Inet
+import com.twitter.conversions.time._
 import com.twitter.finagle._
+import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.stats.{NullStatsReceiver, StatsReceiver}
 import com.twitter.util._
 import io.buoyant.consul.v1
-import io.buoyant.namer.Metadata
 import scala.Function.untupled
+
+private[consul] object LookupCache {
+  val DefaultBackoffs: Stream[Duration] = Backoff.exponentialJittered(10.milliseconds, 5.seconds)
+}
 
 /**
  * A helper supporting service resolution in consul, caching
@@ -34,6 +38,7 @@ private[consul] class LookupCache(
           case ((dcName, domainOption)) =>
             val addr = SvcAddr(
               consulApi,
+              LookupCache.DefaultBackoffs,
               dcName,
               key,
               domainOption,
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -2,11 +2,12 @@ package io.buoyant.namer.consul
 
 import com.twitter.finagle._
 import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.logging.Level
 import com.twitter.util._
 import io.buoyant.consul.v1
 import io.buoyant.namer.Metadata
 import java.net.InetSocketAddress
-
 import scala.util.control.NoStackTrace
 
 private[consul] case class SvcKey(name: String, tag: Option[String]) {
@@ -34,6 +35,7 @@ private[consul] object SvcAddr {
    */
   def apply(
     consulApi: v1.ConsulApi,
+    apiBackoffs: Stream[Duration],
     datacenter: String,
     key: SvcKey,
     domain: Option[String],
@@ -41,7 +43,7 @@ private[consul] object SvcAddr {
     preferServiceAddress: Option[Boolean] = None,
     tagWeights: Map[String, Double] = Map.empty,
     stats: Stats
-  ): Var[Addr] = {
+  )(implicit timer: Timer = DefaultTimer): Var[Addr] = {
     val meta = mkMeta(key, datacenter, domain)
     def getAddresses(index: Option[String]): Future[v1.Indexed[Set[Address]]] =
       consulApi.serviceNodes(
@@ -50,21 +52,18 @@ private[consul] object SvcAddr {
         tag = key.tag,
         blockingIndex = index,
         consistency = consistency,
-        retry = true
+        retry = false
       ).map(indexedToAddresses(preferServiceAddress, tagWeights))
 
     // Start by fetching the service immediately, and then long-poll
     // for service updates.
     Var.async[Addr](Addr.Pending) { state =>
       stats.opens.incr()
-      // last good state - if an error occurs, fall back to this if
-      // a good state was seen.
-      @volatile var lastGood: Option[Addr] = None
       @volatile var stopped: Boolean = false
-      def loop(index0: Option[String]): Future[Unit] = {
+      def loop(blockingIndex: Option[String], backoffs: Stream[Duration], failureLogLevel: Level, currentValueToLog: Addr): Future[Unit] = {
 
         if (stopped) Future.Unit
-        else getAddresses(index0).transform {
+        else getAddresses(blockingIndex).transform {
           case Throw(Failure(Some(cause))) if cause == ServiceRelease =>
             // this exception is raised when we close a watch - thus, it needs
             // to be special-cased so that we don't continue observing that
@@ -76,45 +75,24 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
-          case Throw(Failure(Some(err: ConnectionFailedException))) =>
-            log.warning(
-              "consul datacenter '%s' service '%s' retrying on connection" +
-                " failure: %s",
-              datacenter, key.name, err
-            )
-            // Drop the index, in case it's been reset by a consul restart
-            loop(None)
+
           case Throw(e) =>
-            // an error occurred. if we've previously seen a good state, fall
-            // back to that state; otherwise, fail.
+            // do not update state, log error and continue polling with backoff
             stats.errors.incr()
-            lastGood match {
-              case Some(addr) =>
-                // if we've already seen a good state, fall back to that and
-                // try again.
-                state() = addr
-                log.warning(
-                  "consul datacenter '%s' service '%s' observation error %s," +
-                    " falling back to last good state",
-                  datacenter, key.name, e
-                )
-                loop(index0)
-              case None =>
-                // if no previous good state was seen, treat the exception
-                // as effectively fatal to the service observation.
-                state() = Addr.Failed(e)
-                log.error(
-                  "consul datacenter '%s' service '%s' observation error %s," +
-                    " no previous good state to fall back to!",
-                  datacenter, key.name, e
-                )
-                Future.exception(e)
-            }
+            log.log(
+              failureLogLevel,
+              "consul datacenter '%s' service '%s' observation error %s." +
+                " Last known state is %s",
+              datacenter, key.name, e, currentValueToLog
+            )
+            val backoff #:: nextBackoffs = backoffs
+            // subsequent errors are logged as DEBUG
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
 
           case Return(v1.Indexed(_, None)) =>
             // If consul doesn't return an index, we're in bad shape.
             // TODO: do we want to revert to the last good state here, as well?
-            state() = Addr.Failed(NoIndexException)
+            state.update(Addr.Failed(NoIndexException))
             stats.errors.incr()
             log.error(
               "consul datacenter '%s' service '%s' didn't return an index!",
@@ -122,20 +100,19 @@ private[consul] object SvcAddr {
             )
             Future.exception(NoIndexException)
 
-          case Return(v1.Indexed(addrs, index1)) =>
+          case Return(v1.Indexed(addrs, xConsulIndex)) =>
             stats.updates.incr()
             val addr = addrs match {
               case addrs if addrs.isEmpty => Addr.Neg
               case addrs => Addr.Bound(addrs, meta)
             }
-            lastGood = Some(addr)
-            state() = addr
+            state.update(addr)
 
-            loop(index1)
+            loop(xConsulIndex, apiBackoffs, Level.WARNING, addr)
         }
       }
 
-      val pending = loop(None)
+      val pending = loop(None, apiBackoffs, Level.WARNING, Addr.Pending)
       Closable.make { _ =>
         stopped = true
         stats.closes.incr()
--- a/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
+++ b/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
@@ -8,7 +8,6 @@ import io.buoyant.consul.v1._
 import io.buoyant.namer.{ConfiguredDtabNamer, Metadata}
 import io.buoyant.test.Awaits
 import org.scalatest.FunSuite
-import org.scalatest.exceptions.TestFailedException
 
 class ConsulNamerTest extends FunSuite with Awaits {
 
@@ -81,7 +80,7 @@ class ConsulNamerTest extends FunSuite with Awaits {
     ))
   }
 
-  test("Namer fails if the consul api cannot be reached") {
+  test("Namer stays pending if the consul api cannot be reached") {
     class TestApi extends CatalogApi(null, "/v1") {
       override def serviceNodes(
         serviceName: String,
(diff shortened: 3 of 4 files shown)


edio:
All errors are handled in the same fashion: if previous state exists, we treat error as intermittent and continue watching the service. If no previous state exists, we terminate immediately.

Main differences in behavior
1) On connection exception we no longer try observing with the same blocking index. Assumption is that while consul was unreachable, service catalog state may have changed anyway, so it is safe to retry with index. And the cost is just one single request

2) On request to non-existent DC we no longer monitor address infinitely. We do this only if succeeded in the past.

3) All errors are retried with back-off

4) Only first error in a row for each name is logged as WARNING, rest are logged as DEBUG to reduce noise in logs.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -61,7 +65,7 @@ private[consul] object SvcAddr {
       // a good state was seen.
       @volatile var lastGood: Option[Addr] = None
       @volatile var stopped: Boolean = false
-      def loop(index0: Option[String]): Future[Unit] = {
+      def loop(index0: Option[String], backoffs: Option[Stream[Duration]], nonFatalLogLevel: Level = Level.WARNING): Future[Unit] = {
```
adleong: Can we make this always take a `Stream[Duration]` instead of an `Option[Stream[Duration]]` and pick the default backoffs at the call sites where appropriate?  Similarly, I think we should not provide a default log level here and be more explicit about the log level at the call sites.   Since this is a local method, this should be pretty easy.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -76,29 +80,23 @@ private[consul] object SvcAddr {
...
-                log.warning(
+                // do not update state if we were in good state before error
+                log.log(
+                  nonFatalLogLevel,
                   "consul datacenter '%s' service '%s' observation error %s," +
                     " falling back to last good state",
                   datacenter, key.name, e
                 )
-                loop(index0)
+                sleepAndDo(backoffs.getOrElse(apiBackoffs)) { nextBackoffs =>
```
adleong: I think it's more idiomatic to do something like

```
Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG))
```

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -76,29 +80,23 @@ private[consul] object SvcAddr {
...
+                  // retry and log subsequent errors as debug until state changes
+                  loop(None, Some(nextBackoffs), Level.DEBUG)
+                }
               case None =>
```
adleong: I'm not sure that is makes sense to die here if there is no previous good state, especially given that we are removing the retry filter.  In the case of an error when there is no previous good state, I think we should not set `state()` (e.g. leave it as pending) and do a retry with backoff, just like the case above.  

@hawkw what do you think?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -145,6 +143,12 @@ private[consul] object SvcAddr {
     }
   }
 
+  private[this] def sleepAndDo[A](backoffs: Stream[Duration])(action: Function1[Stream[Duration], Future[A]]): Future[A] = {
```
adleong: I don't think we need this method.

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -106,17 +107,19 @@ class ConsulNamerTest extends FunSuite with Awaits {
     ))
   }
 
-  test("Namer goes pending when dc does not exist") {
+  test("Namer stops watching address when dc does not exist and there's no previous state") {
```
adleong: Given my comment above, I'm not sure this is the desirable behavior...

namer/consul/src/test/scala/io/buoyant/namer/consul/SvcAddrTest.scala
```diff
@@ -0,0 +1,90 @@
+package io.buoyant.namer.consul
+
+import java.util.concurrent.atomic.AtomicInteger
+
```
adleong: group imports together

namer/consul/src/test/scala/io/buoyant/namer/consul/SvcAddrTest.scala
```diff
@@ -0,0 +1,90 @@
...
+      }
+    }
+    // retry 2 times quickly and then hang
+    val quickRetries = new Promise[Unit]
+    lazy val hang: Duration = { quickRetries.setDone(); 999999.seconds }
+    val backoffs: Stream[Duration] = Stream(10.millis, 10.millis) #::: hang #:: Stream.empty
+
+    // when
+    val addr: Var[Addr] = SvcAddr(api, backoffs, "dc1", SvcKey("svc", None), None, None, None, Map.empty, Stats(NullStatsReceiver))
+    addr.changes.respond { s => println(s) }
```
adleong: remove printlns please

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -1,13 +1,18 @@
...
 
-import com.twitter.finagle.Address.Inet
+import com.twitter.conversions.time._
 import com.twitter.finagle._
+import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.stats.{NullStatsReceiver, StatsReceiver}
 import com.twitter.util._
 import io.buoyant.consul.v1
-import io.buoyant.namer.Metadata
+
```
adleong: Linkerd style is to group all imports together (no blank lines) in alphabetical order

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
edio: I was having a specific case in mind: we have security scanner running periodically (I recall seeing someone else among linkerd users with similar setup). Every now and then scanner queries for some random addresses. I'd prefere those to not be accumulated over time causing unnecessary load.
Also requests to nonexistent DCs, I believe, shouldn't be retried (although, they were previously --- in infinite retry filter). It's seems unlikely, that there are envs where DC in consul may suddenly appear after request came and thus transitioning to pending with retries will likely give no benefit ever.

adleong: Fair points.  It's hard to differentiate between an invalid request (such as from a security scanner or for a nonexistent DC) from a transient failure that should be retried.  Do you know what Linkerd's behavior is when it falls into this case?  Will it cache the Addr.Failed and continue to fail subsequent requests without retrying the consul API call?  If so, I don't think that's the behavior that we want since it would effectively make transient failures put Linkerd into a "stuck" state for that name.

Regarding unnecessary load, Linkerd will stop watching these names once they have been idle for 10 minutes, so the load won't be unbounded.  And hopefully the exponential backoff will mean that the added load from these API calls will be not too much.

edio: @adleong ugh, I didn't think of that. No I do not know what is the linkerd behavior.
I'll check this later.

edio: I looked, how Addr.Failed will be handled by Linkerd. Here what happens to `Addr.Failed` in case if mesh interpreter is used:

1. `SvcAddr`: returns `Addr.Failed`
2. `LookupCache`: transforms it into `Activity.Failed[NameTree[Name]]`. Value gets cached through Memoize (sic!)
3. `ConsulNamer` and farther: same
4. `mesh.InterpreterService.toBoundTreeRspEv`: transform to `VarEventStream.End(Throw(e))` (terminal event, containing either value or error)

  wire

5. `mesh.Client.streamActivity`: transforms End event into `Activity.Failed[NameTree[Name]]`
6. `Dst`: transforms to `Activity.Failed[BoundTree]`
7. `DynBoundFactory`: internal state transitions from Pending -> Failed. On `apply()` factory will return `Future.exception` (DunBoundFactory:81)
8. `ServiceFactoryCache`: wraps `DynBoundFactory` decorated with `NoBrokersAvailableException` handling into `IdlingFactory`
9. `IdlingFactory`: returns same `Future.exception`. Value remains in cache for TTL if idle
10. `RoutingService`: invokes `ServiceFactoryCache` and returns `Future.exception` as response. Every invocation resets idle timer, so as long as there are requests, `ServiceFactory` remains "stuck" in cache.

So it seems that
1. Even implementation of consul namer in master can put linkerd in "stuck" state (no previous state in namer, no index in response from consul). Obviously implementation from this PR will do this on any error. To get out of the "stuck" state linkerd has to be restarted or failing activity should be evicted from cache.
2. Considering that activities are memoized in namer, we also put namerd in "stuck" state. Looks like we should never interrupt a loop in activity (except for cases of explicit interruption). Theoretically[*], it can recover on its own, if all observes to activity are gone and activity is closed. Then the next observer will start the loop again.

Please, let me know if I missed something.

[*] during my experiments, I noticed that activity is not always closed as soon as all supposed observers are gone. I'd guess, there's another cache somewhere (maybe in http interface?). I had no time to dig into this however.

I'm going to change code to retry with back-off in case of exception.
I believe I should also change the case, when consul returns no index, but let me know if I misunderstood something. (Btw, why such case even exists? I understand, we can't do long poll on consul w/o index, but can't we just continue polling with some constant delay?)

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -21,6 +22,8 @@ private[consul] object SvcAddr {
   private[this] val ServiceRelease =
     new Exception("service observation released") with NoStackTrace
 
+  implicit private[this] val timer: Timer = DefaultTimer
```
adleong: This will be more testable if you pass the Timer in as a param into apply.  This allows us to pass in a MockTimer in tests.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -59,9 +63,9 @@ private[consul] object SvcAddr {
       stats.opens.incr()
       // last good state - if an error occurs, fall back to this if
```
adleong: I think this comment needs to be updated.  As far as I can tell, `lastGood` is only used for logging purposes now

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -76,40 +77,19 @@ private[consul] object SvcAddr {
...
-                    " no previous good state to fall back to!",
-                  datacenter, key.name, e
-                )
-                Future.exception(e)
-            }
+            log.log(
+              failureLogLevel,
+              "consul datacenter '%s' service '%s' observation error %s." +
+                " Last known state is %s",
+              datacenter, key.name, e, lastGood
```
hawkw: Is the `lastGood` variable still necessary, or can we just use the current state of the activity for logging the last good state?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -76,40 +77,19 @@ private[consul] object SvcAddr {
...
+            )
+            val backoff #:: nextBackoffs = backoffs
+            // subsequent errors are logged as DEBUG
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG))
```
hawkw: Passing the log level so that it's changed to debug on subsequent calls is very elegant, I like that 👍

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
edio: In general I agree, that storing same state in 2 places isn't very pretty.

However I don't see how can we get rid of `lastGood` if we want to log last state. We handle errors inside the loop and there we only have access to `Updatable`. Even conceptually it's tricky: closure with this loop is an argument to `Var`'s "constructor", so to get current value of `Var` one would need to populate a reference to yet unconstructed `Var` to the closure that is used to construct the `Var` instance.

We still can omit the state while logging. Imo it may be usefull during troubleshooting if any sort of outage happens, but it's something super important.

hawkw: I definitely agree that logging the state is likely to be useful, and I think you're right that there's no easy way to avoid having the `lastGood` variable --- just wanted to make sure there wasn't an obvious way to remove it.

Ashald:
I'm trying to get to review this PR but cannot make it out of meetings.. Hope to be done today.

Ashald:
In `LookupCache` we still have `agentConfig` that relies on `ConsulApi`'s retries. I wonder if it makes sense to generalize back-offs approach used in the namer and use it for that call as well. I think I saw someone suggesting getting rid of "InfiniteRetryFilter" for `ConsulApi` and if there were to happen we would need to reconsider how we handle `agentConfig`.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -50,18 +52,17 @@ private[consul] object SvcAddr {
...
 
     // Start by fetching the service immediately, and then long-poll
     // for service updates.
     Var.async[Addr](Addr.Pending) { state =>
       stats.opens.incr()
-      // last good state - if an error occurs, fall back to this if
-      // a good state was seen.
-      @volatile var lastGood: Option[Addr] = None
+      // log last observed state of the service to help in troubleshooting
+      @volatile var lastGood: Addr = Addr.Pending
```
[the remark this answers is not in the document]
edio: Anyway, I read further and I agree that we should avoid storing state in volatile var, especially considering that it's only for logging. Will push update shortly

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -76,40 +77,19 @@ private[consul] object SvcAddr {
             )
```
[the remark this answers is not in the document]
edio: @Ashald  we currently count closes (see Closable).
Ideally, this should correspond  to number of entrances into this branch (however, there's #1820 so it's not that simple). Having counter still may be a good idea, I'm not sure, whether we need to add this counter as part of this PR.

edio:
@Ashald I addressed your comments in ammended commit.
And since we started working on code readability, I also took the liberty to change `state() = ...` to `state.update(...)` as the latter is much more readable imo. I hope that's ok.

PR #1837, opened 2018-02-28
Fix routers logging to the same access log

### LINKED ISSUE #1808: Access log of different router mixed up together
Issue Type:

- [X ] Bug report
- [ ] Feature request

**What happened**:
In our production setting, l5d has two router, one for **ingress** (to app in the same box) and one for **egress**
```
routers:
- protocol: http
  label: svc_egress
  httpAccessLog: /mnt/log/linkerd_outgoing_access.log
  servers:
  - port: 4140
    ip: 0.0.0.0
.... other lines here ...

- protocol: http
  label: svc_ingress
  httpAccessLog: /mnt/log/linkerd_incoming_access.log
  servers:
  - port: 4141
    ip: 0.0.0.0
.... other lines here ...
```

When a request comes into port 4141, I would expect access log goes into _linkerd_incoming_access.log_ but the reality is the entry shows up in _linkerd_outgoing_access.log_

Access log goes to _linkerd_outgoing_access.log_

curl localhost:4141/foo/bar 

L5d 1.3.4

**Environment**:
AWS EC2 box and L5D 1.3.4

siggy:
Multiple routers configured for access logging were incorrectly writing
to the same file.

Modify AccessLogger configuration to ensure each router has a logger
registered with a unique node name.

Fixes #1808

Signed-off-by: Andrew Seigner <siggy@buoyant.io>

### DIFF
--- a/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
+++ b/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
@@ -5,6 +5,7 @@ import com.twitter.finagle.buoyant.h2.{Request, Response}
 import com.twitter.finagle.context.RemoteInfo
 import com.twitter.logging._
 import com.twitter.util.{Time, TimeFormat}
+import io.buoyant.router.RouterLabel
 import java.net.{InetSocketAddress}
 
 case class H2AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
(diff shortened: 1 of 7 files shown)


linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
```diff
@@ -35,6 +35,11 @@ case class H2AccessLogger(log: Logger) extends SimpleFilter[Request, Response] {
 object H2AccessLogger {
 
   object param {
+    case class Label(label: String)
```
adleong: I don't think we need a new label param, we can just use the existing one (com.twitter.finagle.param.Label)

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AccessLogger.scala
```diff
@@ -57,10 +62,11 @@ object H2AccessLogger {
   }
 
   val module: Stackable[ServiceFactory[Request, Response]] =
-    new Stack.Module4[param.File, param.RollPolicy, param.Append, param.RotateCount, ServiceFactory[Request, Response]] {
+    new Stack.Module5[param.Label, param.File, param.RollPolicy, param.Append, param.RotateCount, ServiceFactory[Request, Response]] {
```
adleong: same here, can use com.twitter.finagle.param.Label instead of this one

PR #1842, opened 2018-03-01
Support TLS-encrypted communication with Consul (#1577)

### LINKED ISSUE #1577: Support TLS-encrypted communication with Consul
Consul supports TLS-encrypted communication using HTTPS (see https://www.consul.io/docs/agent/options.html#example-configuration-file-with-tls). When using the `io.l5d.consul` namer there is no way to specify what  CA certificate to use (and client certificate+key path in case of client authentication). Requests are always sent via HTTP, not matter what port is specified.

It would be great if it was possible to specify the TLS params, similarly to the way clients are configured. Example:
```
namers:
- kind: io.l5d.consul
  host: ...
  port: ...
  tls:
    commonName: linkerd.io
    trustCerts:
    - /certificates/cacert.pem
    clientAuth:
      certPath: /certificates/cert.pem
      keyPath: /certificates/key.pem
```

LukaszMarchewka:
Added TLS support for Consul namer.

Fixes #1577

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
@@ -2,7 +2,7 @@ package io.buoyant.namer.consul
 
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.twitter.finagle._
-import com.twitter.finagle.buoyant.{ClientAuth, PathMatcher, TlsClientConfig}
+import com.twitter.finagle.buoyant.TlsClientConfig
 import com.twitter.finagle.service.Retries
 import com.twitter.finagle.tracing.NullTracer
 import io.buoyant.config.types.Port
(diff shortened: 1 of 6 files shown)


wmorgan:
Thanks for the PR, @LukaszMarchewka! We will take a look.

namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
```diff
@@ -52,7 +80,8 @@ case class ConsulConfig(
   consistencyMode: Option[ConsistencyMode] = None,
   failFast: Option[Boolean] = None,
   preferServiceAddress: Option[Boolean] = None,
-  weights: Option[Seq[TagWeight]] = None
+  weights: Option[Seq[TagWeight]] = None,
+  tls: Option[TlsConsulConfig] = None
```
adleong: I think you can use io.buoyant.linkerd.TlsClientConfig here instead of defining a new TlsConsulConfig class.

LukaszMarchewka: I have thought about it. I can use io.buoyant.linkerd.TlsClientConfig if you prefer.

Ashald:
Just realized that this change affects only Consul namer and not a Dtab store. @LukaszMarchewka can you add a TLS option to a Consul Dtab Store as well please? I think the change should be quite similar.

LukaszMarchewka:
@Ashald I'm doing it now :)

LukaszMarchewka:
@Ashald I have added TLS option to a Consul Dtab Store. I have tested it with a Consul.

Ashald:
@LukaszMarchewka thanks for addressing TLS config for Dtab store!

PR #1852, opened 2018-03-04
shutdownGraceMs isn't documented or tested.

### LINKED ISSUE #1807: shutdownGraceMs isn't documented or tested.
From looking at the source code, the admin configuration supports a `shutdownGraceMs` setting that can be read from the configuration file. However, this isn't documented at https://linkerd.io/config/1.3.5/linkerd/index.html#administrative-interface. In particular, the default value isn't documented; apparently the default is 10 seconds: https://github.com/linkerd/linkerd/blob/797cf84901d5f62213c1a18a8b40860e75ecc089/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala#L22.

Neither is it tested that we actually honor this setting.

fantayeneh:
This PR,
- addresses the issue the graceful shutdown period wasn't tested.
- Refactored a little bit the Main class

TODO:
 - Document? 

Fixes #1807 

### DIFF
--- a/linkerd/main/src/main/scala/io/buoyant/linkerd/InetSocketAddresses.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/linkerd/InetSocketAddresses.scala
@@ -1,0 +1,18 @@
+package io.buoyant.linkerd
+
+import java.net.{InetAddress, InetSocketAddress, NetworkInterface}
+
+import scala.collection.JavaConverters._
+
+object InetSocketAddresses {
+
+  private def networkInterfaces: Seq[InetAddress] =
+    NetworkInterface.getNetworkInterfaces.asScala.toSeq
+      .filter(_.isUp)
+      .flatMap(_.getInetAddresses.asScala.toSeq)
+      .filterNot(_.isLoopbackAddress)
+
+  def listeningOn(port: Int): Seq[InetSocketAddress] =
+    networkInterfaces
+      .map(address => new InetSocketAddress(address, port))
+}
--- a/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
@@ -1,14 +1,13 @@
 package io.buoyant.linkerd
 
+import java.io.File
+
 import com.twitter.finagle.Path
 import com.twitter.util._
-import io.buoyant.admin.{App, Build}
+import io.buoyant.admin.{AdminConfig, App, Build}
 import io.buoyant.linkerd.admin.LinkerdAdmin
-import java.io.File
-import java.net.{InetSocketAddress, NetworkInterface}
-import scala.collection.JavaConverters._
+
 import scala.io.Source
-import sun.misc.{Signal, SignalHandler}
 
 /**
  * linkerd main execution.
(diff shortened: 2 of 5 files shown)


linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -28,28 +27,47 @@ object Main extends App {
...
-        closeOnExit(Closable.sequence(
-          Closable.all(routers: _*),
-          Closable.all(telemeters: _*),
-          Closable.all(admin: _*)
-        ))
-        Await.all(routers: _*)
-        Await.all(telemeters: _*)
-        Await.all(admin: _*)
-
+        start(loadLinker(path))(TerminationHook) match {
```
adleong: I find this refactor to be much harder to follow and understand than what was here before.  I'm guessing that you did this refactor to facilitate testing?  Unfortunately, it looks like the test doesn't really test the signal handling at all so I'm not sure it's worth refactoring the code for such a test.

Ideally we would be able to leave the structure of the code the same and test the behavior with `Signal.raise` but I'm not sure if this is possible within a unit test or how the testing framework handles it.  If such a test is not feasible, I think it's okay to resort to manual testing for features like this.

linkerd/main/src/main/scala/io/buoyant/linkerd/InetSocketAddresses.scala
```diff
@@ -0,0 +1,18 @@
+package io.buoyant.linkerd
+
+import java.net.{InetAddress, InetSocketAddress, NetworkInterface}
+
+import scala.collection.JavaConverters._
+
+object InetSocketAddresses {
```
adleong: This utility object is nice!  I might name it differently so as not to confuse it with `java.net.InetSocketAddress`.

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -119,38 +131,14 @@ object Main extends App {
...
-              addr <- addrs
-            } yield {
-              log.info("announcing %s as %s to %s", addr, name.show, announcer.scheme)
-              announcer.announce(addr, name.drop(prefix.size))
-            }
-        }
-        Closable.all(closers: _*)
-    }
-  }
+        val closers = for {
```
adleong: Nice cleanup!

linkerd/main/src/main/scala/io/buoyant/linkerd/TerminationHook.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * Trap termination signals and triggers an App.close for a graceful shutdown.
+   * Shutdown hook is not used because it has, at least, the following problems:
+   * <ul>
+   *   <li>LogManager uses a shutdown hook which makes nothing to be logged during shutdown
+   *   <li>TracerCache uses a shutdown hook to flush
+   * </ul>
+   */
+  def register(shutdown: Duration => Unit, shutdownGrace: Duration): Unit = {
```
adleong: I've got some concerns about this trait but I think it can just go away entirely if the refactor that I objected to above is reverted.

linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
fantayeneh: Yes, I tried hard to come up with something the can be testable but it felt the same to me.
It makes the code unreadable.
I am happy to close the pull request until we find a better strategy  for it.

adleong: I think that would be my preference.  This kind of thing is, unfortunately, pretty tricky to test.  I do appreciate your work on this anyway.

fantayeneh:
NP

Issue #1796, opened 2018-01-31
Feature request: Please add config validation command line function 


peterfroehlich:
It would be really helpful to have a way for linkerd to validate its configuration without spinning up the loadbalancer. 

That means: Validate presence of config files, validate config, validate presence of certs or other included files, report and exit (with fitting return code). This would be immensely useful for most automated (non-containerised) deployments. 

HAproxy, nginx, apache, etc all have a command line parameter to enable this. 

Thanks!

deebo91:
@peterfroehlich, this sounds like a good feature request! Thanks for bringing this to our attention. We will scope out and evaluate what is required.

adleong:
Better config validation is always a good idea!

Linkerd will already exit immediately if the config file is absent or fails to parse.  There are a few things I can see where we can improve:

* For client tls trustCerts we could use the `File` config type instead of string which will validate that the file exists
* Ditto for server tls `certPath`, `keyPath`, `caCertPath`

Are there other validations that you had in mind?  Are you interested in submitting a PR for this?

LukaszMarchewka:
I think an optional parameter "-validate" can solve the request. When the flag is set then Linkerd will exit immediately after loading a config with a proper status and a message.

I can implement it, if you agree.

adleong:
Fixed by https://github.com/linkerd/linkerd/pull/1854

PR #1854, opened 2018-03-06
Add a linkerd config validation by a command line (#1796)

LukaszMarchewka:
Added validation for a configuration.

Solves #1796

### DIFF
--- a/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
@@ -19,6 +19,11 @@ import sun.misc.{Signal, SignalHandler}
  */
 object Main extends App {
 
+  /**
+   * Flag to validate a configuration.
+   */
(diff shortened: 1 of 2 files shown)


LukaszMarchewka:
I have added validation only for a linkerd. If it is ok then i can implement it for a namer too.

LukaszMarchewka:
@adleong I have changed the commit according your suggestions. Could you verify it ?

LukaszMarchewka:
@adleong I have just move some part of a code from the `main` method to the new `run` method (the code is exactly the same). The `main` method is responsible for parameters "parsing" and then starting a linkerd via the `run` method. In my opinion this code is more readable than adding `if` statement into the `main` method. But if you want I can change it.

adleong:
The main method is already quite short so breaking it up hinders readability in my opinion.  And refactors that don't improve the quality of the code make navigating the git history a bit more difficult.  I think this should just be as simple as adding `if (validate()) return` to the existing main.  There are also some formatting changes in this diff that I think should be reverted.  Thanks again for working on this and for considering my feedback.

PR #1858, opened 2018-03-07
Set netty's nio maxCachedBufferSize to 256KB.

chrisgoffinet:
TL;DR: The Java NIO API caches a maximum-sized direct ByteBuffer for each thread, which looks like a native memory leak if you read or write large blocks from many threads.

Per advice from http://www.evanjones.ca/java-bytebuffer-leak.html


olix0r:
Thanks so much, @chrisgoffinet. Have these settings been tested in a prod(-ish) environment yet?

Also, in order to satisfy the [DCO](https://github.com/linkerd/linkerd/blob/master/CONTRIBUTING.md#certificate-of-origin), can you please run:
```shell
$ git commit --amend -s -a --no-edit 
```
And the force push the branch. This will add a signature to the commit message.

chrisgoffinet:
Commit updated. Yes so far HomeAway tried it out and their memory usage has been stable for 5 days. 

See the following thread:
https://discourse.linkerd.io/t/possible-memory-leak/572

PR #1862, opened 2018-03-12
Update Delegator interface to return a single value only

### LINKED ISSUE #1846: Delegator APIs should be one-shot instead of streaming
The delegator UI is powered by a streaming API but only ever uses the first value.  To improve performance, we should switch to a one-shot API instead.

From https://github.com/linkerd/linkerd/pull/1763#issuecomment-357036073

> Change the signature of the delegate method on the Delegator trait to return a Future instead of an Activity. Add a new thrift method to the namerd thrift interface that returns just the current delegation and have the thrift interpreter client use that for the delegate implementation. Add a new grpc method to the namerd mesh interface that returns just the current delegation and have the mesh interpreter client use that for the delegate implementation. I believe there is already a non-streaming version of the delegate http endpoint that the http interpreter client can use.

adleong:
The Delegator interface defines a `delegate` method that returns an Activity.  However, in all cases, only the first value of this Activity was ever used (for the delegator UI).  For complex delegations with fast changing values, a delegation lookup could be quite expensive because many values would be calculated before the Activity was canceled.

We modify the `delegate` method to return a Future instead of an Activity.  This implies changes to all of Namerd's interfaces.

* thriftNamerInterface: We change the client to do a one-shot call with an empty stamp and ignore the stamp in the response instead of long-polling.  For backwards compatibility with long-polling clients, we modify the server to keep requests that have a non-empty stamp pending until cancelled by the client.
* http streaming: No change necessary because the delegate method never took advantage of http streaming.
* mesh: We deprecate the StreamDelegateTree method and modify the implementation to only return a single value on the stream.

I have manually tested all combinations of:
* client version = {1.3.5, HEAD}
* server version = {1.3.5, HEAD}
* interface = {thriftNamerInterface, mesh, http streaming}

Fixes #1846 

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
+++ b/admin/src/main/scala/io/buoyant/admin/names/DelegateApiHandler.scala
@@ -254,7 +254,6 @@ object DelegateApiHandler {
     (dtabTry, pathTry) match {
       case (Return(d), Return(p)) =>
         delegator.delegate(d, p)
-          .toFuture
           .flatMap(JsonDelegateTree.mk).map { tree =>
             val rsp = Response()
             rsp.content = Codec.writeBuf(tree)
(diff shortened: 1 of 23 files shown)


adleong:
@edio

edio:
@adleong I can confirm, that in our environment namerd built from this branch doesn't hang during resolution of a non-existent address and there's no StackOverflowError in cases where it appeared before.

Looking forward for 1.3.7 with this fix. Thanks!

Ashald:
Since @edio verified that it works I'm more than happy with it! 🎉

PR #1863, opened 2018-03-12
Update address set to Neg when consul queries a non-existent datacenter

### LINKED ISSUE #1836: Consul errors cause entire request to time out
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
When a Consul route is found, if the datacenter does not exist, the entire request times out.

Sample dtabs:

    /consul => /#/io.l5d.consul/real-dc;
    /consul => /#/io.l5d.consul/nonexistent-dc;
    /srv => /consul;

Given the above dtabs, I'd expect that when `/#/io.l5d.consul/nonexistent-dc/[SERVICE]` is evaluated it will fail because `nonexistent-dc` doesn't exist. But it should then evaluate `/#/io.l5d.consul/real-dc/[SERVICE]` which should succeed. Instead it times out and doesn't evaluate any rules after the failing one.

Use above dtab. Replace `real-dc` with the name of your existing datacenter.

Nothing in the namerd logs except failed requests to Consul trying to retrieve the nonexistent datacenter.

**Environment**:
linkerd & namerd 1.3.5

Consul namer config

    kind: io.l5d.consul
    host: x.x.x.x
    port: 8500
Kubernetes 1.9.2
Consul 0.6.4
Google Cloud

dadjeibaah:
When the ConsulNamer tries to discover a set of addresses using a consul endpoint that contains a non existent dc, it receives a 500 error from consul. The HTTP request is then retried indefinitely even though the request will never result in a valid HTTP response. This causes bind requests to Linkerd to always timeout. The issue is easily observed when using the dtab playground in Linkerd to evaluate paths to address sets and dtabs contain paths that point to non-existent datacenters in consul.

This PR fixes the issue by checking the HTTP 500 message to evaluate if it is a `No path to datacenter` message. This indicates that there is no `dc` in consul that contains the provided `dc` value. In this case, update the service discovery endpoints to `Addr.Neg` and then retry with a backoff interval.

Change was tested locally through the admin dtab UI and now all dtab requests no longer timeout and give an error back immediately when an invalid dc is used.

fixes #1836 

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -87,9 +87,7 @@ private[consul] object SvcAddr {
                 " Last known state is %s",
               datacenter, key.name, e, currentValueToLog
             )
-            if (e.toString.contains(NonExistentDcMessage)) {
-              state.update(Addr.Neg)
-            }
+            state.update(Addr.Neg)
             val backoff #:: nextBackoffs = backoffs
             // subsequent errors are logged as DEBUG
             Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -77,7 +77,7 @@ private[consul] object SvcAddr {
             Future.Unit
 
           case Throw(e) =>
-            // do not update state, log error and continue polling with backoff
+            // update state with Addr.Neg, log error and continue polling with backoff
             stats.errors.incr()
             log.log(
               failureLogLevel,
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -22,8 +22,6 @@ private[consul] object SvcAddr {
   private[this] val ServiceRelease =
     new Exception("service observation released") with NoStackTrace
 
-  private[this] val NonExistentDcMessage = "No path to datacenter"
-
   case class Stats(stats: StatsReceiver) {
     val opens = stats.counter("opens")
     val closes = stats.counter("closes")
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -22,6 +22,8 @@ private[consul] object SvcAddr {
   private[this] val ServiceRelease =
     new Exception("service observation released") with NoStackTrace
 
+  private[this] val NonExistentDcMessage = "No path to datacenter"
+
   case class Stats(stats: StatsReceiver) {
     val opens = stats.counter("opens")
     val closes = stats.counter("closes")
@@ -85,6 +87,9 @@ private[consul] object SvcAddr {
                 " Last known state is %s",
               datacenter, key.name, e, currentValueToLog
             )
+            if (e.toString.contains(NonExistentDcMessage)) {
+              state.update(Addr.Neg)
+            }
             val backoff #:: nextBackoffs = backoffs
             // subsequent errors are logged as DEBUG
             Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
--- a/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
+++ b/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
@@ -51,7 +51,7 @@ class ConsulNamerTest extends FunSuite with Awaits {
       Future.value(LocalAgent(Config = Some(Config(Domain = Some(domain), Datacenter = datacenter))))
   }
 
-  test("Namer returns Neg while looking up datacenter for the first time") {
+  test("Namer stays pending while looking up datacenter for the first time") {
     class TestCatalogApi extends CatalogApi(null, "/v1") {
       override def serviceNodes(
         serviceName: String,
@@ -505,7 +505,74 @@ class ConsulNamerTest extends FunSuite with Awaits {
 
     assert(state == Activity.Ok(NameTree.Neg))
   }
-  test("Namer falls to Neg state to on serviceNodes failure after preceding successful serviceNode requests") {
+
+  test("Namer state returns Neg then Bound then Neg when a datacenter becomes available and then becomes unavailable") {
+    val datacenterWillBeUnavailable = new Promise[Unit]
+    val datacenterIsAvailable = new Promise[Unit]
+    val scenarioComplete = new Promise[Unit]
+    @volatile var datacenterIsUp = false
+    class TestApi extends CatalogApi(null, "/v1") {
+      override def serviceNodes(
+         serviceName: String,
+         datacenter: Option[String],
+         tag: Option[String],
+         blockingIndex: Option[String],
+         consistency: Option[ConsistencyMode],
+         retry: Boolean
+       ): Future[Indexed[Seq[ServiceNode]]] = blockingIndex match {
+        case Some("0") | None =>
+          if (datacenterIsUp)
+            datacenterIsAvailable before Future.value(
+              Indexed[Seq[ServiceNode]](Seq(testServiceNode, testServiceNode2), Some("1"))
+            )
+          else {
+            datacenterIsUp = true
+            Future.exception(new Exception("datacenter does not exist yet"))
+          }
+        case Some("1") =>
+          datacenterWillBeUnavailable before Future.exception(new Exception("something bad happened again"))
+        case _ =>
+          scenarioComplete.setDone()
+          Future.never
+      }
+    }
+
+    val stats = new InMemoryStatsReceiver
+    val namer = ConsulNamer.untagged(
+      Path.read("/test"),
+      new TestApi(),
+      new TestAgentApi("consul.acme.co"),
+      setHost = false,
+      stats = stats
+    )
+    @volatile var state: Activity.State[NameTree[Name]] = Activity.Pending
+    @volatile var nameTreeUpdates: Int = 0
+    namer.lookup(Path.read("/#/io.l5d.consul/dc1/servicename/residual")).states respond { s =>
+      nameTreeUpdates += 1
+      log.info(s"count: $nameTreeUpdates")
+      state = s
+    }
+
+
+    withClue("before datacenter is available") {
+      assert(state == Activity.Ok(NameTree.Neg))
+
+      datacenterIsAvailable.setDone()
+      eventually {
+        assertOnAddrs(state) { (addrs, _) => assert(addrs.size == 2); () }
+      }
+
+    }
+
+    withClue("during datacenter crash") {
+      datacenterWillBeUnavailable.setDone()
+      eventually {
+        assert(state == Activity.Ok(NameTree.Neg))
+      }
+    }
+  }
+
+  test("Namer falls to Neg state on serviceNodes failure after preceding successful serviceNode requests") {
     val scaleUp = new Promise[Unit]
     val doFail = new Promise[Unit]
     val scaleToEmpty = new Promise[Unit]
@@ -567,7 +634,7 @@ class ConsulNamerTest extends FunSuite with Awaits {
       assert(state == Activity.Ok(NameTree.Neg))
 
       scaleToEmpty.setDone()
-      Await.ready(scenarioComplete)
+      await(scenarioComplete)
       assert(
         state == Activity.Ok(NameTree.Neg),
         "namer did not update after falling back"
--- a/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
+++ b/namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
@@ -51,7 +51,7 @@ class ConsulNamerTest extends FunSuite with Awaits {
       Future.value(LocalAgent(Config = Some(Config(Domain = Some(domain), Datacenter = datacenter))))
   }
 
-  test("Namer stays pending while looking up datacenter for the first time") {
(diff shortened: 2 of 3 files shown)


namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -85,6 +87,9 @@ private[consul] object SvcAddr {
                 " Last known state is %s",
               datacenter, key.name, e, currentValueToLog
             )
+            if (e.toString.contains(NonExistentDcMessage)) {
```
adleong: I think we probably want to include all instances of UnexpectedResponse here.  That should cover more cases and be less brittle.

dadjeibaah: Just for my curiosity, Is there a case where we wouldn't want to update the state when a HTTP 500 error occurs? If we received a 500 other than an invalid datacenter, should we update the state? I briefly looked through the consul agent source code and one error that might return a 500 is `errRPCRateExceeded`. Which I assume is rate limiting. I would think we would want to retry and not update the state to Addr.Neg since a subsequent request would generally end up being a HTTP 200. I may be totally wrong though.

Regardless, I understand that updating the Addr state to `Neg` is not permanent and eventually it would be updated to `Ok` once we get a good value from consul.

adleong: Good question.  Unfortunately, consul doesn't give us a good machine readable way to differentiate.  Either way, you are correct that the retry loop would continue and the state would be updated.

Ashald:
I wonder if we should use https://www.consul.io/api/catalog.html#list-datacenters to collect the list of all known datacenters (maybe even a cumulative one over the whole lifetime of Namerd)  and make a decision about a retry based on if that datacenter exists.

adleong:
@Ashald that's an interesting idea.  Do you have a sense of how that would work, exactly?  My instinct is that relying on just one API call (service) is more robust than combining information from two calls (list datacenters and service) but I'm curious to hear more specifics about your idea.

namer/consul/src/test/scala/io/buoyant/namer/consul/SvcAddrTest.scala
```diff
@@ -113,4 +96,21 @@ class SvcAddrTest extends FunSuite with Matchers {
     Await.ready(retried, 1.second)
     numOfRequests.intValue() should equal(numOfAttempts)
   }
+
+  test("should be Addr.Neg unexpected error occurs") {
+    // givenn
```
adleong: typo

namer/consul/src/test/scala/io/buoyant/namer/consul/SvcAddrTest.scala
```diff
@@ -113,4 +96,21 @@ class SvcAddrTest extends FunSuite with Matchers {
...
+      invoked.setDone()
+      Future.exception(new Throwable("No path to datacenter"))
+    }
+
+    // when
+    val addr: Var[Addr] = SvcAddr(api, hangForLongBackoff, "dc1", SvcKey("svc", None), None, None, None, Map.empty, Stats(NullStatsReceiver))
+    addr.changes.respond(_ => ())
+
+    // then
+    Await.ready(invoked, 1.second)
```
adleong: I think this should be `await`

adleong: oh, I just noticed all the Await.readys in this file.  I think... they should all be `await` instead?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -22,6 +22,8 @@ private[consul] object SvcAddr {
   private[this] val ServiceRelease =
     new Exception("service observation released") with NoStackTrace
 
+  private[this] val NonExistentDcMessage = "No path to datacenter"
```
adleong: unused

namer/consul/src/test/scala/io/buoyant/namer/consul/SvcAddrTest.scala
```diff
@@ -113,4 +96,21 @@ class SvcAddrTest extends FunSuite with Matchers {
...
+    // givenn
+    val invoked = Promise[Unit]()
+    val api = apiStub { (_, _, _, _, _, _) =>
+      invoked.setDone()
+      Future.exception(new Throwable("No path to datacenter"))
+    }
+
+    // when
+    val addr: Var[Addr] = SvcAddr(api, hangForLongBackoff, "dc1", SvcKey("svc", None), None, None, None, Map.empty, Stats(NullStatsReceiver))
+    addr.changes.respond(_ => ())
```
adleong: I think we can remove this line and use `await(addr.changes.toFuture)` below

Ashald:
> that's an interesting idea. Do you have a sense of how that would work, exactly? My instinct is that relying on just one API call (service) is more robust than combining information from two calls (list datacenters and service) but I'm curious to hear more specifics about your idea.

@adleong replying here - cannot find a way to reply in-place.
From our experience, sometimes we observe network issues. And when there is a cross-data center polling, it's not atomic - we have a "polling cycle" per service name plus an overall one for the whole list of services. This in turn causes a sort of inconsistent behavior that causes a lot of confusion as some services are "on" and some are "off", some are flapping, some have latest updates and some are not.

If we were to use this endpoint as some sort of a flag to add some atomicity into the behavior it might help a bit.

adleong:
I'm not sure I follow what you mean. Are you concerned about the Addr getting set to Neg for healthy services when the service API experiences a transient 5XX and suggesting that we could ignore a "Datacenter does not exist" 5XX if we know the DC is present in the DC list response?

This seems like it introduces more opportunities for inconsistency.  I'm not sure it makes sense to trust the DC list response more than the service response if they are inconsistent.  I think my preference would be to set the Addr to Neg on a 5XX and then rely on the retry loop to correct the Addr to a real value on the next retry.

But let me know if I've misunderstood you.

dadjeibaah:
I think there is an additional tradeoff in relying on the the DC list. Right now, the consul namer spends a lot of its time polling consul endpoints. If we were to add a and additional network call to check the "DC list", it would increase the network traffic between Linkerd and the consul. This may not be an issue for a environments that can handle heavy network traffic but I think its good to keep that in mind when considering adding network calls to get info about the state of consul.

Ashald:
First of all, @adleong, my biggest concern that when there is some sort of intermittent network issues between datacenters then some of the cross-DC long-polling processes might fail while others will succeed. Imagine you are in DC `A` and polling DC `B` for 10 services. Due to network issues, polling will succeed for 7 out of 10 services while will fail for remaining 3. From a user perspective, we will observe that Linkerd can handle requests for 7/10 services and it will be confusing (the "atomicity" I was referring to).

As for the concern expressed by @dadjeibaah, I think if we not going to retry when DC is not listed then we might save a lot fo bandwidth.

dadjeibaah:
@Ashald IIUC, you are suggesting something like this:
```scala
private[consul] class DataCenterWatch{
  def apply(consulApi: v1.ConsulApi)
   (implicit timer: Timer = DefaultTimer): Activity[Seq[String]] = {
    Var.async[Activity[Seq[String]]](Activity.pending){state =>
      def loop(backoffs: Stream[Duration]):Future[Unit] = {
        consulApi.datacenters().transform{

          case Throw(e) =>
            log.error("error getting datacenter list")
            Future.sleep(backoff).before(loop(nextBackoffs))

          case Return(dcs) =>
            state.update(Activity.value(dcs))
            loop(backoffs)
        }
        val pending = loop(backoffs)
        Closable.make { _ =>
          stopped = true
          Future.Unit
        }
      }
    }
  }
}


private[this] class UntaggedNamer(lookup: LookupCache, dcWatch: DataCenterWatch, prefix: Path) extends Namer {

    def lookup(path: Path): Activity[NameTree[Name]] =
      path.take(2) match {
        case id@Path.Utf8(dc, service) =>
          val k = SvcKey(service.toLowerCase, None)
          dcWatch().flatMap { dcsList =>
            if (dcsList.contains(dc))
              lookup(dc, k, prefix ++ id, path.drop(2))
            else Activity.value(NameTree.Neg)

          }


        case _ => Activity.value(NameTree.Neg)
      }
  }
```

We add a datacenter list watch to monitor the datacenters available. When we do a lookup during name resolution, we would check to see if the datacenter exists and if it does we go ahead with our `NameTree` lookup, if it doesn't we resolve with `NameTree.Neg`.

adleong:
@Ashald the approach that @dadjeibaah outlines above basically treats the DC list API as the source of truth about DC existence instead of the response from the service API.  Is the DC list API considered more reliable?  In particular, that change would mean that failures from the DC list API could block service resolution.

Overall, my current preference is the way this PR is currently implemented because it's simpler and doesn't involve combining information from multiple sources.  But if the DC list API is considered significantly more reliable or authoritative than the service API, perhaps we can do something like what @dadjeibaah describes.

Ashald:
> IIUC, you are suggesting something like this
@dadjeibaah yeah! And then, _potentially_, later down the road we even can optimize and close watches do datacenters that are not reachable at the moment.

> Overall, my current preference is the way this PR is currently implemented because it's simpler and doesn't involve combining information from multiple sources.
Agree to that. Sorry for delaying the PR with this discussion.

> But if the DC list API is considered significantly more reliable or authoritative than the service API, perhaps we can do something like what @dadjeibaah describes.
I'll try to do a little test to see how quickly it reflects changes.

dadjeibaah:
> I'll try to do a little test to see how quickly it reflects changes.

@Ashald did you by change have the time to test out the changes mentioned above? Curious to see what you have discovered.

Ashald:
@dadjeibaah I'm super sorry but turned our I was very wrong with my assumptions about how it works. It does not update when DC goes down. It's more like "all known DC that we presumably should be able to talk to". Or, in other words, my ides is completely wrong. Sorry for confusing things up. :(

dadjeibaah:
@Ashald its no worries at all! There was totally no confusion whatsoever :) I will fix up the merge conflicts and hopefully can get this merged. Thanks for your feedback!

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -85,6 +85,7 @@ private[consul] object SvcAddr {
                 " Last known state is %s",
               datacenter, key.name, e, currentValueToLog
             )
+            state.update(Addr.Neg)
```
adleong: there is a comment above on line 80 that is no longer accurate and needs to be updated

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -51,7 +51,7 @@ class ConsulNamerTest extends FunSuite with Awaits {
       Future.value(LocalAgent(Config = Some(Config(Domain = Some(domain), Datacenter = datacenter))))
   }
 
-  test("Namer stays pending while looking up datacenter for the first time") {
+  test("Namer returns Neg while looking up datacenter for the first time") {
```
adleong: I don't think the updated test name is correct.  It looks like this tests that the state remains pending while waiting for a response from consul.

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -471,7 +471,7 @@ class ConsulNamerTest extends FunSuite with Awaits {
     ))
   }
```
adleong: I think we should add a test for the case when a DC is created and destroyed.  This would use a TestApi that returns a 500 response for the first request (representing that the DC does not exist yet), a valid 200 response for the second request (representing that the DC has been created), and then a 500 response again for the third request (representing that the DC has been deleted). 

We should check that the state is Neg after the first request, Bound after the second, and Neg again after the third.

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -503,13 +503,76 @@ class ConsulNamerTest extends FunSuite with Awaits {
...
+    @volatile var nameTreeUpdates: Int = 0
+    namer.lookup(Path.read("/#/io.l5d.consul/dc1/servicename/residual")).states respond { s =>
+      nameTreeUpdates += 1
+      log.info(s"count: $nameTreeUpdates")
+      state = s
+    }
+
+
+    withClue("before datacenter is available") {
+      assert(state == Activity.Ok(NameTree.Neg))
```
adleong: I'm not sure, but you might want to wrap this in eventually as well.  Depending on the timing I think this could potentially still be pending

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -503,13 +503,76 @@ class ConsulNamerTest extends FunSuite with Awaits {
...
+          if (datacenterIsUp)
+            datacenterIsAvailable before Future.value(
+              Indexed[Seq[ServiceNode]](Seq(testServiceNode, testServiceNode2), Some("1"))
+            )
+          else {
+            datacenterIsUp = true
+            Future.exception(new Exception("datacenter does not exist yet"))
+          }
+        case Some("1") =>
+          datacenterWillBeUnavailable before Future.exception(new Exception("something bad happened again"))
```
adleong: perhaps use "datacenter was deleted" or something as the exception message

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -503,13 +503,76 @@ class ConsulNamerTest extends FunSuite with Awaits {
...
-    assertOnAddrs(state) { (_, metadata) =>
-      assert(metadata == Addr.Metadata(Metadata.authority -> "servicename.service.dc1.consul.acme.co"))
-      ()
+    assert(state == Activity.Ok(NameTree.Neg))
+  }
+
+  test("Namer state returns Neg then Bound then Neg when a datacenter becomes available and then becomes unavailable") {
+    val datacenterWillBeUnavailable = new Promise[Unit]
+    val datacenterIsAvailable = new Promise[Unit]
+    val scenarioComplete = new Promise[Unit]
```
adleong: I don't think we ever really use this....

Issue #1224, opened 2017-04-18
investigate flakey UsageDataTelemeterEndToEndTest


siggy:
```
./sbt e2e:test integration:compile linkerd-protocol-http/integration:test
...
[info] UsageDataTelemeterEndToEndTest:
[info] - telemeter sends metrics *** FAILED ***
[info]   operation timed out after 2.seconds (UsageDataTelemeterEndToEndTest.scala:84)
```

esbie:
oh, shit. this was likely introduced as a result of #1212

adleong:
Fixed by https://github.com/linkerd/linkerd/pull/1261

PR #1872, opened 2018-03-20
Add a retry budget to flaky tests

### LINKED ISSUE #1225: investigate flakey TlsEndToEndTest
```
./sbt e2e:test integration:compile linkerd-protocol-http/integration:test
...
[info] TlsEndToEndTest:
[info] - client/server works with TLS *** FAILED ***
[info]   operation timed out after 2.seconds (TlsEndToEndTest.scala:60)
```

### LINKED ISSUE #1504: End to end tests intermittently fail due to timeouts

hawkw:
A number of Linkerd's end-to-end tests are known to sometimes fail spuriously on CI (#1224, #1225, #1504, etc). If a test known to be flaky fails, the build is typically restarted, sometimes repeatedly if the test fails twice in a row. This is problematic for the following reasons:
+  This lengthens the CI feedback loop considerably, as we need to recompile Linkerd and re-run *all* the other tests, as well as re-running the failed test.
+ Furthermore, rerunning the whole test suite also creates an opportunity for *other* flaky tests to also fail, prolonging the process.
+ Spurious failures can decrease overall confidence in the test suite.
+ If a new contributor makes a minor change in a pull request and a flaky test fails, the new contributor typically doesn't know which test failures are spurious, and may assume that their changes somehow caused an unrelated part of the codebase to break. This is an uncomfortable experience, particularly if it takes a few hours for someone "in the know" to notice and restart the build.

This PR adds a retry budget to tests known to be flaky. It builds upon ScalaTest's [`Retries`](http://doc.scalatest.org/3.0.0/index.html#org.scalatest.Retries) trait, but adds the notion of a retry *budget*, rather than retrying tests only a single time. This is probably necessary as we've observed some of the flaky tests to fail spuriously multiple times in a row.

My rationale behind this is as follows: obviously we'd all like to write tests that never fail spuriously. With that said, it's proved very difficult to identify the cause behind these flaky test failures, so it doesn't seem like this problem will be fixed easily any time soon. Alternatively, we could consider skipping the known flaky tests entirely on CI.  However, a pretty large number of our E2E tests have been observed to fail spuriously at least occasionally, and I, for one, would be uncomfortable skipping such a large portion of the test suite on CI. Finally, note that the retry budget behaviour will result in more or less the same outcome as our current manual restarting of failed flaky tests, but will shorten the feedback loop significantly, and won't result in broken builds due to flaky test failures.

Fixes #1225. Fixes #1504.

### DIFF
--- a/linkerd/core/src/e2e/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeterEndToEndTest.scala
+++ b/linkerd/core/src/e2e/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeterEndToEndTest.scala
@@ -1,5 +1,8 @@
 package io.buoyant.linkerd.telemeter
 
+import java.net.InetSocketAddress
+import java.text.SimpleDateFormat
+import java.util.Date
 import com.google.protobuf.CodedInputStream
 import com.twitter.conversions.time._
 import com.twitter.finagle.Address.Inet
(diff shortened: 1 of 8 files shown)


linkerd/core/src/e2e/scala/io/buoyant/linkerd/telemeter/UsageDataTelemeterEndToEndTest.scala
```diff
@@ -1,5 +1,8 @@
 package io.buoyant.linkerd.telemeter
 
+import java.net.InetSocketAddress
+import java.text.SimpleDateFormat
+import java.util.Date
```
adleong: un-un-sort please

hawkw:
@adleong Any test that was mentioned in one of the two currently-open flaky tests issues (#1225 and #1504) was marked as retryable, as well as `grpc.NetworkedInteropTest`, which was not mentioned in an issue but has frequently observed to be flaky.

PR #1875, opened 2018-03-23
Restore RequeueFilter in consul namer client stack

edio:
PR #1827 simplifies Consul namer code by removing
retries on Consul API level and relying only on
polling loop in SvcAddr. However before the #1827
Consul API with retry=true was handling also
low-level errors, such as TCP NACK. After #1827
such errors started leaking to high-level polling
loop in SvcAddr. This causes a lot of noise in logs.

This commit restores default finagle Http retry
filter for consul client.

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
@@ -82,9 +82,6 @@ case class ConsulConfig(
     val tlsParams = tls.map(_.params).getOrElse(Stack.Params.empty)
 
     val service = Http.client
-      // Removes the default client requeues module,
-      // (retries are handled in BaseApi.infiniteRetryFilter)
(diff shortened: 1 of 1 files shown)


edio:
This was tested in prod NCBI environment both before and after #1877.
Without #1877  it reduced noise in logs indeed. We never tested #1877  in isolation, but I'd bet it would make no big difference there. And this is my concern.

Initially, we removed retry filter to simplify consul namer. Having only one single loop and handling all errors in a single place made code easier to understand and reason about. But it made high-level consul code to deal with low-level exceptions.

And now we are introducing back another level retries. It is already better than it was in the beginning, as those level of retries handle different errors, and also Requeue filter has budget, it is not infinite retry.

So I still think we should have this fix in the code, but frankly we never tested its effect on the code after #1877 . We can only say, that #1877 and this PR work perfectly together.

edio:
@adleong :smile: I see, it is already merged. It happened, while I was composing my message.

Sorry I didn't react to your comment earlier, had a busy week. Anyway, I believe, merging it is a right decision.

Issue #1873, opened 2018-03-21
Log Spam and Namerd creating too many connections on consul (file descriptors)


cchatfield:
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:
With 1.3.6 (docker pull buoyantio/namerd) I am seeing the observation error spam in the log.

I am also seeing namerd is creating thousands of established connections over time to consul ultimately leading consul to run out of file descriptors. It seemed with 1.3.5 to create about ~200 a day. With 1.3.6 it creates about 500 a day. This ultimately brings down consul and vault rendering an environment unroutable.

**What you expected to happen**:

Limited number of consistent open connections to consul

With 1.3.6 (docker pull buoyantio/namerd) I am seeing the observation error spam in the log.

```
W 0317 03:14:07.864 UTC THREAD28 TraceId:83b2534e4e7d2f12: consul datacenter 'dc1' service 'service-payments' observation error Failure(service observation released, flags=0x02) with RemoteInfo -> Upstream Address: Not Available, Upstream id: Not Available, Downstream Address: consul.service.consul/10.10.1.4:8500, Downstream label: client, Trace Id: 27d1a3600be2b20b.27d1a3600be2b20b<:27d1a3600be2b20b. Last known state is Bound(Set(Inet(/10.12.1.5:31562,Map(endpoint_addr_weight -> 1.0))),Map(authority -> service-payments.service.dc1.consul))
W 0317 03:14:12.869 UTC THREAD27 TraceId:83b2534e4e7d2f12: consul datacenter 'dc1' service 'service-payments' observation error Failure(service observation released, flags=0x02) with RemoteInfo -> Upstream Address: Not Available, Upstream id: Not Available, Downstream Address: consul.service.consul/10.10.1.4:8500, Downstream label: client, Trace Id: b700ed57ab93668d.b700ed57ab93668d<:b700ed57ab93668d. Last known state is Bound(Set(Inet(/10.12.1.5:31562,Map(endpoint_addr_weight -> 1.0))),Map(authority -> service-payments.service.dc1.consul))
[...]

PR #1877, opened 2018-03-27
Fix consul observations leak

edio:
LookupCache uses Memoize to return the same instance of Activity
whenever lookup happens in Consul namer. Although returned activity is
backed by an async and lazy Var (backed by polling SvcAddr), it is
itself not async and lazy, as it is a join result of multiple
Acitivities which are not async and lazy. Hence every observer to a
returned activity triggers new SvcAddr creation and, as a result, new
connection to Consul even if Activity is cached.

This commit reworks Activity creation in LookupCache such that it is
indeed async and lazy. Activity remains pending unless all dependencies
to it are satisfied.

Tested in real environment, where previous version of namer was opening
up to 20K connections to consul agent, causing namerd unresponsiveness,
consul crashes. After the fix it's barely 300 connections.

consul_namer_fix

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
@@ -28,15 +28,18 @@ private[consul] class LookupCache(
 
   private[this] val localDcMoniker = ".local"
 
+
   private[this] val lookupCounter = stats.counter("lookups")
-  private[this] val serviceStats = SvcAddr.Stats(stats.scope("service"))
+  private val service: StatsReceiver = stats.scope("service")
+  private[this] val cachedCounter = service.counter("cached")
+  private[this] val serviceStats = SvcAddr.Stats(service)
 
   private[this] val cachedLookup: (String, SvcKey, Path, Path) => Activity[NameTree[Name]] =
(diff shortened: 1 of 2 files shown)


edio:
Fixes #1873

wmorgan:
@edio nice! Does it make sense to have @cchatfield validate this fix?

edio:
@wmorgan would be great. The more, the better.

Sorry, I don't have time to write tests atm. Probably could do this only next week.

cchatfield:
@wmorgan  I can definitely try it. Do you have a docker container that I can pull?

adleong:
@edio do you have a docker image for this?  if not, I can pull your branch and build one

edio:
@adleong I'm sorry, I don't have a docker image

cchatfield:
@adleong Were you able to build a docker container that I can test?

Thanks

adleong:
@cchatfield I've pushed `buoyantio/namerd:consul_observations_leak` and `buoyantio/linkerd:consul_observations_leak`

cchatfield:
Great.  I'll start it up.  Should be able to have solid results in 24 hrs.

cchatfield:
7 hours and constant at 16 connections. Seems like a solid fix.

cchatfield:
Connections are still stable with a total of 17. This fix works for me. Thanks for figuring it out.

PR #1878, opened 2018-03-28
Install git during circleci setup

adleong:
CircleCI builds were failing because sbt shells out to git but the command was not found.

Install git before invoking sbt.  Also fix some tests that were broken as a result of a bad automatic merge.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
       - run:
           name: Install sbt dependencies
           command: |
-            apt-get update && apt-get install -y git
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -66,6 +66,7 @@ jobs:
       - run:
           name: Install sbt dependencies
           command: |
+            apt-get update && apt-get install -y git
```
olix0r: tioli, but elsewhere we run `apt-get install -y --no-install-recommends` so that it does less work. not sure if that helps here or not

PR #1879, opened 2018-03-29
Fix ByteBuf reference counting

adleong:
The Netty ByteBufs used in Linkerd's H2 implementation are manually reference counted.  There are a number of situations where we fail to call `release()` the correct number of times which results in a ByteBuf leak.  

We make the following changes to prevent these leaks:

* Whenever we would discard the contains of a frame queue, we instead drain the queue and release each of those frames.
* No longer call `release()` on a Netty frame after writing it, instead relying on the H2 frame's releaser to be responsible for releasing the underlying Buf.
* No longer call `retain()` on the underlying Buf when constructing an H2 frame from a Netty frame (except in the case of array backed Bufs).
* Ensure we release frames before dropping them when errors occur.
* When decoding a gRPC message, copy the data out of the frame before releasing it.
* Remove a spurious call to `release()` when adding a trailers frame to a stream.

The end result is that there are only a few cases where `retain()` and `release()` are called:

* When passing ByteBuf to Netty to be written, we `retain()` it first so that it isn't freed when Netty writes it and calls `release()`.  This is desirable because we may be using the ByteBuf elsewhere (for example, in the case of retries) and want to be in control of when the ByteBuf is freed.
* After sending a frame to Netty to be written, we call `release()` on the H2 frame.  If this frame is being used in multiple places then it will be a RefCountedFrame and this call to `release()` will only trigger the frame's releaser when the reference count reaches 0.
* When a RefCountedFrame's reference count reaches 0 or when a non-RefCounted Frame is released, it's releaser will call `release()` on the underlying ByteBuf (and update flow control).  If all Netty writes are complete, this is expected to reduce the ByteBuf's reference count to 0 and free it.
* If a frame would be dropped instead of being sent to Netty for writing, it must be manually `released()`.

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
@@ -68,6 +68,10 @@ object Stream {
   // performed).  This would require some form of buffering, ideally
   // in the dispatcher and server stream
 
+  /**
+   * Read all frames from the queue and release them.  This can be called on a queue after
+   * calling q.fail(e, discard = false) as a safe alternative to q.fail(e, discard = true).
+   */
   def drainFrameQueue(q: AsyncQueue[Frame]): Future[Nothing] = {
     q.poll().flatMap { f =>
       f.release()
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
@@ -2,7 +2,6 @@ package com.twitter.finagle.buoyant.h2
 
 import com.twitter.concurrent.AsyncQueue
 import com.twitter.io.Buf
-import com.twitter.logging.Logger
 import com.twitter.util.{Future, Promise, Return, Throw, Try}
 
 /**
@@ -195,7 +194,6 @@ object Frame {
       private[this] val releaseP = new Promise[Unit]
       def onRelease = releaseP
       def release() = {
-        Logger().trace(s"Data frame ${hashCode()} releasing...")
         val f = release0()
         releaseP.become(f)
         f
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
@@ -69,14 +69,19 @@ object Stream {
   // in the dispatcher and server stream
 
   /**
-   * Read all frames from the queue and release them.  This can be called on a queue after
-   * calling q.fail(e, discard = false) as a safe alternative to q.fail(e, discard = true).
+   * Fail the queue and release all of its Frames.  This can be called as a safe alternative to
+   * q.fail(e, discard = true).
    */
-  def drainFrameQueue(q: AsyncQueue[Frame]): Future[Nothing] = {
-    q.poll().flatMap { f =>
-      f.release()
-      drainFrameQueue(q)
+  def failAndDrainFrameQueue(q: AsyncQueue[Frame], e: Throwable): Unit = {
+    q.fail(e, discard = false)
+    def drain(q: AsyncQueue[Frame]): Future[Nothing] = {
+      q.poll().flatMap { f =>
+        f.release()
+        drain(q)
+      }
     }
+    drain(q)
+    ()
   }
 
   /**
@@ -118,8 +123,7 @@ object Stream {
   private[this] def failOnInterrupt[T](f: Future[T], q: AsyncQueue[Frame]): Future[T] = {
     val p = new Promise[T] with Promise.InterruptHandler {
       override protected def onInterrupt(e: Throwable): Unit = {
-        q.fail(e, discard = false)
-        drainFrameQueue(q)
+        failAndDrainFrameQueue(q, e)
         f.raise(e)
       }
     }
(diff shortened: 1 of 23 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -68,6 +68,17 @@ object Stream {
   // performed).  This would require some form of buffering, ideally
   // in the dispatcher and server stream
 
+  /**
+   * Read all frames from the queue and release them.  This can be called on a queue after
+   * calling q.fail(e, discard = false) as a safe alternative to q.fail(e, discard = true).
+   */
+  def drainFrameQueue(q: AsyncQueue[Frame]): Future[Nothing] = {
+    q.poll().flatMap { f =>
```
olix0r: consider including `q.fail(e, discard = false)` in here -- since this is never called without also failing immediately prior afaict.  then can be renamed to `failFrameQueue`

adleong: good idea

adleong:
By default Netty samples 1% of buffers to check for leaks and logs if leaks are detected.  There are a number of scenarios (particularly when using gRPC) that trigger this leak detection on master that no longer trigger it after this change.

http://netty.io/wiki/reference-counted-objects.html#wiki-h3-11

router/h2/src/main/scala/io/buoyant/router/h2/ClassifierFilter.scala
```diff
@@ -135,9 +135,7 @@ class ClassifierFilter(classifier: H2Classifier) extends SimpleFilter[Request, R
                 // replace it with a new data frame without that flag.
                 // otherwise, the trailers we add to the stream will never
                 // be reached.
-                val frame2 = Frame.Data(frame.buf, eos = false)
-                // release the old data frame, since it will be replaced.
-                frame.release()
```
hawkw: As we discussed previously, releasing the frame here was almost certainly incorrect. This LGTM 👍

router/h2/src/e2e/scala/io/buoyant/router/h2/FlowControlEndToEndTest.scala
```diff
@@ -2,11 +2,9 @@ package io.buoyant.router
 package h2
 
 import com.twitter.finagle.buoyant.h2._
-import com.twitter.io.Buf
 import com.twitter.logging.Level
-import com.twitter.util.{Future, Promise}
+import com.twitter.util.Promise
 import io.buoyant.test.FunSuite
-import scala.collection.mutable.ListBuffer
```
hawkw: Were these imports unused? This looks like an unrelated change at a glance?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/Stream.scala
```diff
@@ -22,8 +22,22 @@ object Stream {
     def close(): Future[Unit]
   }
 
+  private def failAndDrainQueue[T](q: AsyncQueue[Releasable[T]], e: Throwable): Unit = {
```
hawkw: Would be nice if there was a short comment here explaining why this is different from calling `q.fail(..., discard = true)` and why it's used instead.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -68,6 +68,22 @@ object Stream {
   // performed).  This would require some form of buffering, ideally
   // in the dispatcher and server stream
 
+  /**
+   * Fail the queue and release all of its Frames.  This can be called as a safe alternative to
+   * q.fail(e, discard = true).
+   */
+  def failAndDrainFrameQueue(q: AsyncQueue[Frame], e: Throwable): Unit = {
```
hawkw: Seems like there's some duplication between this function and the `failAndDrainQueue` function in `io.buoyant.grpc.runtime.Stream`; is there any reasonable way to factor it out so we can use the same code in both places?

router/h2/src/e2e/scala/io/buoyant/router/h2/RouterEndToEndTest.scala
```diff
@@ -205,5 +205,4 @@ class RouterEndToEndTest
       await(router.close())
     }
   }
-
```
hawkw: Unrelated change?

router/h2/src/e2e/scala/io/buoyant/router/h2/FlowControlEndToEndTest.scala
adleong: yup, an unrelated "clean-as-I-go" change

router/h2/src/e2e/scala/io/buoyant/router/h2/RouterEndToEndTest.scala
adleong: yup, an unrelated "clean-as-I-go" change

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
adleong: Yeah, the functions are basically identical except that one operates on Frames and the other operates on Releasables.  We could consolidate them by having Frame and Releasable implement a common trait or by having `failAndDrainQueue` take a releaser function as an input.  But, given the small size of this method, I think I'd rather avoid adding an abstraction.

hawkw: Oh, I didn't realise the functions were operating on different types; carry on, then!

PR #1883, opened 2018-03-29
Fix race condition in AccessLoggerTest

adleong:
By default, AccessLogger uses a TimeFormat which is computed from the default TimeZone in a static val.  In AccessLoggerTest, we set the default TimeZone to GMT before executing the test.  However, if the static val TimeFormat is computed before the default TimeZone is set, the test will execute with the wrong time zone.  This leads to flaky test failures.

Explicitly pass in a TimeFormat which gets the default TimeZone during the test, guaranteeing that is has already been set.


JustinVenus:
I just this hit this bug.

PR #1884, opened 2018-03-30
Don't log errors on cancelled namerd requests

adleong:
Linkerd and Namerd logs have many spurious error logs when watches are cancelled.  These are especially frequent when using the delegator UI.

Don't log errors which are cancellations.

### DIFF
--- a/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
+++ b/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
@@ -104,6 +104,9 @@ class ThriftNamerClient(
               pending = Future.sleep(sleep).onSuccess(_ => loop(stamp0, backoffs1))
             }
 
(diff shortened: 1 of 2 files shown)


dadjeibaah:
Is there anyway to reproduce this issue? i.e. How was the change tested?

adleong:
Good question!  To test this I ran Linkerd and Namerd locally with the thrift namer interface.  Using Linkerd's delegator interface triggered this issue.  Having Linkerd resolve a request and then shutting down Linkerd also reproduced it.

namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ThriftNamerClient.scala
```diff
@@ -104,6 +104,9 @@ class ThriftNamerClient(
               pending = Future.sleep(sleep).onSuccess(_ => loop(stamp0, backoffs1))
             }
 
+          case Throw(e: Failure) if e.isFlagged(FailureFlags.Interrupted) =>
```
hawkw: Non-blocking, but it might be nice if this repetition could be reduced?

PR #1885, opened 2018-03-30
Support extracting partial segment from path segment

JustinVenus:
I use `io.l5d.serversets` for service discovery which supports named endpoints.  PathMatcher in its current form is not able to extract segments in the presence of an endpoint without also consuming the endpoint as part of the segement.  This makes it difficult to match on services that are advertised as wanting a TLS client and matching the commonName without repeating a lot of TLS client configuration.

1. I'd like to be able to extract a service name when an endpoint is present w/o including endpoint
2. I'd like to be able to use the endpoint name as a sentinel for enabling client TLS
3.  I'd like to be able to cut down on special casing TLS client configuration

Ideally I'd like to have TLS client configuration like the following.
```yaml
  client:
    kind: io.l5d.static
    configs:
    - prefix: "/#/io.l5d.serversets/sd/prod/*/{service}:https"
      tls:
        commonName: "{service}.com"
```

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
@@ -75,21 +75,28 @@ object PathMatcher {
             _extract(pathSegments.tail, exprSegments.tail, vars)
           else if (exprSegment == pathSegment)
             _extract(pathSegments.tail, exprSegments.tail, vars)
-          else if (exprSegment.startsWith("{") && exprSegment.endsWith("}"))
-            _extract(
-              pathSegments.tail,
-              exprSegments.tail,
-              vars + (exprSegment.substring(1, exprSegment.length - 1) -> pathSegment)
-            )
           else
-            (exprSegment.split(":"), pathSegment.split(":")) match {
-              case (Array(a, b), Array(y, z)) if a.startsWith("{") && a.endsWith("}") && b == z =>
-                Some(vars + (a.substring(1, a.length - 1) -> y))
+            """\{([a-zA-Z0-9.:-]+)\}""".r.findAllIn(exprSegment) match {
+              case k if !k.isEmpty && pathSegment.startsWith(exprSegment.take(k.start)) =>
+                var keys = k.toArray
+                keys.foldLeft(exprSegment.replace(".", "\\."))(_re).r.findAllIn(pathSegment) match {
+                  case v if !v.isEmpty =>
+                    _extract(
+                      pathSegments.tail,
+                      exprSegments.tail,
+                      vars ++ keys.map { key =>
+                        key.drop(1).dropRight(1) -> v.group(keys.indexOf(key) + 1)
+                      }.toMap
+                    )
+                  case _ => None
+                }
               case _ => None
             }
         case (_, None) => Some(vars)
         case (None, _) => None
       }
+
+    private[this] def _re(x: String, y: String) = x.replace(y, "(.*)")
 
     override def toString: String = expr
   }
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
@@ -36,6 +36,12 @@ trait PathMatcher {
     extract(path).map(substitutePath(_, pattern))
 }
 
+case class MatchSegment(
+  segment: String,
+  regex: Option[scala.util.matching.Regex],
+  captureKeys: Seq[String]
+)
+
 object PathMatcher {
 
   /**
@@ -59,21 +65,24 @@ object PathMatcher {
   private class Matcher(expr: String) extends PathMatcher {
 
     override def extract(path: Path): Option[Map[String, String]] =
-      _extract(path.showElems, regexSegments, Map.empty)
+      _extract(path.showElems, rawSegments, Map.empty)
 
     private[this] val segmentRegex = """\{([a-zA-Z0-9\.:-]+)\}""".r
 
-    private[this] def mkRegex(x: String, y: String) = x.replace(y, "(.*)")
-
-    private[this] val regexSegments: Seq[(Option[scala.util.matching.Regex], Seq[String])] = {
+    private[this] val rawSegments: Seq[MatchSegment] = {
       expr.split("/").dropWhile(_.isEmpty).map { exprSegment =>
         segmentRegex.findAllIn(exprSegment) match {
           case k if !k.isEmpty =>
-            var keys = k.toArray
-            Some(keys.foldLeft(exprSegment.replace(".", "\\."))(mkRegex).r) -> keys.map { key =>
-              key.drop(1).dropRight(1)
+            val keys = k.toArray
+            // '.' is the only valid Path character that also has a regex meaning.  Any '.' characters must be escaped.
+            val escaped = exprSegment.replace(".", "\\.")
+            // Replace capture keys with regex captures.
+            val withCaptures = keys.foldLeft(escaped)(_.replace(_, "(.*)")).r
+            val keyNames = keys.map { key =>
+              key.drop(1).dropRight(1) // Remove the encasing '{' and '}' characters.
             }.toSeq
-          case _ => None -> Seq(exprSegment)
+            MatchSegment(exprSegment, Some(withCaptures), keyNames)
+          case _ => MatchSegment(exprSegment, None, Seq.empty)
         }
       }
     }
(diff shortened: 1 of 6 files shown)


finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -82,7 +82,11 @@ object PathMatcher {
               vars + (exprSegment.substring(1, exprSegment.length - 1) -> pathSegment)
             )
           else
-            None
+            (exprSegment.split(":"), pathSegment.split(":")) match {
+              case (Array(a, b), Array(y, z)) if a.startsWith("{") && a.endsWith("}") && b == z =>
+                Some(vars + (a.substring(1, a.length - 1) -> y))
+              case _ => None
+            }
```
adleong: I think you need to recursively call `_extract` in order to continue processing the rest of the path

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -82,7 +82,11 @@ object PathMatcher {
               vars + (exprSegment.substring(1, exprSegment.length - 1) -> pathSegment)
             )
           else
-            None
+            (exprSegment.split(":"), pathSegment.split(":")) match {
```
adleong: It's a bit weird to allow captures of the form `{foo}:bar` but not any other form that includes `{foo}` as a substring such as `{foo}-bar` or `www.{foo}.com` for examples.  Could we maybe use a regex to make this a bit more general?

JustinVenus: Thank you for the feedback/suggestions.  I'll work on a more general solution.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -75,18 +75,29 @@ object PathMatcher {
...
-              pathSegments.tail,
-              exprSegments.tail,
-              vars + (exprSegment.substring(1, exprSegment.length - 1) -> pathSegment)
-            )
           else
-            None
+            """\{([a-zA-Z0-9\\.:-]+)\}""".r.findAllIn(exprSegment) match {
+              case k if !k.isEmpty && pathSegment.startsWith(exprSegment.take(k.start)) =>
+                var keys = k.toArray
+                keys.foldLeft(exprSegment.replace(".", "\\."))(_re).r.findAllIn(pathSegment) match {
```
adleong: Just to confirm that I understand, we only escape "." because that's the only valid Path character which also has a regex meaning?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -75,18 +75,29 @@ object PathMatcher {
...
             _extract(pathSegments.tail, exprSegments.tail, vars)
-          else if (exprSegment.startsWith("{") && exprSegment.endsWith("}"))
-            _extract(
-              pathSegments.tail,
-              exprSegments.tail,
-              vars + (exprSegment.substring(1, exprSegment.length - 1) -> pathSegment)
-            )
           else
-            None
+            """\{([a-zA-Z0-9\\.:-]+)\}""".r.findAllIn(exprSegment) match {
```
adleong: I think it would be clearer to factor this code so that it happens when the Matcher is constructed, instead of at _extract time.  I'm thinking something like this:

When the Matcher is constructed and the exprSegments are split out, each one is stored as a Regex and a list of capture names.  For example, `/{foo}-{bar}` is stored as 

```
Seq(
  ExprSegment(regex = "(.*)-(.*)".r, variables = Seq("foo", "bar"))
)
```

This lets us compile all of the regexs ahead of time when the matcher is constructed.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
JustinVenus: Correct.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
JustinVenus: I see what you are saying ... this makes a lot of sense to me.  Thanks for the feedback.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -58,33 +58,47 @@ object PathMatcher {
...
-
     override def extract(path: Path): Option[Map[String, String]] =
-      _extract(path.showElems, exprSegments, Map.empty)
+      _extract(path.showElems, regexSegments, Map.empty)
+
+    private[this] val segmentRegex = """\{([a-zA-Z0-9\.:-]+)\}""".r
+
+    private[this] def mkRegex(x: String, y: String) = x.replace(y, "(.*)")
+
+    private[this] val regexSegments: Seq[(Option[scala.util.matching.Regex], Seq[String])] = {
```
adleong: What do each part of this return type represent?  As written, I think the second part represents the list of capture keys when the segment contains captures but it represents the literal segment when there are no captures.  This reassignment of meaning is confusing.  To make this explicit, consider a case class:

```
case class MatchSegment(
  segment: String,  
  regex: Option[Regex],
  captureKeys: Seq[String]
)
```

or something like

```
sealed trait MatchSegment
case object Wildcard extends MatchSegment
case class Literal(segment: String) extends MatchSegment
case class Captures(regex: Regex, keys: Seq[String]) extends MatchSegment
```

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -58,33 +58,47 @@ object PathMatcher {
...
+      expr.split("/").dropWhile(_.isEmpty).map { exprSegment =>
+        segmentRegex.findAllIn(exprSegment) match {
+          case k if !k.isEmpty =>
+            var keys = k.toArray
```
adleong: I think this `var` should be a `val`

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -58,33 +58,47 @@ object PathMatcher {
...
+            Some(keys.foldLeft(exprSegment.replace(".", "\\."))(mkRegex).r) -> keys.map { key =>
```
adleong: there's a lot going on in this line which makes it hard to understand.  consider breaking it up:

```
// '.' is the only valid Path character that also has a regex meaning.  Any '.' characters must be escaped.
val escaped = exprSegment.replace(".", """\.""") 
// Replace capture keys with regex captures.
val withCaptures = keys.foldLeft(escaped)(_.replace(_, "(.*)")
val keyNames = keys.map { key =>
  key.drop(1).dropRight(1) // Remove the encasing '{' and '}' characters.
}.toSeq
```

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/PathMatcher.scala
```diff
@@ -58,33 +58,47 @@ object PathMatcher {
...
+      (pathSegments.headOption, regexSegments.headOption) match {
+        case (Some(path), Some((None, Seq(expr)))) if (expr == "*") || (expr == path) =>
+          _extract(pathSegments.tail, regexSegments.tail, vars)
+        case (Some(pathSegment), Some((Some(pattern), Seq(segments@_*)))) =>
+          pattern.findAllIn(pathSegment) match {
+            case v if !v.isEmpty =>
+              _extract(
+                pathSegments.tail,
+                regexSegments.tail,
+                vars ++ segments.map { s => s.toString -> v.group(segments.indexOf(s) + 1) }.toMap
```
adleong: What do you think about:

```
segments.zipWithIndex.map { case (s, i) => s -> v.group(i+1) }
```

dadjeibaah:
> We should also update the documentation in linkerd/docs/routers.md that describe match patterns.

I think we should update the routers section with this behavior it hasn't already been done. Unless, this behavior is not new and is only an enhancement.

adleong:
Good point, @dadjeibaah.  Specifically, I think this section: https://github.com/linkerd/linkerd/blob/master/linkerd/docs/routers.md#static-client-config

JustinVenus:
Apologies ... I've been swamped lately.  I'll try to update the doc soon.

PR #1886, opened 2018-03-30
Remove UnpoolHandler

adleong:
UnpoolHandler is a channel handler that copies data from direct byte buffers into heap byte buffers.  This is presumably so that users of this data don't have to worry about reference counting.  However, this is pointless for two reasons:

* We do careful reference counting of the byte buffers anyway
* The Framer, which occurs later in the pipeline, copies the data BACK into a direct byte buffer

Remove the UnpoolHandler.


hawkw:
🙌 :tada:🎊

hawkw:
This is excellent! I had thought removing `UnpoolHandler`. would require significantly more work, but it's awesome that it can just be deleted. I also thought this was likely to be the kind of change that we'd want to be extremely cautious about making, but given your description above it seems like this is fairly unlikely to break anything?

adleong:
It turns out the above failure was due to a badly behaved test.  I've now fixed the test.  @hawkw

PR #1891, opened 2018-04-11
Upgrade Linkerd finagle version to 18.2.0

### LINKED ISSUE #1839: Upgrade finagle dependency to 18.2.0
In order for this issue to be closed, we need to conduct the following tests to ensure there is no regression in Linkerd functionality.

- [x] Route HTTP/1 traffic
- [x] Route GRPC/H2 traffic
- [x] Namerd interpreters work as expected
- [x] Kubernetes namer and other k8s specific features work as expected
- [x] Consul namer functionality
- [x] SSL and TLS features

There are many more features that should be included in this list, if there is anything that has been missed feel free to add it.

dadjeibaah:
The current finagle version in linkerd is [7.1.0](https://github.com/twitter/finagle/releases/tag/finagle-7.1.0). In order to take advantage of critical bug fixes in finagle.

This upgrade impacts Linkerd's H2 implementation as a result of netty's [change](http://netty.io/news/2017/08/25/4-0-51-Final-4-1-15-Final.html) in its Http2 implementation in favor of a more feature complete library. Finagle 18.2.0 relies on `io.netty` version >= 4.1.16. The upgrade also impacts namerd's thrift interpreter caused by deprecation of a number of thrift libraries in finagle.

Linkerd's H2 features have been tested heavily in a local environment as well as the namerd and linkerd integration. Additional tests will be conducted in a k8s environment to ensure no regression has been introduced in Linkerd's k8s feature set.

fixes #1839


olix0r:
@zackangelo You may want to test this branch if you get a chance.

dadjeibaah:
I tested this branch with various test scenarios in minikube that exercised both H1 and H2 routing as well as the namerd thrift interpreter integration and everything looks good.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -87,13 +87,13 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
             case rst: Reset => rst
             case _ => Reset.Cancel
           }
-          if (!closed.get) { writer.reset(id, rst); () }
+          if (!closed.get) { writer.reset(H2FrameStream(id, Http2Stream.State.CLOSED), rst); () }
```
adleong: Are we confident that CLOSED is the right state here rather than one of the half-closed states?  Asking because I honestly have no idea.  (likewise for the other reset calls in this file)

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4ServerDispatcher.scala
```diff
@@ -123,14 +123,17 @@ class Netty4ServerDispatcher(
...
    */
-  override protected[this] def demuxNewStream(f: Http2StreamFrame): Future[Unit] = f match {
+  override protected[this] def demuxNewStream(f: Http2Frame): Future[Unit] = f match {
     case frame: Http2HeadersFrame =>
-      val st = newStreamTransport(frame.streamId)
+      val st = newStreamTransport(frame.stream.id)
       if (st.recv(frame)) serveStream(st)
       Future.Unit
 
+    case settingsFrame: Http2SettingsFrame =>
```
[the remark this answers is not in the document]
dadjeibaah: I believe this passes the settingsFrame forward to the remote so that it can deal with the remote can deal with the frame accordingly. When working through this change, the `Netty4ServerDispatcher` would receive this frame and then throw an exception because saw it as unexpected. I checked to see if the same behavior happened before the Netty upgrade and it seemed to work fine. I believe the new netty version reworked the class hierarchy so that the `settingsFrame was no longer a `Http2StreamFrame` and made it as a subclass of `Http2Frame`.

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -72,7 +60,9 @@ class H2FrameCodec(
           // active before handler added.  The stream was already made
           // active, but ctx may have been null so it wasn't
           // initialized.  https://github.com/netty/netty/issues/4942
-          connectionListener.onStreamActive(stream)
```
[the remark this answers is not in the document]
dadjeibaah: This is still an issue, I ran into this problem and when testing the H1 to H2 upgrade pipeline. The issue referenced above is still open and the only way to get around it is through the workaround used in the [here](https://github.com/linkerd/linkerd/blob/4dd774db15969086c3443a3944905949f867d4e3/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala#L36)

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -401,7 +401,7 @@ class HttpEndToEndTest
     assert(!headers.contains("dtab-local"))
   }
 
-  test("dtab-local header is ignored", Retryable) {
+  test("dtab-local header is no longer preserved in favor of l5d-tab", Retryable) {
```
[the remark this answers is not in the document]
dadjeibaah: Thanks! Will get that fixed!

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -210,78 +207,73 @@ object H2FrameCodec {
...
-      padding: Int,
-      eos: Boolean
-    ): Int = {
-      val data = new DefaultHttp2DataFrame(content.retain(), eos, padding)
-      data.streamId(streamId)
-      ctx.fireChannelRead(data)
-      0 // bytes are marked as consumed via WindowUpdateFrame writes
-    }
+      def requireStream(streamId: Int): Http2FrameStream = {
+        val stream = connection.stream(streamId).getProperty(key).asInstanceOf[Http2FrameStream] match {
```
hawkw: Is there any way to implement this without using `asInstanceOf`, or is that necessary because we're interacting with a Java API?

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -210,78 +207,73 @@ object H2FrameCodec {
...
+          case null =>
```
hawkw: TIOLI: rather than 
```scala
match somethingThatCanBeNull {
  case null => ...
  case thing => thing
}
```
you could also say
```scala
Option(somethingThatCanBeNull).getOrElse(...)
```

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameStream.scala
```diff
@@ -0,0 +1,7 @@
+package io.netty.handler.codec.http2
+
+case class H2FrameStream(streamId: Int, streamState: Http2Stream.State) extends Http2FrameStream {
```
hawkw: Do you think that it would be worthwhile to encode the states a stream can be in as separate types rather than as a field? That way, we could encode valid state transitions at the type level. Consider something like this:
```scala
sealed abstract class H2FrameStream(override val state: Http2Stream.State) extends Http2FrameStream
object H2FrameStream {
  case class Open(id: Int) extends H2FrameStream(Http2Stream.State.OPEN)
  case class Closed(id: Int) extends H2FrameStream(Http2Stream.State.CLOSED)
  case class LocalClosed(id: Int) extends H2FrameStream(Http2Stream.State.HALF_CLOSED_LOCAL)
  case class RemoteClosed(id: Int) extends H2FrameStream(Http2Stream.State.HALF_CLOSED_REMOTE)
}

```
Then you can nicely pattern match on the variants without having to call `.state`, but you still get the `state` and `id` accessors that are required to implement `Http2FrameStream`.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
hawkw: According to the state transition diagram in http://httpwg.org/specs/rfc7540.html#StreamStates, I _believe_ sending or receiving a reset should always transition the stream to CLOSED.

TIOLI: if CLOSED is the only valid transition on a reset, consider encapsulating `writer.reset` in another function that takes an `id` and always constructs a CLOSED stream to pass to `writer.reset`....

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
dadjeibaah: This is necessary because the Java API stores generic values in the `connection.stream` map we have to do the type cast to get `Http2FrameStream`.

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameStream.scala
dadjeibaah: This sounds like a good idea, I am not familiar with how this would work but I can try it out and see what that is like.

config/src/main/scala/io/buoyant/config/types/InetAddressDeserializer.scala
```diff
@@ -1,9 +1,9 @@
 package io.buoyant.config.types
 
 import com.fasterxml.jackson.core.{JsonGenerator, JsonParser}
-import com.fasterxml.jackson.databind.{SerializerProvider, DeserializationContext}
+import com.fasterxml.jackson.databind.{DeserializationContext, SerializerProvider}
 import com.google.common.net.InetAddresses
-import io.buoyant.config.{ConfigSerializer, ConfigDeserializer}
+import io.buoyant.config.{ConfigDeserializer, ConfigSerializer}
```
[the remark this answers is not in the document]
adleong: counterpoint: I actually like these "clean as you go" changes as long as they're not too distracting from the change at hand.  Especially when they're made my autoformatting tools like scalariform or an IDE.  It's a pain to re-break style errors just to get them out of the diff.

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -427,8 +427,10 @@ class HttpEndToEndTest
     req.host = "dog"
     req.headerMap.set("dtab-local", "/a=>/b")
     await(client(req))
-    assert(headers("dtab-local") == "/a=>/b")
-    assert(!headers.contains(dtabWriteHeader))
+    eventually{
+      assert(!headers.contains("dtab-local"))
```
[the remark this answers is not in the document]
dadjeibaah: I didn't quite pin down where exactly this behavior changed. But I can work on that to figure out exactly where.

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
adleong: my personal vote is to LI since this way is more explicit rather than delegating to Option's null handling.

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameStream.scala
adleong: We don't encode any logic about these states and just pass them on to Netty so I don't think there's much to gain from more types here.  I think this probably can be made more concise...

```
case class H2FrameStream(streamId: Int, streamState: Http2Stream.State) extends Http2FrameStream
```

Since the case class will autogenerate vals for streamId and streamState that should satisfy the interface.

adleong:
I've done a couple quick spot tests with client and server TLS as well as a longer running load test and everything looks great!

dadjeibaah:
Not so great news on the h2spec though, I ran that and it looks in the baseline logs we had 12 failed tests but now we have 16 :(

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
dadjeibaah: That's a great idea! I could just change `writer.reset` to accept only and id and just generate a stream object that is in the `CLOSED` state. My only concern is that the method signature in `NettyH2Writer` would look different than i.e. `writer.reset` only accepts an id while `write(...)` accepts a `H2FrameStream` and other parameters.

dadjeibaah:
h2spec was rerun on this PR based on the last two commits. Things are looking great! more eyes and more tests on this PR are always welcome!

linkerd/core/src/test/scala/io/buoyant/linkerd/ClientTest.scala
```diff
@@ -2,12 +2,13 @@ package io.buoyant.linkerd
...
 import com.twitter.finagle.Path
 import com.twitter.finagle.liveness.{FailureAccrualFactory, FailureAccrualPolicy}
-import com.twitter.finagle.loadbalancer.{DefaultBalancerFactory, LoadBalancerFactory}
+import com.twitter.finagle.loadbalancer.{DefaultBalancerFactory, FlagBalancerFactory, LoadBalancerFactory}
 import com.twitter.finagle.ssl.client.SslClientConfiguration
 import com.twitter.finagle.transport.Transport
 import com.twitter.util.Duration
 import io.buoyant.config.Parser
 import io.buoyant.test.FunSuite
+
```
adleong: rm

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -3,8 +3,10 @@ package protocol
 
 import java.io.File
 import java.net.InetSocketAddress
+
```
adleong: rm

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -13,6 +15,7 @@ import com.twitter.util._
 import io.buoyant.test.{Awaits, BudgetedRetries}
 import org.scalatest.tagobjects.Retryable
 import org.scalatest.{FunSuite, MustMatchers, OptionValues}
+
```
adleong: rm

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -398,7 +405,6 @@ class HttpEndToEndTest
       if (header == dtabWriteHeader) assert(headers(header) == "/a=>/b")
       else assert(!headers.contains(header))
     }
-    assert(!headers.contains("dtab-local"))
```
adleong: can this assertion be reverted?

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -427,8 +433,8 @@ class HttpEndToEndTest
     req.host = "dog"
     req.headerMap.set("dtab-local", "/a=>/b")
     await(client(req))
-    assert(headers("dtab-local") == "/a=>/b")
```
adleong: can these assertions be reverted?

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
dadjeibaah: sure!

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
dadjeibaah: and sure!

Issue #1865, opened 2018-03-15
 Configure log rotation policy for HTTP2 access logs


DipeshMitthalal:
# Configure log rotation policy for HTTP2 access logs #

Issue Type:

- [ ] Bug report
- [x] Feature request

Possiblity of configuring HTTP2 accesslog rotation policy via config file

**What you expected to happen**:
Expose the log rotation policy to be configurable

hawkw:
Hi @DipeshMitthalal, thanks for the feature request!

leozc:
This is a good request, for now, to work around we are using logrotated

adleong:
Fixed by #1893

PR #1893, opened 2018-04-12
Exposing rotation policy configuration for http and http2 access logs…

mmrozek:
It is my solution to problem requested here: https://github.com/linkerd/linkerd/issues/1865. In the original request, there are only mention about http2 but I think it will be nice if http has the same configuration.


dadjeibaah:
Thanks for this PR @mmrozek! we will review it as soon as we get the chance.

mmrozek:
Thanks @adleong. Sorry for late response. I tested locally "Hourly" and variations of "1.kilobyte".

PR #1901, opened 2018-04-18
Stop logging harmless reader discarded errors in k8s namer

### LINKED ISSUE #1896: Observed "Writer's reader...discarded" show up when k8s watches are cancelled by Linkerd
## Filing a Linkerd issue ##

Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
When running linkerd in k8s we see lots of innocuous messages that indicate that linkerd cancels k8s watches.

Linkerd should not display these errors as they may be misleading to users

Set up Linkerd in k8s with a config that uses k8s to get endpoints and watch the linkerd logs for the message

adleong:
When Linkerd cancels watches to the k8s API, ReaderDiscarded errors are logged to the console.  These errors are harmless but clutter up the log and make it look as though something is wrong.

Stop logging these errors.

Tested by starting Linkerd configured with a k8s namer.  Send some requests to k8s services though the Linkerd, then shut down Linkerd with SIGINT.  Notice that no ReaderDiscarded errors are logged as Linkerd cancels its API watches and shuts down.

Fixes #1896

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
@@ -2,6 +2,8 @@ package io.buoyant.k8s
 
 import com.twitter.finagle.tracing.NullTracer
 import com.twitter.finagle.{Http, Stack}
+import com.twitter.io.Reader.ReaderDiscarded
+import com.twitter.util.Monitor
 
 /**
  * Meant to be implemented by configuration classes that need to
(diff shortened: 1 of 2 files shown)


k8s/src/main/scala/io/buoyant/k8s/ClientConfig.scala
```diff
@@ -13,6 +15,10 @@ trait ClientConfig {
   protected val DefaultNamespace = "default"
   protected val DefaultPort = 8001
 
+  protected val ReaderDiscardedMonitor = Monitor.mk {
+    case _: ReaderDiscarded => true
```
olix0r: we should probably do the same thing in consul, if we don't already?

Issue #1477, opened 2017-07-07
Disable autoloading of the finagle zipkin tracer in namerd


klingerf:
First reported by @moderation, @rclayton-the-terrible. Relates to #865.

We should update namerd so that it doesn't automatically load the zipkin scribe tracer. If namerd is started without a `telemetry` block, then we should not be attempting to trace any requests. But instead, in the namerd logs at startup, I see:

```
I 0707 19:59:50.286 UTC THREAD97: Tracer: com.twitter.finagle.zipkin.thrift.ScribeZipkinTracer
```

And then after sending enough requests to namerd, I start to see:

```
I 0707 20:02:06.374 UTC THREAD111 TraceId:77689a2c404454a5: FailureAccrualFactory marking connection to "zipkin-tracer" as dead. Remote Address: Inet(localhost/127.0.0.1:1463,Map())
```

The default finagle zipkin tracer is attempting to trace, even though it was not configured via the namerd config file. We should disable it by default.

sfroment:
Even with a telemetry block 
```
telemetry:
    - kind: io.l5d.prometheus
```  
it's trying to connect to zipkin in debug log level I got this error 
```D 0912 11:09:40.812 UTC THREAD68: ChannelStatsHandler caught an exception
java.net.ConnectException: Connection refused: localhost/127.0.0.1:1463
	at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:717)
	at org.jboss.netty.channel.socket.nio.NioClientBoss.connect(NioClientBoss.java:152)
	at org.jboss.netty.channel.socket.nio.NioClientBoss.processSelectedKeys(NioClientBoss.java:105)
	at org.jboss.netty.channel.socket.nio.NioClientBoss.process(NioClientBoss.java:79)
	at org.jboss.netty.channel.socket.nio.AbstractNioSelector.run(AbstractNioSelector.java:337)
	at org.jboss.netty.channel.socket.nio.NioClientBoss.run(NioClientBoss.java:42)
	at org.jboss.netty.util.ThreadRenamingRunnable.run(ThreadRenamingRunnable.java:108)
	at org.jboss.netty.util.internal.DeadLockProofWorker$1.run(DeadLockProofWorker.java:42)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[...]

PR #1902, opened 2018-04-18
 Disable autoloading of the default tracer (#1477)

mmrozek:
I disabled default tracer - switched to NullTracer in all interfaces.

### DIFF
--- a/linkerd/docs/protocol-h2.md
+++ b/linkerd/docs/protocol-h2.md
@@ -8,6 +8,9 @@
 routers:
 - protocol: h2
   h2AccessLog: access.log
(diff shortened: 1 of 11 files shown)


adleong:
Looks like it was an issue with the CI system itself.  I've restarted the CI build.  Sorry about that!

adleong:
This looks like a good fix for Namerd, though I think the issue still affects Linkerd.

adleong:
Ah, nevermind, I see that the original issue refers to Namerd.  This looks great to me!

dadjeibaah:
Adding the related issue.

fixes #1477

Issue #1890, opened 2018-04-10
`idleTimeMs` and `idleTtlSecs` settings don't seem to have any effect on established H2 connections


nikolay-pshenichny:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
Linkerd is not expiring/terminating idle client HTTP2 connections after the timeouts configured in `idleTimeMs` and/or `idleTtlSecs` configuration options.

**What you expected to happen**:
The established Http2 connections from Linkerd to the server are expected to be closed if they were idling for more then the time indicated in the configuration.

**How to reproduce it (as minimally and precisely as possible)**:
1. Set the idleTtlSecs to 60 seconds
2. Make a single request from a grpc client to a grpc server through Linkerd
3. Observe the number of connections in Linkerd for more than 60 seconds. The number will stay == 1
4. Check the `netstat`'s output to see that there's and ESTABLISHED connection from Linkerd to the server.

Docker-compose and configuration files that can help to reproduce the issue - https://github.com/nikolay-pshenichny/linkerd-idleTtlSecs_and_idleTimeMs

**Anything else we need to know?**:

New context (2018-04-10 2:50pm): Linkerd will always keep 1 client open, even if it has been idle for longer than the ttl. The issue is requesting that that behavior be changed or made configurable.
cc: @adleong 

**Environment**:
- linkerd/namerd version, config files:  linkerd v 1.3.7; sample config files can be found using the link above. 
- Platform, version, and config files (Kubernetes, DC/OS, etc): Seeing this in Vagrant; Docker; Kubernetes;
- Cloud provider or hardware configuration:

nikolay-pshenichny:
@adleong , thx for the update!
I haven't done much testing around Http1 and won't be able to confirm that part.
But with Http2 - yes, we are seeing a single connection being present after a single request went through... That connection is never closed.


>My recommendation for a fix would be to change the singleton connection pool that is used by HTTP/2 to close the singleton connection after it has been idle for idleTimeMs. What do you think?

From your description, it sounded like Http1 pool is doing a similar job - ie. the pool will only close the excess connections after the timeout and will hold all other connections. So the Singleton Pool appears to be working as designed. And since it is defined in Finagle-Core, I am not sure I completely understand the idea of fixing it.

From the design perspective of Linkerd, should the connection pools be responsible for tracking and closing idle connections, or is it ok for the connections to expire on their own?

[...]

PR #1903, opened 2018-04-18
Introduce ClientSession configuration section that provides ways to control idleTime and lifeTime of client connections.

nikolay-pshenichny:
This PR introduces a new client configuration section that allows to control `idleTime` and `lifeTime` of the client connections.

Fixes https://github.com/linkerd/linkerd/issues/1890

### DIFF
--- a/linkerd/docs/routers.md
+++ b/linkerd/docs/routers.md
@@ -253,8 +253,12 @@ failFast | `false` | If `true`, connection failures are punished more aggressive
 requeueBudget | see [retry budget](#retry-budget-parameters) | A [requeue budget](#retry-budget-parameters) for connection-level retries.
 failureAccrual | 5 consecutive failures | a [failure accrual policy](#failure-accrual) for all clients created by this router.
 requestAttemptTimeoutMs | no timeout | The timeout, in milliseconds, for each attempt (original or retry) of the request made by this client.
+clientSession | An empty object | see [clientSession](#client-session)
 
 #### Host Connection Pool
+
+This section defines the behavior of [watermark connection pools](https://twitter.github.io/finagle/docs/com/twitter/finagle/client/DefaultPool) on which most of the protocols are relying.
+Note that Http2 protocol uses [SingletonPool](https://github.com/twitter/finagle/blob/master/finagle-core/src/main/scala/com/twitter/finagle/pool/SingletonPool.scala) that maintains a single connection per endpoint and will not be affected by the settings in this section. 
 
 ```yaml
 client:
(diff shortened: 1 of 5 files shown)


nikolay-pshenichny:
@adleong , done.

linkerd/docs/routers.md
```diff
@@ -269,5 +273,14 @@ Key | Default Value | Description
 --- | ------------- | -----------
 minSize | `0` | The minimum number of connections to maintain to each host.
 maxSize | Int.MaxValue | The maximum number of connections to maintain to each host.
-idleTimeMs | forever | The amount of idle time for which a connection is cached in milliseconds.
+idleTimeMs | forever | The amount of idle time for which a connection is cached in milliseconds. Only applied to connections that number greater than minSize, but fewer than maxSize.
```
dadjeibaah: I am trying to understand this comment. How do `HostConnectionPool`'s idleTimeMs differ from `ClientSession`'s idleTimeMs? Do they effectively do the same thing? If they do, then we may no longer need to have this setting in `HostConnectionPool`.

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -693,4 +697,152 @@ class HttpEndToEndTest
...
+
+    } finally {
+      await(client.close())
+      await(fox.server.close())
+      await(server.close())
+      await(router.close())
+    }
+
+
+  }
```
dadjeibaah: I may be wrong, but maybe we could add a test that exercises both `hostConnectionPool` and `clientSession` configs. I am still unclear of how the two different kinds of configurations would work together if that ever is the case.

linkerd/docs/routers.md
adleong: They are similar but slightly different.  This connection timeout is only applied to connections between the minSize and maxSize whereas the clientSessions timeouts apply to all connections.  The connectionPool timeout can be thought of as the amount of time "extra" connections are kept around.

adleong: Although, in the case where minSize=0 (which is the default) then these two timeouts do exactly the same thing.

dadjeibaah: Ahh I see, thanks for the clarification!

adleong:
@nikolay-pshenichny Looks like there's a merge conflict with the latest master.  Would you mind resolving that and then we should be good to merge.

nikolay-pshenichny:
@adleong , I will look into that a little later today.

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
nikolay-pshenichny: @dadjeibaah , I can look into adding more tests for the scenarios when both configurations are used. But I will be able to do that only starting Tuesday (24th) next week.

adleong:
:star: Thank you so much for tracking this down and fixing it, @nikolay-pshenichny!

Issue #1847, opened 2018-03-02
Chunked encoding is being sent to HTTP 1.0 clients


hynek:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

I’ve been explained, that a proper Proxy should always speak HTTP 1.1 up and down, no matter what the client speaks.

However: I’ve run into the case that when a client, that I have no control over (in my case ddclient) requests HTTP 1.0, the request is transformed into an HTTP 1.1 request and the upstream server takes advantage of having HTTP 1.1 and serves a chunked response that the client can’t handle.

**What you expected to happen**:

Either

- don't modify the request (like e.g. HAProxy which I made to talk upstream directly for now)
- de-chunk responses if you know you’re speaking to an HTTP 1.0 client

Ideally both. :)

**How to reproduce it (as minimally and precisely as possible)**:

I’m not sure how to give you an upstream server that reacts to the HTTP version. But in practice it looks like this:

```
GET /nic/update?hostname=XXX&myip=1.2.2.2 HTTP/1.0
Host: XXX
Connection: close
Authorization: Basic XXX

HTTP/1.0 200 OK
Server: gunicorn/19.7.1
Date: Fri, 02 Mar 2018 18:30:05 GMT
Connection: close
Content-Type: text/plain; charset=UTF-8
X-Request-ID: 0399f0a381cd4b7ca8e63fce6d4ff070
X-Served-By: dev-dev-dev

good 1.2.2.2
```

vs

```
GET /nic/update?hostname=XXX&myip=1.2.2.2 HTTP/1.1
Host: XXX
Connection: close
Authorization: Basic XXX

HTTP/1.1 200 OK
Server: gunicorn/19.7.1
Date: Fri, 02 Mar 2018 18:30:34 GMT
Connection: close
Transfer-Encoding: chunked
Content-Type: text/plain; charset=UTF-8
X-Request-ID: d4da77308a6c45a88a3b461b945e9a0d
X-Served-By: dev-dev-dev

C
good 1.2.2.2
0
```

ddclient chokes on the `C` and `0`.

wmorgan:
Thanks for capturing this, @hynek!

adleong:
Fixed by https://github.com/linkerd/linkerd/pull/1904

PR #1904, opened 2018-04-20
De-chunking response if client reguested HTTP 1.0 (#1847)

mmrozek:
Problem
Proxy speaks HTTP 1.1 (in both directions) and it was possible to send a chunked response that the HTTP 1.0 client can't handle.

Solution
Dechunk response and send to the client in HTTP 1.0 compatible form.

Validation
Tests were done on a locally running Linkerd instance + unit tests.

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/ViaHeaderAppenderFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/ViaHeaderAppenderFilter.scala
@@ -7,7 +7,6 @@ import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack}
 
 /**
  * Appends the [Via] (https://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-14#section-9.9) header to the request and response.
- * After the header is appended, the message is upgraded to HTTP/1.1.
  */
 object ViaHeaderAppenderFilter {
 
@@ -39,8 +38,10 @@ object ViaHeaderAppenderFilter {
 
     def apply(req: Request, svc: Service[Request, Response]) = {
       ViaLinkerd.appendViaHeader(req)
-      req.version = Http11
-      if (req.host.isEmpty) {
+
+      // Forwards HTTP/1.0 requests using HTTP/1.0 to minimize the chance that the server will send a response that
+      // uses HTTP/1.1 features, in particular Transfer-Encoding: chunked, that the HTTP/1.0 application can't handle
+      if (req.version == Http11 && req.host.isEmpty) {
         // https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23
         // If the requested URI does not include an Internet host name for the service being
         // requested, then the Host header field MUST be given with an empty value.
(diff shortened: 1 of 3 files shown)


briansmith:
> Dechunk response and send to the client in HTTP 1.0 compatible form.

The problem with dechunking is that it requires buffering the response entire body to calculate the Content-Length. That means either implementing a maximum response body size cutoff that is small or having linkerd potentially consume huge amounts of memory buffering response bodies.

An alternate solution is to avoid upgrading HTTP/1.0 requests; that is, if the original request was HTTP/1.0 then send the request upstream using HTTP/1.0 too. Then the upstream won't chunk the response. This has downsides too, since the benefits of upgrading to HTTP/1.1 wouldn't apply. However, IMO it is still better than having a response body size cut-off or unbounded buffering.

(We should get some agreement on what's best to do between Conduit and Linkerd. The above comments summarize the discussion around the issue when it was brought up for Conduit.)

mmrozek:
@briansmith thank you for a comment. I agree with you that it is probably not the best solution. On the other hand, I am not sure how to interpret [rfc7230](https://tools.ietf.org/html/rfc7230#section-2.6) 

> Intermediaries that process HTTP messages (i.e., all intermediaries
   other than those acting as tunnels) MUST send their own HTTP-version
   in forwarded messages.

briansmith:
RE: rfc6202, it's definitely OK as far as the HTTP spec is concerned to do the de-chunking by buffering. The question is really just the quality-of-implementation issue regarding memory usage.

mmrozek:
Ok @briansmith, I understand your point of view. Do you think that I should change my solution to avoid upgrading HTTP 1.0 requests? How did you solve this issue in Conduit?

dadjeibaah:
@mmrozek are you able to run the e2e tests locally to see if you get the same issue that the CI server is failing on? You can run the specific test by running `sbt 'router-http/e2e:test'` on your machine.

mmrozek:
You're right. I didn't check that - I started only "router-http/test". Now the broken test is corrected.

briansmith:
> @briansmith, I update my PR based on your advice. Please check if it is an acceptable solution.

IMO, this is the best idea for solving this kind of problem. The only question is whether @olix0r or @seanmonstar or @adleong object to the idea. In particular, would we do the same thing in Conduit?

As far as the implementation is concerned, @adleong and @olix0r can review this much better than me.

router/http/src/main/scala/io/buoyant/router/http/ViaHeaderAppenderFilter.scala
```diff
@@ -39,8 +38,7 @@ object ViaHeaderAppenderFilter {
 
     def apply(req: Request, svc: Service[Request, Response]) = {
       ViaLinkerd.appendViaHeader(req)
-      req.version = Http11
-      if (req.host.isEmpty) {
+      if (req.version == Http11 && req.host.isEmpty) {
```
briansmith: Let's add a comment here explaining that we forward HTTP/1.0 requests using HTTP/1.0 to minimize the chance that the server will send a response that uses HTTP/1.1 features, in particular `Transfer-Encoding: chunked`, that the HTTP/1.0 application can't handle.

adleong:
Having Linkerd always upgrade messages to HTTP/1.1 was a deliberate decision.  @olix0r probably has the most context on that decision and if it still makes sense.

mmrozek:
@adleong, could you please give the final decision? I could revert my changes to the previous version but I would like to have clear info how it should work.

mmrozek:
@adleong, I updated PR.

PR #1908, opened 2018-04-23
Fix "ClassCastException" in H2FrameCodec onStreamClosed

### LINKED ISSUE #1905: Handle uncaught exception when streams are closed in H2FrameCodec
When running Linkerd with an H2 client and H2 router, linkerd throws an uncaught exception whenever a stream is closed.

```bash
ERROR 0422 20:57:00.713 PDT finagle/netty4-3: Caught Throwable from listener onStreamClosed.
java.lang.ClassCastException: io.netty.handler.codec.http2.Http2FrameCodec$DefaultHttp2FrameStream cannot be cast to scala.runtime.Nothing$
```

The particular error above is a possible bug caused by the interaction between Netty's Java API and scala. 

The proposed solution is to change:
https://github.com/linkerd/linkerd/blob/9542e3e72a159721306cda232ebffe49fccbc262/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala#L39-L43

In particular this line
```scala
    val stream = stream0.getProperty(streamKey) 
```
to 

```scala
   val stream = stream0.getProperty(streamKey).asInstanceOf[Http2FrameStream]
```

dadjeibaah:
This is a follow-up/bug-fix PR for the finagle upgrade. When a stream is closed between Linkerd and an Http/2 server, we see "ClassCastException" in Linkerd. This PR fixes that issue but properly casting a `Http2Stream` to an `Http2FrameStream` onStreamClosed

Tests were conducted with current master and this branch. The exception no longer shows up with this fix.

fixes #1905

### DIFF
--- a/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
+++ b/finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
@@ -37,7 +37,7 @@ class H2FrameCodec(
     }
 
     override def onStreamClosed(stream0: Http2Stream): Unit = {
-      val stream = stream0.getProperty(streamKey)
+      val stream = stream0.getProperty(streamKey).asInstanceOf[Http2FrameStream]
       if (stream != null) {
         channelCtx.fireUserEventTriggered(Http2FrameStreamEvent.stateChanged(stream).asInstanceOf[AnyRef]); ()
       }


adleong:
@dadjeibaah can you give a bit more detail about what tests were done?

dadjeibaah:
Sure, the bug can be reproduced locally. I set up an H2 server with Linkerd pointing to it. The H2 server responds with a stream of integers up to 11 and closes the connection. After the server responds with a `GOAWAY` frame, Linkerd throws a `ClassCastException` in the logs with the `log.level` set to `INFO`. After the code change, I run the same test with the H2 server and no longer see any errors. To be sure, I ran the test with Linked set a `DEBUG` log level and no errors show up.

PR #1909, opened 2018-04-23
Upgrade to finagle 18.4.0

adleong:
Note that this includes a workaround to https://github.com/twitter/finagle/issues/688 for constructing the assembly jar.  We should remove that workaround once https://github.com/twitter/finagle/issues/688 is fixed.

### DIFF
--- a/linkerd/core/src/test/scala/io/buoyant/linkerd/LoadBalancerTest.scala
+++ b/linkerd/core/src/test/scala/io/buoyant/linkerd/LoadBalancerTest.scala
@@ -1,6 +1,6 @@
 package io.buoyant.linkerd
 
-import com.twitter.finagle.loadbalancer.{DefaultBalancerFactory, FlagBalancerFactory, LoadBalancerFactory}
+import com.twitter.finagle.loadbalancer.{FlagBalancerFactory, LoadBalancerFactory}
 import com.twitter.finagle.Path
 import io.buoyant.router.StackRouter.Client.PerClientParams
 import org.scalatest.FunSuite
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
@@ -28,7 +28,6 @@ private[consul] class LookupCache(
 
   private[this] val localDcMoniker = ".local"
 
-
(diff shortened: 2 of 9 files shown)


linkerd/core/src/test/scala/io/buoyant/linkerd/LoadBalancerTest.scala
```diff
@@ -1,6 +1,6 @@
 package io.buoyant.linkerd
 
-import com.twitter.finagle.loadbalancer.{DefaultBalancerFactory, FlagBalancerFactory, LoadBalancerFactory}
+import com.twitter.finagle.loadbalancer.{FlagBalancerFactory, LoadBalancerFactory}
```
olix0r: is `FlagBalancerFactory` actually used? does it just need to be in scope or something?

project/Base.scala
```diff
@@ -133,6 +133,26 @@ class Base extends Build {
        |""".stripMargin.split("\n").toSeq
 
   val nodeModulesRE = ".*/node_modules/.*".r
+
+  // This is a workaround for https://github.com/twitter/finagle/issues/688 and should be removed
+  // once that issue is fixed.
```
olix0r: would you mind briefly describing the problem/workaround.

```
/**
 * finagle-thrift-18.04 incorrectly exports symbols from libthrift-0.10.0.
 * In order to build assembly packages, we chose an arbitrary copy of these files
 * on the assumption that they are identical.
 *
 * This can be safely removed once twitter/finagle#688 is resolved.
 */
```

... or something

linkerd/core/src/test/scala/io/buoyant/linkerd/LoadBalancerTest.scala
adleong: it's used on line 27 and line 48

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -28,7 +28,6 @@ private[consul] class LookupCache(
 
   private[this] val localDcMoniker = ".local"
 
-
   private[this] val lookupCounter = stats.counter("lookups")
```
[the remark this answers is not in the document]
adleong: Yeah

PR #1918, opened 2018-04-25
Expose k8s namer internal state

### LINKED ISSUE #1894: Admin endpoint that exposes internal k8s namer state
Linkerd namers should expose an admin endpoint which provides data about the internal state of that namer including:

adleong:
It can be difficult to know which watches the Kubernetes namer has cached, which of them are active, and if they are receiving updates from the Kubernetes API.  This has made issues with the Kubernetes namer and/or the Kubernetes API extremely difficult to diagnose and debug.

We instrument the Activities in the Kubernetes namer to keep track of metadata about themselves.  We also instrument calls to the Kubernetes API.  All of this metadata is exposed though an admin endpoint.

screen shot 2018-04-24 at 3 28 49 pm

screen shot 2018-04-24 at 3 29 31 pm

screen shot 2018-04-25 at 11 49 30 am

Fixes #1894


k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -121,9 +125,12 @@ abstract class EndpointsNamer(
   labelName: Option[String] = None,
   backoff: Stream[Duration] = EndpointsNamer.DefaultBackoff
 )(implicit timer: Timer = DefaultTimer)
-  extends Namer {
+  extends Namer with Admin.WithHandlers {
   import EndpointsNamer._
 
+  // Metadata about the state of portMap watches.
+  private[this] val portMapInstrumentation = mutable.Map[(String, String, Option[String]), (InstrumentedActivity[NumberedPortMap], WatchState[v1.Service, v1.ServiceWatch])]()
```
dadjeibaah: This value looks like it contains a lot of information that isn't easily recognizable at first glance. Do you think it would be helpful for this to be a class of some sort?

namer/core/src/main/scala/io/buoyant/namer/InstrumentedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+  lastStoppedAt: Option[String],
+  lastUpdatedAt: Option[String],
+  value: T
+)
+
+/**
+ * InstrumentedVar is a Var.async that also records metadata about itself such as whether it is
+ * running and the value of the most recent update.
+ */
+class InstrumentedVar[T](init: T, update: Updatable[T] => Closable) {
```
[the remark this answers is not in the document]
adleong: I don't think util has anything like this.  We could consider contributing it back, but I'm not sure if they'd be interested.

namer/core/src/main/scala/io/buoyant/namer/InstrumentedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+case class VarState[T](
+  running: Boolean,
+  lastStartedAt: Option[String],
+  lastStoppedAt: Option[String],
+  lastUpdatedAt: Option[String],
+  value: T
+)
+
+/**
+ * InstrumentedVar is a Var.async that also records metadata about itself such as whether it is
```
olix0r: this type doesn't actually implement Var... perhaps say it _holds` a Var.async?

namer/core/src/main/scala/io/buoyant/namer/InstrumentedVar.scala
```diff
@@ -0,0 +1,116 @@
...
+  @volatile private[this] var _value: T = init
+  @volatile private[this] var _lastUpdatedAt: Option[Time] = None
+
+  val underlying: Var[T] = Var.async(init)(instrumentedUpdate)
+
+  def running: Boolean = _running
+  def lastStartedAt: Option[Time] = _lastStartedAt
+  def lastStoppedAt: Option[Time] = _lastStoppedAt
+  def lastUpdatedAt: Option[Time] = _lastUpdatedAt
+  def value: T = _value
```
olix0r: Do you think it's worth writing unit tests for InstrumenedVar?

k8s/src/main/scala/io/buoyant/k8s/EndpointsNamer.scala
```diff
@@ -391,4 +435,49 @@ object EndpointsNamer {
...
+            "watch" -> watchState
+          )
+      }
+      val endpointsState = endpoints.map {
+        case (ServiceKey(ns, svc, label), InstrumentedEndpoints(act, watchState)) =>
+          val labelStr = label match {
+            case Some(s) => s":$s"
+            case None => ""
+          }
+          s"$ns/$svc$labelStr" -> Map(
```
olix0r: tioli braces might make this a little clearer i.e. `${ns}/${svc}${labelStr}`

PR #1919, opened 2018-04-25
Add a 'Forwarded' header to the initial header frame of an H2 request

### LINKED ISSUE #1892: h2 router forwarded-for header
Issue Type:

- [ ] Bug report
- [x] Feature request

**What happened**:
Whilst using linkerd as a gRPC ingress controller, I was expecting to be able to determine the client's IP address, much in the same way it's done with HTTP1.1: https://linkerd.io/config/1.3.7/linkerd/index.html#adding-the-forwarded-header

The IP address was nowhere to be found :(

Create a h2 router suitable for routing gRPC calls and try view the headers - l5d headers are present, but no client information.

**Environment**:
linkerd 1.3.7
Kubernetes 1.9.4
Google Cloud but the same thing happens locally.

dadjeibaah:
Linkerd allows for a `Forwarded` header to be configured on each request that passes through Linkerd for Http1 requests. The same behavior is not available when sending Http/2 traffic from a client to a destination service.

This PR allows a router to be configured to add a forwarded header for Http/2 requests similar to what Linkerd does for Http/1.

Tests were carried out locally. I set up an Http/2 server that logs all requests headers from an incoming request. With the new change, this is the result of the test.

```bash
:method: GET
:path: /
:scheme: http
:authority: localhost:4142
accept: */*
accept-encoding: gzip, deflate
user-agent: nghttp2/1.30.0
forwarded: by="127.0.0.1:4142";for="127.0.0.1:58747";authority=localhost:4142;proto=http <-- new header field
l5d-dst-service: /svc/localhost:4142
via: h2 linkerd
l5d-dst-client: /$/inet/127.1/8888
l5d-ctx-trace: r/MQS83qX4UA9S5rBTBvlAD1LmsFMG+UAAAAAAAAAAA=
l5d-reqid: 00f52e6b05306f94
````

fixes #1892

### CODE REVIEW

**Comment by yoitsro** [comment]
Hey @dadjeibaah. Thanks for this. Out of curiosity, have you managed to test this against a remote IP too? Also, does this work with IPv6 and what would the header look like for IPv6 addresses?

**Comment by dadjeibaah** [comment]
@yoitsro good question. I haven't tested this with remote IPs. One way I could test that is to deploy linkerd in a k8s environment with the forwarded header configuration set. Does that suffice as an appropriate test for remote IPs? I would be happy to get any suggestions on what you think would be best. As for your question about this working with IPv6 addresses, the answer is yes. It should work exactly like the Http/1 feature which does display IPv6 addresses in the `Forwarded` header field. I think I can set up a test for that as well.

File: linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AddForwardedHeaderConfig.scala [hunk 1]
```diff
@@ -0,0 +1,68 @@
+package io.buoyant.linkerd.protocol.h2
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.{ServiceFactory, Stack, Stackable}
+import com.twitter.finagle.buoyant.h2.{Request, Response}
+import io.buoyant.router.http.{AddForwardedHeader, LabelerConfig}
+
+case class H2AddForwardedHeaderConfig(
```
adleong: I think everything in the AddForwardedHeaderConfig and AddForwardedHeaderConfig.module is request/response type agnostic.  This means that we should be able to just have one of these (not separate ones for H1/H2) and it can live in http-base.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/H2AddForwardedHeaderConfig.scala
```diff
@@ -0,0 +1,68 @@
...
+    val default = Param(None)
+  }
+
+  /**
+   * Configures parameters for `AddForwardedHeader`.
+   *
+   * Because `AddForwardedHeaderConfig` types may depend on stack
+   * parameters (for instance, `Server.RouterLabel`)
+   */
+  val module: Stackable[ServiceFactory[Request, Response]] =
```
adleong: can parameterize this on the request/response types:

```
def module[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] =
```

project/LinkerdBuild.scala
```diff
@@ -65,7 +65,8 @@ object LinkerdBuild extends Base {
       .settings(coverageExcludedPackages := ".*XXX_.*")
 
     val baseHttp = projectDir("router/base-http")
-      .dependsOn(core)
+      .dependsOn(core, Finagle.h2)
```
adleong: I think it would be nice to avoid depending on http and H2 in http-base if possible

router/base-http/src/main/scala/io/buoyant/router/http/AddForwardedHeader.scala
```diff
@@ -42,13 +43,35 @@ class AddForwardedHeader(byLabel: () => String, forLabel: () => String) extends
   }
 }
 
-object AddForwardedHeader {
```
adleong: I think rather than move this into http-base, we should split it into two: one in router/http and one in router/h2

adleong: possibly with common functionality factored into a trait or util object that can stay in http-base

router/base-http/src/main/scala/io/buoyant/router/http/AddForwardedHeader.scala
```diff
@@ -42,13 +43,35 @@ class AddForwardedHeader(byLabel: () => String, forLabel: () => String) extends
...
+class H2AddForwardedHeader(byLabel: () => String, forLabel: () => String) extends SimpleFilter[H2Request, H2Response] {
```
adleong: These filters are protocol specific so they should be moved to router/http and router/h2 respectively.

router/base-http/src/main/scala/io/buoyant/router/http/AddForwardedHeader.scala
```diff
@@ -151,35 +174,81 @@ object AddForwardedHeader {
...
-      val filter = new AddForwardedHeader(byl, forl)
-      self.apply(conn).map(filter.andThen(_))
+  object H1 {
+    case class Enabled(enabled: Boolean)
+    implicit object Param extends Stack.Param[Enabled] {
+      // The RFC indicates that this feature should be disabled by default.
+      val default = Enabled(false)
+    }
+
+    class LabelingProxy(
```
adleong: LabelingProxy can be made generic by accepting the filter as a parameter:

```
class LabelingProxy[Req, Rep](
  byLabeler: Labeler.By,
  forLabeler: Labeler.For,
  addForwardedHeader: Filter[Req, Rep, Req, Rep],
  underlying: ServiceFactory[Req, Rep]
)
```

router/base-http/src/main/scala/io/buoyant/router/http/AddForwardedHeader.scala
```diff
@@ -151,35 +174,81 @@ object AddForwardedHeader {
...
+        // connection. This means that a randomized labeler, for
+        // instance, may reuse labels throughout a client's lifetime.
+        val byl = byLabeler(conn.localAddress)
+        val forl = forLabeler(conn.remoteAddress)
+        val filter = new AddForwardedHeader(byl, forl)
+        self.apply(conn).map(filter.andThen(_))
+      }
     }
+
+    val module: Stackable[ServiceFactory[Request, Response]] =
```
adleong: The module is protocol specific (and I think this is correct since it should instantiate the appropriate AddForwarededHeader filter to pass to LabelingProxy) so should live in router/http and router/h2

yoitsro:
@dadjeibaah I think that would suffice as an appropriate test. 

And yeah - I guess it's always good to have tests for something where people may depend on the format of this string. I know I'm gonna have to parse it in Node.js. Again, thank you :)

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
```diff
@@ -67,6 +67,7 @@ class H2Initializer extends ProtocolInitializer.Simple {
       .replace(H2TraceInitializer.role, H2TraceInitializer.serverModule)
       .prepend(LinkerdHeaders.Ctx.serverModule)
       .prepend(h2.ErrorReseter.module)
+      .insertBefore(H2AddForwardedHeader.module.role, AddForwardedHeaderConfig.module[Request, Response])
```
adleong: Just curious if scala can infer these type parameters if you omit them?

router/http/src/main/scala/io/buoyant/router/http/AddForwardedHeader.scala
```diff
@@ -1,11 +1,13 @@
 package io.buoyant.router.http
 
+import java.net._
```
adleong: alphasort, no blank lines in imports

router/http/src/test/scala/io/buoyant/router/http/AddForwardedHeaderTest.scala
```diff
@@ -51,8 +51,8 @@ class AddForwardedHeaderTest extends FunSuite {
     assert(forwards.toSeq == Seq("for=8.8.4.4", "by=_http;for=8.8.8.8"))
   }
 
-  test("AddForwardedHeader.module obfuscates nodes by default") {
-    val params = Stack.Params.empty + AddForwardedHeader.Enabled(true)
+  test("OAddForwardedHeader.module obfuscates nodes by default") {
```
adleong: typo in test name

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2AddForwardedHeaderConfigTest.scala
```diff
@@ -0,0 +1,150 @@
+package io.buoyant.linkerd.protocol.h2
+
+import com.fasterxml.jackson.databind.exc.InvalidTypeIdException
+import com.twitter.finagle.Stack
+import io.buoyant.config.Parser
+import io.buoyant.router.RouterLabel
+import io.buoyant.router.http.{AddForwardedHeaderConfig, ForwardedHeaderLabeler}
+import io.buoyant.test.FunSuite
+
+class H2AddForwardedHeaderConfigTest extends FunSuite {
```
adleong: Is any of this H2 specific?  Isn't this just tests of the AddForwardedHeaderConfig, which lives in http-base?

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/AddForwardedHeaderConfigTest.scala
```diff
@@ -4,7 +4,7 @@ import com.fasterxml.jackson.databind.exc.InvalidTypeIdException
 import com.twitter.finagle.Stack
 import io.buoyant.config.Parser
 import io.buoyant.router.RouterLabel
-import io.buoyant.router.http.AddForwardedHeader
+import io.buoyant.router.http.{AddForwardedHeaderConfig, ForwardedHeaderLabeler}
 import io.buoyant.test.FunSuite
 
 class AddForwardedHeaderConfigTest extends FunSuite {
```
adleong: should this move to the http-base project?

router/h2/src/test/scala/io/buoyant/router/h2/H2AddForwardedHeaderTest.scala
```diff
@@ -0,0 +1,113 @@
...
+import com.twitter.finagle._
+import com.twitter.finagle.buoyant.h2.{Method, Request, Response, Stream}
+import com.twitter.util.{Future, Time}
+import io.buoyant.router.http.ForwardedHeaderLabeler
+import io.buoyant.test.FunSuite
+
+class H2AddForwardedHeaderTest extends FunSuite {
+  val OkSvc = Service.mk[Request, Response] { req =>
+
+    val rsp = Response(req.headers, Stream.const("Hello, World!"))
```
adleong: Maybe use Stream.empty so that you don't need to worry about releasing the data frames

router/h2/src/test/scala/io/buoyant/router/h2/H2AddForwardedHeaderTest.scala
```diff
@@ -0,0 +1,113 @@
...
+    Future.value(rsp)
+  }
+
+  val OkStack = H2AddForwardedHeader.module
+    .toStack(Stack.Leaf(Stack.Role("endpoint"), ServiceFactory.const(OkSvc)))
+
+  def mkReq(authority: String = "svc/name", proto: String = "http") =
+    Request(proto, Method.Get, authority, "Some/Path", Stream.const("Hello, World!"))
```
adleong: Maybe use Stream.empty so that you don't need to worry about releasing the data frames

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
dadjeibaah: Good point, I can try that out.

linkerd/protocol/h2/src/test/scala/io/buoyant/linkerd/protocol/h2/H2AddForwardedHeaderConfigTest.scala
dadjeibaah: That is correct, thanks for catching that. It is definitely redundant to have to test for this all over again.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
dadjeibaah: So, I tried to omit the types but Scala infers that the type is `com.twitter.finagle.Stackable[com.twitter.finagle.ServiceFactory[Nothing,Nothing]]`. I tried to mess with the `module` method definition to add upper type bounds but it was not successful. The `Request` and Response` type classes for HTTP/1 and HTTP/2 do not share a parent class.

adleong: Weird!  I would have expected type unification to be able to infer the type parameters since this is being called on a `Stack[ServiceFactory[Request, Response]]`.  Anyway, this is fine, I was just curious.

router/core/src/main/scala/io/buoyant/router/LabelerConfig.scala
```diff
@@ -0,0 +1,68 @@
+package io.buoyant.router
+
```
adleong: Can this file be moved into linkerd/core?  Since it's a config file it probably belongs in linkerd rather than router.  This used to be part of AddForwardedHeaderConfig which is also moving to linkerd/core so I think it makes sense for this to move into linkerd/core as well.

dadjeibaah: Just made the change and looks like it is possible. 👍

Issue #1868, opened 2018-03-17
Feature: Marathon services failing health checks should fail over in name tree resolution


hawkw:
## Filing a Linkerd issue ##

Issue Type:

- [ ] Bug report
- [x] Feature request

Currently, the `useHealthChecks` option to the Marathon namer filters out services failing health checks by removing them from the returned bound name. If all replicas of a service are failing health checks, then the returned bound name will be Empty. Dtab alternation will not trigger on an empty name, so if a dtab is configured with an alternate path, Linkerd will route to the empty name, resulting a "no hosts available" exception. If the returned name was Neg instead, then Linkerd would fail over to the alternate path instead.

In order to implement failover, we may want to consider adding an option to treat this case as though the service for which all replicas are failing does not exist in service discovery.

jayeve:
Looking to take a stab at this one. There are certainly engineers more familiar with this specific codebase; specifically @siggy and @adleong. I'm fairly proficient with scala code and can certainly adopt the patterns I've seen used in the linkerd repo; however, I would benefit tremendously if either of you could provide some insight as to where the relevant changes need to be made to address this issue. Thank you in advance for any guidance you may provide. It would be great to contribute to an open source project like this one!

hawkw:
I think it might be worthwhile to consider making this change in a way that can benefit more than just the Marathon namer; similar failover-on-empty-addrs might be useful with other service discovery backends as well?

adleong:
I'm not sure about that.  Finagle explicitly differentiates between Empty (the service exists but has 0 instances) and Neg (the service does not exist).  I'm not sure if we always want to fallback on empty services.

For this issue in particular, take a look at `AppIdNamer.scala`.  We would want to return a `NameTree.Neg` in the case that the address set is empty.  You can find an example of this type of behavior in the Kubernetes namer in `EndpointsNamer.scala` in the `mkNameTree` method.

hawkw:
@adleong I'm thinking this behaviour should be opt-in with a configuration setting --- sorry if that wasn't clear in the issue description!

adleong:
Fixed by https://github.com/linkerd/linkerd/pull/1921

PR #1921, opened 2018-04-26
Handle Marathon app id exists but no available replicas

jayeve:
**Subject**
Return `NameTree.Neg` on no available bound addrs
 
**Problem**
Currently, the Marathon namer will always return a Bound Finagle Name for any Marathon app id that exists. This holds true for app ids that exist but do not have available replicas (no hosts with routable addresses). This could be indicative of a case where all replicas of an app id are unhealthy, non-existent, or otherwise inaccessible. In this case, the Marathon namer should return `NameTree.Neg` to reflect an inaccessible app.
 
**Solution**
This change models the lookup behavior in the [Consul LookupCache](https://github.com/linkerd/linkerd/blob/34e7efb584c297858e7dc1d6133cda8b9cfa41ca/namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala#L45-L52). Any watched `Addr` that is empty will update state to `Addr.Neg`. This change will be picked up in the watched `Activity`, and result in propagating a `NameTree.Neg` (negative result, representing the inaccessible app). In addition, this change will propagate `Addr.Pending` and `Addr.Failed` results through `Activity.Pending` and `Activity.Failed` results, respectively, instead of always returning a Bound Name for any app id determined to exist.

**Validation**
This change includes Unit Test coverage of the new code pathways:
- empty address list in a watched `Addr` results in `NameTree.Neg` propagating from lookup
- A failed result in a watched `Addr` results in a `Activity.Failed` propagating from lookup
- A Pending result in a watched `Addr` results in an `Activity.Pending` propagating from lookup

**Fixes** 
#1868

### DIFF
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
@@ -42,7 +42,7 @@ class AppIdNamer(
       // each time the map of all Apps updates, find the
       // shortest-matching part of `path` that exists as an App ID.
       val possibleIds = (1 to lowercasePath.size).map(lowercasePath.take(_))
-      appsActivity.map { apps =>
+      appsActivity.flatMap { apps =>
         Trace.recordBinary("marathon.path", path.show)
         val found = possibleIds.collectFirst {
           case app if apps(app) => {
@@ -50,19 +50,21 @@ class AppIdNamer(
             val residual = path.drop(app.size)
             val id = prefix ++ app
             val addr = getAndMonitorAddr(app)
-            addr.map {
-              case Addr.Bound(_, _) => NameTree.Leaf(Name.Bound(addr, id, residual))
-              case Addr.Pending => NameTree.Leaf(Name.Bound(addr, id, residual))
-              case Addr.Failed(_) => NameTree.Neg
-              case Addr.Neg => NameTree.Neg
+
+            val observation = addr.map {
+              case Addr.Neg => Activity.Ok(NameTree.Neg)
+              case Addr.Pending => Activity.Pending
+              case Addr.Failed(why) => Activity.Failed(why)
+              case Addr.Bound(_, _) => Activity.Ok(NameTree.Leaf(Name.Bound(addr, id, residual)))
             }
+            Activity.apply(observation)
           }
         }
 
         Trace.recordBinary("marathon.found", found.isDefined)
         found match {
-          case Some(nameTree) => nameTree.sample()
-          case None => NameTree.Neg
+          case Some(nameTree) => nameTree
+          case None => Activity.value(NameTree.Neg)
         }
       }
     }
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
@@ -45,7 +45,7 @@ class AppIdNamer(
       appsActivity.flatMap { apps =>
         Trace.recordBinary("marathon.path", path.show)
         val found = possibleIds.collectFirst {
-          case app if apps(app) => {
+          case app if apps(app) =>
             Trace.recordBinary("marathon.appId", app.show)
             val residual = path.drop(app.size)
             val id = prefix ++ app
@@ -58,7 +58,6 @@ class AppIdNamer(
               case Addr.Bound(_, _) => Activity.Ok(NameTree.Leaf(Name.Bound(addr, id, residual)))
             }
             Activity(observation)
-          }
         }
 
         Trace.recordBinary("marathon.found", found.isDefined)
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
@@ -57,7 +57,7 @@ class AppIdNamer(
               case Addr.Failed(why) => Activity.Failed(why)
               case Addr.Bound(_, _) => Activity.Ok(NameTree.Leaf(Name.Bound(addr, id, residual)))
             }
-            Activity.apply(observation)
+            Activity(observation)
           }
         }
 
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
@@ -45,16 +45,23 @@ class AppIdNamer(
       appsActivity.map { apps =>
         Trace.recordBinary("marathon.path", path.show)
         val found = possibleIds.collectFirst {
-          case app if apps(app) =>
+          case app if apps(app) => {
             Trace.recordBinary("marathon.appId", app.show)
             val residual = path.drop(app.size)
             val id = prefix ++ app
             val addr = getAndMonitorAddr(app)
-            Name.Bound(addr, id, residual)
+            addr.map {
+              case Addr.Bound(_, _) => NameTree.Leaf(Name.Bound(addr, id, residual))
+              case Addr.Pending => NameTree.Leaf(Name.Bound(addr, id, residual))
+              case Addr.Failed(_) => NameTree.Neg
+              case Addr.Neg => NameTree.Neg
+            }
+          }
         }
+
         Trace.recordBinary("marathon.found", found.isDefined)
         found match {
-          case Some(name) => NameTree.Leaf(name)
+          case Some(nameTree) => nameTree.sample()
           case None => NameTree.Neg
         }
       }
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
@@ -111,7 +111,7 @@ class AppIdNamer(
             pending = api.getAddrs(app).respond {
               case Return(addrs) =>
                 initialized = true
-                addr() = Addr.Bound(addrs)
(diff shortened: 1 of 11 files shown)


marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -45,16 +45,23 @@ class AppIdNamer(
       appsActivity.map { apps =>
         Trace.recordBinary("marathon.path", path.show)
         val found = possibleIds.collectFirst {
-          case app if apps(app) =>
+          case app if apps(app) => {
```
adleong: No need for braces on cases 😉

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -45,16 +45,23 @@ class AppIdNamer(
       appsActivity.map { apps =>
```
adleong: To make the types work out, I think you want to change this `map` to `flatMap` and have the function return an `Activity[NameTree[Name]]` (see commend below)

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -45,16 +45,23 @@ class AppIdNamer(
...
         Trace.recordBinary("marathon.path", path.show)
         val found = possibleIds.collectFirst {
-          case app if apps(app) =>
+          case app if apps(app) => {
             Trace.recordBinary("marathon.appId", app.show)
             val residual = path.drop(app.size)
             val id = prefix ++ app
             val addr = getAndMonitorAddr(app)
-            Name.Bound(addr, id, residual)
+            addr.map {
```
adleong: This is a `Var[NameTree[Name]]` but I think it needs to be `Activity[NameTree[Name]]` to match with the flatMap (see comment above).

You can use `addr.map` to map Addr states into `Activity.State` and get a `Var[Activity.State[NameTree[Name]]]`.  Then you can use `Activity.apply` to turn that into an `Activity[NameTree[Name]]`.

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -45,16 +45,23 @@ class AppIdNamer(
...
+              case Addr.Bound(_, _) => NameTree.Leaf(Name.Bound(addr, id, residual))
+              case Addr.Pending => NameTree.Leaf(Name.Bound(addr, id, residual))
```
adleong: e.g. Addr.Pending should map to Activity.State.Pending.

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -45,16 +45,23 @@ class AppIdNamer(
...
+              case Addr.Failed(_) => NameTree.Neg
```
adleong: And probably Addr.Failed should map to Activity.State.Failed

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -45,16 +45,23 @@ class AppIdNamer(
...
+              case Addr.Neg => NameTree.Neg
+            }
+          }
         }
+
         Trace.recordBinary("marathon.found", found.isDefined)
         found match {
-          case Some(name) => NameTree.Leaf(name)
+          case Some(nameTree) => nameTree.sample()
```
adleong: using flatMap above lets us avoid needing `sample()` here.

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -42,20 +42,29 @@ class AppIdNamer(
       // each time the map of all Apps updates, find the
       // shortest-matching part of `path` that exists as an App ID.
       val possibleIds = (1 to lowercasePath.size).map(lowercasePath.take(_))
-      appsActivity.map { apps =>
+      appsActivity.flatMap { apps =>
         Trace.recordBinary("marathon.path", path.show)
         val found = possibleIds.collectFirst {
-          case app if apps(app) =>
+          case app if apps(app) => {
```
adleong: No need for braces here

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -42,20 +42,29 @@ class AppIdNamer(
...
             Trace.recordBinary("marathon.appId", app.show)
             val residual = path.drop(app.size)
             val id = prefix ++ app
             val addr = getAndMonitorAddr(app)
-            Name.Bound(addr, id, residual)
+
+            val observation = addr.map {
+              case Addr.Neg => Activity.Ok(NameTree.Neg)
+              case Addr.Pending => Activity.Ok(NameTree.Leaf(Name.Bound(addr, id, residual)))
```
adleong: I think this should be Activity.Pending

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
```diff
@@ -42,20 +42,29 @@ class AppIdNamer(
...
+              case Addr.Failed(why) => Activity.Failed(why)
+              case Addr.Bound(_, _) => Activity.Ok(NameTree.Leaf(Name.Bound(addr, id, residual)))
+            }
+            Activity.apply(observation)
```
adleong: This can be shortened to just `Activity(observation)`

marathon/src/main/scala/io/buoyant/marathon/v2/AppIdNamer.scala
jayeve: As `Activity.Pending`, there are quite a few failures
```
[info] - Namer handles looking up /app/id *** FAILED ***              
[info]   com.twitter.util.Activity$Pending$@294151d6 did not equal Ok(Leaf(Path(io.l5d.marathon,service,name))) (AppIdNamerTest.scala:68)    
[info] - Namer handles looking up /app/id case-insensitive *** FAILED ***                                                                    
[info]   com.twitter.util.Activity$Pending$@294151d6 did not equal Ok(Leaf(Path(io.l5d.marathon,service,name))) (AppIdNamerTest.scala:80)    
[info] - Namer updates when blocking call from getAppIds returns *** FAILED ***                                                              
[info]   com.twitter.util.Activity$Pending$@294151d6 did not equal Ok(Leaf(Path(io.l5d.marathon,servicename))) (AppIdNamerTest.scala:93)                
[info] - Addrs update when blocking call for getAddrs returns *** FAILED ***                                                                 
[info]   com.twitter.util.Activity$Pending$@294151d6 is not a NameTree.Leaf[Name.Bound] (AppIdNamerTest.scala:140)                           
[info] - Namer recovers if marathon api fails initially *** FAILED ***                                                                       
[info]   com.twitter.util.Activity$Pending$@294151d6 did not equal Ok(Leaf(Path(io.l5d.marathon,foo,bar))) (AppIdNamerTest.scala:171)        
[info] - Namer returns a cached address when the marathon api goes down *** FAILED ***                                                       
[info]   com.twitter.util.Activity$Pending$@294151d6 did not equal Ok(Leaf(Path(io.l5d.marathon,foo,bar))) (AppIdNamerTest.scala:192)        
[info] - Namer returns failure if getAddrs() fails due to marathon api being down *** FAILED ***                                             
[info]   Failed(com.twitter.finagle.ChannelWriteException) is not a NameTree.Leaf[Name] (AppIdNamerTest.scala:111)                           
[info] - Namer recovers if getAddrs() fails initially *** FAILED ***  
[info]   Failed(com.twitter.finagle.ChannelWriteException) is not a NameTree.Leaf[Name] (AppIdNamerTest.scala:111)
```

As `Activity.Ok(NameTree.Leaf(Name.Bound(...)))` there are still failures, but fewer
```
[info] - Namer returns failure if getAddrs() fails due to marathon api being down *** FAILED ***                                             
[info]   Failed(com.twitter.finagle.ChannelWriteException) is not a NameTree.Leaf[Name] (AppIdNamerTest.scala:111)                           
[info] - Namer recovers if getAddrs() fails initially *** FAILED ***  
[info]   Failed(com.twitter.finagle.ChannelWriteException) is not a NameTree.Leaf[Name] (AppIdNamerTest.scala:111)
```

Still investigating why that is. It may require changing the tests. Afterall, previously all `found` appIds would be bound regardless of the `addr` value

marathon/src/test/scala/io/buoyant/marathon/v2/AppIdNamerTest.scala
```diff
@@ -135,19 +155,28 @@ class AppIdNamerTest extends FunSuite with Awaits {
     @volatile var state: Activity.State[NameTree[Name]] = Activity.Pending
     namer.lookup(Path.read("/servicename/residual")).states.respond(state = _)
 
-    state match {
```
adleong: should we assert that `state == Activity.Pending` here?

marathon/src/test/scala/io/buoyant/marathon/v2/AppIdNamerTest.scala
```diff
@@ -103,12 +127,8 @@ class AppIdNamerTest extends FunSuite with Awaits {
   }
 
   def assertAddrsFail(state: Activity.State[NameTree[Name]]) = state match {
```
adleong: now that this no longer checks the Addr, I think this should be renamed.

adleong:
These test changes look good to me!

PR #1933, opened 2018-05-04
add a note that highlights transformer and client tls behavior

### LINKED ISSUE #1930: TLS prefix should include all transformers used in linkerd config
**What happened**:
A user ran into TLS issues when using Client TLS with a prefix configuration. Below is a portion of the config that caused the problem.

```YAML
namers:
    - kind: io.l5d.k8s
      prefix: /io.l5d.k8s.out
      host: localhost
      port: 8001
      transformers:
      - kind: io.l5d.k8s.daemonset
        namespace: default
        port: incoming
        service: l5d
        hostNetwork: true
routers:
    - protocol: http
      label: outgoing
      dtab: |
        /k8s          =>   /#/io.l5d.k8s.out;
        /portNsSvc    =>   /#/portNsSvcToK8s;
        /host         =>   /portNsSvc/http/default;
        /host         =>   /portNsSvc/http;
        /svc          =>   /$/io.buoyant.http.domainToPathPfx/host;
servers:
      - port: 4140
        ip: 0.0.0.0
      client:
        kind: io.l5d.static
        configs:
        - prefix: /#/io.l5d.k8s.out
          tls:
            commonName: linkerd
            trustCerts:
            - /io.buoyant/linkerd/tls/ca.crt
``` 

The issue is that when linkerd is building client names for routing, transformers present in the linkerd configuration will be prepended to a client's dtab name. At the end of the identification phase, in the example above, Linkerd will have a name in the form of:
```
/io.l5d.k8s.daemonset/default/incoming/l5d/#/io.l5d.k8s.out/<k8s service name>...
```
This name does not match the client prefix name specified in the client TLS section. Because of this, linkerd unable to initiate TLS for outgoing clients.

This behavior is not really documented anywhere and it is pretty subtle. So, we should update the docs to highlight this.

dadjeibaah:
There is a small 'gotcha' users can encounter when using client TLS with kind `io.l5d.static` with transformers.  when linkerd is building client names for routing, transformers present in the linkerd configuration will be prepended to a client's dtab name. At the end of the identification phase, this name will have a different prefix compared to the one set in the client TLS section.

This PR highlights this behavior so that users understand that they may run into this scenario.

fixes #1930


linkerd/docs/client_tls.md
```diff
@@ -87,3 +87,10 @@ routers:
           certPath: /certificates/cert.pem
           keyPath: /certificates/key.pem
 ```
+
+<aside class="notice">
+When using any kind of transformer with client TLS using `io.l5d.static`, ensure that the client
+TLS's `prefix` field starts with the first transformer that is applied to a client name in the dtab.
+This will ensure that TLS is initiated between linkerd and any downstream service that matches the
+client prefix.
```
adleong: I think this could use a little more information about how to determine the full client name including the transformer prefix.  Probably the easiest way would be click the client name to expand it in the Linkerd admin dashboard.  An example would probably be helpful.

dadjeibaah: Good point, I could add information that points to the link to the client name. Hopefully, that is available in the admin endpoint.

linkerd/docs/client_tls.md
```diff
@@ -87,3 +87,44 @@ routers:
...
+      kind: default
+      transformers:
+      - kind: io.l5d.port
+        port: 4141
+  dtab: | 
+    /svc => /#/io.l5d.fs;
+  client:
+    kind: io.l5d.static
+    configs:
+    - prefix: /#/io.l5d.fs/{service}
```
adleong: Should this be `/%/io.l5d.port/4141/#/io.l5d.fs/{service}`?

linkerd/docs/client_tls.md
```diff
@@ -87,3 +87,44 @@ routers:
           certPath: /certificates/cert.pem
           keyPath: /certificates/key.pem
 ```
+
+### Client TLS and transformers
+
+When using any kind of transformer with client TLS using `io.l5d.static`, ensure that the client
```
adleong: I think this can benefit from a bit more background:

> If you use transformers, each transformer will prepend a transformer prefix to the client name to indicate that it has been transformed.  Transformer prefixes always start with `/%`.  When using the `io.l5d.static` client configuration, ensure that the client prefix includes the transformer prefix.  For example....

linkerd/docs/client_tls.md
```diff
@@ -87,3 +87,44 @@ routers:
...
+        commonName: "{service}.linkerd.io"
+        trustCerts:
+        - /certificates/cacert.pem
+        clientAuth:
+          certPath: /certificates/cert.pem
+          keyPath: /certificates/key.pem
+```
+
+The full client name generated by Linkerd for service `hello` would look like 
+`%/io.l5d.port/#/io.l5d.fs/4141/hello`. Therefore to set up TLS for the service `hello`, the prefix
```
adleong: missing a leading slash

adleong: shouldn't the port number be part of the port transformer prefix?

linkerd/docs/client_tls.md
```diff
@@ -87,3 +87,44 @@ routers:
...
+      tls:
+        commonName: "{service}.linkerd.io"
+        trustCerts:
+        - /certificates/cacert.pem
+        clientAuth:
+          certPath: /certificates/cert.pem
+          keyPath: /certificates/key.pem
+```
+
+The full client name generated by Linkerd for service `hello` would look like
```
adleong: Maybe give a bit more explanation:

> Since the base client name is `/#/io.l5d.fs/hello` and the port transformer applies `/%/io.l5d.port/4141` as a transformer prefix, the transformed client name is `/%/io.l5d.port/4141/#/io.l5d.fs/hello`.

linkerd/docs/client_tls.md
```diff
@@ -87,3 +87,44 @@ routers:
...
+        clientAuth:
+          certPath: /certificates/cert.pem
+          keyPath: /certificates/key.pem
+```
+
+The full client name generated by Linkerd for service `hello` would look like 
+`%/io.l5d.port/#/io.l5d.fs/4141/hello`. Therefore to set up TLS for the service `hello`, the prefix 
+should be `%/io.l5d.port/#/io.l5d.fs/4141`
+
+When multiple transformers are used for a Linkerd instance, it may be difficult to determine the
```
adleong: I'd remove the bit about multiple transformers... it can be difficult even if there is a single transformer.

linkerd/docs/client_tls.md
```diff
@@ -87,3 +87,44 @@ routers:
...
+
+The full client name generated by Linkerd for service `hello` would look like 
+`%/io.l5d.port/#/io.l5d.fs/4141/hello`. Therefore to set up TLS for the service `hello`, the prefix 
+should be `%/io.l5d.port/#/io.l5d.fs/4141`
+
+When multiple transformers are used for a Linkerd instance, it may be difficult to determine the 
+exact full client name Linkerd uses to send TLS traffic to the downstream service. A quick way to 
+find out this information is to look through Linkerd's admin dashboard UI. On the initial dashboard 
+page, the `client` section displays active clients used to connect to downstream services. 
+Clicking on any of the client names in that section will reveal the full client name. The text that
```
adleong: > will reveal the full client name including any transformer prefixes.

linkerd/docs/client_tls.md
dadjeibaah: That is correct. I will make update the config example.

dadjeibaah: A strange thing encountered was that the client name in the UI did not have a beginning forward slash. The user that reported this issue mentioned that they added a prefix that began with `%/`. From my understanding, it should start with the `%/` but I am not entirely sure.

linkerd/docs/client_tls.md
dadjeibaah: In the admin UI, the port number came after the fs namer prefix. It might be that I did not set up the config properly.

Issue #1906, opened 2018-04-23
memory (and possibly observations) leak in io.l5.mesh interface


edio:
## Filing a Linkerd issue ##

Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

Over time namerd memory usage increases if `io.l5d.mesh` interface is used

**What you expected to happen**:

Memory usage won't increase

**How to reproduce it (as minimally and precisely as possible)**:
1. run consul in dev mode
```
consul agent -dev
```

2. run namerd
namerd.yaml
```
admin:
  ip: 127.0.0.1
  port: 9001
namers:
- kind: io.l5d.consul
  prefix: /consul
interfaces:
- kind: io.l5d.mesh
  ip: 0.0.0.0
  port: 4321
telemetry:
  - kind: io.l5d.influxdb
storage:
  kind: io.l5d.inMemory
  namespaces:
    default: |
      /service => /#/consul/dc1;
```
```
./namerd-1.3.7-exec namerd.yaml
```
3. run the following code (all dependencies should be easy to figure out, but let me know if you want me to publish a gradle project)
```
import com.twitter.conversions.time._
import com.twitter.finagle.buoyant.H2
import com.twitter.finagle.service.Backoff
import com.twitter.finagle.util.DefaultTimer
import com.twitter.finagle.{Dtab, Name, Path}
import com.twitter.util.{Await, Closable, Duration, Future, Time, TimeoutException, Timer}
import io.buoyant.interpreter.mesh.Client

object Scenario extends Closable {
  implicit val timer = DefaultTimer
  private[this] val name = Name.Path(Path.read("/$/inet/localhost/4321"))
  private[this] val namespace = Path.read("/default")
  private[this] val label = "io.l5d.mesh"
  private[this] val backoffs = Backoff.exponentialJittered(100.millis, 1.second)
  private[this] val client = H2.client.withParams(H2.client.params).newService(name, label)
  private[this] val interpreter = Client(namespace, client, backoffs, DefaultTimer)

  private[this] val opTimeout = 2.seconds
  private[this] val minPeriod = 10.millis

  override def close(deadline: Time): Future[Unit] = client.close(deadline)
    .raiseWithin(opTimeout, new TimeoutException("failed to close within timeout"))

  def execute(): Future[Unit] = Future.sleep(minPeriod)
    .join(interpreter.bind(Dtab.empty, Path.read(s"/service/consul")).values.toFuture()
[...]

PR #1934, opened 2018-05-04
Rework H2 stream cancellation

adleong:
There does not exist a reliable mechanism for the consumer of an H2 stream to cancel/reset it.  For example, consider a client has received a response stream from the server but wishes to terminate the stream without reading to the end.  The client can attempt to read a frame from the response stream and immediately interrupt the read, which may reset the stream, but this is timing dependent and will not always succeed.  The client may also use the request stream to issue a reset but this is awkward because the caller may not have kept a reference to the request stream, especially if the request stream is complete.  

We refactor the Stream interface to remove `reset` from `Stream.Writer` and add `cancel` to `Stream`, as well as an `onCancel` Future.  Consumers of a Stream may call `cancel` to indicate that they wish to terminate the stream.  Producers to the stream may listen to the `onCancel` Future to be notified of a consumer initiated stream termination.

We also substantially reworked Netty4StreamTransport to make the handling of resets clearer.  Rather than having dependently nested stream states, we now keep an independent state for the send (local) side and the recv (remote) side.  We also centralized the reset logic as much as possible into a `reset` method.  

A note to reviewers: Netty4StreamTransport is enough of a rewrite that I recommend reading the old and new versions of this file separately instead of trying to make sense of the diff.  I will also provide further context about changes as inline comments on this PR.

This change unlocks further improvements to the interface for cancellation in gRPC.  In turn that will allow us to add proper cancellation to the io.l5d.mesh gRPC API to address #1906


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -88,16 +110,12 @@ object Stream {
...
-
     /**
      * Write an object to a Stream so that it may be read as a Frame
      * (i.e. onto an underlying transport). The returned future is not
      * satisfied until the frame is written and released.
      */
     def write(frame: Frame): Future[Unit]
-
-    def reset(err: Reset): Unit
-    def close(): Unit
```
adleong: This method had no purpose.  To close a Stream, simply write a frame with eos=true.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -162,24 +186,14 @@ object Stream {
   def const(s: String): Stream =
     const(Buf.Utf8(s))
 
-  def empty(): Stream with Writer = empty(new AsyncQueue[Frame](1))
-
-  def empty(frameQ: AsyncQueue[Frame]): Stream with Writer =
```
adleong: An empty Stream should never have frames written to it or read from it.  The only reason for this queue was to allow cancellation by polling the queue and then immediately interrupting it.  We have a better mechanism for cancellation now (the cancel method).

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/StreamProxy.scala
```diff
@@ -27,7 +31,6 @@ abstract class StreamProxy(underlying: Stream) extends Stream {
 class StreamOnFrame(underlying: Stream, onFrame: Try[Frame] => Unit)
   extends StreamProxy(underlying) {
   override def read(): Future[Frame] = underlying.read().respond(onFrame)
-  override def toString: String = s"StreamProxy($underlying, onFrame=$onFrame)"
```
adleong: These made for very messy logs when logging StreamState.

finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
```diff
@@ -313,8 +311,8 @@ class Netty4StreamTransportTest extends FunSuite {
     assert(!transport.onReset.isDefined)
     assert(!transport.isClosed)
 
-    reqStream.reset(Reset.InternalError)
-    assert(await(endF.liftToTry) == Throw(StreamError.Local(Reset.InternalError)))
```
adleong: We now use the StreamError type only in `onReset` to signal to the dispatcher if a reset needs to be written to the remote.

finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
```diff
@@ -351,31 +349,24 @@ class Netty4StreamTransportTest extends FunSuite {
...
-    ctx.rspF.raise(Failure("foo").flagged(Failure.Rejected))
+    ctx.rspF.raise(Reset.Refused)
     assert(await(ctx.rspF.liftToTry) == Throw(Reset.Refused))
   }
 
-  test("client: response cancelation resets stream: Exception") {
-    val ctx = new ClientCtx {}
-    assert(!ctx.rspF.isDefined)
-    ctx.rspF.raise(new Exception("ugh"))
-    assert(await(ctx.rspF.liftToTry) == Throw(Reset.InternalError))
```
adleong: Reset.Cancel seems like a better default when a Future is interrupted than Reset.InternalError to me.

finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
```diff
@@ -510,7 +499,7 @@ class Netty4StreamTransportTest extends FunSuite {
     assert(rspStreamQ.offer(Netty4Message.Trailers(trailers)))
     ctx.synchronized {
       assert(written.head == new DefaultHttp2HeadersFrame(trailers, true)
-        .stream(H2FrameStream(id, Http2Stream.State.OPEN)))
+        .stream(H2FrameStream(id, Http2Stream.State.HALF_CLOSED_LOCAL)))
```
adleong: We now change the stream state _before_ writing the frame.  I think that's okay???

finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
```diff
@@ -654,7 +643,7 @@ class Netty4StreamTransportTest extends FunSuite {
     val ctx = new ServerCtx {}
     import ctx._
 
-    assert(transport.recv({
+    assert(!transport.recv({
```
adleong: These frames have illegal headers and are not admitted.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
```diff
@@ -114,7 +114,7 @@ object ServerDispatcher {
               case s: GrpcStatus => s
               case e => GrpcStatus.Internal(e.getMessage)
             }
-            frames.write(status.toTrailers).onSuccess(_ => frames.close())
```
adleong: Writing a trailers frame is sufficient to close the stream.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
```diff
@@ -33,6 +49,12 @@ trait Stream {
    */
   def onEnd: Future[Unit]
 
+  /**
+   * Satisfied when the consumer of a stream cancels it by calling
+   * `cancel()`.
+   */
+  def onCancel: Future[Reset]
```
olix0r: Can this fail?  Does cancelling this future do anything?

If "no" to both of these, consider using an interface like `def onCancel(f: Reset => Unit)` and then have impls like `override def onCancel(f: Reset => Unit) { cancelP.onSuccess(f) }`.

This would have the benefit of hiding some of the complicated Future semantics from this part of the interface if they're not necessary.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
```diff
@@ -89,6 +89,9 @@ class BufferedStream(underlying: Stream, bufferCapacity: Long = 16383) { buffere
             }
           }
         }
+        stream.onCancel.onSuccess { rst =>
+          underlying.cancel(rst)
+        }
```
olix0r: I'm not clear enough on the semantics of `fork()` to know that this is safe...  This means that if any reader cancels then all other readers will also be reset? How does this interact with retries?  Does this need to refcount underlying?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Stream.scala
olix0r: Also, if we'd ever want to join/select over onCancels, the callback approach would be suboptimal.  worth asking the question though

adleong: good suggestion, thanks

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
adleong: The semantics of resets when retries are involved aren't really externally specified so it's somewhat up to us.  I think it makes sense to reset everything once a single reader cancels.

Consider when the client which sends a small but unterminated stream.  If the server resets that stream, I would interpret that to be a signal to the client to stop producing.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -306,208 +211,170 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
+        if (reset(err, local = false)) {
+          statsReceiver.remoteResetCount.incr()
+          true
+        } else false
 
+      /*
+       * HEADERS FRAME with EOS (TRAILERS)
+       */
       case hdrs: Http2HeadersFrame if hdrs.isEndStream =>
+        //Such a Frame may be received while in RecvPending (headers
```
olix0r: space

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -306,208 +211,170 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
-            } else recv(data)
-
-          case LocalClosed(remote@RemoteStreaming()) =>
-            if (stateRef.compareAndSet(state, Closed(Reset.NoError))) {
-              if (recvFrame(toFrame(data), remote)) {
-                remote.close()
+          // Pending
+          case StreamState(sendState, RecvPending) =>
+            log.debug("[%s] data frame received while in RecvPending", prefix)
+            reset(Reset.InternalError, local = true)
```
olix0r: this might be considered a ProtocolError?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/BufferedStream.scala
olix0r: On the other hand, consider that resets can be used to signal things like rate limiting.

I think this behavior is Okay for now, but this may warrant some more thought in general.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -306,208 +211,170 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
-              if (stateRef.compareAndSet(state, Closed(Reset.NoError))) {
-                remote.setMessage(msg)
-                resetP.setDone()
+              if (stateRef.compareAndSet(state, StreamState(sendState, RecvClosed))) {
+                stream.onCancel.onSuccess { rst =>
+                  // If the recvMsg stream is cancelled, reset the stream.
+                  reset(rst, local = true); ()
+                }
+                recvMsg.setValue(msg)
+                if (sendState == SendClosed) {
```
olix0r: i imagine we don't actually need to set the onCancel handler when sendState is SendClosed? (because this basically means we're in full-close///

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -306,208 +211,170 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
...
-          case RemoteOpen(remote@RemoteStreaming()) =>
-            if (resetFromLocal(remote, Reset.InternalError)) false
-            else recv(hdrs)
-
-          case Open(remote@RemotePending()) =>
+          // Pending
+          case StreamState(sendState, RecvPending) =>
             val q = new AsyncQueue[Frame]
-            val msg = mkRecvMsg(hdrs.headers, Stream(q))
+            val stream = Stream(q)
```
olix0r: here (and throughout), might be helpful to call this `recvStream` -- it can be a bit confusing which stream we're talking about at times.

Issue #1411, opened 2017-06-16
proxy loop issue in linkerd


edio:
Hello,

linkerd suffers from a so called _proxy loop_ issue, when it tries to proxy request where target address is the proxy itself or if request eventually reaches the proxy after interim hops.

STR:
1. run linkerd with the following router configured
```
routers:
- protocol: http
  label: proxyissue
  dstPrefix: /http
  servers:
  - port: 4140
    ip: 0.0.0.0
  dtab: |
      /http=>/$/io.buoyant.hostportPfx/split;
      /split=>/$/inet;
```
2. make a request to `localhost:4140` using executed linkerd instance as a proxy

AR: linkerd starts spawning threads until original request timeouts, OOM happens or request gets too large (due to large `Via` header value)
ER: I'd expect linkerd to detect proxy loop and return 400 with explanation that proxy loop is detected

Some thoughts on approaches to resolve the issue.

Linkerd already adds `Via` header. So 2 approaches:

1) actual loop detection algorithm
Con: we'd need then to update linkerd to put some unique identifier instead of `linkerd $version` to `Via` header. Potentially slow

2) heuristic: `Via` header value length <= threshold 
Con: can give false positives if infrastructure is built that way, that large number of hops is occasionally required. But practically hardly ever possible.

I checked if it's possible to workaround issue with delegations or plugins, and that's what I've found:

1) plugin
generally not possible due to https://github.com/linkerd/linkerd/issues/1255
If at least namer was allowed to check headers, we might return some `/failure#` from it, while having delegation `/failure#=>!`

2) dtab
this one could fix the issue for loops of length 1
Something like this 
```/inet#=>/$/inet;
/inet#/localhost/4140=>!;
```
It is by no means a general solution, because there might be many names which are resolved to `127.0.0.1`
So the only thing we can do atm, is to add delegation `/inet#/*/4140=>!` and we should remember then, that `4140` is reserved even if no linkerd is running on host.

Both dtab and plugins won't allow returning code other than 502.

So it seems, that it is linkerd core, that should contain logic for proxy loop detection.

wmorgan:
Thanks @edio! Great writeup and thank you for all the alternatives. We'll take a look.

adleong:
related to #87

zaharidichev:
[...]

PR #1939, opened 2018-05-09
Intercept requests to provide request destination information

### LINKED ISSUE #1732: Ability to see how linkerd would resolve a request
Issue Type:

- [ ] Bug report
- [x] Feature request

As [discussed in Slack briefly](https://linkerd.slack.com/archives/C0JV5E7BR/p1507222184000355) before, it would be extremely useful (I'd go so far as to say it's very important for us) to be able to query linkerd to see how it would route a request for a given path.

We have seen a number of cases where linkerd has gotten in a "stuck" state and continues to route requests to endpoints which have been replaced. This is possibly related to #1730 but I am not _sure_ if they're the same. In any case that is not what this ticket is about. Until we can track down the root cause of these issues, we have built a script that can identify _which_ linkerd pods are in such a state and restart them. To do this, it currently relies on using the `/delegator.json` endpoint of the admin UI, but I understand this bypasses the local caches linkerd would normally use when routing traffic. #1731 also prevents this from being completely useful.

What I would like is some form of endpoint I can call with a path, and it will return the set of addresses to which it may route such a request.

Again sorry for the brief description; very happy to elaborate on this further if helpful.

dadjeibaah:
One way to obtain endpoint information for a particular service is to use the `delegator.json` or the dtab admin UI. The `delegator.json` requires that a user sends a `POST` request to the `delegator.json` endpoint with JSON containing the router label, the dtab the linkerd instance is using, and a service name. An alternative way get endpoint information is to access the dtab admin UI to find out where a request may end up.

Troubleshooting services using these techniques can sometimes be a challenge in Linkerd, especially when debugging linkerd in a complex service mesh. 

This PR introduces a feature that helps users obtain service-specific information by issuing an HTTP `TRACE` request with an `l5d-add-context` header. Linkerd forwards the request to a downstream service associated with the service name. When Linkerd receives a response, it will append its router context to the response body and send the response back to the calling client. The router context includes the service name used to fulfill the request, the client name Linkerd ends up binding to, the set of  IP addresses used for load balancing, the actual IP address Linkerd forwards the request to and the dtab resolution steps that show how Linkerd identifies a service name.

Usage:

_Assuming we have Linkerd in a linker-linker set up using transformers and a service that responds with `hi` to a `TRACE` request_

```bash 
$ http_proxy=http://localhost:4140 curl -s -X TRACE -H "Max-Forwards:3" -H "l5d-add-context:true" http://cat

* Rebuilt URL to: http://cat/
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 4140 (#0)
> TRACE http://cat/ HTTP/1.1
> Host: cat
> User-Agent: curl/7.58.0
> Accept: */*
> Proxy-Connection: Keep-Alive
> Max-Forwards:3
> l5d-add-context:true
>
< HTTP/1.1 200 OK
< Via: 1.1 linkerd, 1.1 linkerd
< l5d-success-class: 1.0
< Content-Length: 738
<
hi

--- Router: incoming ---
request duration: 7 ms
service name: /svc/cat
client name: /%/io.l5d.localhost/#/io.l5d.fs/cat
addresses: [127.0.0.1:8888]
selected address: 127.0.0.1:8888
dtab resolution:
  /svc/cat
  /#/io.l5d.fs/cat (/svc=>/#/io.l5d.fs)
  /%/io.l5d.localhost/#/io.l5d.fs/cat (SubnetLocalTransformer)

--- Router: outgoing ---
request duration: 19 ms
service name: /svc/cat
client name: /%/io.l5d.port/4141/#/io.l5d.fs/cat
addresses: [127.0.0.1:4141]
selected address: 127.0.0.1:4141
dtab resolution:
  /svc/cat
  /b/cat (/svc=>2.00*/a & /b)
  /bb/cat (/b=>/bb)
  /#/io.l5d.fs/cat (/bb=>/#/io.l5d.fs)
  /%/io.l5d.port/4141/#/io.l5d.fs/cat (DelegatingNameTreeTransformer$)
```
(Read from bottom up)
We observe that the `TRACE` request we goes through a linkerd instance on the outgoing router then to the incoming router of the linkerd instance and then finally reaches the cat service.

Integration and unit tests were created to verify this new behavior. I also stood up a `client -> linkerd -> server` environment to carry out tests.

fixes #1732


adleong:
Awesome!  Before I review the code, I'll review the output:

```
$ curl -s localhost:4140/linkerd -H "l5d-req-evaluate: 1" | jq
{
  "identification": "/svc/linkerd",
  "selectedAddress": "/104.24.112.2:80",
  "addresses": [
    "/104.24.113.2:80",
    "/104.24.112.2:80"
  ],
  "dtabResolution": {
    "type": "alt",
    "path": "/svc/linkerd",
    "alt": [
      {
        "type": "delegate",
        "path": "/example/linkerd",
        "dentry": {
          "prefix": "/svc",
          "dst": "/example"
        },
        "delegate": {
          "type": "neg",
          "path": "/#/io.l5d.fs/linkerd",
          "dentry": {
            "prefix": "/example",
            "dst": "/#/io.l5d.fs"
          }
        }
      },
      {
        "type": "delegate",
        "path": "/example/default",
        "dentry": {
          "prefix": "/svc/*",
          "dst": "/example/default"
        },
        "delegate": {
          "type": "leaf",
          "path": "/#/io.l5d.fs/default",
          "dentry": {
            "prefix": "/example",
            "dst": "/#/io.l5d.fs"
          },
          "bound": {
            "addr": {
              "type": "bound",
              "addrs": [
                {
                  "ip": "104.24.113.2",
                  "port": 80,
                  "meta": {}
                },
                {
                  "ip": "104.24.112.2",
                  "port": 80,
                  "meta": {}
                }
              ],
              "meta": {}
            },
            "id": "/#/io.l5d.fs/default",
            "path": "/"
          }
        }
      }
    ]
  }
}
```

I think it would be awesome to have both human readable and json outputs.  What do you think about human readable by default but json if the `Accept: application/json` header is passed?  

We should also add the client name (`/#/io.l5d.fs/default`) as a top level field probably, rather than it just being in the delegate tree.

Delegate trees, especially in json format, are pretty confusing.  I wonder if we can distill this into something more easily consumed.  Perhaps by walking the path from the root of the delegate tree to the bound leaf we can output just the delegation steps that lead to it.  e.g.:
```
/svc/linkerd
/example/default     (/svc/* => /example/default)
/#/io.l5d.fs/default (/example => /#/io.l5d.fs)
```
And/or we can print a list of rejected alternatives:
```
/#/io.l5d.fs/linkerd
```

dadjeibaah:
Having both human readable and a JSON formatted response sound like great ideas. What do you think about not having the JSON format output and just rely on the human readable format? I can think of users wanting to use JSON in scripts as a way to monitor the client state in linkerd. I am thinking the non-JSON format might be sufficient for such use cases. I don't think it would be hard to implement both, I just feel it might be good to have only one way of getting request evaluation info.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -47,6 +47,7 @@ class HttpInitializer extends ProtocolInitializer.Simple {
       // ensure the client-stack framing filter is placed below the stats filter
       // so that any malframed responses it fails are counted as errors
       .insertAfter(FailureAccrualFactory.role, FramingFilter.clientModule)
+      .insertAfter(FailureAccrualFactory.role, RequestEvaluator.module)
```
adleong: I just noticed that if you try to use the request evaluator on an address that Linkerd cannot connect to, the connection fails before getting to the evaluator:

```
$ curl localhost:4140 -H "Host: foo" -H "l5d-req-evaluate: 1"
connection timed out: /1.2.3.4:8888 at remote address: /1.2.3.4:8888. Remote Info: Not Available⏎
```

Perhaps if we move the Request higher the stack (above connection establishment) but below the load balancer, this will work?  Perhaps `.insertAfter(LoadBalancerFactory.role)`?

adleong:
I haven't looked at the code yet but I noticed a few issues with the output so:

```
$ curl localhost:4140 -H "Host: default" -H "l5d-req-evaluate: true"

identification: /#/io.l5d.fs/default
selectedAddress: /127.0.0.1:9990
addresses: /127.0.0.1:9990,/127.0.0.2:9990,/127.0.0.3:9990
Dtab Resolution:
/svc/default
1.0 * /x/default
/a/default
/#/io.l5d.fs/default
    ⏎
```

It looks like the identification line is actually printing the client name instead of the service name.  I think this should be: `/svc/default` as the service name and `/#/io.l5d.fs/default` as the client name.

It looks like there's some leading a trailing whitespace.  The formatting of the addresses list could use some encasing characters and spacing the make it more readable.  Consider formatting like `[127.0.0.1:9990, 127.0.0.2:9990, 127.0.0.3:9990]` or something similar.  The dtab resolution could also use some indenting or something to distinguish the resolution output from the field names.

adleong:
```
$ curl localhost:4140 -H "Host: default" -H "l5d-req-evaluate: true"

identification: /#/io.l5d.fs/default
selectedAddress: /127.0.0.1:4140
addresses: /127.0.0.1:4140
Dtab Resolution:
/svc/default
1.0 * /x/default
/a/default
/#/io.l5d.fs/default
    ⏎
```

When I use a dtab with a union, the output appears to be incorrect.  The above comes from this dtab:

```
/b/* => /$/inet/127.1/4140;
/a => /#/io.l5d.fs;
/x => /a & /b;
/svc => /x;
```

The `/b` path of the union has been chosen (thus the final address of 127.1:4140) but the delegation and client name still show the `/a` path being chosen which leads to `/#/io.l5d.fs/default`.

olix0r:
> By adding an `l5d-req-evaluate` header to a request, ...

I'm concerned that this approach, in isolation, violates HTTP semantics.  Caching layers, etc, need to be able to determine how one request differs from another and this cannot reliably be done merely by the presence of one header or another.  In my view, this is application somewhat different from that of dtab overrides because it alters the application semantics entirely.

Ideally, we would use an alternate HTTP verb like `TRACE`. If we can't do that reliably, I think we need to be pretty rigorous about understanding the implications of this implementation.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+
+        val lbSet = DstBoundCtx.current match {
+          case None => Addr.Neg
+          case Some(addrSet) => addrSet.name.addr.sample()
+        }
+
+        val addresses = lbSet match {
+          case Addr.Neg | Addr.Failed(_) | Addr.Pending => None
+          case Addr.Bound(a, _) => Some(
+            a.map { address: Address =>
```
adleong: this can be more condensed I think:

```
a.map {
  case inetAddr: Address.Inet => ...
  case _ =>
}
```

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+        dtab: BaseDtab,
+        interpreter: DstBindingFactory.Namer,
+        next: ServiceFactory[Request, Response]
+      ): ServiceFactory[Request, Response] = new RequestEvaluator(endpoint, interpreter, dtab)
+        .andThen(next)
+    }
+
+  def formatDTree(
+    dTree: Option[DelegateTree[Name.Bound]],
+    output: Any,
```
adleong: is this unused?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+        endpoint: EndpointAddr,
+        dtab: BaseDtab,
+        interpreter: DstBindingFactory.Namer,
+        next: ServiceFactory[Request, Response]
+      ): ServiceFactory[Request, Response] = new RequestEvaluator(endpoint, interpreter, dtab)
+        .andThen(next)
+    }
+
+  def formatDTree(
+    dTree: Option[DelegateTree[Name.Bound]],
```
adleong: all recursive calls to this method are called with a Some here.  Could this just be a `DelegateTree[Name.Bound]` and have the Option handling outside of the method?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+      case Some(Transformation(path, name, value, remainingTree)) =>
+        formatDTree(Some(remainingTree), output, tree :+ value.path.show)
+
+      case Some(Delegate(path, dentry, remainingTree)) =>
+        formatDTree(Some(remainingTree), output, tree :+ path.show)
+
+      case Some(Union(path, dentry, remainingWeightedTrees@_*)) =>
+        remainingWeightedTrees.map { wd =>
+          formatDTree(Some(wd.tree), output, tree :+ s"${wd.weight} * ${path.show}")
+        }.filter(!_.isEmpty).head
```
adleong: this looks like it always takes the first branch of a union

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+              case Return(id) => id
+            },
+            selectedEndpoint,
+            addresses,
+            tree
+          )
+
+          resp.contentType = req.contentType.getOrElse(MediaType.PlainText)
+          resp.contentString = RequestEvaluator
+            .writeContentString(resp.contentType, evaluatedRequest)
```
adleong: I was never super clear on this... should this key off the request's `accept` header instead?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+import com.twitter.util.{Future, Return, Throw, Try}
+import io.buoyant.admin.DelegationJsonCodec
+import io.buoyant.config.Parser
+import io.buoyant.namer.DelegateTree._
+import io.buoyant.namer.{DelegateTree, Delegator}
+import io.buoyant.router.RoutingFactory
+import io.buoyant.router.RoutingFactory.BaseDtab
+import io.buoyant.router.context.{DstBoundCtx, DstPathCtx}
+
+case class EvaluatedRequest(
```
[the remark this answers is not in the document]
adleong: ah, nevermind, I see this exists so that it can be serialized to json.

linkerd/core/src/test/scala/io/buoyant/linkerd/RequestEvaluatorTest.scala
```diff
@@ -0,0 +1,106 @@
...
+  val jsonMapper = Parser.jsonObjectMapper(Nil)
+
+  val successMessage = "request succeeded"
+  val testService = Service.mk[Request, Response] { req =>
+    val rsp = Response()
+    rsp.content = Buf.Utf8(successMessage)
+    Future.value(rsp)
+  }
+
+  val testStack = DstPathCtx.Setter.module.toStack(
```
adleong: I think this can be
`DstPathCtx.Setter.module +: RequestEvaluator.module +: Stack.Leaf(...)`

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+
+class RequestEvaluator(
+  endpoint: EndpointAddr,
+  namer: DstBindingFactory.Namer,
+  dtab: BaseDtab
+) extends SimpleFilter[Request, Response] {
+
+  private val EvaluatorHeaderName = "l5d-req-evaluate"
```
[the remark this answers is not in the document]
hawkw: IIRC it's also more highly optimized by the Scala compiler (at least for fields rather than functions)

olix0r:
@adleong I suspect that it's going to be extremely hard to implement this feature, as described here, in a way that doesn't violate the spec. I'm wary of reaching for a silver bullet to like cache control -- there are going to be a slew of other complications.  For instance, would this feature cause bodies to be returned for HEAD requests?

Do we actually need to alter the response body?  Can this information be encoded in headers? Alternatively, can we buffer this information somewhere such that the user can retrieve this information _after_ the request has completed?

olix0r:
Another approach, along the lines of `TRACE`, would be to introduce a linkerd-specific verb for this.

dadjeibaah:
Thanks, @olix0r for the highlighting the potential spec violation. I wasn't aware of that. 😌 I am curious to learn more about this violation. Do you mind pointing me to where  [RFC 2616](https://tools.ietf.org/html/rfc2616#page-38) highlights this spec?  (I am assuming that the spec info resides in this RFC).

I like the idea of creating a custom HTTP verb for this feature as it gives the evaluator its domain to do its work. I am not sure if Linkerd does not automatically proxy custom HTTP methods. However, I don't think that would be difficult to change if it ends up doing that. I read that proxies and caches treat custom HTTP methods as unsafe and therefore do not cache them. So, maybe that handles the caching problem. Are there any other compatibility issues that we need to worry about when dealing with custom HTTP methods, proxies and caches?

adleong:
I like the idea of a custom Linkerd verb.  I just checked and it looks like Linkerd has no problem proxying custom verbs.  One limitation of this approach is that it doesn't work when you want to evaluate a request and you're using an identifier that keys off of the method such as the io.l5d.methodAndHost identifier.

We could return this information in a header or other side channel but I like the idea that this doesn't actually proxy the request.  It acts as a sort of "dry-run" mode.  "Don't proxy this request, just tell me about it."

A custom verb seems like the best compromise to me.  It's still useful in most cases but without any egregious spec violations (as far as I know 😬).

adleong:
I see this as a feature that has two main use-cases: when getting started with Linkerd and when debugging routing issues.

When getting started with Linkerd, there is a probably a tight feedback loop where the user makes frequent changes to the Linkerd config/dtab, restarts Linkerd, and experiments.  In this case, the user is probably logged into the machine where Linkerd is running and may be testing requests by using curl to localhost.  This feature is useful here because it closes the gap between the request itself and the delegator UI.  

When debugging routing issues, Linkerd is probably already set up and running, potentially serving production traffic.  Changing Linkerd config is costly because it requires a restart.  In this case, the user wants to get more information about how Linkerd would route a request in order to verify that routing is happening as expected or to diagnose why it is not.  In this case the test request may come from curl or from another application and it my pass through intermediaries such as proxies or load balancers before getting to Linkerd.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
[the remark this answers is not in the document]
dadjeibaah: Well, since the new implementation won't rely on JSON, we no longer need the case class, and I have changed it to a method.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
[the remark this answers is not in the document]
dadjeibaah: That makes sense. I will add that.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+  private val JsonCharSet = ";charset=UTF-8"
```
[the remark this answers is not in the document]
dadjeibaah: Thanks for finding that. I will remove it.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        val selectedEndpoint = endpoint.addr match {
+          case inetAddr: Address.Inet => inetAddr.addr.toString
+          case _ => ""
+        }
+
+        val lbSet = DstBoundCtx.current match {
+          case None => Addr.Neg
+          case Some(addrSet) => addrSet.name.addr.sample()
```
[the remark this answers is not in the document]
dadjeibaah: The `Future[Addr]` is pretty neat! Using this made it much easier to read the code IMO.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
dadjeibaah: You're right. I think I could also have `Addr.Bound...` be the first case. I could also condense `case Addr.Neg | Addr.Failed(_) | Addr.Pending` to be the default value for `lbSet`. That way, we no longer need strict type matching used on line 67.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,165 @@
...
+          case delegator: Delegator => delegator.delegate(dtab.dtab(), identificationPath)
+            .map(Some(_))
+          case _ => Future.None
+        }
+
+        dtreeF.map { dtree =>
+          val resp = Response()
+          val tree = RequestEvaluator.formatDTree(dtree, req.contentType, List.empty)
+          val evaluatedRequest = EvaluatedRequest(
+            Try(tree.last) match {
```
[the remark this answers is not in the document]
dadjeibaah: I think I had `Try...` because I couldn't figure out how to get the client name without using the `DelegateTree`. With your help, I was able to get the client name from the `DstBoundCtx`.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
dadjeibaah: You're right. I mistook the `Content-Type` header for the `Accept` header.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
dadjeibaah: I changed the recursive code a little bit to be clearer. I also removed the use of `Option`.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
dadjeibaah: Yep. I will make the change.

linkerd/core/src/test/scala/io/buoyant/linkerd/RequestEvaluatorTest.scala
dadjeibaah: Agreed this is much clearer.

adleong:
This is so cool!  Some quick notes on the updated output:

```
$ curl localhost:4140 -X TRACE -H "Host: foo" -H "l5d-max-depth: 2"
dog

--- Router: incoming ---
service name: /svc/foo
client name: /%/io.l5d.localhost/#/io.l5d.fs/foo
selected address: 127.0.0.1:8888
addresses: [127.0.0.1:8888]
dtab resolution:
  /svc/foo
  /
  /%/io.l5d.localhost/#/io.l5d.fs/foo

--- Router: outgoing ---
service name: /svc/foo
client name: /%/io.l5d.port/4141/#/io.l5d.fs/foo
selected address: 127.0.0.1:4141
addresses: [127.0.0.1:4141, 127.0.0.2:4141, 127.0.0.3:4141]
dtab resolution:
  /svc/foo
  /
  /%/io.l5d.port/4141/#/io.l5d.fs/foo⏎
```

* I would put addresses before selected address so that the order of service -> client -> addresses -> selected goes from more general to more specific
* What do you think of adding a `timestamp` field?  This gives some timing data about each hop
* Looks like the dtab resolution output is a bit broken with transformers
* Consider printing the dentry alongside each step in the delegation (like the dtab UI does)
* Missing newline at the end of the output

dadjeibaah:
Great suggestions @adleong.
> * I would put addresses before selected address so that the order of service -> client -> addresses -> selected goes from more general to more specific

that's a good point. I will do that.

> * What do you think of adding a `timestamp` field?  This gives some timing data about each hop

I think that could be useful. It would give users information about how long each `TRACE` request takes. At one point, I ran the tracer and it seemed to be taking longer than expected. I can imagine there could be serious latency that may be encountered while doing this.

> * Looks like the dtab resolution output is a bit broken with transformers
Darn! I will take a look to see what's going on.

> * Consider printing the dentry alongside each step in the delegation (like the dtab UI does) 

Consider it done. 

> * Missing newline at the end of the output

Right now, the code strips empty watch space at the start and end of the final output. Do we not want this behavior? Is a newline expected on plaintext responses from servers?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+import com.twitter.finagle.naming.buoyant.DstBindingFactory
+import com.twitter.util._
+import io.buoyant.namer.DelegateTree._
+import io.buoyant.namer.{DelegateTree, Delegator}
+import io.buoyant.router.RouterLabel
+import io.buoyant.router.RoutingFactory
+import io.buoyant.router.RoutingFactory.BaseDtab
+import io.buoyant.router.context.{DstBoundCtx, DstPathCtx}
+
+class RequestEvaluator(
```
adleong: We should think about what we want to call this.  I don't think Evaluator is quite right.  This is a kind of tracing but we should be careful not to conflate it with the type of tracing already in Linkerd (zipkin etc).  

Perhaps a name related to ActiveTracing (as opposed to the "passive" tracing that zipkin does).  Or maybe RoutingInfoFilter... hmmm...

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+  endpoint: EndpointAddr,
+  namer: DstBindingFactory.Namer,
+  dtab: BaseDtab,
+  label: String
```
adleong: is this the routerLabel?  let's call it that if so

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+) extends SimpleFilter[Request, Response] {
+
+  private[this] val RequestTracerMaxDepthHeader = "l5d-max-depth"
+
+  private[this] def printEvaluatedRequest(
```
adleong: I'd explicitly annotate the return type of String here to signal that this method isn't side-effecting.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+    prependText: String,
+    routerLabel: String,
```
adleong: I think we can omit this here and access the classes constructor parameter directly.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+         |client name: $clientName
+         |selected address: $selectedAddress
+         |addresses: [${addresses.getOrElse(Set.empty).mkString(", ")}]
+         |dtab resolution:
+         |${dtabResolution.map("  " + _).mkString("\n")}
+    """.stripMargin
+    ).mkString("\n").trim
+  }
+
+  private[this] def formatDTree(
```
adleong: perhaps formatDelegation?  since we're not printing the tree, just one path through it

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+    dTree match {
+      case Leaf(path, _, bound) if bound.id == searchPath =>
+        tree :+ path.show
+      case Transformation(_, _, value, remainingTree) =>
+        formatDTree(remainingTree, tree :+ value.path.show, searchPath)
+      case Delegate(path, _, remainingTree) =>
+        formatDTree(remainingTree, tree :+ path.show, searchPath)
+      case Union(path, _, remainingWeightedTrees@_*) =>
+        remainingWeightedTrees.map { wd =>
+          formatDTree(wd.tree, tree :+ s"${wd.weight} * ${path.show}", searchPath)
```
adleong: I think this is not quite right.  Technically the weight belongs to the subtree (wd.tree) instead of this node (path).  This causes delegations like 
```
  /svc/default
  1.0 * /x/default
  /b/default
  /$/inet/127.1/4140
```
when really it should be

```
  /svc/default
  /x/default
  1.0 * /b/default
  /$/inet/127.1/4140
```

Regardless, I'm not sure if it's even valuable to print the weight at all.  We can probably just omit it.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+        }.filter(!_.isEmpty).head
```
adleong: protip:

`head` is unsafe because it fails on empty lists.  prefer `headOption`:

`.filter(!_.isEmpty).headOption.toList.flatten`

filter+headOption can be combined into `find`:

`.find(!_.isEmpty).toList.flatten`

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+      case Alt(path, _, remainingTrees@_*) =>
+        remainingTrees.map { d =>
+          formatDTree(d, tree :+ path.show, searchPath)
+        }.filter(!_.isEmpty).head
+      case _ => List.empty
+    }
+  }
+
+  private[this] def getRouterCtx(prevResp: Response) = {
```
adleong: I'm a bit confused about the name.  What is RouterCtx?  The return type is Future[Response], right?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+
+    val dtreeF = namer.interpreter match {
+      case delegator: Delegator => delegator.delegate(dtab.dtab(), serviceName)
+        .map(Some(_))
+      case _ => Future.None
+    }
+
+    dtreeF.joinWith(addresses) {
+      case (dTree: Option[DelegateTree[Name.Bound]], addrSet: Option[Set[String]]) =>
+        val resp = Response()
```
adleong: consider just appending to prevResp.contentString instead of creating a new Response

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+        resp
+    }
+  }
+
+  override def apply(
+    req: Request,
+    svc: Service[Request, Response]
+  ): Future[Response] = {
+
+    val maxTraceDepth = req.headerMap.get(RequestTracerMaxDepthHeader).map(_.toInt)
```
adleong: Need some error handling here:

```
$ curl localhost:4140 -X TRACE -H "Host: foo" -H "l5d-max-depth: alex"
For input string: "alex"⏎
```

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
```diff
@@ -0,0 +1,168 @@
...
+    val httpMethod = req.method
```
adleong: do we need this val or can we just inline `req.method`?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
dadjeibaah: Makes sense. Will make that change.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
dadjeibaah: Good point.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestEvaluator.scala
dadjeibaah: I liked the name you mentioned earlier, RequestTracer. We could call it RequestActiveTracer (RAT haha) so we don't confuse it with passive tracing.

adleong:
This is looking really awesome
```
$ curl -X TRACE localhost:4140 -H "Host: default" -H "l5d-max-depth: 10"
dog

--- Router: http ---
service name: /svc/default
client name: /#/io.l5d.fs/default
addresses: [127.0.0.1:8888]
selected address: 127.0.0.1:8888
dtab resolution:
  /svc/default (/\x2f=>~)
  /b/default (/svc=>2.00*/a & /b)
  /bb/default (/b=>/bb)
  /#/io.l5d.fs/default (/bb=>/#/io.l5d.fs)
```

A few quick nits on the output:
* Timestamps would be awesome
* Looks like the first line in the delegation is printing the `Dentry.nop` which looks weird
* Probably don't need 2 blank lines between blocks

dadjeibaah:
I think my latest commit has a different output from what you have. The waits are no longer displayed as well as the `Dentry.nop`. I forgot to add the timestamps. Watch for my next commit.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+import com.twitter.finagle.naming.buoyant.DstBindingFactory
+import com.twitter.util._
+import io.buoyant.namer.DelegateTree._
+import io.buoyant.namer.{DelegateTree, Delegator}
+import io.buoyant.router.RouterLabel
+import io.buoyant.router.RoutingFactory
+import io.buoyant.router.RoutingFactory.BaseDtab
+import io.buoyant.router.context.{DstBoundCtx, DstPathCtx}
+
+class RequestActiveTracer(
```
adleong: This class could use some scaladoc here.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+        case DelegateTree.Delegate(_, d, _) if d == Dentry.nop => s"$path"
+        case _ => s"$path ($dentry)"
+      }
+
+    }
+  }
+
+  private[this] val RequestTracerMaxDepthHeader = "l5d-max-depth"
+
+  private[this] def printEvaluatedRequest(
```
adleong: I'd hesitate to call this `print` since it doesn't print.  And it seems like we're moving away from the `evaluate` terminology.

`formatRoutingContext` or something maybe?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+         |addresses: [${addresses.getOrElse(Set.empty).mkString(", ")}]
+         |selected address: $selectedAddress
+         |dtab resolution:
+         |${dtabResolution.map("  " + _).mkString("\n")}
+    """.stripMargin
+    ).mkString("\n").trim.concat("\n")
+  }
+
+  /**
+   * formatDelegation prints out path the proxy used to resolve the client name.
```
adleong: s/prints out/returns ?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+   * It 'walks' a DelegateTree of Name.Bound and finds a DelegateLeaf that matches
+   * the client path provided.
+   * @param dTree
```
adleong: fill out these scaladoc properties with descriptions or remove them entirely

adleong: but filling them out might be good because it's not super clear what they are.  `tree` is some kind of accumulator, I assume?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+   * @param tree
+   * @param searchPath
+   * @return
+   */
+  private[this] def formatDelegation(
+    dTree: DelegateTree[Name.Bound],
+    tree: List[Node],
+    searchPath: Path
+  ): List[Node] = {
+    dTree match {
```
adleong: this method is pretty dense.  breaking it up with comments or vals with intermediate values might help make it more readable.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+      case l@Leaf(leafPath, dentry, bound) if bound.id == searchPath =>
+        tree.lastOption.map {
```
adleong: this is to perform a transformation on the last node of `tree`?  personally, I'd find this easier to follow if it were something like

```
val lastNode = tree.lastOption.collect {
  case Node(...) => Node(...)
}
tree.dropRight(1) ++ lastNode
```

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+          case Node(nodePath, nodeDentry, Transformation(_, _, _, _)) =>
+            tree.dropRight(1) :+ Node(
+              nodePath,
+              dentry.show,
```
adleong: This dentry juggling we have to do here is pretty confusing and could use a comment

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+    dtreeF.joinWith(addresses) {
+      case (dTree: Option[DelegateTree[Name.Bound]], addrSet: Option[Set[String]]) =>
+        val tree = formatDelegation(dTree.getOrElse(DelegateTree.Empty(Path.empty, Dentry.nop)), List.empty, clientPath)
+        prevResp.contentString = printEvaluatedRequest(
+          prevResp.contentString,
+          serviceName.show,
+          clientPath.show,
+          selectedEndpoint,
+          addrSet,
+          tree.foldLeft(List[String]())((a, b) => a :+ b.toString)
```
adleong: is this just `tree.map(_.toString)`?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+        }.find(!_.isEmpty).toList.flatten
+      case a@Alt(path, dentry, remainingTrees@_*) =>
+        remainingTrees.map { d =>
+          formatDelegation(d, tree :+ Node(path.show, dentry.show, a), searchPath)
+        }.find(!_.isEmpty).toList.flatten
+      case _ => List.empty
+    }
+  }
+
+  private[this] def getRequestTraceResponse(prevResp: Response) = {
```
adleong: might just call this `response` instead of `prevResp` since we're mutating it and not creating a new one

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,198 @@
...
+    req: Request,
+    svc: Service[Request, Response]
+  ): Future[Response] = {
+
+    val maxTraceDepth = Try(req.headerMap.get(RequestTracerMaxDepthHeader).map(_.toInt))
+
+    (req.method, maxTraceDepth) match {
+      case (Method.Trace, Throw(_: NumberFormatException)) =>
+        val resp = Response()
+        resp.contentString = s"Invalid value for $RequestTracerMaxDepthHeader header"
```
adleong: should resp have a non 200 status code? 400?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
dadjeibaah: That's right. `tree` is an accumulator and eventually will represent the path of the `dTree` that was are searching. I will flesh out the scaladocs for this method to highlight that and give descriptions to the param.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
dadjeibaah: When we are building out the path, and we get to a leaf. We need to check to see if the parent of the leaf we are evaluating is a transformation. If it is a transformation, we switch the set the `Dentry` of the transformation to be the `Dentry` of the leaf and then, set the `Dentry` of the leaf to the `Dentry` of the transformation. I think your suggestion makes it clearer. Seperating the creation of the node from the modification of the tree path is a good call.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
dadjeibaah: Yea. It doesn't make sense for this to be 200.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+import com.twitter.util._
+import io.buoyant.namer.DelegateTree._
+import io.buoyant.namer.{DelegateTree, Delegator}
+import io.buoyant.router.RouterLabel
+import io.buoyant.router.RoutingFactory
+import io.buoyant.router.RoutingFactory.BaseDtab
+import io.buoyant.router.context.{DstBoundCtx, DstPathCtx}
+
+/**
+ * Intercepts Http TRACE requests with an l5d-max-depth header.
```
adleong: update comment

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+    override def toString: String = {
+      dTreeNode match {
+        case DelegateTree.Delegate(_, d, _) if d == Dentry.nop => s"$path"
+        case _ => s"$path ($dentry)"
+      }
+
+    }
+  }
+
+  private[this] val MaxForwardsHeader = "Max-Forwards"
```
adleong: can use `com.twitter.finagle.http.Fields.MaxForwards`

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+          case None => Future.value(Response(req))
+
+          // returns a response with router context
+          case _ =>
+            val stopwatch = Stopwatch.start()
+            getRequestTraceResponse(Response(req), stopwatch)
+        }
+
+      case (Method.Trace, Return(Some(num)), isAddRouterCtxt) if num > 0 =>
+        // Decrement Max-Fowards header
```
adleong: typo

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+
+    (req.method, maxTraceDepth, addRouterCtxt) match {
+      case (Method.Trace, Throw(_: NumberFormatException), _) =>
+        // Max-Forwards header is unparseable
+        val resp = Response(Status.BadRequest)
+        resp.contentString = s"Invalid value for $MaxForwardsHeader header"
+        Future.value(resp)
+      case (Method.Trace, Return(Some(0)), isAddRouterCtxt) =>
+        // Max-Forwards header is 0 and the l5d-add-context header is present
+        isAddRouterCtxt match {
```
adleong: can collapse the matches:

```
case (Method.Trace, Return(Some(0)), Some(isAddRouterCtx)) if StringUtil.toBoolean(isAddRouterCtx) =>
  ...
case (Method.Trace, Return(Some(0)), _) =>
```

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+          // returns a response with no router context
+          case None => Future.value(Response(req))
+
+          // returns a response with router context
+          case _ =>
```
adleong: We probably want to be more strict than allowing any value to be treated as `true`.  `com.twitter.finagle.util.StringUtil.toBoolean` seems like a good thing to use here.  It will map "1" and "t" and "true" to true and everything else to false.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+    val maxTraceDepth = Try(req.headerMap.get(MaxForwardsHeader).map(_.toInt))
+    val addRouterCtxt = req.headerMap.get(AddRouterContextHeader)
+
+    (req.method, maxTraceDepth, addRouterCtxt) match {
+      case (Method.Trace, Throw(_: NumberFormatException), _) =>
+        // Max-Forwards header is unparseable
+        val resp = Response(Status.BadRequest)
+        resp.contentString = s"Invalid value for $MaxForwardsHeader header"
+        Future.value(resp)
+      case (Method.Trace, Return(Some(0)), isAddRouterCtxt) =>
```
adleong: we usually abbreviate context to ctx.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+    }
+  }
+
+  override def apply(
+    req: Request,
+    svc: Service[Request, Response]
+  ): Future[Response] = {
+
+    val maxTraceDepth = Try(req.headerMap.get(MaxForwardsHeader).map(_.toInt))
+    val addRouterCtxt = req.headerMap.get(AddRouterContextHeader)
```
adleong: consider 
```
req.headerMap.get(...) match {
  case Some(s) => StringUtil.toBoolean(s)
  case None => false
}
```

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+
+        isAddRouterCtxt match {
+          // Forward requests without adding router context to response
+          case None => svc(req).ensure {
+            req.headerMap.set(MaxForwardsHeader, num.toString); ()
+          }
+          case _ =>
+            // Forward requests and router context to response
+            val stopwatch = Stopwatch.start()
+            svc(req).flatMap((prevResp: Response) => getRequestTraceResponse(prevResp, stopwatch)).ensure {
```
adleong: can this just be `svc(req).flatMap(getRequestTraceResponse(_, stopwatch)).ensure {`

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+              req.headerMap.set(MaxForwardsHeader, num.toString); ()
+            }
+        }
+      case (_, _, _) => svc(req)
```
adleong: I think if we get a TRACE with l5d-add-context but no Max-Forwards, we should still add the context.  I think we basically treat this as Max-Forwards = infinity.

dadjeibaah: Yea I thought about that as well. Could we run into a scenario were we have Linkerd set up in a linker to linker fashion and when a request with no `Max-Forwards` could eventually make the request hang in a linker to linker loop? I guess the timeout exception might make this a non-issue.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
dadjeibaah: Darn, I thought I renamed that variable. using the underscore is even better.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
adleong: Yes, although this is already an issue with Linkerd: https://github.com/linkerd/linkerd/issues/1411

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
dadjeibaah: `StringUtil.toBoolean` returns false by default if the string being converted is not "1", 
"t" or "true". I am thinking maybe `req.headerMap.get(AddRouterContextHeader).map(StringUtil.toBoolean)` should suffice?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,263 @@
...
+import io.buoyant.router.RoutingFactory.BaseDtab
+import io.buoyant.router.context.{DstBoundCtx, DstPathCtx}
+
+/**
+ * Intercepts Http TRACE requests with a Max-Forwards and l5d-add-context header.
+ * Returns identification, delegation, and service address information
+ * for a given router as well as any other TRACE responses received from downstream
+ * services.
+ *
+ * A router that receives a tracer request with an l5d-max-depth
```
adleong: update comment: no such thing as l5d-max-depth anymore

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,263 @@
...
+ * @param routerLabel name of router
+ */
+class RequestActiveTracer(
+  endpoint: EndpointAddr,
+  namers: DstBindingFactory.Namer,
+  dtab: BaseDtab,
+  routerLabel: String
+) extends SimpleFilter[Request, Response] {
+
+  private[this] case class PathNode(
```
adleong: this represents one step in the delegation, rigth?  should this be called DelegationNode or something similar?

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,263 @@
...
+      |addresses: [${addresses.getOrElse(Set.empty).mkString(", ")}]
+      |selected address: $selectedAddress
+      |dtab resolution:
+      |${dtabResolution.map("  " + _).mkString("\n")}
+    """.stripMargin
+    ).mkString("\n").trim.concat("\n")
+  }
+
+  /**
+   * Returns a list of [[PathNode]] that represents a delegate tree path that identifies how a
```
adleong: Scaladoc links using `[[` and `]]` have a tendency to break artifact publishing.  I don't know why, but I've had to remove them in the past.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,263 @@
...
+        }.find(!_.isEmpty).toList.flatten
+      case a@Alt(path, dentry, remainingTrees@_*) =>
+        remainingTrees.map { d =>
+          formatDelegation(d, nodes :+ PathNode(path.show, dentry.show, a), clientName)
+        }.find(!_.isEmpty).toList.flatten
+      case _ => List.empty
+    }
+  }
+
+  private def getRequestTraceResponse(resp: Response, stopwatch: Stopwatch.Elapsed) = {
```
adleong: try to be consistent about `private` vs `private[this]` unless there's a reason this method needs to be accessed from another instance.  I prefer using the smallest visibility scope necessary.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,263 @@
...
+    }
+  }
+
+  override def apply(
+    req: Request,
+    svc: Service[Request, Response]
+  ): Future[Response] = {
+
+    val maxForwards = Try(req.headerMap.get(MaxForwards).map(_.toInt))
+    val isAddRouterCtx = req.headerMap.get(AddRouterContextHeader).map(StringUtil.toBoolean).getOrElse(false)
```
adleong: TIOLI I used to use this `opt.map(f).getOrElse(default)` pattern a lot but more recently I'm starting to favor

```
opt match {
  case Some(x) => f(x)
  case None => default
}
```

They're identical in effect but I find the second more readable.  Either way is fine, just wanted to point out the alternative.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,263 @@
...
+
+    (req.method, maxForwards, isAddRouterCtx) match {
+      case (Method.Trace, Throw(_: NumberFormatException), _) =>
+        // Max-Forwards header is unparseable
+        val resp = Response(Status.BadRequest)
+        resp.contentString = s"Invalid value for $MaxForwards header"
+        Future.value(resp)
+      case (Method.Trace, Return(Some(0)), addRouterCtx) if addRouterCtx =>
```
adleong: could be `case (Method.Trace, ..., true) =>`

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,263 @@
...
+        // Max-Forwards header is 0 and the l5d-add-context header is present
+        // returns a response with router context
+        val stopwatch = Stopwatch.start()
+        getRequestTraceResponse(Response(req), stopwatch)
+      case (Method.Trace, Return(Some(0)), _) =>
```
adleong: could be `case (Method.Trace, ..., false) =>`

adleong: ditto for remaining cases

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
...
+import com.twitter.util._
+import io.buoyant.namer.DelegateTree._
+import io.buoyant.namer.{DelegateTree, Delegator}
+import io.buoyant.router.RouterLabel
+import io.buoyant.router.RoutingFactory
+import io.buoyant.router.RoutingFactory.BaseDtab
+import io.buoyant.router.context.{DstBoundCtx, DstPathCtx}
+
+/**
+ * Intercepts Http TRACE requests with a Max-Forwards and l5d-add-context header.
```
olix0r: I would expect the header be named `l5d-max-forwards` to be consistent with other linkerd-specific headers. Is the `max-forwards` header standardized more widely?

adleong: Yes, from https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

>The final recipient is either the origin server or the first proxy or gateway to receive a Max-Forwards value of zero (0) in the request (see section 14.31).

dadjeibaah: Yea, `Max-Forwards is the default header associated with the `TRACE` method according to the spec. The header is what proxies use to figure out if they should forward a trace request or just reflect it back to the client.

adleong: > The Max-Forwards request-header field provides a mechanism with the TRACE (section 9.8) and OPTIONS (section 9.2) methods to limit the number of proxies or gateways that can forward the request to the next inbound server. This can be useful when the client is attempting to trace a request chain which appears to be failing or looping in mid-chain.

olix0r: Excellent! thank you. it might be helpful to include that RFC link at the bottom of the docstring.

olix0r:
@dadjeibaah Would you mind sharing the output of `curl -v` for the example configuration you provide in the description?

dadjeibaah:
Sure, I can do that.

linkerd/core/src/main/scala/io/buoyant/linkerd/RequestActiveTracer.scala
```diff
@@ -0,0 +1,265 @@
+package io.buoyant.linkerd
+
+import com.twitter.finagle._
+import com.twitter.finagle.client.Transporter.EndpointAddr
+import com.twitter.finagle.http._
```
olix0r: I would expect an HTTP-specifc module like this to live in an HTTP-specific subproject (and not core). Can this be moved? If not, why not? How will we offer similar support for H2?

olix0r:
Last (I think) comment:

```
< HTTP/1.1 200 OK
< Via: 1.1 linkerd, 1.1 linkerd
< l5d-success-class: 1.0
< Content-Length: 738
<
```

The response _should_ include a `content-type` header. Technically, according to the [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18), the response _must_ include a `Date` header:

Are there any other HTTP-correctness things we should do?

adleong:
@olix0r The response here may be generated by either the final server (if max-forwards is sufficiently large) or by Linkerd (if max-forwards reaches 0).  In the case where Linkerd generates the response, it would make sense for us to set `Content-Type: text/plain`.

In the case where the final server generates the response, we should probably just leave the response headers alone.  The wrinkle being that by appending router context to the response body, we may be violating the content-type that the server sent.  I'm not sure what to do about that.

dadjeibaah:
According to the RFC the `Content-Type` needs to be `message/http`. I am not really sure what that means really.

dadjeibaah:
@adleong, we could set the `Content-Type` if we check and see if it isn't present, otherwise, we can send back whatever the final server sent?

dadjeibaah:
@olix0r, according to the [RFC](https://tools.ietf.org/html/rfc7231#section-4.3.8) we would need to send  

> a 200 (OK) response with a Content-Type of "message/http"

A "message/http" `content-type` would need to be encoded in either `7bit`, `8bit` or `binary`. I think for now we could release with this feature and then down the line, we could encode a `TRACE` request and set the `Content-Type` to make sure we are matching the spec. I will create a ticket so we can track that.

PR #1940, opened 2018-05-10
Fix flakey retry test

### LINKED ISSUE #1925: ClassifiedRetryFilterTest is flakey
This test sometimes fails with:

```
[info] ClassifiedRetryFilterTest:
[info] - retries *** FAILED ***
[info]   Some("1") did not equal Some("3") (ClassifiedRetryFilterTest.scala:94)
```

adleong:
ClassifiedRetryTest sometimes fails.  I suspect this is because if the test runs slowly enough then the classification timeout is exceeded and requests that are supposed to be retried will not be.

Use `Time.withCurrentTimeFrozen` so that even if the test runs slowly, the classification timeout will never be exceeded.

I cannot reproduce the test failure anymore, but this is my best guess as to why it fails sometimes.

Fixes #1925 


adleong:
@dadjeibaah maybe!  If you notice a flakey test, please file an issue.

PR #1942, opened 2018-05-11
enable TLS protocols configuration

dvulpe:
Hi,

I started using Linkerd and noticed the server TLS configuration provides an option for configuring the list of allowed ciphers. Unfortunately if one configures the Linkerd router with a strong cipher which is TLS 1.2 specific (like ECDHE-RSA-AES128-GCM-SHA256) the Linkerd client would fail to connect to it as it doesn't attempt to use TLSv1.2.

This patch adds an integration test which demonstrates that configuring a finagle server end with a TLSv1.2 specific cipher  and the client with the TLSv1.2 protocol the connection is successful as well as enables the protocol configuration option for both server and client ends.

Dan

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
@@ -1,12 +1,14 @@
 package com.twitter.finagle.buoyant
 
+import java.io._
+
 import com.twitter.finagle.Stack
 import com.twitter.finagle.netty4.ssl.client.Netty4ClientEngineFactory
-import com.twitter.finagle.ssl.{KeyCredentials, TrustCredentials}
 import com.twitter.finagle.ssl.client.{SslClientConfiguration, SslClientEngineFactory}
+import com.twitter.finagle.ssl.{KeyCredentials, Protocols, TrustCredentials}
 import com.twitter.finagle.transport.Transport
 import com.twitter.io.StreamIO
-import java.io._
+
 import scala.util.control.NoStackTrace
 
 case class TlsClientConfig(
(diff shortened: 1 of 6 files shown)


dadjeibaah:
Thanks for submitting this @dvulpe! We will take a look at your PR.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
```diff
@@ -1,33 +1,38 @@
...
       Stack.Params.empty + Transport.ClientSsl(None)
-    case TlsClientConfig(_, Some(true), _, _, clientAuth) =>
+    case TlsClientConfig(_, Some(true), _, _, clientAuth, enabledProtocols) =>
       val tlsConfig = SslClientConfiguration(
         trustCredentials = TrustCredentials.Insecure,
-        keyCredentials = keyCredentials(clientAuth)
+        keyCredentials = keyCredentials(clientAuth),
+        protocols = enabledProtocols.map(Protocols.Enabled).getOrElse(Protocols.Unspecified)
       )
+      val factory = Netty4ClientEngineFactory()
```
adleong: I think this is unused

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpTlsEndToEndTest.scala
```diff
@@ -69,6 +70,44 @@ class HttpTlsEndToEndTest extends FunSuite with BudgetedRetries {
 
     assert(rsp.status == Status.Ok)
   }
+
+  test("client/server works with TLSv1.2 only cipher", Retryable) {
```
adleong: Would you mind updating this e2e test such that the request passes through a Linkerd router with the appropriate protocols configured?

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpTlsEndToEndTest.scala
```diff
@@ -4,11 +4,12 @@ import java.io.{File, InputStream}
 import java.net.InetSocketAddress
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
 import java.nio.file.{Files, Paths}
+
```
dadjeibaah: please remove spaces and alphasort imports

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpTlsEndToEndTest.scala
dvulpe: Thanks for the feedback.

I think this test boundary is around the Finagle integration rather than L5D. 
In my opinion this test should only fail when something in Finagle (newer version?) would change. 

If you think this kind of test would add value, we could look at introducing it in another part of the project, I tend to add as fewer as possible as they can fail for multiple reasons which makes debugging really hard and slow down the feedback loop.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
dvulpe: Fixing

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpTlsEndToEndTest.scala
dvulpe: Fixing

adleong:
Hi @dvulpe!  The CI failures here are due to a memory issue in the Linkerd tests that has just been fixed on master.  If you merge latest master the issue should be resolved.  Sorry about that!

dvulpe:
Hi, 

Is there anything else needed for this PR or is it ok to be merged into master?

Thanks,
Dan

Issue #1922, opened 2018-04-28
Enable plugins to emit stat into L5D stat trees for Announcer


leozc:
Issue Type:

- [ ] Bug report
- [ X] Feature request

We are writing multiple plugins for our L5D mesh, there are needs to initialize a finagle client to perform RPC, however there is no way to submit the stat into L5D metrics system.

Suggestion: Is it possible to enable this user scenario?

``` 
def getClient(withRequestTimeoutMS: Int =  1000 , addr: InetSocketAddress, path:Path): Service[http.Request, http.Response]  = {
    val monitor: Monitor = new Monitor {
      def handle(t: Throwable): Boolean = {
        logger.error(t, t.getMessage, t.getStackTrace)
        true
      }
    }
    val metrics = MetricsTree() // possible to get the metrics tree from global 
    val stats = new MetricsTreeStatsReceiver(metrics) // register this stat object to be collected?
    Http.client
      .withStatsReceiver(stats)
      .withHttpStats
      .withRequestTimeout(Duration.fromMilliseconds(withRequestTimeoutMS))
      .withMonitor(monitor)
      .newService(s"${addr.getHostName()}:${addr.getPort()}", "hc_client")

  }
```

edio:
If the intention is to see stats under `admin/metrics.json`, then it is already possible with linkerd.

Usually (I'd say, always, but maybe some type of plugins miss that still) plugin factory method receives `Stack.Params` as an argument, and one can obtain `StatsReceiver` from it:

```
val stats = params[param.Stats].statsReceiver
```

adleong:
@leozc does @edio's comment give you what you need?

leozc:
I am working on plugins base on`AnnouncerInitializer` and 'IdentifierInitializer'  - I don't see the factory method sreceive Stack.Params.

adleong:
@leozc should we close this now that params are being passed to Announcer plugins?  Or are there other types of plugins that need this change as well?

leozc:
We need to work on the identifier type as well - should I open another bug
or reuse this one ?

I am happy to work on the identifier plugin improvement as well.

adleong:
@leozc whichever is easier for you is fine with me.  Looking forward to your PR.

leozc:
Working perfectly - after picking up 1.4.1 and rebuild - now my announcer can emit the stat to the router it attaches to.
e.g.
  "rt/svc_ingress/announcer_hc_cln/http/response_size.p9990" : 166,
[...]

PR #1944, opened 2018-05-13
inject stack.param into announcer init flow

leozc:
This will allow announcer to access to the run-time parameters -including stat tree.

This is a POC related to https://github.com/linkerd/linkerd/issues/1922

### DIFF
--- a/linkerd/announcer/serversets/src/main/scala/io/buoyant/linkerd/announcer/serversets/ServersetsAnnouncerInitializer.scala
+++ b/linkerd/announcer/serversets/src/main/scala/io/buoyant/linkerd/announcer/serversets/ServersetsAnnouncerInitializer.scala
@@ -1,7 +1,7 @@
 package io.buoyant.linkerd.announcer.serversets
 
 import com.fasterxml.jackson.annotation.JsonIgnore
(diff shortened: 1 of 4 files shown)


leozc:
```
  @JsonIgnore
  def router(params: Stack.Params): Router = {
    val prms = params ++ routerParams
    val param.Label(label) = prms[param.Label]
    val announcers = _announcers.toSeq.flatten.map { announcer =>
      announcer.prefix -> announcer.mk(params)
    }
    protocol.router.configured(prms)
      .serving(servers.map(_.mk(protocol, label)))
      .withAnnouncers(announcers)
  }

```

Will do another push when I get back

leozc:
Added missing file - Route.scala.

leozc:
Done

JustinVenus:
@adleong the announcer feature is pretty useful for the group I work with.  We run our linkerd's on the host OS and announce via serversets so we can discover them for prometheus metrics.  We also been kicking around the idea of using the announcer interface (with a plugin) to register linkerd's with Amazon application load balancers.

```quote
On a somewhat unrelated note, I've been wondering if people actually find announcers useful and I'd love to chat with you about them if you use them.
```

PR #1952, opened 2018-05-17
Fixes #1552 - Rework TLS client configuration trust store to use a CA…

### LINKED ISSUE #1552: Remove dependence on temp files for TLS certificates
We experienced an issue where several of our linkerd instances started failing with this error message: 

```
Aug 02 14:53:29 store-app1-i.sjc1sl.bigcommerce.net linkerd[25890]: Caused by: java.security.cert.CertificateException: could not find certificate file: /tmp/certCollection3425953256568330028.tmp
```

It turns out that it was trying to read a file out of the temp directory that had been cleaned out by a cron job that deletes files in the server's temp directory that hadn't been modified after a week. 

If these files are required for an extended period of time, linkerd shouldn't store them in the system's temp directory. 

The dependence on these temp files appears to be introduced in this PR for mutual TLS: https://github.com/linkerd/linkerd/pull/1319/files#diff-c23f3271d49c726272f66f9c10e24b3aR29

dvulpe:
… bundle instead of dropping all ca certs to a bundle in the temp folder. This simplifies the client TLS code, removes a dependency on the temporary folder and aligns client TLS configuration with server TLS configuration

Breaking change: this modifies trustCerts from an optional list of strings to an optional string

Signed-off-by: Dan Vulpe <dvulpe@gmail.com>

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
@@ -1,13 +1,11 @@
 package com.twitter.finagle.buoyant
 
 import java.io._
-
 import com.twitter.finagle.Stack
 import com.twitter.finagle.netty4.ssl.client.Netty4ClientEngineFactory
-import com.twitter.finagle.ssl.client.{SslClientConfiguration, SslClientEngineFactory}
 import com.twitter.finagle.ssl.{KeyCredentials, Protocols, TrustCredentials}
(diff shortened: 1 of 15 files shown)


dadjeibaah:
Interested in getting thoughts about using `@deprecated` annotations in Linkerd. I know finagle uses them and have encountered them while performing upgrades. I think it is a nice reminder to have for users. WDYT?

adleong:
I think deprecated annotations are intended for library fields to signal to library users that those library fields are deprecated.  Given that Linkerd isn't a library, I'm not sure that it makes sense to use the scala deprecated annotation.

We should, however, warn in the logs when users use a config property that is deprecated.

PR #1953, opened 2018-05-17
Ensure H2 client writes l5d-ctx headers

### LINKED ISSUE #1945: gRPC server not receiving l5d-ctx-dtab header from Linkerd
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:

Linkerd seems to be dropping the `l5d-dtab` header on gRPC requests.

For example, external request to Service A, and Service A calls Service B.  The external request contains a `l5d-dtab` header selecting specific versions of Service A and Service B.

We see the request routed to Service A per `l5d-dtab` but the call to Service B routes to the default version of Service B.

Found in the Service A logs that the server did not receive a `l5d-ctx-dtab` header from Linkerd and consequently could not pass along the header in the call to Service B.

Expected the destination server to receive the original client's dtab in a `l5d-ctx-dtab` header.

1. Namerd dtab:

```
/srv=>/#/io.l5d.k8s/m6a/grpc;
/svc=>/$/io.buoyant.http.domainToPathPfx/grpc;
/grpc/Search=>/srv/search-v1-0-6|/srv/search-v1-0-6;
/grpc/ChannelList=>/srv/channellist-v1-0-3|/srv/channellist-v1-0-3;
```

2. `ld5-dtab` header set by grpc-java client set on a search request:

```
metadata.put(Metadata.Key.of("l5d-dtab", Metadata.ASCII_STRING_MARSHALLER), "/grpc/Search => /srv/search-edge; /grpc/ChannelList => /srv/channellist-edge");
```

3.  Server logs indicate request routed to search-edge and channellist-v1-0.3 pods.  Expected search-edge and channellist-edge.

4.  Request headers logged by search-edge:

```
headers=Metadata(grpc-accept-encoding=gzip, devicetimezone=America/New_York, l5d-dst-service=/svc/search.Search, l5d-reqid=ed03c39fd4c1891a, l5d-dst-client=/#/io.l5d.k8s/m6a/grpc/search-edge, via=h2 linkerd, grpc-trace-bin=, l5d-ctx-trace=XI9zgVoKC6LtA8Of1MGJGu0Dw5/UwYkaAAAAAAAAAAA=, content-type=application/grpc, user-agent=grpc-java-netty/1.8.0, l5d-dst-residual=/search)
```

**Environment**:

- linkerd/namerd version, config files: See config below:  linkerd 1.3.7, namerd 1.3.7
- Platform, version, and config files (Kubernetes, DC/OS, etc):  Kubernetes 1.7
- Cloud provider or hardware configuration:  N/A

Relevant configs:

```
---
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  labels:
    app: l5d
  name: l5d
spec:
  template:
    metadata:
      labels:
        app: l5d
    spec:
      volumes:
      - name: l5d-config
        configMap:
          name: "l5d-config"
      containers:
      - name: l5d
        resources:
          limits:
            cpu: 500m
            memory: 1G
        image: buoyantio/linkerd:1.3.7
        env:
        - name: POD_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
        args:
        - /io.buoyant/linkerd/config/config.yaml
        ports:
        - name: incoming
          containerPort: 4140
        - name: admin
          containerPort: 9990
        volumeMounts:
        - name: "l5d-config"
          mountPath: "/io.buoyant/linkerd/config"
          readOnly: true
      - name: kubectl
        image: buoyantio/kubectl:v1.4.0
        args:
        - "proxy"
        - "-p"
        - "8001"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: l5d-config
data:
  config.yaml: |-
    admin:
      ip: 0.0.0.0
      port: 9990
    usage:
      orgId: linkerd-m6a
      enabled: false
    routers:
    - label: grpc-incoming
      protocol: h2
      experimental: true
      identifier:
        kind: io.l5d.header.path
        segments: 1
      interpreter:
        kind: io.l5d.namerd
        dst: /$/inet/namerd.m6a-service-mesh.svc.cluster.local/4100
        namespace: internal
      servers:
      - port: 4140
        ip: 0.0.0.0
      service:
        responseClassifier:
          kind: io.l5d.h2.grpc.default
```

adleong:
HTTP/2 requests sent by Linkerd are missing the `l5d-ctx-*` headers such as `l5d-ctx-dtab` and `l5d-ctx-trace`.

When the H2 client stack is configured, the PrepConn module is replaced with the DelayedRelease module.  However, since the DelayedReleaseModule has the PrepFactory role, this resulted in the client stack no longer having a PrepConn role which caused the dtab client module to not be added.  We fix this by simply changing the role of the DelayedReleaseModule to PrepConn so that the role does not change when the PrepConn module is replaced.

Fixes #1945 

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/DelayedReleaseService.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/DelayedReleaseService.scala
@@ -7,7 +7,7 @@ import com.twitter.util.{Future, Promise, Return, Time}
 import java.util.concurrent.atomic.AtomicBoolean
 
 object DelayedRelease {
-  val role = StackClient.Role.prepFactory
+  val role = StackClient.Role.prepConn
   val description = "Prevents an H2 service from being closed until its response stream completes"
   val module: Stackable[ServiceFactory[Request, Response]] =
     new Stack.Module0[ServiceFactory[Request, Response]] {
--- a/linkerd/protocol/h2/src/e2e/scala/io/buoyant/linkerd/protocol/h2/H2EndToEndTest.scala
+++ b/linkerd/protocol/h2/src/e2e/scala/io/buoyant/linkerd/protocol/h2/H2EndToEndTest.scala
@@ -7,7 +7,7 @@ import com.twitter.finagle.param.Stats
 import com.twitter.finagle.Path
 import com.twitter.finagle.service.ExpiringService
 import com.twitter.finagle.stats.InMemoryStatsReceiver
-import com.twitter.util.Duration
+import com.twitter.util.{Duration, Future}
 import io.buoyant.linkerd.{Linker, Router}
 import io.buoyant.linkerd.protocol.H2Initializer
 import io.buoyant.router.StackRouter.Client.PerClientParams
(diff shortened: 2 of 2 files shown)


dadjeibaah:
Interesting, do we ever use a `PrepFactory` module in the H2 client stack? What are the uses of a `PrepFactory` module?

adleong:
Actually it looks like put the DelayedReleaseModule in 2 different places: replacing the prepFactory role in the path stack and replacing the prepConn role in the client stack.  I actually... have no idea why we put it in both.  Here's the PR where it got added: https://github.com/linkerd/linkerd/pull/1444/files

I can't really think of a reason why it needs to be in both stacks but I'm nervous about changing something like that without a lot of testing.

PR #1954, opened 2018-05-18
Avoid adding node_modules to admin.jar

robertpanzer:
It looks like somewhere between 1.3.6 and 1.3.7 the node_modules folder was accidentally added to the admin_2.12.jar resulting in a much bigger artifact:
```
494574 /Users/rpanzer/.m2/repository/io/buoyant/admin_2.12/1.3.6/admin_2.12-1.3.6.jar
12980121 /Users/rpanzer/.m2/repository/io/buoyant/admin_2.12/1.3.7/admin_2.12-1.3.7.jar
```
While the settings for the full assembly seem to filter these out again, they are still somewhat disturbing when creating own assemblies/fat jars.

This PR changes the base settings for every module to also exclude any node_modules directories.

Disclaimer: My knowledge of sbt is barely existing, if at all.

### DIFF
--- a/project/Base.scala
+++ b/project/Base.scala
@@ -106,7 +106,15 @@ class Base extends Build {
         Some("snapshots" at nexus + "content/repositories/snapshots")
       else
         Some("releases"  at nexus + "service/local/staging/deploy/maven2")
-    }
+    },
+    mappings in (Compile, packageBin) ~= { (ms: Seq[(File,String)]) => ms filter {
+      case (file,toPath) => {
+        file.getPath match {
+          case nodeModulesRE() => false
+          case _ => true
+        }
+      }
+    }}
   )
 
   val aggregateSettings = Seq(


project/Base.scala
```diff
@@ -106,7 +106,15 @@ class Base extends Build {
         Some("snapshots" at nexus + "content/repositories/snapshots")
       else
         Some("releases"  at nexus + "service/local/staging/deploy/maven2")
-    }
+    },
+    mappings in (Compile, packageBin) ~= { (ms: Seq[(File,String)]) => ms filter {
+      case (file,toPath) => {
+        file.getPath match {
```
robertpanzer: Used a pattern match instead of a simple direct boolean expression because it might be necessary to also filter out other files across all modules.

dadjeibaah:
Hey @robertpanzer thanks for submitting this PR! I was trying to see if I could reproduce this locally prior to this commit but I don't seem to be running into the issue. I may be doing something wrong 😬. I pulled down latest master and ran `./sbt admin/publishLocal` to publish the `admin_2.12.jar` to my local ivy cache. I then ran `jar tf admin_2.12.jar | grep -i node_modules` to see if I could find any node module directories in the jar but didn't find any. Do you see the same results?

adleong:
I do see the node_modules when building locally:

```
$ tar tf .ivy2/local/io.buoyant/admin_2.12/1.4.0-SNAPSHOT/jars/admin_2.12.jar | grep -i node_modules | wc -l
    9796
```

robertpanzer:
Awesome!
To check I also downloaded the artifacts of  1.3.6 and 1.3.7 of admin_2.12 and their size differs significantly. (And 1.4.0 has the same size as 1.3.7)

adleong:
It looks like the the node_modules directory will get included in the jar if it happens to be present in the source tree when the jar is built.  So the fact that it started showing up in 1.3.7 is just due to whoever built that release and what was on their computer at the time.  This change seems like a great way to guard against that.

robertpanzer:
Sure! 
That means I have to change the commit message to add the „Signed off by...“ line?

robertpanzer:
Just updated the commit message.

adleong:
Thanks!

PR #1956, opened 2018-05-19
 Instrument watches in the io.l5d.mesh interpreter

adleong:
It can be difficult to know which watches the mesh interpreter has cached, which of them are active, and if they are receiving updates from the Namerd. This has made issues with the Namerd extremely difficult to diagnose and debug.

We instrument the Activities in the mesh interpreter to keep track of metadata about themselves. This metadata is exposed though an admin endpoint `/interpreter_state/io.l5d.mesh/default.json`.

screen shot 2018-05-18 at 5 34 34 pm


interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -58,16 +82,41 @@ object Client {
...
     }
 
     /**
      * When observed, streams dtabs from the mesh server and
      */
     override lazy val dtab: Activity[Dtab] = {
-      val open = () => delegator.streamDtab(mkDtabReq(root))
-      streamActivity(open, decodeDtab, backoffs, timer)
+      dtabCacheMu.synchronized {
+        if (dtabCache != null) {
```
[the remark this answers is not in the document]
adleong: bind and resolve use pattern matching to destructure the Option from map.get.  In this case we're just doing a null test.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -49,7 +51,29 @@ object Client {
...
+      act: InstrumentedVar[Addr],
+      stream: StreamState[mesh.ReplicasReq, mesh.Replicas]
+    )
+    private case class InstrumentedDtab(
+      act: InstrumentedActivity[Dtab],
+      stream: StreamState[mesh.DtabReq, mesh.DtabRsp]
+    )
+
+    private[this] val bindCacheMu = new {}
+    @volatile private[this] var bindCache = Map.empty[(Dtab, Path), InstrumentedBind]
```
olix0r: It would be good, as a followup, to investigate which caching techniques are friendliest to the jvm (i.e. Map, mutable.Map, CHM, etc)

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -105,6 +167,63 @@ object Client {
...
+            }
+        }
+        val resolveState = resolves.map {
+          case (path, InstrumentedResolve(act, stream)) =>
+            path.show -> Map(
+              "state" -> act.stateSnapshot(),
+              "watch" -> stream
+            )
+        }
+        val dtabState = Option(dtab).map { dtab =>
```
olix0r: Is it possible for `dtab` to be `null`? if not, it would be clearer as `Some(...)`

adleong: it is possible for dtab to be null.

PR #1960, opened 2018-05-22
Add clientCaChainPath client auth TLS config setting

### LINKED ISSUE #1957: Linkerd >= 1.4.0 no longer validates intermediate certs presented by servers during mTLS
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

With linkerd >= 1.4.0, requests to servers that have intermediates in their cert file now fail. It looks like the client-side no longer follows the presented cert chain when it attempts to validate the server's certs against the client's configured `trustCerts`. This causes the client to not send along its certs, leading to an internal `SSLV3_ALERT_HANDSHAKE_FAILURE` exception.

The client-side linkerd should walk the cert chain and test each issuer against those in `trustCerts`, not just the leaf cert.

https://github.com/joeyb/linkerd-cert-issue

**Environment**:
linkerd >= 1.4.0

adleong:
There is no way to specify a supporting CA certificate chain to support a client certificate when Linkerd is using client auth.  If a certificate chain is provided in the file specified by `certPath`, only the first certificate in the file is served.

Add a new client auth TLS config option `clientCaChainPath`.  This allows you to provide a supporting CA certificate chain for the client cert.

Fixes #1957 

### DIFF
--- a/linkerd/docs/client_tls.md
+++ b/linkerd/docs/client_tls.md
@@ -66,10 +66,11 @@ protocols         | unspecified                                | The list of TLS
 
 If present, a client auth object must contain two properties:
 
-Key      | Default Value | Description
----------|---------------|-------------
-certPath | _required_    | File path to the TLS certificate file.
-keyPath  | _required_    | File path to the TLS key file.  Must be in PKCS#8 format.
+Key               | Default Value | Description
+------------------|---------------|-------------
+certPath          | _required_    | File path to the TLS certificate file.
+clientCaChainPath | none          | Path to a file containing a CA certificate chain to support the client certificate.
+keyPath           | _required_    | File path to the TLS key file.  Must be in PKCS#8 format.
 
 <aside class="warning">
 Setting `disableValidation: true` will force the use of the JDK SSL provider which does not support client auth. Therefore, `disableValidation: true` and `clientAuth` are incompatible.
--- a/linkerd/examples/mutualTls.yaml
+++ b/linkerd/examples/mutualTls.yaml
@@ -7,6 +7,15 @@ routers:
 - protocol: http
   dtab: |
     /svc => /#/io.l5d.fs
+  client:
+    tls:
+      commonName: foo
+      trustCerts:
+      - /foo/cacert.pem
+      clientAuth:
+        certPath: /foo/clientCert.pem
+        keyPath: /foo/clientKey.pem
+        clientCaChainPath: /foo/intermediateCa.pem
   servers:
   - port: 4140
     # Expect the incoming connection to be encryped.
@@ -16,4 +25,4 @@ routers:
       # clients must provide a valid certificate
       requireClientAuth: true 
       # the authority used to validate client certificates
-      caCertPath: /foo/cacert.pem 
+      caCertPath: /foo/cacert.pem
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
@@ -75,9 +75,16 @@ case class TlsClientConfig(
 
   private[this] def keyCredentials(clientAuth: Option[ClientAuth]): KeyCredentials =
     clientAuth match {
-      case Some(ClientAuth(cert, key)) => KeyCredentials.CertAndKey(new File(cert), new File(key))
+      case Some(ClientAuth(cert, None, key)) =>
+        KeyCredentials.CertAndKey(new File(cert), new File(key))
+      case Some(ClientAuth(cert, Some(clientCaChain), key)) =>
+        KeyCredentials.CertKeyAndChain(new File(cert), new File(key), new File(clientCaChain))
       case None => KeyCredentials.Unspecified
     }
 }
 
-case class ClientAuth(certPath: String, keyPath: String)
+case class ClientAuth(
+  certPath: String,
+  clientCaChainPath: Option[String],
+  keyPath: String
+)


briansmith:
This is a very reasonable approach. In the past other products (e.g. Apache) have found that having separate files for the end-entity certificate and the certificate chain have pros and cons. The pro is that you can easily replace the end-entity certificate from the same CA, as long as the certificate chain never changes, just by replacing one file. The con is that "as long as the certificate chain never changes" is a very error-prone condition. Even if you request a new certificate from the same CA, they might use a different intermediate certificate.

Another con is that people may just ignore the clientCaChainPath and often things will "just work" because of AIA fetching. Hoewver, the AIA fetching causes the client to have to do extra networking to fetch the intermediate certificates.

IMO, a less error prone design would be to change the processing of `certPath` to allow the intermediates to be extracted from it, if there is more than one certificate in the file.

The other error that is common is for the user to include the *entire* certificate chain in the file, including the root, and then the web server serves the root certificate in the TLS handshake too. It is worth documenting, at least, that the cert chain should omit the root certificate, for efficiency reasons.

Alternatively, you could document that the cert chain should contain at least one root certificate. Then you could, after loading the cert chain, verify that the certificate is valid and issue a warning (at least) if it isn't valid, and then strip any root (self-signed) certificates before sending the cert chain in the TLS handshake. This is basically what I'm planning to do in Conduit, but it is more work because it requires dealing with certificate verification APIs.

adleong:
Thanks, @briansmith, this is very helpful.  I do like the idea of extracting intermediates from the cert file but this is difficult to do with Finagle's existing API which expects a file for the cert and a file for the chain.  If we do pre-processing on the cert files and pass new files to Finagle, any changes to the original cert files would not get picked up by Finagle.  We would need to also watch the original files and update the files that we created on any change.  Doable, but potentially more complex and error prone.

briansmith:
OK, keep in mind that if you have two files and you watch both of them, then you must deal with race conditions where one file is updated and is out of sync with the other file. For example, the user might replace the end-entity certificate file and then some time might pass before they replace the intermediate certificate chain file. If the configuration is reloaded between those two times then the end-entity certificate won't match the intermediate and the configuration will be broken until the second file is reloaded. Again, not a show-stopper but something to be aware of.

linkerd/docs/client_tls.md
```diff
@@ -66,10 +66,11 @@ protocols         | unspecified                                | The list of TLS
...
 If present, a client auth object must contain two properties:
 
-Key      | Default Value | Description
----------|---------------|-------------
-certPath | _required_    | File path to the TLS certificate file.
-keyPath  | _required_    | File path to the TLS key file.  Must be in PKCS#8 format.
+Key               | Default Value | Description
+------------------|---------------|-------------
+certPath          | _required_    | File path to the TLS certificate file.
+clientCaChainPath | none          | Path to a file containing a CA certificate chain to support the client certificate.
```
briansmith: Why does just this setting have the word "client" in its name and in its description, but the other settings don't? It seems like "client" is redundant here and could/should be removed.

linkerd/examples/mutualTls.yaml
```diff
@@ -7,6 +7,15 @@ routers:
...
     /svc => /#/io.l5d.fs
+  client:
+    tls:
+      commonName: foo
+      trustCerts:
+      - /foo/cacert.pem
+      clientAuth:
+        certPath: /foo/clientCert.pem
+        keyPath: /foo/clientKey.pem
+        clientCaChainPath: /foo/intermediateCa.pem
```
briansmith: Same thing here. It seems like `client` in the name is inconsistent with the other parts of the `clientAuth` setting.

linkerd/examples/mutualTls.yaml
```diff
@@ -16,4 +25,4 @@ routers:
       # clients must provide a valid certificate
       requireClientAuth: true 
       # the authority used to validate client certificates
-      caCertPath: /foo/cacert.pem 
+      caCertPath: /foo/cacert.pem
```
briansmith: Should this have `client` in its name like `requireClientAuth` does?

linkerd/examples/mutualTls.yaml
```diff
@@ -16,4 +25,4 @@ routers:
       # clients must provide a valid certificate
```
briansmith: Shouldn't we have a way of setting the intermediate certificate chain above for the server cert too, like we do for the client cert? In other words, `(cert, intermediates, key)` should always go together.

briansmith:
> OK, keep in mind that if you have two files and you watch both of them, then you must deal with race conditions where one file is updated and is out of sync with the other file. For example, the user might replace the end-entity certificate file and then some time might pass before they replace the intermediate certificate chain file. If the configuration is reloaded between those two times then the end-entity certificate won't match the intermediate and the configuration will be broken until the second file is reloaded. Again, not a show-stopper but something to be aware of.

I guess we must already handle this, because we have the same issue with the private key file being updated out-of-sync with the end-entity certificate file.

linkerd/docs/client_tls.md
adleong: This is to mirror the `serverCaChainPath`.  Of course, that might be a poor name also.  My intention was to try and avoid confusion between the CA that the server uses to validate the client identity (`caCertPath` in the server tls struct) and the CA that issued the server's cert (`serverCaChainPath` in the server tls struct).  

Likewise I wanted to differentiate between the CA that the client uses to validate the server identity (`trustCerts` in the client tls struct) and the CA that issued the client's cert (`clientCaChainPath` in the client auth section of the client tls struct).

Any suggestions for better names for `clientCaChainPath` and `serverCaChainPath` would be appreciated.  The names of any other properties are more difficult to change because they have already been released and require a breaking config change to rename.

briansmith: I don't have strong feelings about this. Since both already have "keyPath" and "certPath" it makes sense to me that both would use the same name for this third thing too. Perhaps "intermediateCertsPath" might be a reasonable name, since that implies that (1) these aren't the trusted roots, and (2) the roots shouldn't be included in the chain.

adleong: Great suggestion, thanks!

Issue #1963, opened 2018-05-26
Enable plugins to emit stat into L5D stat trees for Identifier 


leozc:
Issue Type:

- [ ] Bug report
- [ X] Feature request

This is the alternative code path for feature request similar to #1922 

1.  #1922  is done for `Announcer`
2. this bug is to track for `Identifier`

Here are the comments copied from #1922
We are writing multiple plugins for our L5D mesh, there are needs to initialize a finagle client to perform RPC, however there is no way to submit the stat into L5D metrics system.

Suggestion: Is it possible to enable this user scenario?

``` 
def getClient(withRequestTimeoutMS: Int =  1000 , addr: InetSocketAddress, path:Path): Service[http.Request, http.Response]  = {
    val monitor: Monitor = new Monitor {
      def handle(t: Throwable): Boolean = {
        logger.error(t, t.getMessage, t.getStackTrace)
        true
      }
    }
    val metrics = MetricsTree() // possible to get the metrics tree from global 
    val stats = new MetricsTreeStatsReceiver(metrics) // register this stat object to be collected?
    Http.client
      .withStatsReceiver(stats)
      .withHttpStats
      .withRequestTimeout(Duration.fromMilliseconds(withRequestTimeoutMS))
      .withMonitor(monitor)
      .newService(s"${addr.getHostName()}:${addr.getPort()}", "hc_client")

  }
```

leozc:
```
https://github.com/linkerd/linkerd/blob/fe822302d38f249533d1af326ea06efd842bf9b4/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala#L247
```
Here should be where the identifier being construed

leozc:
PR: https://github.com/linkerd/linkerd/pull/1964

leozc:
Reenable - pull the latest linkerd and debug through and found out the statReceiveer is NullStatReceiver:

Here is the corresponding section from `http://linkerd:9990/admin/registry.json`
```
            "NackAdmissionFilter" : {
              "statsReceiver" : "NullStatsReceiver/common_auth/l5d_authIdentifier",
              "window" : "2.minutes",
              "nackRateThreshold" : "0.5"
            },
```

adleong:
@leozc Please take a look at https://github.com/linkerd/linkerd/compare/master...alex/params and let me know if that fixes it for you

leozc:
[...]

PR #1964, opened 2018-05-27
Enable any identifier implements HttpIdentifierConfig to access Stack.params from its top level Router

leozc:
Firstcut for solving #1963 - asking for feedback

**Goal: Enable any identifier implements  `HttpIdentifierConfig` to access Stack.params of its top level Router**

What have been done:
1. pass `Stack.Params` from router to identifier.
2. perform plumbing through and add the new param to all subclasses

Question: 
Is it safe to pass down the whole `Stack.Param` (I think it is fine) - in my case, I just need to use the `StatsReceiver`

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -255,5 +255,4 @@ case class HttpConfig(
 
   @JsonIgnore
   override def routerParams = routerParamsPartial.maybeWith(combinedIdentifier(routerParamsPartial))
-
 }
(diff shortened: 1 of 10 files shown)


wmorgan:
Thanks for doing this, @leozc! We'll take a look.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -244,7 +244,7 @@ case class HttpConfig(
     .maybeWith(httpAccessLogAppend.map(AccessLogger.param.Append.apply))
     .maybeWith(httpAccessLogRotateCount.map(AccessLogger.param.RotateCount.apply))
     .maybeWith(loggerParam)
-    .maybeWith(combinedIdentifier)
+    .maybeWith(combinedIdentifier(routerParams))
```
adleong: is this infinitely recursive?  should we pass all of the params EXCEPT the identifier param to the identifier?

leozc: Ya silly mistake - fixed and ran test - all passed

PR #1976, opened 2018-06-06
Improve the No Hosts Available Exception message

### LINKED ISSUE #1948: Improve NoBrokersAvailableException
The infamous NoBrokersAvailableException is cryptic and doesn't provide much useful information to use for debugging.  This exception should give a better human readable explanation of the problem as well as important contextual information such as:

* The identified name that is being resolved
* A list of client names that were attempted

adleong:
When Linkerd is unable to route a request, it returns a cryptic "No Hosts Available Exception" error message.

We update this error message to give more contextual information about the request and why it could not be routed.

Old output:
```
$ curl localhost:4140 -H "Host: alex" -H "l5d-dtab: /svc/alex => /svc/leong"
No hosts are available for /svc/alex, Dtab.base=[], Dtab.local=[/svc/alex=>/svc/leong]. Remote Info: Not Available
```

New output:
```
$ curl -v localhost:4140 -H "Host: alex" -H "l5d-dtab: /svc/alex => /svc/leong"
* Rebuilt URL to: localhost:4140/
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 4140 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 4140 (#0)
> GET / HTTP/1.1
> Host: alex
> User-Agent: curl/7.54.0
> Accept: */*
> l5d-dtab: /svc/alex => /svc/leong
>
< HTTP/1.1 400 Bad Request
< l5d-err: Unable+to+route+request%21%0A%0Aservice+name%3A+%2Fsvc%2Falex%0Aresolutions+considered%3A%0A++%2F%23%2Fio.l5d.fs%2Fleong+%28neg%29%0A++%2F%23%2Fio.l5d.fs%2Fblork+%28neg%29%0A++%2F%23%2Fio.l5d.fs%2Falex+%28neg%29%0A++%2F%23%2Fio.l5d.fs%2Fblork+%28neg%29%0Adtab%3A%0A++%2Fsvc%2F*+%3D%3E+%2F%23%2Fio.l5d.fs%2Fblork%0A++%2Fsvc+%3D%3E+%2F%23%2Fio.l5d.fs%0Abase+dtab%3A%0A%0Aoverride+dtab%3A%0A++%2Fsvc%2Falex+%3D%3E+%2Fsvc%2Fleong%0A
< Content-Length: 294
< Content-Type: text/plain
<
Unable to route request!

service name: /svc/alex
resolutions considered:
  /#/io.l5d.fs/leong (neg)
  /#/io.l5d.fs/blork (neg)
  /#/io.l5d.fs/alex (neg)
  /#/io.l5d.fs/blork (neg)
dtab:
  /svc/* => /#/io.l5d.fs/blork
  /svc => /#/io.l5d.fs
base dtab:

override dtab:
  /svc/alex => /svc/leong
```

Fixes #1948 

### DIFF
--- a/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
+++ b/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
@@ -141,7 +141,7 @@ class HttpEndToEndTest
       }
 
       get("ralph-machio") { rsp =>
-        assert(rsp.status == Status.BadRequest)
+        assert(rsp.status == Status.BadGateway)
         assert(rsp.headerMap.contains(Headers.Err.Key))
         ()
       }
--- a/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
+++ b/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
@@ -141,7 +141,7 @@ class HttpEndToEndTest
       }
 
       get("ralph-machio") { rsp =>
-        assert(rsp.status == Status.BadGateway)
+        assert(rsp.status == Status.BadRequest)
         assert(rsp.headerMap.contains(Headers.Err.Key))
         ()
       }
--- a/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
+++ b/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
@@ -3,7 +3,7 @@ package com.twitter.finagle.buoyant.linkerd
 import com.twitter.finagle.{Dtab => FDtab, Status => _, _}
 import com.twitter.finagle.buoyant.{Dst => BuoyantDst}
 import com.twitter.finagle.context.{Contexts, Deadline => FDeadline}
-import com.twitter.finagle.http._
+import com.twitter.finagle.http.{param => hparam, _}
 import com.twitter.finagle.tracing._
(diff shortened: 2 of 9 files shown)


olix0r:
@adleong would you mind including the full response (i.e. `curl -v`) in the example?

adleong:
Updated.

Ashald:
This sounds like an amazing feature! Although, I'm a bit concerned with the status code change. We have number of things that rely on the fact that Linkerd returns 502 when it cannot route a request so this would be a breaking change for us. It'd be great if it was configurable instead.

olix0r:
> We have number of things that rely on the fact that Linkerd returns 502 when it cannot route a request so this would be a breaking change for us. It'd be great if it was configurable instead.

I have a strong preference to avoid configuration for this sort of thing.

@adleong Why is it appropriate to return a 400?

Ashald:
I personally find `502` to be a pretty good match. While I understand the reasoning behind 

> Inability to route is almost like a 404

I think the fact that Linkerd serves as a gateway/proxy is of bigger importance here. It's not necessarily a user error - while user indeed may have requested a service that does not exist, it's also possible that it's just all instances of requested service failed this very moment. If we were able to distinguish between an unbound name and and a bound one with`Addr.Neg` inside then we could've used more appropriate status codes (400 and 502 respectively seem as good matches in such case, IMHO).

Last but not least, 4XX are usually non-retryable while 5XX are. And given the above, I think it's safer to report a retryable error in such case if we're not sure why exactly we were not able to route the request.

olix0r:
I'd prefer, at the least, that we limit this change to only address output formatting. If we want to dig into error/response codes as a followup, I think that would also be fine.

adleong:
Thanks for the feedback!  I've reverted the status code back to 502.

Ashald:
Thank you for the amazing feature!

PR #1980, opened 2018-06-10
Fixes #1552 - Rework TLS client configuration trust store to use a CA…

### LINKED ISSUE #1552: Remove dependence on temp files for TLS certificates

dvulpe:
… bundle instead of dropping all ca certs to a bundle in the temp folder. This simplifies the client TLS code, removes a dependency on the temporary folder and aligns client TLS configuration with server TLS configuration.

TLS client configuration still supports the former trustCerts option which if it's used it will cause a warning to be logged.

Sample from the logs:
```
W 0610 17:49:05.788 BST THREAD1: trustCerts configuration option is deprecated, please consider using trustCertsBundle
```

Signed-off-by: Dan Vulpe <dvulpe@gmail.com>

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
@@ -1,11 +1,10 @@
 package com.twitter.finagle.buoyant
 
 import java.io._
-
 import com.twitter.finagle.Stack
 import com.twitter.finagle.netty4.ssl.client.Netty4ClientEngineFactory
-import com.twitter.finagle.ssl.client.{SslClientConfiguration, SslClientEngineFactory}
 import com.twitter.finagle.ssl.{KeyCredentials, Protocols, TrustCredentials}
+import com.twitter.finagle.ssl.client.{SslClientConfiguration, SslClientEngineFactory}
 import com.twitter.finagle.transport.Transport
 import com.twitter.io.StreamIO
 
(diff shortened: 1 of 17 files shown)


adleong:
Thank you, @dvulpe!  Removing our dependency on temp files will be so awesome.  I'm really looking forward to reviewing this.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/TlsClientConfig.scala
```diff
@@ -31,33 +31,17 @@ case class TlsClientConfig(
...
-        val certCollection = File.createTempFile("certCollection", null)
-        // open the cert paths as Streams...
-        val f = new FileOutputStream(certCollection)
-        for {
-          cert <- certs
-          certStream = new FileInputStream(cert)
-        } { // ...and copy the certs into the cert collection
-          // TODO: can this be made more concise with scala.io?
-          StreamIO.copy(certStream, f)
+    case TlsClientConfig(_, _, Some(cn), certs, bundle, clientAuth, enabledProtocols) =>
```
adleong: I'd prefer to be a bit more defensive here and throw an IllegalArgumentException if trustCerts and trustCertsBundle are both set.  Then we can handle them in separate cases:

```
case TlsClientConfig(_, _, Some(cn), Some(certs), Some(bundle), _, _) =>
  throw ...
case TlsClientConfig(_, _, Some(cn), Some(certs), None, _, _) =>
  // use certs
case TlsClientConfig(_, _, Some(cn), None, Some(bundle), _, _) =>
  // use bundle
case TlsClientConfig(_, _, Some(cn), None, None, _, _) =>
  // use Unspecified
```

dvulpe:
Nice one, thanks! I look forward to my next contribution to this great project!

PR #1988, opened 2018-06-13
Refactor formatDelegation into DelegateTree.find for easier reuse

adleong:
The `formatDelegation` method returns the delegation steps toward a particular leaf in a delegation tree.  This is generally useful but cannot be reused because it's a private member of `RouterContextFormatter`.

We refactor this method into `DelegateTree` itself, renaming it to `find` so that it can be more easily reused.  We also simplify the method, removing the need for an accumulator and add a test.

### DIFF
--- a/linkerd/core/src/main/scala/io/buoyant/linkerd/RouterContextFormatter.scala
+++ b/linkerd/core/src/main/scala/io/buoyant/linkerd/RouterContextFormatter.scala
@@ -69,9 +69,9 @@ object RouterContextFormatter {
         ) match {
             case Some(delegation) => delegation.map {
               case (path, "") =>
-                s"  ${path.show}"
+                s"    ${path.show}"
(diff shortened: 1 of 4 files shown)


linkerd/core/src/main/scala/io/buoyant/linkerd/RouterContextFormatter.scala
```diff
@@ -11,18 +11,8 @@ import io.buoyant.router.RoutingFactory.BaseDtab
 
 object RouterContextFormatter {
```
dadjeibaah: Unused import "import io.buoyant.namer.DelegateTree._". It might have been left over from the initial commit of this code.

adleong: good catch

PR #1992, opened 2018-06-14
Handle close terminated HTTP/1.0 responses

### LINKED ISSUE #1978: Responses from HTTP 1.0 servers get stuck
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:

Routed a request to a server that responds with HTTP 1.0 without `content-length` header, e.g. a simple Python server.
Since Linkerd 1.4.1 the response body doesn't reach the client anymore, only the headers arrive.

I expect to get the full response.

Start Linkerd with this configuration:

```yaml
admin:
  port: 9990
  ip: 0.0.0.0

routers:
- protocol: http
  dtab: |
    /svc => /$/inet/localhost/8082;
  httpAccessLog: logs/access.log
  label: rt
  servers:
  - port: 4140
    ip: 0.0.0.0
```

Start this python script with `./simple-server.py 8082`:

```python
#!/usr/bin/env python
from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
import SocketServer

port_number = 80

class S(BaseHTTPRequestHandler):
    def _set_headers(self):
        self.send_response(200)
        self.send_header('Content-type', 'text/plain')
        self.end_headers()

    def do_GET(self):
        self._set_headers()
        self.wfile.write("hello, from port: " + `port_number`)

def run(server_class=HTTPServer, handler_class=S, port=80):
    global port_number
    port_number = port
    server_address = ('', port_number)
    httpd = server_class(server_address, handler_class)
    print 'Starting httpd...'
    httpd.serve_forever()

if __name__ == "__main__":
    from sys import argv

    if len(argv) == 2:
        run(port=int(argv[1]))
    else:
        run()
```

Then start the plain linkerd-exec version 1.4.1 with this configuration file and run this call:

```
$ curl -v localhost:4140                                                
* Rebuilt URL to: localhost:4140/
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 4140 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 4140 (#0)
> GET / HTTP/1.1
> Host: localhost:4140
> User-Agent: curl/7.54.0
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Via: 1.0 linkerd
< l5d-success-class: 1.0
< Date: Fri, 08 Jun 2018 08:40:38 GMT
* HTTP/1.0 connection set to keep alive!
< Connection: keep-alive
< Server: BaseHTTP/0.3 Python/2.7.15
< Content-type: text/plain
<
```

curl is stuck now and doesn't terminate.
With 1.4.0 the full response arrives at the client.

Netstat looks like this while the call is blocked:

```
$ netstat -an|grep 8082
tcp4       0      0  *.8082                 *.*                    LISTEN
$ netstat -an|grep 4140
tcp4       0      0  127.0.0.1.4140         127.0.0.1.59227        ESTABLISHED
tcp4       0      0  127.0.0.1.59227        127.0.0.1.4140         ESTABLISHED
tcp4       0      0  *.4140                 *.*                    LISTEN 
```

It looks like the simple server has correctly closed the connection.

**Environment**:
Linkerd 1.4.1 with the config shown above
Plain OS X, no fancy container stuff or orchestrators
MacBook 2013

adleong:
When Linkerd receives an HTTP/1.0 response that is terminated by closing the connection, the response will not have a Content-Length header.  This causes Linkerd to interpret the response as chunk encoded.  Since Linkerd keeps persistent connections open to its clients, clients will not have a content-length header or connection termination to signal the end of the response.  This causes them to hang, waiting for the response to complete.

This issue was introduced in #1904 when we stopped upgrading responses to HTTP/1.1.

For HTTP/1.0 responses with no transfer-coding and no content-length, we fully buffer the response and set the content-length header. 

Fixes #1978 

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/ContentLengthFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/ContentLengthFilter.scala
@@ -1,0 +1,57 @@
+package io.buoyant.router.http
+
+import com.twitter.finagle.Stack.Module0
+import com.twitter.finagle.{Service, ServiceFactory, SimpleFilter, Stack}
+import com.twitter.finagle.http.{Fields, Request, Response, Version}
+import com.twitter.io.Reader
+import com.twitter.util.Future
+
+/**
+ * When Linkerd receives an HTTP/1.0 response that is terminated by closing the connection, the
+ * request will not have a Content-Length header.  This causes Linkerd to interpret the response
+ * as chunk encoded.  To avoid sending chunk encoded responses to HTTP/1.0 clients, we fully buffer
+ * such responses and set the Content-Length header.
+ */
+object ContentLengthFilter {
+
+  object module extends Module0[ServiceFactory[Request, Response]] {
+
+    override def role: Stack.Role = Stack.Role("content-length")
+
+    override def description: String = "Add a content-length header for HTTP/1.0 responses that need one"
+
+    override def make(next: ServiceFactory[Request, Response]): ServiceFactory[Request, Response] =
+      filter.andThen(next)
+  }
+
+  object filter extends SimpleFilter[Request, Response] {
+
+    def hasTransferEncoding(response: Response): Boolean =
+      response.headerMap.get(Fields.TransferEncoding) match {
+        case None => false
+        case Some(te) if te.toLowerCase == "identity" => false
+        case Some(_) => true
+      }
+
+    override def apply(
+      request: Request,
+      service: Service[Request, Response]
+    ): Future[Response] = {
+      service(request).flatMap { response =>
+
+        if (response.version == Version.Http10
+          && !hasTransferEncoding(response)
+          && response.contentLength.isEmpty) {
+          Reader.readAll(response.reader).map { buf =>
+            response.setChunked(false)
+            response.contentLength = buf.length
+            response.content = buf
+            response
+          }
+        } else {
+          Future.value(response)
+        }
+      }
+    }
+  }
+}
--- a/router/http/src/main/scala/io/buoyant/router/http/ContentLengthFilter.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/ContentLengthFilter.scala
@@ -8,7 +8,7 @@ import com.twitter.util.Future
 
 /**
  * When Linkerd receives an HTTP/1.0 response that is terminated by closing the connection, the
- * request will not have a Content-Length header.  This causes Linkerd to interpret the response
+ * response will not have a Content-Length header.  This causes Linkerd to interpret the response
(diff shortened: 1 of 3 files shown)


router/http/src/main/scala/io/buoyant/router/http/ContentLengthFilter.scala
```diff
@@ -0,0 +1,57 @@
...
+    override def make(next: ServiceFactory[Request, Response]): ServiceFactory[Request, Response] =
+      filter.andThen(next)
+  }
+
+  object filter extends SimpleFilter[Request, Response] {
+
+    def hasTransferEncoding(response: Response): Boolean =
+      response.headerMap.get(Fields.TransferEncoding) match {
+        case None => false
+        case Some(te) if te.toLowerCase == "identity" => false
```
dadjeibaah: According to [RFC 7230, section 3.3.1: Transfer-Encoding](https://tools.ietf.org/html/rfc7230#section-3.3.1), Transfer-Encoding header values can be a comma separated string. Is it possible we may run into a value like `chunked, identity`? If so, this may not satisfy the condition. I am not entirely sure if `chunked, identity` is an actual field value that would be generated by an Http/1.0 client.

adleong: From https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html

> 2.If a Transfer-Encoding header field (section 14.41) is present and has any value other than "identity", then the transfer-length is defined by use of the "chunked" transfer-coding (section 3.6), unless the message is terminated by closing the connection.

adleong:
Somewhat related functionality (upgrading responses to HTTP/1.1) used to be in the ViaHeaderAppenderFilter.  However, I think it's actually more clear to separate functionality out into separate filters rather than jamming it all together in one.

router/http/src/main/scala/io/buoyant/router/http/ContentLengthFilter.scala
```diff
@@ -0,0 +1,57 @@
...
+    override def apply(
+      request: Request,
+      service: Service[Request, Response]
+    ): Future[Response] = {
+      service(request).flatMap { response =>
+
+        if (response.version == Version.Http10
+          && !hasTransferEncoding(response)
+          && response.contentLength.isEmpty) {
+          Reader.readAll(response.reader).map { buf =>
```
olix0r: this indentation is a bit hard to read

PR #1994, opened 2018-06-15
Implement a gRPC standards-compliant response classifier

### LINKED ISSUE #1993: gRPC default response classifier is incomplete
Issue Type:

- [ ] Bug report
- [X] Feature request

The `io.l5d.h2.grpc.default` response classifier is incomplete in its classification of retryable gRPC responses. According to the gRPC specifications, the following events are considered retryable:

* [gRPC `UNAVAILABLE` response](https://github.com/grpc/proposal/blob/master/A6-client-retries.md)
* [HTTP/2 429, 502, 503, and 504 responses](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md)
* [HTTP/2 `RST_STREAM:REFUSED_STREAM`](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors)

The `default` response classifier only retries the first case, where a gRPC server explicitly issues an `UNAVAILABLE` response. This happens when requests are received by the server while it is in the process of shutting down. gRPC`UNAVAILABLE` responses can also be returned explicitly by the service developer to signal availability.

The second case occurs when intermediate HTTP/2 forward or reverse proxies fail, but don't understand the gRPC protocol. Raw HTTP/2 responses are returned instead.

I found the third case occurs when two L5d instances sit between a client and server (one on each side of the connection), and the server-side L5d encounters an error talking to the Server. We saw this in a partially initialized Kubernetes pod, where L5d was running but the service container was in a crash-backoff loop. In this scenario, the client-side L5d observed a `RST_STREAM:REFUSED_STREAM` when the client tried to connect. The HTTP/2 connection between the client and server L5ds was fine, but the individual streams within that connection were failing.

rmichela:
The `io.l5d.h2.grpc.default` response classifier is incomplete in its classification of retryable gRPC responses.

This commit adds the `io.l5d.h2.grpc.compliant` response classifier, with an expanded set of retryable gRPC responses.
- Status code 14 (`Unavailable`) is considered retryable - same as default
- HTTP/2 RST_STREAM:REFUSED_STREAM is considered retryable - per https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#errors
- HTTP/2 429, 502, 503, and 504 responses are considered retryable - per https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md

Implements #1993

Signed-off-by Ryan Michela <rmichela@salesforce.com>

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
@@ -140,6 +140,9 @@ object Response {
   def apply(headers: Headers, stream: Stream): Response =
     Impl(headers, stream)
 
+  def unapply(arg: Response): Option[Status] =
+    Some(arg.status)
+
   private case class Impl(
     headers: Headers,
     stream: Stream
--- a/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
+++ b/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
@@ -1,5 +1,7 @@
 package io.buoyant.linkerd.protocol.h2.grpc
 
+import com.twitter.finagle.buoyant.h2.Status.{ClientError, ServerError}
+import com.twitter.finagle.buoyant.h2.{Reset, Response, Status}
 import com.twitter.finagle.buoyant.h2.service.{H2Classifier, H2ReqRep, H2ReqRepFrame}
 import com.twitter.finagle.service.ResponseClass
 import com.twitter.util.{Return, Throw}
@@ -9,6 +11,8 @@ import io.buoyant.grpc.runtime.GrpcStatus.{Ok, Unavailable}
 trait GrpcClassifier extends H2Classifier {
 
   def retryable(status: GrpcStatus): Boolean
+  def retryable(status: Status): Boolean = false
+  def retryable(throwable: Throwable): Boolean = false
 
   /**
    * Since GRPC sends status codes in the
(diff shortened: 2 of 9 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
```diff
@@ -140,6 +140,9 @@ object Response {
   def apply(headers: Headers, stream: Stream): Response =
     Impl(headers, stream)
 
+  def unapply(arg: Response): Option[Status] =
```
dadjeibaah: Since `arg.status` throws an exception instead of returning `null`  based on the `arg.status` definition
```scala
override def status = headers.get(Headers.Status).headOption match {
      case Some(code) => Status.fromCode(code.toInt)
      case None => throw new IllegalArgumentException(s"missing ${Headers.Status} header")
    }
```

Would it work to do this instead?
```scala
def unapply(arg: Response): Option[Status] =
    Some(arg.status)
```

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/grpc/GrpcClassifier.scala
```diff
@@ -66,6 +95,35 @@ object GrpcClassifiers {
...
+    }
+
+    override def retryable(status: Status): Boolean =
+      status == Status.TooManyRequests ||
+        status == Status.ServiceUnavailable ||
+        status == Status.BadGateway ||
+        status == Status.GatewayTimeout
+
+    override def retryable(throwable: Throwable): Boolean = throwable match {
+      case t: Reset if t == Reset.Refused => true
```
dadjeibaah: I think this can be shortened to:
```scala
   case Reset.Refused => true
```

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
rmichela: `Some(arg.status)` works.

PR #1997, opened 2018-06-17
Disable statistics exported by the JVM for hsprefdata

### LINKED ISSUE #1996: Unexplainable pauses for Linkerd
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:
When observing p99, p99.9 and max latencies we have seen very high >1s latencies. When having linkerd connected to other linkerd in a mesh, and running a few nodes this has large impact across production environment.

No random spikes requiring a restart of linkerd.

Run linkerd on a partition that is spinning disk with I/O

java 8u171

**Environment**:
1.4.0
bare metal

This is happens for two reasons:
* JVM by default exports statistics by mmap-ing a file in /tmp (hsperfdata). On Linux, modifying a memory mapped file can block until disk I/O completes, which can be hundreds of milliseconds. Since the JVM modifies these statistics during garbage collection and safepoints, this causes pauses that are hundreds of milliseconds long
* If you have GC logs enabled for Linkerd and its writing to a partition that is sensitive to latency, the JVM currently blocks on `write()` syscall which is in the critical path for GC as well.

References:
* http://www.evanjones.ca/jvm-mmap-pause.html
* https://engineering.linkedin.com/blog/2016/02/eliminating-large-jvm-gc-pauses-caused-by-background-io-traffic

chrisgoffinet:
The JVM by default exports statistics by mmap-ing a file in /tmp (hsperfdata).
On Linux, modifying a memory mapped file can block until disk I/O completes, which can be hundreds of milliseconds. Since the JVM modifies these statistics during garbage collection and safepoints, this causes pauses that are hundreds of milliseconds long on spinning disks.

This will break tools that read this file, such as jcmd, jstat, jconsole. If you *really* need those, and you are running /tmp backed by tmpfs or an SSD, it's okay to enable this feature.

References:

http://www.evanjones.ca/jvm-mmap-pause.html
https://engineering.linkedin.com/blog/2016/02/eliminating-large-jvm-gc-pauses-caused-by-background-io-traffic

Note: this is a problem for writing gc logs as well (out of scope), so make sure if you are enabling GC logs for Linkerd, to write them to an SSD or tmpfs.

Fixes #1996

Signed-off-by: Chris Goffinet <chris@threecomma.io>

### DIFF
--- a/project/LinkerdBuild.scala
+++ b/project/LinkerdBuild.scala
@@ -256,7 +256,6 @@ object LinkerdBuild extends Base {
        |   -Dio.netty.allocator.numHeapArenas=${FINAGLE_WORKERS:-8}      \
        |   -Dio.netty.allocator.numDirectArenas=${FINAGLE_WORKERS:-8}    \
        |   -Dcom.twitter.finagle.netty4.numWorkers=${FINAGLE_WORKERS:-8} \
-       |   -XX:+PerfDisableSharedMem                                     \
        |   ${LOCAL_JVM_OPTIONS:-}                                        "
        |""".stripMargin
 
@@ -348,6 +347,7 @@ object LinkerdBuild extends Base {
       """|#!/bin/sh
          |
          |jars="$0"
+         |HSPREF_SETTING=$([ ! -z "$ENABLE_HSPREF" ] && echo "" || echo "-XX:+PerfDisableSharedMem")
          |if [ -n "$NAMERD_HOME" ] && [ -d $NAMERD_HOME/plugins ]; then
          |  for jar in $NAMERD_HOME/plugins/*.jar ; do
          |    jars="$jars:$jar"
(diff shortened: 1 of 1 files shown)


wmorgan:
Wild. Thanks @chrisgoffinet.

chrisgoffinet:
Yes I'll get that ready


Best,
Chris On Mon, Jun 18, 2018 at 1:48pm, Alex Leong < notifications@github.com [notifications@github.com] > wrote:
@adleong commented on this pull request.

Awesome! Should we have a separate variable for this so that it can be disabled when features like jmap are desired?

chrisgoffinet:
@adleong Sorry for the delay on this, got busy. If you want to enable the old behavior you can set `ENABLE_HSPREF` to `true` or any value. When it's not set it will disable this feature (recommended) for production use.

PR #2005, opened 2018-06-19
Expose interpreter state for interpreters with transformers

### LINKED ISSUE #1999: Namerd interpreter_state diagnostics unavailable if used with a transformer
If the Namerd interpreter is used with a transformer, the interpreter_state diagnostics endpoint will not be available.  This is because the transformer wraps the interpreter but does not expose the admin handlers of the underlying interpreter.

E.g.

```
     interpreter:
        kind: io.l5d.namerd
        dst: /$/inet/namerd/4100
        namespace: default
        transformers:
        - kind: io.l5d.k8s.daemonset
          namespace: test
          port: incoming
          service: l5d
```

### LINKED ISSUE #2000: Interpreter diagnostic endpoints clobber each other
If Linkerd is configured with multiple routers, each with the io.l5d.namerd interpreter pointing to the same dtab namespace, then each interpreter will expose a diagnostics endpoint on the same URL.  This causes one diagnostics endpoint to override the other.  

These URLs should probably be prefixed by router name to avoid collisions.

adleong:
If an interpreter is configured with a transformer, the `interpreter_state` admin endpoint is not available for that interpreter.  Additionally, if interpreters are configured on multiple routers, each interpreter will attempt to register a handler at the same URL and override the others.

We update transformers to pass through the admin handlers of the underlying interpreter so that the `interpreter_state` remains available.  We also update the URL structure of the `interpreter_state` endpoint to be prefixed with the router name.  This avoids any cross-router conflicts.

Fixes #1999 
Fixes #2000

### DIFF
--- a/project/LinkerdBuild.scala
+++ b/project/LinkerdBuild.scala
@@ -72,7 +72,7 @@ object LinkerdBuild extends Base {
 
   object Namer {
     val core = projectDir("namer/core")
-      .dependsOn(configCore)
+      .dependsOn(configCore, admin)
       .withLib(Deps.jacksonCore)
       .withTests()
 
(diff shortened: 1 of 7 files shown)


project/LinkerdBuild.scala
```diff
@@ -585,11 +585,12 @@ object LinkerdBuild extends Base {
       .dependsOn(Protocol.thrift % "test", Interpreter.perHost % "test")
 
     val main = projectDir("linkerd/main")
-      .dependsOn(admin, configCore, core)
+      .dependsOn(admin, configCore, core, Protocol.http % "e2e", Interpreter.namerd % "e2e", Interpreter.perHost % "e2e")
       .withTwitterLib(Deps.twitterServer)
```
dadjeibaah: What is the reason for adding `Protocol.http % "e2e", Interpreter.namerd % "e2e", Interpreter.perHost % "e2e"`? Where this tests that weren't added to the End to End test suite?

adleong: I added an end-to-end test in linkerd/main which uses the http protocol, the namer interpreter, and the per-host interpreter.  Therefore I need to depend on those projects in the e2e configuration.

Issue #1990, opened 2018-06-14
Cycling Kubernetes pods and scraping Prometheus metrics causing latency issues


milesbxf:
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:

We deploy Linkerd as a daemonset (to around 500 nodes) on Kubernetes, and we're scraping metrics using Prometheus.

We cycled every pod in the cluster (around 3400 in total) whilst doing some routine maintenance. We disabled Prometheus scraping prior to doing this (as part of unrelated work) and reenabled it the next day.

After scraping was reenabled latency of requests sent via Linkerd increased by around 100-150% and some requests started failing. We saw errors in the Linkerd logs like this:

```
D 0525 14:42:21.804 UTC THREAD32 TraceId:5ff10920f019ca62: Failed mid-stream. Terminating stream, closing connection
com.twitter.finagle.ChannelClosedException: ChannelException at remote address: /10.224.140.9:80. Remote Info: Not Available
    at com.twitter.finagle.netty4.transport.ChannelTransport$$anon$1.channelInactive(ChannelTransport.scala:188)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:245)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:231)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:224)
    at io.netty.channel.ChannelInboundHandlerAdapter.channelInactive(ChannelInboundHandlerAdapter.java:75)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:245)

D 0525 14:17:43.098 UTC THREAD28 TraceId:611a81b0939bcb4b: Failed mid-stream. Terminating stream, closing connection
com.twitter.io.Reader$ReaderDiscarded: This writer's reader has been discarded
  at com.twitter.finagle.netty4.http.StreamTransports$$anon$1.discard(StreamTransports.scala:71)
  at com.twitter.finagle.http.DelayedReleaseService$$anon$2$$anon$3.discard(DelayedReleaseService.scala:58)
  at io.buoyant.router.Http$Router$.$anonfun$responseDiscarder$1(Http.scala:45)
[...]

PR #2006, opened 2018-06-19
Improve Prometheus telemeter efficiency

adleong:
When the number of services is very large, the Prometheus metrics endpoint can become extremely slow.  (See https://github.com/linkerd/linkerd/issues/1990)

We improve the efficiency by 
* skipping empty histograms
* using StringBuilder directly to avoid allocating as many intermediate strings
* escaping labels earlier to avoid repeating the work of escaping for each metric

### DIFF
--- a/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
+++ b/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
@@ -44,14 +44,19 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
   private[this] val labelValDisallowedChars = """(\\|\"|\n)""".r
   private[this] def escapeLabelVal(key: String) = labelValDisallowedChars.replaceAllIn(key, """\\\\""")
 
-  private[this] def formatLabels(labels: Seq[(String, String)]): String =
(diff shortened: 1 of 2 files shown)


chrisgoffinet:
@adleong I just tested this, I had 800 services in Linkerd. The scrape time went from 400-600ms down to 68ms.

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -44,14 +44,19 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
...
-          s"""${escapeLabelKey(k)}="${escapeLabelVal(v)}""""
-      }.mkString("{", ", ", "}")
-    } else {
-      ""
+      val lastIndex = labels.size - 1
+      sb.append("{")
+      for (((k, v), i) <- labels.zipWithIndex) {
+        sb.append(k)
+        sb.append("=\"")
+        sb.append(v)
```
olix0r: If I understand correctly, this change means that all key/vals in `labels0` now skip `escapeLabelKey` and `escapeLabelVal`. Do we need to take another pass over `labels0` to do escaping?

adleong: that's correct but all the label names are defined as literals in this file and none of them contain any characters that need escaping.

PR #2010, opened 2018-06-20
Use a separate threadpool for the admin server

adleong:
Linkerd's admin server uses the same worker threadpool as Linkerd uses for serving requests in the data plane.  This means that if there are many admin requests or admin requests take a long time to complete, this can tie up all of the available threads and delay processing of data plane requests.  This can be observed as an increase in tail latency when there are a large number of stats and there are frequent requests to the stats endpoint.

We use a separate worker pool for the admin server, with a configurable number of worker threads (2 by default).  This prevents admin requests from tying up data plane workers.  To improve visibility, we also expose stats about the admin server itself in the metrics.

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
+++ b/admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
@@ -1,6 +1,7 @@
 package io.buoyant.admin
 
 import com.twitter.finagle.buoyant.TlsServerConfig
+import com.twitter.finagle.stats.StatsReceiver
 import java.net.{InetAddress, InetSocketAddress}
 import io.buoyant.config.types.Port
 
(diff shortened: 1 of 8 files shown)


admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
```diff
@@ -9,13 +10,14 @@ case class AdminConfig(
...
+  workerThreads: Option[Int] = None
 ) {
 
-  def mk(defaultAddr: InetSocketAddress): Admin = {
+  def mk(defaultAddr: InetSocketAddress, stats: StatsReceiver): Admin = {
     val adminIp = ip.getOrElse(defaultAddr.getAddress)
     val adminPort = port.map(_.port).getOrElse(defaultAddr.getPort)
     val addr = new InetSocketAddress(adminIp, adminPort)
-    new Admin(addr, tls)
+    new Admin(addr, tls, workerThreads.getOrElse(2), stats)
```
olix0r: Does this need to be validated? It seems that we only should really support values on ~ [1-64]

PR #2011, opened 2018-06-21
Add clear context functionality for H2 requests

### LINKED ISSUE #2002: When a Linkerd error occurs, send back an empty stream  response with no l5d-* header context.
As a follow up to issue #2001, Linkerd should ensure it clears the `l5d-err` as well as any response that comes up when an error in Linkerd occurs. Right now, setting `clearContext: true` in an H2 linkerd router config returns the error response in H2 stream.  

The correct behavior is to send an empty stream with no `l5d-*` headers sent in the response.

dadjeibaah:
When `clearContext: true` is set to true in a Linkerd config, H2 requests fail to clear any `l5d-` and also fails to clear any data frames that may reveal the internal state of Linkerd. 

This PR clears any `l5d-*` headers present in an H2 response and clears any data frames when a response with an `l5d-err` is encountered.

Tests were done locally, using a local H2 and Linkerd set to `clearContext: true`.

fixes #2002

### DIFF
--- a/linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
+++ b/linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
@@ -451,14 +451,16 @@ object LinkerdHeaders {
       clearLinkerdHeaders(req.headers)
       service(req).map { resp =>
         val rsp = resp.dup()
-        if (rsp.headers.contains(Err.Key)) {
-          Stream.readToEnd(rsp.stream)
-          rsp.headers.remove("content-length")
-          rsp.headers.remove("content-type")
-          clearLinkerdHeaders(rsp.headers)
-          Response(rsp.headers, Stream.empty())
+        val headers = rsp.headers
+        if (headers.contains(Err.Key)) {
+          // Reads and discards all frames from the stream to avoid H2 frame leaks
+          val _ = Stream.readToEnd(rsp.stream)
+          headers.remove("content-length")
+          headers.remove("content-type")
+          clearLinkerdHeaders(headers)
+          Response(headers, Stream.empty())
         } else {
-          clearLinkerdHeaders(rsp.headers)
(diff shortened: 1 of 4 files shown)


linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
```diff
@@ -435,13 +435,31 @@ object LinkerdHeaders {
...
-          req.headers.remove(k)
+          headers.remove(k)
+        }
+      }
+    }
+
+    def apply(req: Request, service: Service[Request, Response]) = {
+      clearLinkerdHeaders(req.headers)
+      service(req).map { resp =>
+        val rsp = resp.dup()
```
adleong: I notice that we never use `rsp` except to hold its headers.  This is basically equivalent but it might be a bit more clear:
```
val headers = resp.headers.dup()
...
clearLinkerdHeaders(headers)
Response(headers, Stream.empty())
```

linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
```diff
@@ -435,13 +435,31 @@ object LinkerdHeaders {
...
+        if (rsp.headers.contains(Err.Key)) {
+          Stream.readToEnd(rsp.stream)
```
adleong: maybe write this as `val _ = Stream.readToEnd(rsp.stream)` to explicitly signal that we're doing a fire-and-forget future here.  Or add a comment.

linkerd/protocol/h2/src/main/scala/com/twitter/finagle/buoyant/h2/LinkerdHeaders.scala
dadjeibaah: Yea, that is much more straightforward. Thanks for the suggestion!

PR #2015, opened 2018-06-25
Instrument the consul namer

### LINKED ISSUE #1911: Admin endpoint that exposes internal consul namer state
Linkerd consul namer should expose an admin endpoint which provides data about the internal state of that namer including:

ivanopagano:
### Problem
Calls to the consul namer are not available for examination to monitor the service state

### Solution
Introduce some instrumentation to the namer calls and record available information, including the underlying service requests and response, timing to serve the requests, final result sent to the namer caller.  
This includes caching of the namer calls, whose values, keyed on service path, will be available through an administration endpoint over http, as json.

The solution uses a similar approach to how other `Namer`s state is exposed

### Validation
the aggregated test-suite for the project passes all checks

Local tests were made based on the example module consul config

Fixes #1911 

Signed-off-by: Ivano Pagano ivano.pagano@scalac.io


namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulNamer.scala
```diff
@@ -17,8 +21,16 @@ object ConsulNamer {
     weights: Map[String, Double] = Map.empty,
     stats: StatsReceiver = NullStatsReceiver
   ): Namer = {
-    val lookup = new LookupCache(consulApi, agentApi, setHost, consistency, preferServiceAddress, weights, stats)
-    new TaggedNamer(lookup, prefix)
+    val taggingParser = (path: Path) =>
```
adleong: Is it possible to make to avoid this refactor in favor of a smaller change?  For instance, to keep the TaggedNamer and UntaggedNamer classes but have them both extend a trait which adds the handlerPrefix and adminHandler implementations.  I think this lets us avoid needing to create and pass this parser function.

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -33,59 +39,72 @@ private[consul] class LookupCache(
...
-    cachedLookup(dc, key, id, residual)
+  def apply(servicePath: Path): Option[Activity[NameTree[Name]]] = {
+    val bound =
+      Option(lookupStatus.get(servicePath)) orElse
+        pathParser(servicePath).map {
+          case (dc, key, id, residual) =>
+            log.debug("consul lookup: %s %s", dc, id.show)
+            lookupCounter.incr()
+            val binding = lookup(dc, key, id, residual)
+            lookupStatusMu.synchronized { lookupStatus.put(servicePath, binding) }
```
adleong: I think we need to expand the scope of this synchronized block to include the creation of the Activity (ie the call to lookup) and also add a second check to see if the entry is in the cache inside the synchronized block.

This is to avoid the race condition where two different Activities could be returned for the same path.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -42,18 +42,26 @@ private[consul] object SvcAddr {
     consistency: Option[v1.ConsistencyMode] = None,
     preferServiceAddress: Option[Boolean] = None,
     tagWeights: Map[String, Double] = Map.empty,
-    stats: Stats
+    stats: Stats,
+    stateWatch: Option[PollState[http.Request, v1.IndexedServiceNodes]] = None
```
adleong: Why is this an Option?  Isn't it always called with Some?

adleong:
I've tested this locally with these steps:

1. Run consul: `consul agent -dev`
2. Run Linkerd: `./sbt linkerd-examples/consul:run`
3. Issue request: `curl localhost:4140 -H "Host: consul"`
4. Check state endpoint: `open http://localhost:9990/namer_state/io.l5d.consul.json`

I get this serialization error:

```
com.fasterxml.jackson.databind.JsonMappingException: No serializer found for class com.twitter.collection.RecordSchema$Record and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) (through reference chain: scala.collection.convert.Wrappers$MapWrapper["/dc1/consul"]->scala.collection.convert.Wrappers$MapWrapper["poll"]->io.buoyant.namer.consul.PollState["request"]->com.twitter.finagle.http.Request$Impl["response"]->com.twitter.finagle.http.Response$Impl["ctx"])
```

namer/consul/src/main/scala/io/buoyant/namer/consul/InstrumentedWatch.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.twitter.finagle.http
+import com.twitter.util.{Return, Throw, Time, Try, Future}
+import io.buoyant.consul.v1
+
+/**
+ * PollState holds metadata about the calls on consul api.
+ * This class is intended to be serialized.
+ *
```
adleong: extra blank line here is not needed

namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulNamer.scala
ivanopagano: Indeed it would be possible, but only if the actual "parser" is passed to the `LookupCache.apply` call each time by the corresponding subclass.

This is subjective, but this way the essential difference between the two Namers is actually captured by that function and nothing more.

Even more, the function is passed to build the specific instance of `LookupCache` because it's there that the Path is now parsed, to avoid passing both the parsed Path elements and the whole Path (needed as a cache key) to the cache lookup call.

I'm not totally against the different subclasses, I'm just pointing out things to consider to decide what's the best solution.

Let me know what you think about this

adleong: I agree with you on all points.  However, my preference is to avoid refactors like this unless
a) The refactor is clearly superior or b) the refactor is inextricably connected to the change at hand.

Otherwise, I'd prefer to keep this PR focused on the new functionality and relegate any non-essential refactors to separate PRs.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
ivanopagano: I must've been overly-cautious to start with, considering it would be called from different places. Removing now

namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulNamer.scala
ivanopagano: Cool then, I'll see to it 👍

adleong:
Here is the output I got when testing locally:

```
{
  "/dc1/consul": {
    "state": {
      "running": true,
      "lastStartedAt": "2018-06-26 22:23:25 +0000",
      "lastUpdatedAt": "2018-06-26 22:23:25 +0000",
      "value": {
        "value": {
          "addr": {
            "changes": {}
          },
          "id": "/#/io.l5d.consul/dc1/consul",
          "path": "/"
        }
      }
    },
    "poll": {
      "request": {
        "method": "GET",
        "uri": "/v1/catalog/service/consul?index=6&dc=dc1",
        "headers": {
          "X-B3-SpanId": "42276b1cdd0d006e",
          "Finagle-Ctx-com.twitter.finagle.Retries": "0",
          "Host": "localhost:8500",
          "X-B3-TraceId": "42276b1cdd0d006e"
        }
      },
      "lastRequestAt": "2018-06-26 22:23:25 +0000",
      "response": {
        "value": [
          {
            "Node": "Alexs-MBP.localdomain",
            "Address": "127.0.0.1",
            "ServiceID": "consul",
            "ServiceName": "consul",
            "ServiceTags": [],
            "ServiceAddress": "",
            "ServicePort": 8300
          }
        ],
        "index": "6"
      },
      "lastResponseAt": "2018-06-26 22:23:25 +0000"
    }
  }
}
```

adleong:
One thing that I notice is that we're using an `InstrumentedActivity[NameTree[Name.Bound]]`.  The problem with that is that the value type of the InstrumentedActivity (`NameTree[Name.Bound]`) doesn't serialize nicely because it contains an embedded `Var[Addr]`.  That's why we see:

```
          "addr": {
            "changes": {}
          },
```

instead of something useful.  Rather than storing and exposing an `InstrumentedActivity[NameTree[Name.Bound]]` I think we should instead store an expose an `InstrumentedVar[Addr]`.  ie the result of SvcAddr.apply.

adleong:
I'm also not sure if there's any value in recording the HTTP request headers in the API calls.  There isn't any relevant information stored in these headers and I think they may just be a distraction.  Consider the approach taken in `InstramentedWatch.scala` where only the method and uri are stored (and they are stored a string rather than as a case class).

ivanopagano:
This is what I get from the endpoint after my latest changes
```json
{
    "/dc1/consul": {
        "poll": {
            "lastRequestAt": "2018-06-27 18:31:40 +0000",
            "lastResponseAt": "2018-06-27 18:31:40 +0000",
            "request": "GET /v1/catalog/service/consul?index=153&dc=dc1",
            "response": {
                "index": "153",
                "value": [
                    {
                        "Address": "127.0.0.1",
                        "Node": "l5d-test",
                        "ServiceAddress": "127.0.0.1",
                        "ServiceID": "consul",
                        "ServiceName": "consul",
                        "ServicePort": 9999,
                        "ServiceTags": [
                            "primary",
                            "v1"
                        ]
                    }
                ]
            }
        },
        "state": {
            "lastStartedAt": "2018-06-27 18:26:47 +0000",
            "lastUpdatedAt": "2018-06-27 18:31:40 +0000",
            "running": true,
            "value": {
                "addrs": [
                    {
                        "addr": "127.0.0.1:9999",
                        "metadata": {
                            "endpoint_addr_weight": 1.0
                        }
                    }
                ],
                "metadata": {}
            }
        }
    },
    "/dc1/play": {
        "poll": {
            "lastRequestAt": "2018-06-27 18:30:53 +0000",
            "lastResponseAt": "2018-06-27 18:30:53 +0000",
            "request": "GET /v1/catalog/service/play?index=66&dc=dc1",
            "response": {
                "index": "66",
                "value": [
                    {
                        "Address": "192.168.0.148",
                        "Node": "l5d-test",
                        "ServiceAddress": "127.0.0.1",
                        "ServiceID": "play",
                        "ServiceName": "play",
                        "ServicePort": 9000,
                        "ServiceTags": [
                            "primary",
                            "v1"
                        ]
                    }
                ]
            }
        },
        "state": {
            "lastStartedAt": "2018-06-27 18:30:53 +0000",
            "lastUpdatedAt": "2018-06-27 18:30:53 +0000",
            "running": true,
            "value": {
                "addrs": [
                    {
                        "addr": "127.0.0.1:9000",
                        "metadata": {
                            "endpoint_addr_weight": 1.0
                        }
                    }
                ],
                "metadata": {}
            }
        }
    }
}
```

namer/consul/src/main/scala/io/buoyant/namer/consul/package.scala
```diff
@@ -1,10 +1,26 @@
...
 
 package object consul {
   private[consul]type VarUp[T] = Var[T] with Updatable[T]
   private[consul]type ActUp[T] = VarUp[Activity.State[T]]
   val log = Logger.get("io.buoyant.namer.consul")
+
+  /** wraps a path and keeps additional consul-specific structure */
+  private[consul] case class ConsulPath(
+    raw: Path,
+    scheme: Option[PathScheme] = None
```
adleong: Why is this an Option?  As far as I can tell it is always Some.

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -72,21 +106,23 @@ private[consul] class LookupCache(
...
+        val cached = Option(lookupStatus.get(servicePath.raw))
+        val address = cached match {
+          case Some(InstrumentedBind(addr, _)) =>
+            Future(addr)
+          case None =>
+            log.debug("consul lookup: %s %s", dc, id.show)
+            lookupCounter.incr()
+            lookupAddress(servicePath.raw, dc, key)
+        }
+        convertToName(address, id, residual)
```
adleong: I'm not sure about this refactor. With this change, only the Var[Addr] is cached, the Activity[NameTree[Name.Bound]] is recreated on each call.

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -33,37 +38,66 @@ private[consul] class LookupCache(
...
+   */
+  private[this] val lookupStatusMu = new {}
+  private[this] val lookupStatus = new ConcurrentHashMap[Path, InstrumentedBind]()
+
+  private[consul] def status: Map[Path, InstrumentedBind] = lookupStatus.asScala.toMap
+
+  private[this] val lookupAddress: (Path, String, SvcKey) => Future[InstrumentedVar[Addr]] =
+    (raw, dc, key) => {
+      val pollState = SvcAddr.mkConsulPollState
+      val addrFuture: Future[InstrumentedVar[Addr]] = resolveDc(dc).join(domain).map {
```
adleong: I think we want to reorder this so that the work of creating this Var[Addr] is only done if it's definitely not in cache.

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -33,37 +38,66 @@ private[consul] class LookupCache(
...
+            weights,
+            serviceStats,
+            pollState
+          )
+      }
 
+      //if the lookup succeeded we want to cache the result, double checking for race conditions
+      addrFuture.map {
+        addr =>
+          var cachedAddr: InstrumentedVar[Addr] = addr
```
adleong: should be able to do this without a `var`

namer/consul/src/main/scala/io/buoyant/namer/consul/package.scala
ivanopagano: Now that I think of it, after the latest changes it doesn't makes sense anymore, I'll fix

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
ivanopagano: indeed, that's why the function doing that is memoized (i.e. `convertToName`)

To be honset, I'm not entirely sure if memoization is a good fit for an input parameter of type `Future[InstrumentVar[Addr]]` actually.

Would you like it better if we make 2 actual caches, one for the namer call (holding the `Activity[...]`) and another for the admin endpoint calls (holding the `InstrumentedBind`, and thus the original `InstrumentedVar[Addr]`)?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
ivanopagano: this would happen only if there's a race condition to compute the address, and never again once the cache is filled

are you suggesting to synchronize the whole method with the mutex for the cache and double check at the beginning of the method?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
ivanopagano: I've rewritten it to use a `Var[...]` to be able to make the result visible outside the synchronized thunk

I'm not familiar enough with `twitter-util` to evaluate if there could be issues related to `Var` lazy semantics. wdyt?

Did you had anything else in mind?

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
ivanopagano: Now that I think of it, are you suggest to instead memoize the "whole LookupCache" call instead of simply `convertToName`?

That wouldn't sound so bad in fact

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -33,60 +38,94 @@ private[consul] class LookupCache(
...
+      }
+      Activity(observation)
+    }
 
-        cachedCounter.incr()
-        new Activity(observation)
+  private[this] val compute: (Path, String, SvcKey, Path, Path) => Activity[NameTree[Name]] =
+    untupled(Memoize[(Path, String, SvcKey, Path, Path), Activity[NameTree[Name]]] {
+      case (raw, dc, key, id, residual) => {
+        val cached = Option(lookupStatus.get(raw))
```
adleong: This cache within a cache is confusing to reason about.  I think this lookup will always fail because lookupAddress will insert this item into cache.

ivanopagano: I didn't actually trust enough `Memoize` to avoid race conditions
I've reviewed the source code and I'll remove this redundant check

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -33,60 +38,85 @@ private[consul] class LookupCache(
...
+
+      //if the lookup succeeded we want to cache the result, double checking for race conditions
+      addrFuture.map {
+        addr =>
+          lookupStatusMu.synchronized {
+            if (!lookupStatus.containsKey(raw)) {
+              lookupStatus.put(raw, InstrumentedAddr(addr, pollState))
+              cachedCounter.incr()
+            }
+            lookupStatus.get(raw).addr
```
adleong: I think this is always going to be the same value as was just put into cache so this can simply be `addr`.  In fact, since this is the original value of the future anyway, this whole `map` block can be replaced by an `onSuccess` block.

ivanopagano: will do

adleong:
@ivanopagano this is looking great!  I think once you change `Option[PathScheme]` in `ConsulPath` to not be an option and update the call from map to onSuccess in LookupCache this will be ready to merge.

namer/consul/src/main/scala/io/buoyant/namer/consul/InstrumentedWatch.scala
```diff
@@ -0,0 +1,54 @@
...
+        response = None
+    }
+  }
+
+  // These fields exist to be serialized.
+  protected var request: Option[Req] = None
+  protected var lastRequestAt: Option[String] = None
+  protected var response: Option[Rep] = None
+  protected var lastResponseAt: Option[String] = None
+  protected var error: Option[Throwable] = None
```
adleong: I just noticed this in the consul dtab store PR but it applies here too:

arbitrary throwables cannot be serialized. Let's make this an Option[String] and in the Throw case of recordResponse, we should populate this with the throwable's message.

ivanopagano: I'll put myself to it after dinner

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -33,60 +38,79 @@ private[consul] class LookupCache(
...
+            domainOption,
+            consistency = consistency,
+            preferServiceAddress = preferServiceAddress,
+            weights,
+            serviceStats,
+            pollState
+          )
+      }
+
+      //if the lookup succeeded we want to cache the result, double checking for race conditions
```
adleong: This comment is no longer accurate, there's not "double checking" anymore.  I think this is fine because this function is only ever called from inside of a memoize block, but we should fix the comment.

ivanopagano: done and fixed the previous comment on the concurrent map definition

ivanopagano:
Thanks for being patient and for the careful review

namer/consul/src/main/scala/io/buoyant/namer/consul/LookupCache.scala
```diff
@@ -6,10 +6,15 @@ import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.stats.{NullStatsReceiver, StatsReceiver}
 import com.twitter.util._
 import io.buoyant.consul.v1
+import io.buoyant.namer.{InstrumentedActivity, InstrumentedVar}
```
dadjeibaah: Unused import `InstrumentedActivity`

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -14,6 +15,11 @@ import com.twitter.util.{NonFatal => _, _}
 import io.buoyant.consul.log
 import scala.util.control.NonFatal
 
+// a thunked version of the api call such that we can peek at the request before making the call
```
dadjeibaah: Is this supposed to read as chunked maybe?

PR #2016, opened 2018-06-26
Use existing opentracing context

### LINKED ISSUE #1447: Add a "l5d-trace" user header which can be used to manipulate the "l5d-ctx-trace" header
Per the request in #1428, we should consider adding an additional [user header](https://linkerd.io/config/head/linkerd/index.html#user-headers) to set the trace id, so that users don't have to set the "l5d-ctx-trace" header, which requires trace id encoding. Instead, we could support a plaintext "l5d-trace" header on the request, which will be rewritten into the "l5d-ctx-trace" header on the response. I picture this working a lot like the "l5d-dtab" and "l5d-ctx-dtab" headers, described [here](https://github.com/linkerd/linkerd/blob/master/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala#L14).

Editorial note: I'm not sure if it would be better for this header to be called "l5d-trace" and require a root:parent:id tuple, or to instead to be called "l5d-reqid" with just the root id, which would be more similar to the "l5d-reqid" header that's we're already setting on the response.

jmalvarezf:
Hi,

As we wanted to have the ability for linkerd to propagate the opentracing context that was originated outside linkerd, and it wasn't clearly in your roadmap, we went ahead and modified linkerd to do this. 
I'm a scala newbie, so the code may be less than optimal. Any change you feel would fit better, feel more than free to tell me. The change consists basically in adding an if when getting tracing context to get opentracing headers, and to propagate the context in the same headers in the next step. 
This is very important for us to be able to have the full trace for a request, as it is originated outside the platform that is running linkerd as service mesh.

We have tested this in our deployment and it's working as expected.

This should fix #1447 with standard opentracing headers

Thanks!

### DIFF
--- a/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
+++ b/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
@@ -35,16 +35,16 @@ import scala.collection.breakOut
  *   - `l5d-dst-concrete`: the concrete client name after delegation
  *   - `l5d-dst-residual`: an optional residual path remaining after delegation
  *   - `l5d-reqid`: a token that may be used to correlate requests in
- *                  a callgraph across services and linkerd instances
+ * a callgraph across services and linkerd instances
  *
  * And in addition to the context headers, linkerd may emit the following
  * headers on outgoing responses:
  *
  *   - `l5d-err`: indicates a linkerd-generated error. Error responses
- *                that do not have this header are application errors.
+ * that do not have this header are application errors.
  *   - `l5d-retryable`: if true, the request for this response is known to be
- *                safe to retry (for example, because it was not delivered to
- *                its destination)
+ * safe to retry (for example, because it was not delivered to
+ * its destination)
  */
 object Headers {
   val Prefix = "l5d-"
@@ -70,6 +70,7 @@ object Headers {
         val description = "Extracts linkerd context from http headers"
 
         val deadline = new Deadline.ServerFilter
+
         def dtab(maxHeaderSize: Int) = new Dtab.ServerFilter(maxHeaderSize)
 
         def make(maxHeaderSize: hparam.MaxHeaderSize, next: ServiceFactory[Request, Response]) =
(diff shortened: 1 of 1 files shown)


adleong:
Thanks for this, @jmalvarezf!  I agree with you that Linkerd is long overdue to play more nicely with other tracing systems like OpenTracing.  Unfortunately, there are a few complications.

So far, Linkerd has always used its own `l5d-ctx-trace` header for trace propagation and just blindly forwarded other headers such as the `x-b3` ones.  This means that Linkerd tracing can peacefully co-exist in a system that also does `x-b3` style tracing.  The two systems will basically be invisible to each other.  

With your change, Linkerd will read the `x-b3` headers and prefer them to the `l5d-ctx-trace` header.  It will also write out both styles of headers.  The problem with this is that it's a backwards incompatible change.  In cases where both headers are present, this changes Linkerd's behavior to use the `x-b3` header instead of the `l5d-ctx-trace` one.

Even switching the precedence so that Linkerd prefers the `l5d-ctx-trace` header over the `x-b3` headers doesn't solve the problem fully.  In cases where the request is sent through a tracing aware service that increments the `x-b3` span header, that span would get lost at the next Linkerd because Linkerd would ignore it in favor of the `l5d-ctx-trace` header.

The complexity compounds even further if you want to support other tracing propogation headers such as jaeger's `uber-trace-id` or lightstep's `ot-tracer-` headers.

The TLDR is that I think this needs to be something configurable so that users can decide which tracing systems they want Linkerd to participate in and which they want Linkerd to ignore.  Would you be willing to hold off on this PR while I put together a Linkerd trace propagation plugin interface?  This would then allow you to resubmit your change as a trace propagation plugin.

I hope this context is useful, please let me know what you think!

jmalvarezf:
Hi @adleong,

Yes, I think it makes total sense! 
I will rewrite this when you have the interface ready. Do you want me to close this? Or you prefer to leave it open? 

Thank you!

adleong:
@jmalvarezf I'll leave it up to you.  Either way we can close this one once the replacement is ready.

adleong:
Please see https://github.com/linkerd/linkerd/pull/2025 and let me know what you think.

jmalvarezf:
I'll try to adapt this to the interface you propose during the following month when I have enough time and I'll let you know. Thank you!

adleong:
@jmalvarezf the new interface has merged.  Please let me know if you need any help with this new interface!

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -286,32 +290,47 @@ object Headers {
           service(req)
         }
       }
+
     }
 
     object Trace {
       val Key = Prefix + "trace"
+      val OpentracingSpanHeader = "x-b3-spanid";
```
adriancole: this has nothing to do with opentracing by the way. OpenTracing includes no library support for B3 unless something changed recently. [census does](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#propagation) however. B3 is defined here https://github.com/openzipkin/b3-propagation

jmalvarezf:
You are right. This is not opentracing but zipkin. Ill close this and start working on the interface already merged.

Thanks

adriancole:
cheers!

PR #2020, opened 2018-06-27
Upgrade to Finagle 18.5.0 and switch from Buf to ByteBuf

adleong:
Upgrade to Finagle 18.5.0.  In this version of Finagle, certain utility methods for converting between Buf and ByteBuf were removed.  This gives us the perfect opportunity to change the type of the `buf` field of `Frame.Data` from Buf to ByteBuf.

Netty uses ByteBuf as its byte buffer type throughout.  In our H2 implementation we wrap Netty's ByteBufs in Buf, Finagle's byte buffer type.  Buf is type that provides a nice API and abstracts over a variety of byte buffer types such as Array[Byte], ByteBuffer, and ByteBuf.  However, since we only use byte buffers from Netty, our Bufs always contain a Netty ByteBuf.  Not only does this make the abstraction unnecessary but the abstraction is also leaky because we require that the Bufs contain a ByteBuf in order for Netty's ByteBuf reference counting to work correctly.

We change `Frame.Data` to hold a Netty ByteBuf directly instead of a Buf which wraps a ByteBuf.  This avoids the need to convert between the two and eliminates a pointless abstraction.

### DIFF
--- a/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
+++ b/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
@@ -1,9 +1,9 @@
 package io.buoyant.grpc.runtime
 
-import com.twitter.concurrent.{AsyncMutex, Permit}
+import com.twitter.concurrent.AsyncMutex
 import com.twitter.finagle.buoyant.h2
-import com.twitter.io.Buf
-import com.twitter.util.{Future, Return, Promise, Throw, Try}
+import com.twitter.util.{Future, Return, Throw, Try}
+import io.netty.buffer.{ByteBuf, CompositeByteBuf, Unpooled}
 import java.nio.ByteBuffer
 import scala.util.control.NoStackTrace
 
@@ -16,7 +16,7 @@ import scala.util.control.NoStackTrace
  * invoked serially, i.e. when there are no pending `recv()` calls on
  * the stream.
  */
-private[runtime] trait DecodingStream[+T] extends Stream[T] {
+private[runtime] trait DecodingStream[T] extends Stream[T] {
   import DecodingStream._
 
   protected[this] def frames: h2.Stream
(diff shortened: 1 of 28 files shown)


grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
```diff
@@ -76,7 +76,7 @@ object ServerDispatcher {
 
     private[this] def acceptUnary[Req](codec: Codec[Req], req: h2.Request): Future[Req] =
```
dadjeibaah: unused import on Line 5.

project/Deps.scala
```diff
@@ -8,14 +8,14 @@ object Deps {
 
   // process lifecycle
```
dadjeibaah: Should we update `addSbtPlugin("com.twitter" %% "scrooge-sbt-plugin" % "18.4.0")` in `plugin.sbt` as well?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -275,102 +350,11 @@ object DecodingStream {
...
-        val msgbb = bb.duplicate()
-        msgbb.limit(end)
-        val msg = decoder(msgbb)
-        Some(Stream.Releasable(msg, release))
-      }
-      // Update the unparsed buffer to point after the parsed message.
-      bb.position(end)
-      Decoded(RecvState.Buffer(None, Buf.ByteBuffer.Owned(bb), nextReleaser), msg)
-    } else Decoded(RecvState.Buffer(Some(hdr), Buf.ByteBuffer.Owned(bb), releaser), None)
+  private case class Header(compressed: Boolean, size: Int)
```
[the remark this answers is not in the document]
adleong: this is an inner class of `object DecodingStream` which is a singleton.  So `private` and `private[this]` are equivalent, I think.

project/Deps.scala
adleong: good catch

PR #2025, opened 2018-06-29
Make trace propagation pluggable

adleong:
Linkerd uses the `l5d-ctx-trace` header to propagate trace context while ignoring trace headers from other tracing systems such as the zipkin `x-b3`.  While this allows Linkerd's tracing to co-exist with other tracing systems without interfering with them, it does not allow Linkerd to continue the trace initiated by another system.

We make Linkerd's trace propagation a pluggable module.  A trace propagation module decides how to read the trace context from a request when Linkerd receives one, and how to write the trace context onto a request when Linkerd sends one.  This PR includes only one implementation of this module: io.l5d.default which implements the existing behavior using the l5d-ctx-trace header.


linkerd/core/src/main/scala/io/buoyant/linkerd/TracePropagator.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Write the trace id onto a request.
+   */
+  def setContext(req: Req, traceId: TraceId): Unit
+}
+
+object TracePropagator {
+
+  def NilTracePropagagtor[Req] = new TracePropagator[Req] {
```
dadjeibaah: s/NilTracePropagagtor/NilTracePropagator

linkerd/core/src/main/scala/io/buoyant/linkerd/TracePropagator.scala
```diff
@@ -0,0 +1,42 @@
...
+   * Read the trace id from the request, if it has one.
+   */
+  def traceId(req: Req): Option[TraceId]
+
+  /**
+   * Return a sampler which decides if the given request should be sampled, based on properties
+   * of the request.  If None is returned, the decision of whether to sample the request is deferred
+   * to the tracer.
+   */
+  def sampler(req: Req): Option[Sampler]
```
olix0r: Presumably, we just hand the request right back to the sampler. Can this API be simplified?

adleong: Not quite.  Sampler takes a Long as input, not a request.  We feed the request's trace id into the sampler.  I think we want to ensure Sampler is used here rather than just returning Option[Boolean] to ensure that sampling is based on the trace id and is properly salted.

olix0r: Why do we need to pass a request to the interface? so that it can modify headers? should the method name be updated to indicate it's usually side-effecty?

PR #2027, opened 2018-06-29
instrument consul dtabs

### LINKED ISSUE #1915: Admin endpoint that exposes internal consul dtab store state
Namerd's consul dtab store should expose an admin endpoint which provides data about the internal state of that store including:

ivanopagano:
Expose the internal consul dtabs state

The admin API have no access to the internal requests for dtabs based on the consul kv store

Introduce some instrumentation to the consul dtabs store calls and record available information, including the underlying service requests and response, timing to serve the requests, final result sent to the caller.
This includes caching of the dtabs calls related to consul, whose values, keyed on the requested namespace, will be available through an administration endpoint over http, as a json response.

The solution uses a similar approach to how Namers internal state is exposed

The current aggregated test-suite for the project passes all checks  
Local test were made for the http endpoint

Fixes #1915 

Signed-off-by: Ivano Pagano ivano.pagano@scalac.io


ivanopagano:
I'd need a hand to figure out how to add the new Admin Handlers exposed now by the `ConsulDtabStore` to the standard admin service

Looking at the codebase I seem to understand that Namers endpoints are added by default, but it seems instead that it doesn't work the same for the DtabStore

ivanopagano:
I'm also wondering if it's useful for the `ConsulDtabStore.list` calls to be recorded and shown to the admin endpoints.

For what I understand the registered namespaces are already available as keys to the dtabs map, and I don't see any particular value in a separate endpoint

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -212,6 +225,51 @@ class ConsulDtabStore(
...
-    Activity(run).stabilize
+    dtabStatus.putIfAbsent(ns, InstrumentedDtab(run, pollState))
+    Activity(run.underlying).stabilize
   }
+
+  val handlerPrefix = root.show.drop(1) // drop leading "/"
+
+  override def adminHandlers: Seq[Admin.Handler] = Seq(
+    Admin.Handler(
+      s"/namer_dtabs_store/${handlerPrefix}.json",
```
adleong: This results in a URL like `/namer_dtabs_store/namerd/dtabs.json`.  I we want to have a URL like `/namer_dtabs_store/namerd/default.json` where `default` is the name of dtab namespace.  This should be as simple as using `ns` instead of `root` here.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -212,6 +225,51 @@ class ConsulDtabStore(
...
+      new ConsulDtabStoreHandler(dtabStatus.asScala.toMap)
+    )
+  )
 }
+
+private[consul] case class InstrumentedDtab(
+  act: InstrumentedVar[Activity.State[Option[VersionedDtab]]],
```
adleong: I think we should also add a custom serializer for VersionedDtab so that this gets printed more nicely in the output.  In particular, we would want to serialize the version byte array as a string:

```
      "value": {
        "dtab": "/svc=>/#/io.l5d.fs;/svc/cat=>/#/io.l5d.fs/dog | /#/io.l5d.fs/cat",
        "version": "MjAw"
      }
```

consul/src/main/scala/io/buoyant/consul/v1/Instrumentation.scala
```diff
@@ -0,0 +1,56 @@
...
+        response = None
+    }
+  }
+
+  // These fields exist to be serialized.
+  protected var request: Option[Req] = None
+  protected var lastRequestAt: Option[String] = None
+  protected var response: Option[Rep] = None
+  protected var lastResponseAt: Option[String] = None
+  protected var error: Option[Throwable] = None
```
adleong: arbitrary throwables cannot be serialized.  Let's make this an `Option[String]` and in the Throw case of recordResponse, we should populate this with the throwable's message.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
adleong: wait nevermind, I think what you have here with using `root` is actually correct.  Ignore my previous comment.

adleong: how about `storage` for the first segment?  so that this becomes `/storage/dtabs.json`?

dadjeibaah: I am currently working on the k8s version of this implementation and was thinking about how the endpoint would look like in the k8s scenario. I like the `/storage/dtab.json`. What about including the storage type as well in the link like `/storage/io.l5d.consul/dtabs.json` or `/storage/consul/dtabs.json`?

adleong: including the type is less important here than it is for namers or interpreters because there can only be one storage module configured in Namerd (whereas there may be many namers configured).

dadjeibaah: Ahh I see, I was under the impression that you could have multiple storage modules in namerd.

consul/src/main/scala/io/buoyant/consul/v1/Instrumentation.scala
ivanopagano: will this correction need to be done on `io.buoyant.namerd.iface.PollState`, too?

the original code was there

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
ivanopagano: fine for me

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
ivanopagano: I shamelessly stole the one for k8s from @dadjeibaah (thanks)

adleong:
Now that #2015 has merged, can you merge latest master here?

linkerd/docs/namer.md
```diff
@@ -194,9 +194,13 @@ namers:
 ```
 
 Linkerd provides support for service discovery via [Consul](https://www.consul.io/). 
+
 The internal state of the Consul namer can be viewed at the
 admin endpoint: `/namer_state/<prefix>.json`.
 
+The internal store of the Consul dtabs can be viewed at the
```
adleong: This documentation belongs in namerd/docs/storage.md

ivanopagano: I'll see to it

PR #2028, opened 2018-06-29
Expose k8s dtab storage state on admin endpoint

### LINKED ISSUE #1916: Admin endpoint that exposes internal k8s dtab store state
Namerd's k8s dtab store should expose an admin endpoint which provides data about the internal state of that store including:

dadjeibaah:
When using Namerd with k8s dtab storage, it can be difficult to observe the internal dtab state that is used by Linkerd instances. This can make things difficult to debug when Namerd is used with Linkerd.

This PR exposes the internal dtab states that Namerd has stored in k8s.

screen shot 2018-07-03 at 2 23 51 pm

fixes #1916

### DIFF
--- a/namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
+++ b/namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
@@ -1,14 +1,17 @@
 package io.buoyant.namerd.storage
 
-import com.twitter.finagle.{Http, Dtab => FDtab}
+import com.twitter.finagle.http.{Request, Response, Status}
+import com.twitter.finagle.{Http, Service, Dtab => FDtab}
 import com.twitter.io.Buf
 import com.twitter.logging.Logger
 import com.twitter.util._
+import io.buoyant.admin.Admin
+import io.buoyant.config.Parser
 import io.buoyant.k8s.ObjectMeta
+import io.buoyant.namer.InstrumentedActivity
 import io.buoyant.namerd.DtabStore.{DtabNamespaceAlreadyExistsException, DtabNamespaceDoesNotExistException, DtabVersionMismatchException}
 import io.buoyant.namerd.storage.kubernetes._
-import io.buoyant.namerd.{Ns, DtabStore, VersionedDtab}
-import java.util.concurrent.atomic.AtomicReference
+import io.buoyant.namerd.{DtabStore, Ns, VersionedDtab}
 import scala.collection.breakOut
 
 /**
@@ -29,7 +32,7 @@ import scala.collection.breakOut
  *                  differs from Dtab namespaces)
  */
 class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
-  extends DtabStore {
+  extends DtabStore with Admin.WithHandlers {
   private[this] val log = Logger()
 
   // NOTE: we construct two clients, one for the streaming watch requests, one for everything else.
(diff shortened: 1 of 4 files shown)


namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
```diff
@@ -69,7 +72,7 @@ class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
   // semantics. If those are required, we would probably need to implement a synchronized means of
   // reading `act` and using that value to feed to `witness`, rather than taking the `foldLeft`
   // approach as below.
-  private[this] val act: Activity[NsMap] =
+  private[this] val act: InstrumentedActivity[NsMap] =
     watchApi.dtabs.activity(dtabListToNsMap) {
```
adleong: We'll also want to create a WatchState, pass it into activity, and serialize it the admin handler.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
```diff
@@ -38,6 +41,8 @@ class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
   val api: NsApi = Api(client.newService(dst)).withNamespace(namespace)
   val watchApi: NsApi = Api(client.newService(dst)).withNamespace(namespace)
 
+  private[this] var instrumentedDtabStorage: mutable.Seq[InstrumentedDtabStorage] = mutable.Seq[InstrumentedDtabStorage]()
```
adleong: I don't think this needs to be a var OR a mutable.Seq.  What if this was just a val and `val act` was defined in terms of this?

```
val instrumentedDtabStorage = {
  val watchState = ...
  val act = watchApi.dtabs.activity...
}
val act = instrumentedDtabStorage.underlying
```

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
```diff
@@ -162,4 +174,34 @@ class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
...
+
+  private[K8sDtabStore] class DtabStoreStateHandler extends Service[Request, Response] {
+    private[this] val mapper = Parser.jsonObjectMapper(Nil)
+    mapper.registerModule(DtabCodec.module)
+
+    override def apply(request: Request): Future[Response] = {
+      val state = instrumentedDtabStorage.map { dtabWatch =>
+        Map(
+          "state" -> dtabWatch.act.stateSnapshot().map {
+            case Some(ns) => ns.collect {
```
adleong: is this a total function?  could we use map instead of collect?

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
dadjeibaah: That's a good point. I can make that change.

namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala
```diff
@@ -52,6 +56,10 @@ class K8sDtabStore(client: Http.Client, dst: String, namespace: String)
     (name(dtab), versionedDtab)
   }
 
+  private[this] case class InstrumentedDtabStorage(
```
adleong: I think this class is now unused.

dadjeibaah: whoops! good catch!

Issue #2030, opened 2018-07-02
Consul namer returns Single IP if node provides a hostname


leozc:
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:
Consul namer returns a single IP address even the hostname in the node is backed by multiple address (e.g. ELB A record)

**What you expected to happen**:
Return a list of IPs of the DNS record and hence it can be used by client side load balancing in later stage

**How to reproduce it (as minimally and precisely as possible)**:
1. consul contains a single return, which the host is a host name => e.g. microsoft.com
2. Namerd points to consul
3. Use Dtab tool to resolve the path.

**Anything else we need to know?**:
IMHO, it is an interesting context, should we ever use host name in the host filed in Consul? DNS is a separate level redirection hierarchy this brings additional implication.
 
**Environment**:
- linkerd/namerd version, config files: latest version
- Platform, version, and config files (Kubernetes, DC/OS, etc): EC2 ony
- Cloud provider or hardware configuration: AWS

leozc:
May be worth to note  InetAddress.getAllByName does return IPv4 and IPv6 address entries for the same host - we may want to add a parameter to indicate which class of address to use.

```
$ export _JAVA_OPTIONS="-Djava.net.preferIPv4Stack=true"
$ scala
Picked up _JAVA_OPTIONS: -Djava.net.preferIPv4Stack=true
Welcome to Scala 2.12.5 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_172).
Type in expressions for evaluation. Or try :help.

scala> import java.net.{InetAddress, InetSocketAddress, UnknownHostException}
import java.net.{InetAddress, InetSocketAddress, UnknownHostException}

scala> InetAddress.getAllByName("api.mixerbox.com").toTraversable.map(singleIP => new InetSocketAddress(singleIP, 443))
res0: Traversable[java.net.InetSocketAddress] = ArrayBuffer(api.mixerbox.com/54.88.94.148:443, api.mixerbox.com/54.152.138.221:443, api.mixerbox.com/107.23.90.88:443)
```

without the IPv4 only Option
```
scala> InetAddress.getAllByName("api.mixerbox.com").toTraversable.map(singleIP => new InetSocketAddress(singleIP, 443))

scala> InetAddress.getAllByName("api.mixerbox.com").toTraversable.map(singleIP => new InetSocketAddress(singleIP, 443))
[...]

PR #2031, opened 2018-07-02
fix consul namer, which returns all IPs intead of one

leozc:
A fix for Linkerd Issue https://github.com/linkerd/linkerd/issues/2030

Essentially instead of return a single IP we return a list of the IPs for the host.

Potential problem: as discussed in issue [2030](https://github.com/linkerd/linkerd/issues/2030), depends on the JVM setup, it is possible to return both IPv4 and IPv6 addresses of a DNS.

I am not sure what the best way to solve this beside adding additional parameters somewhere (possibly on Namer level) to define the DNS resolving policy..

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -7,7 +7,8 @@ import com.twitter.logging.Level
 import com.twitter.util._
 import io.buoyant.consul.v1
 import io.buoyant.namer.{Metadata, InstrumentedVar}
-import java.net.InetSocketAddress
+import java.net.{InetAddress, InetSocketAddress}
+
 import scala.util.control.NoStackTrace
 
 private[consul] case class SvcKey(name: String, tag: Option[String]) {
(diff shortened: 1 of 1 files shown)


leozc:
Before:
screen shot 2018-07-01 at 10 41 45 pm

After:
screen shot 2018-07-01 at 10 23 29 pm

dadjeibaah:
@leozc thanks for submitting this PR. This is a very interesting case. I do agree that having to rely on the JVM’s DNS resolver brings an additional layer of redirection to the consul namer. Maybe we could make this option configurable so that it is explicit for other consul namer users? 

In regards to the behavior of the DNS returning IPv4 vs IPv6, IMO, this should be configurable at the JVM level and not the Namer level. However, I think Linkerd already has the `-Djava.net.preferIPv4Stack` JVM setting set to true in [LinkerdBuild.scala](https://github.com/linkerd/linkerd/blob/master/project/LinkerdBuild.scala#L236). So this might be a non-issue but may require additional testing.

adleong:
Can you add the DCO signoff?  It also looks like CI got stuck on your PR _again_ but I think I've figured out a way to force it to re-run.  I'll do that once you add the DCO signoff.

leozc:
@adleong  DCO done

adleong:
@leozc looks like the bot isn't quite happy: `Expected "Leo Liang <leozc@leozc.com>", but got "leozc <leozc@leozc.com>".`

leozc:
Please try again...

Issue #2033, opened 2018-07-03
Add config param to disable the /admin/shutdown endpoint


robertpanzer:
Issue Type:

- [ ] Bug report
- [x] Feature request

**What happened**:

Currently it is possible to shutdown any Linkerd instance by simple posting to the `/admin/shutdown` endpoint.
While I can imagine that many scenarios appreciate this behavior, it might not always be the case and could lead to a decrease in availability if attackers repeatedly post to this endpoint.

**What you expected to happen**:

It would be nice to have a configuration parameter that allows to disable this endpoint.

**How to reproduce it (as minimally and precisely as possible)**:

Start Linkerd locally and run the following command:

```
$ curl -v -X POST localhost:9990/admin/shutdown
```

Linkerd will shut down asap.

**Anything else we need to know?**:

**Environment**:
- linkerd/namerd version, config files: Latest version, any config that enables the admin server
- Platform, version, and config files (Kubernetes, DC/OS, etc): OS X
- Cloud provider or hardware configuration:

robertpanzer:
I could try to create a PR for this feature in case it is considered worth to be added.
Would it make sense to simply have a parameter `enableShutdown` or so with default `true` that allows to disable this endpoint?

```yaml
admin:
  port: 9990
  ip: 0.0.0.0
  shutdownEnabled: false
```

Or should there be a more general approach that would allow to disable any admin endpoint by defining a blacklist?

```yaml
admin:
  port: 9990
  ip: 0.0.0.0
  endpointBlackList:
  - /admin/shutdown
  - /admin/threads
```

What do you think?

adleong:
Good question, @robertpanzer.  Here's a list of the default admin endpoints:

```
/
/admin/announcer
/admin/contention
/admin/files/
/admin/lint
/admin/lint.json
/admin/metrics.json
/admin/metrics/usage
/admin/ping
/admin/pprof/contention
/admin/pprof/heap
/admin/pprof/profile
/admin/registry.json
/admin/server_info
/admin/shutdown
/admin/threads
/admin/threads.json
/admin/tracing
/bound-names.json
/client_state.json
/config.json
/delegator
/delegator.json
/favicon.ico
/files/
/help
/logging
/logging.json
```

Most of these endpoints are pretty harmful if an attacker gains access to them.  At the very least, it leaks much data about Linkerd and its internal state.  It's recommended that the admin server not be externally accessible for this reason.

[...]

PR #2034, opened 2018-07-03
#2033 Add config param to disable shutdown endpoint

robertpanzer:
This PR implements the first solution suggested in the comment to #2033.
It adds a config param that allows to disable the `/admin/shutdown` endpoint.

With this PR, if I configure Linkerd like this:

```yaml
admin:
  port: 9990
  ip: 0.0.0.0
  shutdownEnabled: false
```

and post to the `/admin/shutdown` endpoint, the request fails with 404:

```
$ curl -v -X POST localhost:9990/admin/shutdown
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 9990 (#0)
> POST /admin/shutdown HTTP/1.1
> Host: localhost:9990
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Content-Length: 0
< 
```

If the second solution is preferred I can also provide a PR for it.

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/Admin.scala
+++ b/admin/src/main/scala/io/buoyant/admin/Admin.scala
@@ -117,7 +117,7 @@ object Admin {
   }
 }
 
-class Admin(val address: InetSocketAddress, tlsCfg: Option[TlsServerConfig], workers: Int, stats: StatsReceiver) {
+class Admin(val address: InetSocketAddress, tlsCfg: Option[TlsServerConfig], workers: Int, stats: StatsReceiver, securityConfig: Option[AdminSecurityConfig]) {
   import Admin._
 
   private[this] val notFoundView = new NotFoundView()
@@ -140,7 +140,11 @@ class Admin(val address: InetSocketAddress, tlsCfg: Option[TlsServerConfig], wor
         log.debug("admin: %s => %s", url, service.getClass.getName)
         muxer.withHandler(url, service)
     }
-    HeadFilter andThen notFoundView andThen muxer
+    val adminService = HeadFilter andThen notFoundView andThen muxer
+    securityConfig match {
+      case Some(cfg) => cfg.mkFilter andThen adminService
+      case None => adminService
+    }
   }
 
   def serve(app: TApp, extHandlers: Seq[Handler]): ListeningServer =
--- a/admin/src/main/scala/io/buoyant/admin/Admin.scala
+++ b/admin/src/main/scala/io/buoyant/admin/Admin.scala
@@ -117,7 +117,13 @@ object Admin {
   }
 }
 
-class Admin(val address: InetSocketAddress, tlsCfg: Option[TlsServerConfig], workers: Int, stats: StatsReceiver, securityConfig: Option[AdminSecurityConfig]) {
+class Admin(
+  val address: InetSocketAddress,
+  tlsCfg: Option[TlsServerConfig],
+  workers: Int,
+  stats: StatsReceiver,
+  securityConfig: Option[AdminSecurityConfig]) {
+  
   import Admin._
 
   private[this] val notFoundView = new NotFoundView()
--- a/admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
+++ b/admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
@@ -29,31 +29,22 @@ object AdminSecurityConfig {
 }
 
 case class AdminSecurityConfig(
-  uiEnabled: Option[Boolean] = Some(true),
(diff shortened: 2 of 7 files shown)


robertpanzer:
As suggested in #2033 I changed the PR to allow explicit whitelisting of any admin URLs.
If the admin config remains unchanged, all URLs are accessible to preserve the current behavior.

But it is also possible to configure the allowed URLs:
```yaml
admin:
  security:
    uiEnabled: true
    pathWhitelist:
    - ^/threads.json$
```

This way all endpoints for the UI are allowed plus the endpoint `/threads.json` is whitelisted explicitly.
If a URL is not on the whitelist, the new `SecurityFilter` will respond with 404.

The default for the UI endpoints is `true` in case the `security` option is configured. (If the `security` option isn't given, everything is accessible anyway)

robertpanzer:
As suggested on #2033 I split the defined multiple whitelists for ui, control and diagnostics and also added a blacklist parameter.

linkerd/docs/config.md
```diff
@@ -144,6 +145,36 @@ This administrative interface was originally based on
...
+    - ^/logging[.]json$
+```
+
+The security configuration supports these parameters allowing fine grained control over the administrative endpoints that are available.
+
+Key | Default Value | Description
+--- | ------------- | -----------
+uiEnabled | true | Configures if all endpoints required to run the administrative ui are available
+controlEnabled | true | Configures if the endpoints to control Linkerd are available, i.e. `/admin/shutdown` and `/logging.json`
+diagnosticsEnabled | true | Configures if all endpoints are available.
```
adleong: As described here, I think setting "diagnosticsEnabled: true" and "controlEnabled: false" would still allow access to the control endpoints, which doesn't seem like the right behavior.  I think diagnostics should be everything else, not included in ui or control.

What do you think about this algorithm:
1. determine which category a request is for: ui, control, or diagnostics
2. if that category is enabled and the path doesn't match the blacklist, the request is allowed
3. if that category is disabled but the path matches the whitelist and not the blacklist, the request is allowed
4. otherwise the request is rejected

admin/src/main/scala/io/buoyant/admin/AdminConfig.scala
```diff
@@ -3,12 +3,15 @@ package io.buoyant.admin
 import com.twitter.finagle.buoyant.TlsServerConfig
 import com.twitter.finagle.stats.StatsReceiver
 import java.net.{InetAddress, InetSocketAddress}
+
```
adleong: style nit: no blank lines in imports, please

linkerd/docs/config.md
robertpanzer: Oh yes, that’s a good point. 
What do you think about adding the ui and control endpoints to the blacklist if the corresponding parameter is set to false?

I am unsure atm how to find out the category.

robertpanzer: I’ll try to implement this tomorrow when I have a fresh mind :)

robertpanzer:
@adleong I think I implemented according to your proposal.
It would be great if you could have another look.

adleong:
@robertpanzer this is looking great!  I have a few suggestions but I know there is some time pressure to get this merged quickly.  In the interest of time, I've pushed a branch with my suggested changes: https://github.com/linkerd/linkerd/compare/alex/disable-shutdown

Please take a look and merge those changes into this PR if you agree.

adleong:
You will also need to add the DCO signoff to all of your commits before we can merge this.

robertpanzer:
Thanks for helping out @adleong! 🎉 
I merged your branch and squashed my commits.

admin/src/main/scala/io/buoyant/admin/Admin.scala
```diff
@@ -117,7 +117,7 @@ object Admin {
   }
 }
 
-class Admin(val address: InetSocketAddress, tlsCfg: Option[TlsServerConfig], workers: Int, stats: StatsReceiver) {
+class Admin(val address: InetSocketAddress, tlsCfg: Option[TlsServerConfig], workers: Int, stats: StatsReceiver, securityConfig: Option[AdminSecurityConfig]) {
```
dadjeibaah: nit: you probably could format the class definition for readability to something like:

```scala
class Admin(
  val address: InetSocketAddress,
  tlsCfg: Option[TlsServerConfig],
  workers: Int, 
  stats: StatsReceiver, 
  securityConfig: Option[AdminSecurityConfig]) {
...
```

robertpanzer: Updated the PR according to your suggestion.

PR #2051, opened 2018-07-09
Improve gRPC cancellation

### LINKED ISSUE #1906: memory (and possibly observations) leak in io.l5.mesh interface

adleong:
Fixes #1906

There is no good method for the consumer of a gRPC stream to signal that they are no longer interested in reading from the stream and that the producer should stop producing.  In particular, for the io.l5d.mesh Namerd interface, this means that when Linkerd receives a gRPC stream from Namerd for a name, it will continue to receive updates for that name forever, even if it becomes no longer observed.  This can result in a memory leak over time.

We update the gRPC stream API to add a `reset` method which the consumer can use to signal cancellation.  We update the gRPC client and server dispatches to connect the gRPC resets to the H2 stream cancellation mechanism added in https://github.com/linkerd/linkerd/pull/1934.  Much of the state management in DecodingStream is no longer necessary since we are simply passing cancellations through to the underlying H2 stream.

We also update the io.l5d.mesh Namerd client to cancel the stream when the corresponding Activity is no longer observed.

This was tested by using the script from #1906 and verifying that memory usage was stable.


grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/DecodingStream.scala
```diff
@@ -30,7 +31,7 @@ private[runtime] trait DecodingStream[T] extends Stream[T] {
    * This state holds a `Releaser`, which manages when HTTP/2 frames
```
dadjeibaah: Is this still considered a state of the stream or is it now just functioning as a buffer? Maybe the comments need to be updated?

adleong: it's still state and it still holds a releaser, even if it's less state than we were storing before.

PR #2054, opened 2018-07-10
Fix NPE during validation

### LINKED ISSUE #2052: Regression: Validate=true
We just upgraded from 1.4.0 to 1.4.4 and it seems there is a regression in the config validation mode.
We did not change anything in the config between 1.4.0 and 1.4.4, and 1.4.4 starts up normal with the config. 
But validate fails: 

```
# /opt/linkerd/linkerd-latest -validate=true /etc/linkerd/production.yaml
-XX:+AggressiveOpts -XX:+AlwaysPreTouch -XX:+CMSClassUnloadingEnabled -XX:CMSInitiatingOccupancyFraction=70 -XX:+CMSParallelRemarkEnabled -XX:+CMSScavengeBeforeRemark -XX:InitialHeapSize=33554432 -XX:MaxHeapSize=1073741824 -XX:MaxNewSize=357916672 -XX:MaxTenuringThreshold=6 -XX:OldPLABSize=16 -XX:+PerfDisableSharedMem -XX:+PrintCommandLineFlags -XX:+ScavengeBeforeFullGC -XX:-TieredCompilation -XX:+UseCMSInitiatingOccupancyOnly -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseConcMarkSweepGC -XX:+UseParNewGC -XX:+UseStringDeduplication
Jul 10, 2018 11:33:01 AM com.twitter.finagle.http.HttpMuxer$ $anonfun$new$1
INFO: HttpMuxer[/admin/metrics.json] = com.twitter.finagle.stats.MetricsExporter(<function1>)
Jul 10, 2018 11:33:01 AM com.twitter.finagle.http.HttpMuxer$ $anonfun$new$1
INFO: HttpMuxer[/admin/per_host_metrics.json] = com.twitter.finagle.stats.HostMetricsExporter(<function1>)
I 0710 11:33:02.052 CEST THREAD1: linkerd 1.4.4 (rev=4fafcadc2eca0c4a2c227150c4cce64ecfed87a7) built at 20180706-161126
I 0710 11:33:02.532 CEST THREAD1: Finagle version 18.5.0 (rev=225a244e96935278721ea40a54aa2c2e53412f21) built at 20180508-105457
java.lang.NullPointerException
	at io.buoyant.linkerd.Main$.defaultCloseGracePeriod(Main.scala:169)
	at com.twitter.app.App.nonExitingMain(App.scala:351)
	at com.twitter.app.App.nonExitingMain$(App.scala:315)
	at io.buoyant.linkerd.Main$.nonExitingMain(Main.scala:20)
	at com.twitter.app.App.main(App.scala:304)
	at com.twitter.app.App.main$(App.scala:302)
	at io.buoyant.linkerd.Main$.main(Main.scala:20)
	at io.buoyant.linkerd.Main.main(Main.scala)
Exception thrown in main on startup

```

How to recreate: 
Try to validate the example config from the [documentation] (https://linkerd.io/config/1.4.2/linkerd/index.html#introduction) (don'T forget to create a dir called 'disco' in your cwd). 
It failes with the same error. 

Environment:
- Linkerd 1.4.4 on Centos 7.5.1804

adleong:
When the `validate` flag is set, Linkerd returns from its main method early and does not set the shutdownGracePeriod.  This leads to a NullPointerException.

Move the setting of shutdownGracePeriod up to before the early return.

Fixes #2052 

### DIFF
--- a/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
+++ b/linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
@@ -36,6 +36,7 @@ object Main extends App {
     args match {
       case Array(path) =>
         val config = loadLinker(path)
+        shutdownGracePeriod = config.admin.flatMap(_.shutdownGraceMs)
         if (validate())
           return
 
@@ -45,7 +46,6 @@ object Main extends App {
         val routers = linker.routers.map(initRouter(_))
 
         log.info("initialized")
-        shutdownGracePeriod = config.admin.flatMap(_.shutdownGraceMs)
         registerTerminationSignalHandler(shutdownGracePeriod)
         closeOnExit(Closable.sequence(
           Closable.all(routers: _*),


linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -36,6 +36,7 @@ object Main extends App {
     args match {
       case Array(path) =>
         val config = loadLinker(path)
+        shutdownGracePeriod = config.admin.flatMap(_.shutdownGraceMs)
```
olix0r: why does this need to be a mutable `var` at all?

adleong: We need to override a def `defaultCloseGracePeriod` with this value.  However, the value is loaded from the config file in the main method.  This means that when main is invoked and the config is read, this value must be stored somewhere so that it can be returned from `defaultCloseGracePeriod`.  Attempting to call `defaultCloseGracePeriod` before main has run will result in an NPE.

PR #2055, opened 2018-07-10
Discard chunked response when running diagnostic tracing

### LINKED ISSUE #2038: l5d-add-context TRACE: ignore server esponse if chunked response is returned
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
I've tried to do an -XTRACE on a server that always responds with chunked messages, expecting a trace output - but I've received HTTP 500.
```
* Rebuilt URL to: http://some-host.ozon.ru/
*   Trying ip-here...
* TCP_NODELAY set
* Connected to some-host.ozon.ru (ip-here) port 80 (#0)
> TRACE / HTTP/1.1
> Host: some-host.ozon.ru
> User-Agent: curl/7.60.0
> Accept: */*
> l5d-add-context: 1
> Content-Type: text/json
>
< HTTP/1.1 502 Bad Gateway
< l5d-err: Cannot+set+content+on+Chunked+message
< Content-Length: 37
< Content-Type: text/plain
<
* Connection #0 to host some-host.ozon.ru left intact
Cannot set content on Chunked message
```

Tracing info should be returned

`curl -v -XTRACE -H 'l5d-add-context: 1' -H 'Host: some-host.ozon.ru' -H 'Content-Type: text/json' 'http://some-host.ozon.ru'`
where `some-host.ozon.ru` returns chunked response.

I understand that the response of server is forwarded by proxy in case there's another linkerd proxy behind the resolved IP - but it's a good idea to discard a response if it's chunked and return what proxy had resolved up to this point instead of crashing to HTTP 500.

**Environment**:
- linkerd/namerd version, config files: v1.4.2

dadjeibaah:
When running a diagnostic trace to a server that responds with chunked messages, Linkerd returns an HTTP 500 Bad Gateway response because the `DiagnosticTracer` filter is unable to handle chunked responses. 

This PR fixes the issue by discarded the contents of the chunked response and sending a place holder message like:

```bash
Transfer-Encoding: chunked
HTTP 200 - OK
Diagnostic trace encountered chunked response. Response content discarded.
``` 

The message prints the response headers and status code just in case some additional contextual information is intended to be propagated by any chunked response server present along the trace path.

Integration tests were added to as well local tests were done on to vet this PR.

fixes #2038 

### DIFF
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/DiagnosticTracer.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/DiagnosticTracer.scala
@@ -58,8 +58,8 @@ class DiagnosticTracer(
 
     routerCtxF.map { ctx =>
       if (resp.isChunked) {
-        resp.setChunked(false)
         resp.reader.discard()
+        resp.setChunked(false)
         resp.contentString = ChunkedResponsePlaceholderMsg(resp.headerMap, resp.status)
       }
       val content = s"${resp.contentString.trim}\n\n$ctx"
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/DiagnosticTracer.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/DiagnosticTracer.scala
@@ -35,6 +35,15 @@ class DiagnosticTracer(
   routerLabel: String
 ) extends SimpleFilter[Request, Response] {
   private[this] val AddRouterContextHeader = "l5d-add-context"
+
+  private[this] def ChunkedResponsePlaceholderMsg(headers: HeaderMap, status: Status) = {
+    val headerStr = headers.map {
+      case (headerKey, headerValue) => s"$headerKey: $headerValue"
+    }.mkString("\n")
+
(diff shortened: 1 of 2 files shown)


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/http/DiagnosticTracer.scala
```diff
@@ -48,6 +57,10 @@ class DiagnosticTracer(
     )
 
     routerCtxF.map { ctx =>
+      if (resp.isChunked) {
+        resp.setChunked(false)
```
adleong: do you also need to do `resp.reader.discard()`?

dadjeibaah: Yea, you are right! I got mixed up with the description of `resp.setChunked(...)`

```scala
/**
   * Manipulate the `Message` content mode.
   *
   * If `chunked` is `true`, any existing content will be discarded and further attempts
   * to manipulate the synchronous content will result in an `IllegalStateException`.
   *
   * If `chunked` is `false`, the synchronous content methods will become available
   * and the `Reader`/`Writer` of the message will be ignored by Finagle.
   */

I will add that in.
```

adleong: based on that comment, does the reader.discard need to happen before setting chunked to false.  Otherwise the reader.discard will be ignored by finagle?

PR #2056, opened 2018-07-11
Pass router param inputs to identifiers

### LINKED ISSUE #1963: Enable plugins to emit stat into L5D stat trees for Identifier

adleong:
Stack params are passed to identifier plugins to configure them.  However, only params from the router config are passed to the identifier.  In particular, this means that the stats receiver is not available to the identifier because it is not a router config param.

Pass stack params as input to the router params.  This allows the stats receiver to be passed into the router params.

Fixes #1963 

### DIFF
--- a/linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
+++ b/linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
@@ -1,12 +1,14 @@
 package io.buoyant.linkerd.protocol.http
 
+import com.twitter.finagle.buoyant.Dst
 import com.twitter.finagle.http.{Method, Request}
 import com.twitter.finagle.{Dtab, Path, Stack}
+import com.twitter.util.Future
 import io.buoyant.config.Parser
-import io.buoyant.linkerd.RouterConfig
-import io.buoyant.linkerd.protocol.{HttpConfig, HttpInitializer}
+import io.buoyant.linkerd.{IdentifierInitializer, RouterConfig}
+import io.buoyant.linkerd.protocol.{HttpConfig, HttpIdentifierConfig, HttpInitializer}
 import io.buoyant.router.Http
-import io.buoyant.router.RoutingFactory.IdentifiedRequest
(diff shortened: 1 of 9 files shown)


linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
```diff
@@ -113,7 +113,7 @@ class HttpConfigTest extends FunSuite with Awaits {
                   |- port: 5000
       """.stripMargin
     val config = parse(yaml)
-    val identifier = config.routerParams[Http.param.HttpIdentifier]
+    val identifier = config.routerParams(Stack.Params.empty)[Http.param.HttpIdentifier]
       .id(Path.read("/svc"), () => Dtab.empty)
     val req = Request(Method.Get, "/one/two/three")
     assert(
```
dadjeibaah: Should we add a test that validates that we are receiving a `StatsReceiver` in an identifier to show that it propagates through initialization?

PR #2071, opened 2018-07-16
Zipkin trace propagation plugin

jmalvarezf:
Hi,
This is the first implementation of the interface that you provided to use Zipkin B3 headers to continue an existing trace context in case it exists. If not, fallbacks to the default behaviour. 
As I told you in the other PR, I'm not a scala expert, so please bear with me and feel free to make any amends you find are useful.

I've tested the HTTP implementation. We do not have an H2 env where to test the H2 Tracer, but I provide it for completeness.

Thank you,

Jose Maria.

### DIFF
--- a/admin/src/main/scala/io/buoyant/admin/Admin.scala
+++ b/admin/src/main/scala/io/buoyant/admin/Admin.scala
@@ -122,8 +122,9 @@ class Admin(
   tlsCfg: Option[TlsServerConfig],
   workers: Int,
   stats: StatsReceiver,
-  securityConfig: Option[AdminSecurityConfig]) {
-  
+  securityConfig: Option[AdminSecurityConfig]
+) {
+
   import Admin._
 
   private[this] val notFoundView = new NotFoundView()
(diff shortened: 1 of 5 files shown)


adleong:
Thanks, @jmalvarezf!  This looks great, I'll try to test it today.

One quick question: I notice that this propagator looks for the x-b3 headers but if they're not present it falls back to reading the l5d-ctx headers.  Is this fallback behavior actually desirable?  At first glance it seems like this might make things more confusing if Linkerd could potentially be loading its trace context from one of two different places.

jmalvarezf:
Hi @adleong,

Yes, that was the intention. If you think this is more confusing than useful, I can redo that part to only propagate via Zipkin headers.

Thank you,

jmalvarezf:
I have also a commit that would address a problem with some tracing libraries that need all b3 headers (there are two missing when forwarding the request) so if you confirm that you don't want that behavior I would address both.

Thanks!

jmalvarezf:
Hi @adleong,

Thanks for your comments!! I've added them all, along with the other corrections (no fallback behaviour, and b3 headers correction).

Best regards,

Jose Maria.

jmalvarezf:
Hi @adleong,

Thank you for your patience! I've edited the commits with the sign-off. Is that enough?

Thank you,

jmalvarezf:
Ok, now it is properly done. 

Thanks!

jmalvarezf:
We discovered a bug in the flags/sampled headers. The code now includes the patch.

adleong:
Thank you for this great contribution, @jmalvarezf!

PR #2084, opened 2018-07-25
Add instrumentation for fs namer

### LINKED ISSUE #1912: Admin endpoint that exposes internal fs namer state
Linkerd filesystem namer should expose an admin endpoint which provides data about the internal state of that namer including:

adleong:
It can be difficult to determine the cause of the problem when service discovery information is out of date.  

For the io.l5d.fs namer, we add instrumentation so that we can see the current state of the file system watches and the most recent events received from the OS.  This state can be viewed at `/namer_state/io.l5d.fs.json` in the admin server.

screen shot 2018-07-25 at 1 42 22 pm

Fixes #1912 


adleong:
It's possible but the idea is to expose the underlying state as accurately as possible.  The file contents are just raw strings so I think that's how we should display them.

PR #2085, opened 2018-07-25
Always clear the x-forwarded-client-cert header

### LINKED ISSUE #2068: x-forwarded-client-cert in external request isn't (always) filtered out
- [x] Bug report

**What happened**:

linkerd doesn't remove the `x-forwarded-client-cert` header field from inbound requests.

When linkerd receives a `x-forwarded-client-cert` header field in a request, it must remove it or it must reject the request. It must never forward the `x-forwarded-client-cert` header field because that would allow any client to forge the identity of other clients if the inner application is relying on the `x-forwarded-client-cert` header field. (Also some applications may look at the `x-forwarded-client-cert` to determine if the request was sent over HTTPS or HTTP.)

This could be a security-sensitive issue for any application relying on the `x-forwarded-client-cert` being accurate.

This filtering probably must happen regardless of the setting that enables linkerd's injection of the header field.

There may be other header fields that need to be always replaced/filtered in a similar way.

adleong:
When Linkerd receives a x-forwarded-client-cert header field in a request, it must remove it or it must reject the request. It must never forward the x-forwarded-client-cert header field because that would allow any client to forge the identity of other clients if the inner application is relying on the x-forwarded-client-cert header field. (Also some applications may look at the x-forwarded-client-cert to determine if the request was sent over HTTPS or HTTP.)

Always clear any incoming x-forwarded-client-cert header.

Fixes #2068 

### DIFF
--- a/linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/ForwardClientCertTest.scala
+++ b/linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/ForwardClientCertTest.scala
@@ -102,7 +102,8 @@ class ForwardClientCertTest extends FunSuite {
       val rsp = {
         val req = Request()
         req.host = "clifford"
-        req.headerMap(ForwardClientCertFilter.Header) = "totally_bogus"
+        req.headerMap.add(ForwardClientCertFilter.Header, "Hash=17595BEB34A925D9D5D74E581A47883B6969DF6102227ED4770F5121458684EF;SAN=https://buoyant.io;DNS=upstream;DNS=linkerd;Subject=\"C=US,CN=upstream\"")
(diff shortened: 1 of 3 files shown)


linkerd/protocol/http/src/integration/scala/io/buoyant/linkerd/protocol/ForwardClientCertTest.scala
```diff
@@ -69,4 +70,49 @@ class ForwardClientCertTest extends FunSuite {
...
+    val router = linker.routers.head.initialize()
+    val server = router.servers.head.serve()
+
+    val client = Upstream.mk(server)
+
+    try {
+      val rsp = {
+        val req = Request()
+        req.host = "clifford"
+        req.headerMap(ForwardClientCertFilter.Header) = "totally_bogus"
```
briansmith: It would be good to have a test where the value is syntactically valid to ensure that it is removed even when it looks good, and not only when it is "totally bugus."

briansmith: It would also be good to have a test that shows that if there are multiple instances of the header field that all of them are removed.

PR #2091, opened 2018-08-02
Add an io.l5d.destination interface to support Linkerd2

dadjeibaah:
There is no set way for Linkerd2 proxies to use Namerd service discovery in a Kubernetes daemonset environment. Linkerd 2 uses its control plane's gRPC destination API, while Linkerd 1.x uses Namerd for its service discovery and control plane. To enable a Linkerd2-proxy to work alongside Namerd and Linkerd 1.x, Namerd needs to implement the gRPC destination API and use its namers to provide service discovery information to an l5d2 proxy.

This PR adds a new Namerd `io.l5d.destination` interface that implements the API and resolves services names using configured namers. When a destination request reaches the interface, a gRPC stream is created and Namerd creates an Activity that interprets a given destination path into a set of addresses. When a stream is created, the interface immediately sends a `NoEndpoints` message followed by an `Add` message if there are addresses available for a service name. As addresses are discovered and/or removed, the interface calculates the delta between previously discovered and the current address and sends `Add` and `Remove` gRPC messages back to the client. If there is no address change, i.e., if there is nothing to add or remove, we assume that the service associated with that service exists but does not have endpoints available, and the interface sends a `NoEndpoints` message with `exists: true`.

The interface has been tested with `controller/script/destination-client` in the [`linkerd2`](https://github.com/linkerd/linkerd2/tree/master/controller/script/destination-client) repo. 

One major caveat with the `io.l5d.destination` is that does not provide reach resource labels that Linkerd2 uses for metrics. This means that metrics generated by a proxy backed Namerd could potentially be sparse and more not give in-depth request telemetry when compared to Linkerd2's destination API.

### DIFF
--- a/namerd/iface/destination/src/main/resources/pull-destination-proto.sh
+++ b/namerd/iface/destination/src/main/resources/pull-destination-proto.sh
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
-cd ../protobuf
+cd ../protobuf || exit
+DESTINATION_PROTO=$(curl https://raw.githubusercontent.com/linkerd/linkerd2-proxy-api/master/proto/destination.proto)
+NET_PROTO=$(curl https://raw.githubusercontent.com/linkerd/linkerd2-proxy-api/master/proto/net.proto)
+
 echo "Copying destination.proto from linkerd2 repo"
-printf "//This file is added using pull-destination-proto.sh. DO NOT EDIT!\n$(curl https://raw.githubusercontent.com/linkerd/linkerd2-proxy-api/master/proto/destination.proto)" > destination.proto
+printf "//This file is added using pull-destination-proto.sh. DO NOT EDIT!\\n%s" "$DESTINATION_PROTO" > destination.proto
 
 
 echo "Copying net.proto from linkerd2 repo"
-printf "//This file is added using pull-destination-proto.sh. DO NOT EDIT!\n$(curl https://raw.githubusercontent.com/linkerd/linkerd2-proxy-api/master/proto/net.proto)" > net.proto
+printf "//This file is added using pull-destination-proto.sh. DO NOT EDIT!\\n%s" "$NET_PROTO" > net.proto
 
--- a/namerd/iface/destination/src/main/resources/pull-destination-proto.sh
+++ b/namerd/iface/destination/src/main/resources/pull-destination-proto.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+cd ../protobuf
+echo "Copying destination.proto from linkerd2 repo"
+printf "//This file is added using pull-destination-proto.sh. DO NOT EDIT!\n$(curl https://raw.githubusercontent.com/linkerd/linkerd2-proxy-api/master/proto/destination.proto)" > destination.proto
+
+
+echo "Copying net.proto from linkerd2 repo"
+printf "//This file is added using pull-destination-proto.sh. DO NOT EDIT!\n$(curl https://raw.githubusercontent.com/linkerd/linkerd2-proxy-api/master/proto/net.proto)" > net.proto
+
--- a/namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
+++ b/namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
@@ -1,10 +1,10 @@
 package io.buoyant.namerd.iface.destination
 
 import com.twitter.finagle.naming.NameInterpreter
-import com.twitter.finagle.{Address, _}
+import com.twitter.finagle._
 import com.twitter.io.Buf
 import com.twitter.logging.Logger
-import com.twitter.util.{Event, _}
+import com.twitter.util._
 import io.buoyant.grpc.runtime.VarEventStream.{Ev, Val}
 import io.buoyant.grpc.runtime.{GrpcStatus, Stream, VarEventStream}
 import io.linkerd.proxy.destination.Destination
@@ -22,40 +22,38 @@ class DestinationService(
 ) extends Destination {
   import DestinationService._
 
-  //Helper trait to capture Failed Addresses
-  sealed trait FailedAddress extends Set[Address]
-  case class FAddress(e: Throwable) extends FailedAddress {
-    override def contains(elem: Address): Boolean = false
-
-    override def +(elem: Address): Set[Address] = Set.empty
-
-    override def -(elem: Address): Set[Address] = Set.empty
-
-    override def iterator: Iterator[Address] = Iterator.empty
-  }
+  type TrySet[T] = Try[Set[T]]
 
   //Diff implementation for Address Set
-  class AddressDiffable extends Diffable[Set] {
+  class SetDiffable extends Diffable[TrySet] {
     override def diff[T](
-      left: Set[T],
-      right: Set[T]
-    ): Diff[Set, T] =
+      left: TrySet[T],
+      right: TrySet[T]
+    ): Diff[TrySet, T] =
       (left, right) match {
-        case (FAddress(e), _) => AddressDiff(Set.empty, Set.empty, Some(e))
-        case (_, FAddress(e)) => AddressDiff(Set.empty, Set.empty, Some(e))
-        case _ => AddressDiff(right -- left, left -- right)
+        case (Throw(e), _) => FailedDiff(e)
+        case (_, Throw(e)) => FailedDiff(e)
+        case (Return(l), Return(r)) => SetDiff(r -- l, l -- r)
       }
 
-    override def empty[T]: Set[T] = Set.empty
+    override def empty[T]: TrySet[T] = Return(Set.empty)
   }
 
-  case class AddressDiff[T](add: Set[T], remove: Set[T], exc: Option[Throwable] = None) extends Diff[Set, T] {
-    override def patch(coll: Set[T]): Set[T] = coll ++ add -- remove
+  case class FailedDiff[T](throwable: Throwable) extends Diff[TrySet, T] {
+    override def patch(coll: TrySet[T]): TrySet[T] = Throw(throwable)
 
-    override def map[U](f: T => U): Diff[Set, U] = AddressDiff(add.map(f), remove.map(f))
+    override def map[U](f: T => U): Diff[TrySet, U] = FailedDiff(throwable)
+  }
+  case class SetDiff[T](add: Set[T], remove: Set[T]) extends Diff[TrySet, T] {
+    override def patch(coll: TrySet[T]): TrySet[T] = coll match {
+      case t@Throw(_) => t
+      case Return(set) => Return(set ++ add -- remove)
+    }
+
+    override def map[U](f: T => U): Diff[TrySet, U] = SetDiff(add.map(f), remove.map(f))
   }
 
-  implicit val _addressDiffable = new AddressDiffable
+  implicit val _addressDiffable = new SetDiffable
 
   private[this] val log = Logger.get(getClass.getName)
   override def get(req: GetDestination): Stream[Update] = {
@@ -71,12 +69,18 @@ class DestinationService(
         val res = Stream.mk[Update]
         res.close(GrpcStatus.Unimplemented("Empty request path"))
         return res
+      /*
+      * Linkerd 2 proxies sends request paths in kubeDNS format.
+      * e.g. service.default.cluster.local. Linkerd 1.x uses the format /svc/service_name.
+      * This line gets the service name before the first dot in the request path destination request
+      * path and uses that first string value against the configured Name interpreter. This assumes that your
+      * k8s namer already handles how to reach endpoints in specific namespaces. */
       case Some(p) => Buf.Utf8(p.substring(0, p.indexOf(".")))
     }
 
     val name = Path(Seq(Buf.Utf8(pathPfx), path): _*)
     val events = interpreter.bind(Dtab.empty, name).map(_.eval).flatMap {
-      case None => Activity.value(Var.value(Addr.Bound()))
+      case None => Activity.value(Var.value(Addr.Neg))
       case Some(boundSet) =>
         val set = Var.collect(boundSet.map(_.addr).toList).map(foldAddr)
         Activity.value(set)
@@ -85,19 +89,19 @@ class DestinationService(
       case Activity.Pending => Var(Addr.Pending)
       case Activity.Ok(t) => t
     }.changes.collect {
-      case Addr.Failed(e) => FAddress(e)
-      case Addr.Bound(addresses, _) => addresses
+      case Addr.Failed(e) => Throw(e)
+      case Addr.Neg => Throw(new Throwable(s"endpoints for service ${name.show} doesn't exist"))
+      case Addr.Bound(addresses, _) => Return(addresses)
     }
 
-    val updateEvents = eventFlatMap[Diff[Set, Address], Ev[Update]](events.diff, {
-      case AddressDiff(add, remove, exc) =>
-        val noEndpointsUpdate = exc match {
-          case Some(_) => Seq(Val(mkNoEndpointsUpdate(false)))
-          case None => Nil
-        }
+    val updateEvents = eventFlatMap[Diff[TrySet, Address], Ev[Update]](events.diff, {
+      case FailedDiff(e) =>
+        log.debug(s"address set encountered FailedDiff: $e")
+        Seq(Val(mkNoEndpointsUpdate(false)))
+      case SetDiff(add, remove) =>
         val addUpdate = if (add.nonEmpty) Seq(Val(mkAddUpdate(add))) else Nil
         val removeUpdate = if (remove.nonEmpty) Seq(Val(mkRemoveUpdate(remove))) else Nil
-        noEndpointsUpdate ++ addUpdate ++ removeUpdate
+        addUpdate ++ removeUpdate
     })
     val totalUpdates = prependEventOnCondition[Ev[Update]](updateEvents, Val(mkNoEndpointsUpdate(true)),
       {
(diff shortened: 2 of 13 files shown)


namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+              if (add.isEmpty && remove.isEmpty) stream.send(mkNoEndpointsUpdate(true)); ()
+          }
+      }
+    }
+    stream
+  }
+
+  private[this] def foldAddr(addrs: List[Addr]): Addr = {
+    addrs.fold(Addr.Pending) {
+      case (Addr.Pending, Addr.Pending) => Addr.Pending
```
adleong: Consider something like:

```
// non-pending states have priority
case (Addr.Pending, x) => x 
case (x, Addr.Pending) => x
// combine bound
case (Addr.Bound(prev, _), Addr.Bound(cur, _)) => ...
// Bound has priority over other states
case (x@Addr.Bound(_, _), _) => x
case (_, x@Addr.Bound(_, _)) => x
// otherwise, just use the first state (e.g. in a list of negs or fails)
case (x, _) => x
```

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+      case (Addr.Pending, Addr.Neg) => Addr.Neg
+      case (Addr.Pending, Addr.Bound(addresses, _)) => Addr.Bound(addresses)
+      case (Addr.Bound(prevAddr, _), Addr.Bound(curAddr, _)) => Addr.Bound(prevAddr ++ curAddr)
+      case _ => Addr.Bound(Set.empty[Address])
+    }
+  }
+
+  private[this] def updates(
+    name: Path,
```
adleong: is name unused?

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+  override def get(req: GetDestination): Stream[Update] = {
+    log.info(s"stream initiated from request: $req")
+    if (!req.scheme.contains("k8s")) {
+      val res = Stream.mk[Update]
+      res.close(GrpcStatus.Unimplemented("Unknown scheme: " + req.scheme.getOrElse("unknown")))
+      return res
+    }
+
+    // This assumes that our dtab begins with the /svc prefix
+    val name = Path(Seq(Buf.Utf8("svc")) ++ req.path.map(Buf.Utf8(_)): _*)
```
adleong: could make this configurable if you wanted

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+
+    val stream = Stream.mk[Update]
+    interpreter.bind(Dtab.empty, name).run.changes.respond {
```
adleong: using respond here and below is a problem because respond creates an observation that we never close.  The really nice thing about VarEventStream is that it handles creating and closing observations properly.  I'd highly recommend trying to massage this into a final Var[Update] using combinators like map and then passing that to VarEventStream than trying to use respond and managing the observations yourself.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+      case Activity.Pending => // Immediately send no endpoints when stream starts successfully
```
adleong: I don't think this is right, if the Activity is pending we should just not send any update.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+        stream.send(mkNoEndpointsUpdate(false)); ()
+      case Activity.Failed(e) => // Send no endpoints message when we fail to get endpoints
+        log.info(s"address lookup failed: $e")
+        stream.send(mkNoEndpointsUpdate(false)); ()
```
adleong: In this case we probably want to stream.close with some error code

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+      case Activity.Ok(t) => t.eval match {
+        case None => // Service exists but cannot find addresses for service.
```
adleong: I think this is the case that will get if we have NameTree.Neg.  So really this means that the service does not exist.  NoEndpoints(false) should be sent here, I think.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+          stream.send(mkNoEndpointsUpdate(true)); ()
+        case Some(value) => // Evaluate what diff of addresses and send add, remove and update requests.
+          val _ = updates(name, value).diff.respond {
+            case AddressDiff(add, remove) =>
+              if (add.nonEmpty) stream.send(mkAddUpdate(add)); ()
```
adleong: This is a wrinkle when trying to convert Event[AddressDiff] to Event[Update] because one AddressDiff can turn into potentially two Updates.  There's no built-in Event combinator that helps us with this, but we can write our own:

```
def eventFlatMap[T, U](ev: Event[T], f: T => Seq[U]): Event[U] = new Event[U] {
  def register(w: Witness[U]): Closable = {
    ev.respond { t => 
      for(u <- f(t)) w.notify(u)
    }
  }
}
```

and then we can do:

```
val diffs: Event[AddressDiff] = ...
val updates = eventFlatMap(diffs, { case AddressDiff(add, remove) => 
  val addUpdate = if (add.nonEmpty) Seq(mkAddUpdate(add)) else Nil
  val removeUpdate = if (remove.nonEmpty) Seq(mkRemoveUpdate(remove)) else Nil
  addUpdate ++ removeUpdate
}
```

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+              if (remove.nonEmpty) stream.send(mkRemoveUpdate(remove)); ()
+
+              /* If we don't have endpoints to add and remove that means we did not have addresses
```
adleong: I don't think this is right.  if there are no adds and no removes, then that means "no change".  in other words, no update.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,150 @@
...
+                 To begin with, therefore the service exists but we do not have address to update.
+                 Send a no endpoints message with exists set to true
+               */
+
+              if (add.isEmpty && remove.isEmpty) stream.send(mkNoEndpointsUpdate(true)); ()
+          }
+      }
+    }
+    stream
```
adleong: I think overall the pipeline will look something like:

```
Activity[NameTree[Name.Bound]]
Activity[Option[Set[Name.Bound]] //eval
Activity[Var[Set[Addr]]] // Var.collect
Activity[Var[Addr.Bound]] // foldAddr
Var[Addr] // flatMap  (e.g. Activity.Pending -> Var(Addr.Pending))
Event[Addr] // changes
Event[Addr.Bound] // collect, dropping types other than Bound
Event[Set[Address]] // map
Event[Diff] // diff
Event[Update]  // eventFlatMap
```

One wrinkle that you'll run into is how to prepend a NoEndpoints(true) before the first Add in the event stream.  A custom combinator could do the trick:

```
def prependEvent[T](ev: Event[T], init: T): Event[T] = new Event[T] = {
  def register(w: Witness[T]): Closable = {
    w.notify(init)
    ev.respond(w.notify)
  }
}
```

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+import java.net.{InetAddress, InetSocketAddress}
+import java.util.concurrent.atomic.AtomicBoolean
+
+class DestinationService(
+  pathPfx: String,
+  interpreter: NameInterpreter
+) extends Destination {
+  import DestinationService._
+
+  //Helper trait to capture Failed Addresses
```
adleong: I think this can be cleaned up if you define `type TrySet[T] = Try[Set[T]]` and turn your Var[Addr] into an Event[TrySet[Address]].  This avoids the need for the FAddress class.

You would then define a Diffable[TrySet] which pattern matches on the left and right TrySets.  If both are return, it returns a SetDiff as usual.  If either of them are Throw then it returns a new class called Failed[T] or something like that which extends Diff[TrySet, T].

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+
+    override def +(elem: Address): Set[Address] = Set.empty
+
+    override def -(elem: Address): Set[Address] = Set.empty
+
+    override def iterator: Iterator[Address] = Iterator.empty
+  }
+
+  //Diff implementation for Address Set
+  class AddressDiffable extends Diffable[Set] {
```
adleong: Should this just be called SetDiffable?

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+      (left, right) match {
+        case (FAddress(e), _) => AddressDiff(Set.empty, Set.empty, Some(e))
+        case (_, FAddress(e)) => AddressDiff(Set.empty, Set.empty, Some(e))
+        case _ => AddressDiff(right -- left, left -- right)
+      }
+
+    override def empty[T]: Set[T] = Set.empty
+  }
+
+  case class AddressDiff[T](add: Set[T], remove: Set[T], exc: Option[Throwable] = None) extends Diff[Set, T] {
```
adleong: SetDiff?

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+      res.close(GrpcStatus.Unimplemented("Unknown scheme: " + req.scheme.getOrElse("unknown")))
+      return res
+    }
+
+    val path = req.path match {
+      case None =>
+        val res = Stream.mk[Update]
+        res.close(GrpcStatus.Unimplemented("Empty request path"))
+        return res
+      case Some(p) => Buf.Utf8(p.substring(0, p.indexOf(".")))
```
adleong: this line should have a comment explaining what we're doing here and why.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    val name = Path(Seq(Buf.Utf8(pathPfx), path): _*)
+    val events = interpreter.bind(Dtab.empty, name).map(_.eval).flatMap {
+      case None => Activity.value(Var.value(Addr.Bound()))
```
adleong: Should this be Addr.Neg?

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+      })
+    VarEventStream(totalUpdates)
+  }
+
+  private[this] def foldAddr(addrs: List[Addr]): Addr = {
+    addrs.fold(Addr.Pending) {
+      // nonPending take priority
+      case (Addr.Pending, nonPending) => nonPending
+      case (nonPending, Addr.Pending) => nonPending
+      // Bound addresses combined
```
adleong: I think you're missing the case where both addrs are bound and need to be combined.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+    def register(w: Witness[U]): Closable = {
+      ev.respond { t =>
+        for (u <- f(t)) w.notify(u)
+      }
+    }
+  }
+
+  //Prepends an init event to an already existing event only once.
+  private[destination] def prependEventOnCondition[T](ev: Event[T], init: T, f: T => Boolean): Event[T] = new Event[T] {
+    @volatile var prependOnce: AtomicBoolean = new AtomicBoolean(false)
```
adleong: dont want a var of AtomicBoolean.  a val of AtomicBoolean is mutable using its mutating methods.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+    def register(w: Witness[T]): Closable = {
+      ev.respond { t =>
+        if (!prependOnce.getAndSet(true) && f(t)) {
```
adleong: won't this set the AB even if f(t) is false?

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
+package io.buoyant.namerd.iface.destination
+
+import com.twitter.finagle.naming.NameInterpreter
+import com.twitter.finagle.{Address, _}
```
adleong: wildcard import should already cover Address, I think?

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,190 @@
...
+import com.twitter.io.Buf
+import com.twitter.logging.Logger
+import com.twitter.util.{Event, _}
```
adleong: wildcard import should already cover Event, I think?

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
dadjeibaah: Good point, my initial thinking was that this needed to be a _happens once_ event but I can see that calling the `.getAndSet` method is too early and doesn't align with the description of the method.

namerd/iface/destination/src/main/resources/pull-destination-proto.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
adleong: I'm no bash expert but https://www.shellcheck.net/ gives a few warnings about this script

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
```diff
@@ -0,0 +1,196 @@
...
+      }
+    }
+  }
+
+  //Prepends an init event to an already existing event only once.
+  private[destination] def prependEventOnCondition[T](ev: Event[T], init: T, f: T => Boolean): Event[T] = new Event[T] {
+    val prependOnce: AtomicBoolean = new AtomicBoolean(false)
+    def register(w: Witness[T]): Closable = {
+      ev.respond { t =>
+        if (f(t) && !prependOnce.get) {
```
adleong: this can allow multiple threads in before the AB is set to true.  Consider something like:

```
// if this is already set, we can skip ahead
if (!prependOnce.get) { 
  // check f(t) first and if it returns true then do CAS
  if (f(t) && prependOnce.compareAndSet(false, true)) {
    w.notify(init)
  }
}
w.notify(t)
```

namerd/iface/destination/src/main/resources/pull-destination-proto.sh
hawkw: (Sorry for the drive-by review, but) I kind of wonder if this is something that we can do from sbt rather than with a bash script?

dadjeibaah: @hawkw. I tried adding this LinkerdBuild.scala, but a bash script was easier. I can continue researching on how to do this through sbt and create a separate PR.

namerd/iface/destination/src/main/scala/io/buoyant/namerd/iface/destination/DestinationService.scala
dadjeibaah: Thanks, I think I ran into this while running my tests on k8s but couldn't figure out what was causing it. This change fixed it.

namerd/iface/destination/src/main/resources/pull-destination-proto.sh
adleong: it doesn't seem to me like figuring out how to add this to sbt is a good use of time

PR #2099, opened 2018-08-11
Exclude Marathon tasks in health-check grace period (#2098)

### LINKED ISSUE #2098: Marathon app instances receive traffic during health check grace period
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
I have linkerd configured with the `useHealthCheck` flag set to `true`.
I scaled up one of the apps in my DC/OS cluster. Its health-check grace period is set to 60 seconds, this is because it takes quite long for this app to start up. Unfortunately, the `useHealthCheck` switch only affects apps that have a failed health check, not those that still have an unknown state, so Linkerd routed requests to the new instance of my app, even though it wasn't ready.

No requests should be routed to a service until Marathon marks it as healthy. Instances in a health check grace period should not receive any requests.

Configure a marathon namer with `useHealthCheck:true`. Run a Marathon app with a long health-check grace period. Keep sending requests to the app via linkerd. Observe what happens when you scale up such app.

**Environment**:
- linkerd/namerd version, config files: linkerd 1.3.1
- Platform, version, and config files (Kubernetes, DC/OS, etc): DC/OS Enterprise 1.11

rogoman:
This fixes #2098. The Marathon namer has a `useHealthCheck` option that allows for excluding tasks that Marathon claims to be unhealthy. However, if a Marathon app has a health-check grace period defined, a freshly started task will show up as running, but without any health-related status.
In this case the Marathon API returns a `TASK_RUNNING` status, but there aren't any `healthCheckResults` present in the response. Obviously, Linkerd should not route any request to such task and should wait until it shows up as healthy.

My change adds a check if a Marathon app has any health checks defined (a non-empty `healthChecks` attribute of the `app` record), and, if so, expect the same number of `healthCheckResults` to be positive before routing any request to such task.

I have added unit tests that include the `healthChecks` attribute in the Marathon API response and verify that the Marathon namer behavior is as expected.

Signed-off-by: Tomasz Rogozik <trogozik@applause.com>

### DIFF
--- a/marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
+++ b/marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
@@ -44,6 +44,8 @@ object Api {
     state: Option[String]
   )
 
+  final case class HealthCheck()
+
   final case class TaskIpAddress(
     ipAddress: Option[String]
   )
@@ -64,6 +66,7 @@ object Api {
   final case class App(
     id: Option[String],
     ipAddress: Option[AppIpAddress],
+    healthChecks: Option[Seq[HealthCheck]],
     tasks: Option[Seq[Task]]
   )
 
@@ -73,6 +76,11 @@ object Api {
 
   final case class AppRsp(
     app: Option[App] = None
+  )
+
+  final case class TaskWithHealthCheckInfo(
+    task: Task,
+    numberOfHealthChecks: Int
   )
 
   object discoveryPort {
@@ -92,9 +100,9 @@ object Api {
   }
 
   object healthyHostPort {
-    def unapply(task: Task): Option[(String, Int)] = {
-      if (isHealthy(task)) {
-        hostPort.unapply(task)
+    def unapply(extendedTask: TaskWithHealthCheckInfo): Option[(String, Int)] = {
+      if (isHealthy(extendedTask)) {
+        hostPort.unapply(extendedTask.task)
       } else {
         None
       }
@@ -108,13 +116,21 @@ object Api {
   }
 
   object healthyIpAddress {
-    def unapply(task: Task): Option[String] = {
-      if (isHealthy(task)) {
-        ipAddress.unapply(task)
+    def unapply(extendedTask: TaskWithHealthCheckInfo): Option[String] = {
+      if (isHealthy(extendedTask)) {
+        ipAddress.unapply(extendedTask.task)
       } else {
         None
       }
     }
+  }
+
+  object healthCheckCount {
+    def unapply(healthChecks: Option[Seq[HealthCheck]]): Option[Int] =
+      healthChecks match {
+        case Some(healthCheckSeq) => Some(healthCheckSeq.size)
+        case None => Some(0)
+      }
   }
 
   def apply(client: Client, uriPrefix: String, useHealthCheck: Boolean): Api =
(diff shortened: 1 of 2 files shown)


rogoman:
Please note that I only learned the basics of Scala for the purpose of this PR, so any suggestions to my coding style or hints on how to implement things easier are highly appreciated.

dadjeibaah:
Thanks for submitting this @rogoman! We will take a look and recommend any suggestions we have.

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -108,15 +116,23 @@ object Api {
...
+      if (isHealthy(extendedTask)) {
+        ipAddress.unapply(extendedTask.task)
       } else {
         None
       }
     }
   }
 
+  object healthCheckCount {
+    def unapply(healthChecks: Option[Seq[HealthCheck]]): Option[Int] =
```
dadjeibaah: scala's `map` function on the `Option` can be pretty handy in these situations. You can achieve the same `Option[Int]` method result like this:
```scala
def unapply(healthChecks: Option[Seq[HealthCheck]]): Option[Int] = healthChecks.map(_.size)
```

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
```diff
@@ -146,33 +162,40 @@ object Api {
...
-        tasks.collect {
-          case ipAddress(host) if !useHealthCheck => Address(host, port)
-          case healthyIpAddress(host) => Address(host, port)
-        }.toSet
-      case Some(App(_, _, Some(tasks))) =>
-        tasks.collect {
-          case hostPort(host, port) if !useHealthCheck => Address(host, port)
-          case healthyHostPort(host, port) => Address(host, port)
-        }.toSet
+      case Some(App(_, discoveryPort(port), healthCheckCount(count), Some(tasks))) =>
```
dadjeibaah: I think we can leave this method largely unchanged if we incorporate the number of health checks into the `Task` case class.

marathon/src/test/scala/io/buoyant/marathon/v2/ApiTest.scala
```diff
@@ -33,6 +33,30 @@ class ApiTest extends FunSuite with Awaits with Exceptions {
       "app": {
         "id": "/foo",
         "cmd": "foo cmd",
+        "healthChecks": [
```
dadjeibaah: Kudos for writing unit tests! 😄

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
adleong: This is an unapply method that we actually want to always match, so it should always return Some, as it does here.

A potentially less confusing way to implement this would be as a plain method:
```
def healthCheckCount(healthChecks: Option[Seq[HealthCheck]]): Int = healthChecks match {
  case Some(checks) => checks.size
  case None => 0
}
```

and then invoking this method directly instead of in as part of a case destructuring.

But this section of the code relies heavily on unapply methods so take it or leave it.

marathon/src/main/scala/io/buoyant/marathon/v2/Api.scala
adleong: @rogoman correct me if I'm wrong but I believe that Task is a class that matches the Marathon API and API responses are automatically deserialized to it.  This means that we can't add the health check count to that class.

rogoman: @adleong yes. Adding the number of health checks to the original `Task` case class was the easiest to implement, but I worried that it's going to be confusing for anyone reading the code and trying to find a match with what can be found in the Marathon API.

rogoman:
@dadjeibaah not introducing this additional class was my first idea but I worried that if I added a new attribute to the original `Task` case class, it wouldn't match what Marathon API returns anymore. @dadjeibaah, @adleong it's your call. I can implement any case you choose.

rogoman:
@adleong as for manual testing, I have built a Docker image using the instructions from `BUILD.md` and was planning to deploy such image to one of our test clusters to perform the test that led me to filing the original issue. I apologize for not doing it upfront, but was hoping to get a general "thumbs-up" from you guys before performing any extensive testing.

rogoman:
The original test that I performed was deploying a service that I can mark certain instances of to be unhealthy. See the chart below (logs of incoming HTTP requests grouped by app instance):

linkerd-settings

Originally it was running 3 instances (2 healthy). I scaled it up to 5 instances (3 should be healthy) but during the health check grace period one of the additional instances (that was failing the health check but still in the grace period) received traffic as well.

Now this is the situation after my fix:

log_explorer___datadog

Initially I had 3 healthy instances. I scaled them up to 5, out of which 4 were supposed to be healthy. During the grace period none of the requests were routed to any of the new instances. Once the fourth instance showed up as healthy, it started receiving requests. Same thing happened after scaling up to 8 instances (7 were supposed to be healthy). During the health-check grace period none of the new instances received any traffic. Once Marathon decided they're healthy, they started receiving traffic.

This is just visual data, to help understand the procedure. Of course I checked the logs of all the instances just to make sure there were no requests routed when there shouldn't be. None of the unhealthy instances received any requests, ever.

rogoman:
Thanks for the approval, @adleong . Is there anything else I need to do before this is merged to master?

dadjeibaah:
You are all set @rogoman I will merge this in right now.

PR #2103, opened 2018-08-15
Update Istio features with a deprecation notice.

### LINKED ISSUE #2092: Question: Current state of Istio integration?
I'm wondering what the current state of integration with the Istio control plane is, with that project having just gone 1.0 and, I'm assuming, including breaking changes.

Is continuing the Istio integration work started in 1.1.1 and continued in 1.3.1 an ongoing project goal?

dadjeibaah:
The Istio integration started off as an experiment to provide support for running Linkerd together with Istio's control plane. However, Linkerd's Istio integration has not seen any significant use as discussed in #2092.

This PR adds deprecation notices to docs as well as adds a warning when any of the Istio configurations are loaded at startup.

fixes #2092

### DIFF
--- a/interpreter/istio/src/main/scala/io/buoyant/interpreter/k8s/istio/IstioInterpreterInitializer.scala
+++ b/interpreter/istio/src/main/scala/io/buoyant/interpreter/k8s/istio/IstioInterpreterInitializer.scala
@@ -33,8 +33,7 @@ case class IstioInterpreterConfig(
   @JsonIgnore
   override val experimentalRequired = true
 
-  @JsonIgnore
-  val _ = Logger.get(this.getClass.getName).warning("Istio K8S Interpreter has been deprecated since version 1.4.7")
+  Logger.get(this.getClass.getName).warning("Istio K8S Interpreter has been deprecated since version 1.4.7")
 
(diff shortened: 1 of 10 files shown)


interpreter/istio/src/main/scala/io/buoyant/interpreter/k8s/istio/IstioInterpreterInitializer.scala
```diff
@@ -32,6 +33,9 @@ case class IstioInterpreterConfig(
   @JsonIgnore
   override val experimentalRequired = true
 
+  @JsonIgnore
+  val _ = Logger.get(this.getClass.getName).warning("Istio K8S Interpreter has been deprecated since version 1.4.7")
```
adleong: do you need to assign this to a dummy val?  Can you just do `Logger.get(...).warning(...)`?

dadjeibaah: Ahh, it looks like we don't. I thought that since this was a function of type `Unit` we would need to capture the expression in a dummy variable to avoid the `discarded Unit` error. TIL.

PR #2119, opened 2018-09-05
Add GC logging by default

### LINKED ISSUE #2110: Turn GC logging on by default
GC logging is an essential source of information when debugging Linkerd performance.  We should enable GC logging by default by adding the following settings:

```
-XX:+PrintGCDetails
-XX:+PrintGCDateStamps
-XX:+PrintHeapAtGC
-XX:+PrintTenuringDistribution
-XX:+PrintGCApplicationStoppedTime
-XX:+PrintPromotionFailure
-Xloggc:/var/log/linkerd/gc.log
-XX:+UseGCLogFileRotation
-XX:NumberOfGCLogFiles=10
-XX:GCLogFileSize=10M
```

dadjeibaah:
GC logging is necessary for debugging JVM issues in Linkerd. This PR adds flags to the Linkerd exec script to enable GC logging. It also adds a line in the exec script that creates a directory `var/log/linkerd` for `gc.log` files if it doesn't exist. it also allows for the GC log directory to be configurable using the environment variable `GC_LOG`

fixes #2110

### DIFF
--- a/project/Base.scala
+++ b/project/Base.scala
@@ -170,6 +170,7 @@ class Base extends Build {
       val home = s"/${organization.value}/${name.value}/${version.value}"
       val exec = s"$home/${configuration.value}-exec"
       from(dockerJavaImage.value)
+      run("[ -d /var/log/linkerd ]", "||", "mkdir", "-p", "/var/log/linkerd")
       run("mkdir", "-p", home)
       workDir(home)
       env(envPrefix+"HOME", home)
--- a/project/Base.scala
+++ b/project/Base.scala
@@ -170,7 +170,6 @@ class Base extends Build {
       val home = s"/${organization.value}/${name.value}/${version.value}"
       val exec = s"$home/${configuration.value}-exec"
       from(dockerJavaImage.value)
-      run("[ -d /var/log/linkerd ]", "||", "mkdir", "-p", "/var/log/linkerd")
       run("mkdir", "-p", home)
       workDir(home)
       env(envPrefix+"HOME", home)
--- a/project/LinkerdBuild.scala
+++ b/project/LinkerdBuild.scala
@@ -373,10 +373,16 @@ object LinkerdBuild extends Base {
          |
          |# Configure GC logging directory
          |if [ -z "$GC_LOG" ]; then
-         |  GC_LOG="/var/log/linkerd"
+         |  GC_LOG="/var/log/namerd"
          |fi
          |
          |[ -d "$GC_LOG" ] || mkdir -p "$GC_LOG"
+         |
+         |if [ ! \( -w "$GC_LOG" \) ]; then
+         |  echo "GC_LOG must be set to a directory that user [$USER] has write permissions on.\
+         |  Unable to use [$GC_LOG] for GC logging."
+         |fi
+         |
          |""" +
       execScriptJvmOptions +
       """|exec "${JAVA_HOME:-/usr}/bin/java" -XX:+PrintCommandLineFlags \
@@ -442,10 +448,16 @@ object LinkerdBuild extends Base {
          |
          |# Configure GC logging directory
          |if [ -z "$GC_LOG" ]; then
-         |  GC_LOG="/var/log/linkerd"
+         |  GC_LOG="/var/log/namerd"
          |fi
          |
          |[ -d "$GC_LOG" ] || mkdir -p "$GC_LOG"
+         |
+         |if [ ! \( -w "$GC_LOG" \) ]; then
+         |  echo "GC_LOG must be set to a directory that user [$USER] has write permissions on.\
+         |  Unable to use [$GC_LOG] for GC logging."
+         |fi
+         |
          |""" +
       execScriptJvmOptions +
       """|if read -t 0; then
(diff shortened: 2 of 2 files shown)


project/LinkerdBuild.scala
```diff
@@ -257,6 +257,16 @@ object LinkerdBuild extends Base {
        |   -Dio.netty.allocator.numHeapArenas=${FINAGLE_WORKERS:-8}      \
        |   -Dio.netty.allocator.numDirectArenas=${FINAGLE_WORKERS:-8}    \
        |   -Dcom.twitter.finagle.netty4.numWorkers=${FINAGLE_WORKERS:-8} \
+       |   -XX:+PrintGCDetails                                           \
+       |   -XX:+PrintGCDateStamps                                        \
+       |   -XX:+PrintHeapAtGC                                            \
+       |   -XX:+PrintTenuringDistribution                                \
+       |   -XX:+PrintGCApplicationStoppedTime                            \
+       |   -XX:+PrintPromotionFailure                                    \
+       |   -Xloggc:/var/log/linkerd/gc.log                               \
```
adleong: do you think we should use an env var here so this is overridable?

project/Base.scala
```diff
@@ -194,6 +194,7 @@ class Base extends Build {
       val home = s"/${organization.value}/${name.value}/${version.value}"
       val exec = s"$home/${configuration.value}-exec"
       from(dockerJavaImage.value)
+      run("[ -d /var/log/linkerd ]", "||", "mkdir", "-p", "/var/log/linkerd")
```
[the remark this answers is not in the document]
adleong: What about the case where users are running the executable directly (not in docker)?

project/LinkerdBuild.scala
dadjeibaah: Good point. I'll get that set up in the next commit.

project/LinkerdBuild.scala
```diff
@@ -360,6 +370,13 @@ object LinkerdBuild extends Base {
          |  done
          |fi
          |export MALLOC_ARENA_MAX=2
+         |
+         |# Configure GC logging directory
+         |if [ -z "$GC_LOG" ]; then
+         |  GC_LOG="/var/log/linkerd"
+         |fi
+         |
+         |[ -d "$GC_LOG" ] || mkdir -p "$GC_LOG"
```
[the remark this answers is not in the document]
dadjeibaah: Yea. When running this as a non-root user the JVM gives a permission denied error when trying to create logs. Running this in docker works just fine since docker runs the process as root. I can add a warning message in the script.

adleong:
For reference, here is the output when I run this as non-root:

```
$ linkerd/target/scala-2.12/linkerd-1.4.6-SNAPSHOT-exec linkerd/examples/fs.yaml
mkdir: /var/log/linkerd: Permission denied
GC_LOG must be set to a directory that user [alex] has write permissions on.  Unable to use [/var/log/linkerd] for GC logging.
-XX:+AggressiveOpts -XX:+AlwaysPreTouch -XX:+CMSClassUnloadingEnabled -XX:CMSInitiatingOccupancyFraction=70 -XX:+CMSParallelRemarkEnabled -XX:+CMSScavengeBeforeRemark -XX:GCLogFileSize=10485760 -XX:InitialHeapSize=33554432 -XX:MaxHeapSize=1073741824 -XX:MaxNewSize=348966912 -XX:MaxTenuringThreshold=6 -XX:NumberOfGCLogFiles=10 -XX:OldPLABSize=16 -XX:+PerfDisableSharedMem -XX:+PrintCommandLineFlags -XX:+PrintGC -XX:+PrintGCApplicationStoppedTime -XX:+PrintGCDateStamps -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintHeapAtGC -XX:+PrintPromotionFailure -XX:+PrintTenuringDistribution -XX:+ScavengeBeforeFullGC -XX:-TieredCompilation -XX:+UseCMSInitiatingOccupancyOnly -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseConcMarkSweepGC -XX:+UseGCLogFileRotation -XX:+UseParNewGC -XX:+UseStringDeduplication
Java HotSpot(TM) 64-Bit Server VM warning: Cannot open file /var/log/linkerd/gc.log due to No such file or directory
```

Linkerd then continues to start up normally (though presumably without gc logging).

It's maybe a bit confusing that the error here is broken across three different messages (the mkdir one, the GC_LOG echo, and the Cannot open file one) but it's probably okay.

project/LinkerdBuild.scala
```diff
@@ -360,6 +361,31 @@ object LinkerdBuild extends Base {
          |  done
          |fi
          |export MALLOC_ARENA_MAX=2
+         |
+         |# Configure GC logging directory
+         |if [ -z "$GC_LOG" ]; then
+         |  GC_LOG="/var/log/namerd"
+         |fi
+         |
+         |mkdir -p "$GC_LOG" && touch "$GC_LOG/gc.log"
```
dadjeibaah: Wanted to make a note about the the line `touch "$GC_LOG/gc.log"`. It looks like even `mkdir -p $GC_LOG` will have an exit status of 0 if the directory already exists and has strict permissions. We actually need to do an additional check to see if our process can create a file in the directory since that is what the JVM will do when its writing logs to that directory.

PR #2130, opened 2018-09-26
Configure adding clientId to TTwitter requests

corhere:
There are Thrift clients for many languages but few support TTwitter
upgrade. Some Thrift servers use the TTwitter clientId header to apply
rate-limiting quotas on clients. Thrift clients that cannot speak
TTwitter are unable to make requests against such servers.

Linkerd already supports upgrading the protocol to the server, even when
the client does not talk TTwitter. This PR adds a Thrift client
configuration option to set the clientId on outgoing TTwitter requests
when no clientId is set on the incoming request. This allows a
TTwitter-unaware client to use a service that requires the clientId be
set by proxying the requests through an appropriately-configured
linkerd.

Signed-off-by: Cory Snider <cory.snider@fitbit.com>

### DIFF
--- a/linkerd/docs/protocol-thrift.md
+++ b/linkerd/docs/protocol-thrift.md
@@ -53,3 +53,4 @@ Key | Default Value | Description
 --- | ------------- | -----------
 thriftFramed | `true` | If `true`, a framed thrift transport is used for outgoing requests; otherwise, a buffered transport is used. Typically this setting matches the router's servers' `thriftFramed` param.
 attemptTTwitterUpgrade | `false` | Controls whether thrift protocol upgrade should be attempted.
+clientId | no clientId | If set, the given clientId is added to upgraded outgoing protocol requests when one is not given in the incoming client request.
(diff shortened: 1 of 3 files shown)


corhere:
Those are excellent questions! It took me a while to find the answers as I have not used Scala nor Finagle before, but the client id of all outgoing requests is set to the configured value. That being said, it seems that an unmodified linkerd 1.4.6 (downloaded straight from the GitHub release) does not copy the client id from an incoming TTwitter request into the outgoing client request either. Is there an underlying linkerd bug or am I doing something wrong?

I tested the behaviour by modifying [`ThriftClient` and `ThriftServer` from `finagle-example`](https://github.com/twitter/finagle/tree/master/finagle-example/src/main/scala/com/twitter/finagle/example/thrift) to make a request with a client id and respond with the contents of the request's client id, respectively. Then I put linkerd in the middle.

- [ThriftClient.scala](https://github.com/linkerd/linkerd/files/2439906/ThriftClient.scala.txt)
- [ThriftServer.scala](https://github.com/linkerd/linkerd/files/2439910/ThriftServer.scala.txt)
- [linkerd.yaml](https://github.com/linkerd/linkerd/files/2439915/linkerd.yaml.txt)

When I connected them directly, the `ThriftClient` client id was echoed back, as expected. But when I put linkerd 1.4.6 in the middle, the server responded with "no clientId".

adleong:
I think you're right about the current behavior not propagating the client id.  I'd consider that to be a bug and this PR is a great opportunity to fix it!

adleong:
I'm closing this due to inactivity.  Please re-open if you would like to resume.

Bachmann1234:
Just out of curiosity what was left to do in this PR?

dadjeibaah:
@Bachmann1234 it's been a while so can't really remember. But I would assume opening up the PR and testing it out to see if the current code fixes the issue would be worthwhile. Are you interested in working on it?

Bachmann1234:
Not 100% sure I have the bandwidth right now so if the original person comes back awesome. But I wanted to make a note in case I end up having a strong desire to pick something new up.

PR #2133, opened 2018-09-30
[SPIKE] Expose stats scoped by the caller service

fantayeneh:

This PR not completed by far is to start the discussion on how we can expose
stats scoped by the caller service, which will help us to use linkerd as the source of truth for all service to service call metrics. 

The idea is stolen from Matt Klein talk on Envoy. 
svc2svc

Let me know if you're interested in this, or if there is a better way of achieving the same result.

Things to think about are 
- strategy to define who the caller is (maybe tracing headers can be used for this)
- how to protect linkerd from the explosion of this stats

Fanta

### DIFF
--- a/router/http/src/main/scala/io/buoyant/router/http/stats/ServiceToServiceStats.scala
+++ b/router/http/src/main/scala/io/buoyant/router/http/stats/ServiceToServiceStats.scala
@@ -1,0 +1,26 @@
+package io.buoyant.router.http.stats
+
+import com.twitter.finagle.http.Response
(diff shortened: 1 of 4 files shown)


adleong:
Thanks @fantayeneh this is super interesting.  This is pretty similar to some of the work that's being done in Linkerd2 to get per-source metrics.

I think a good first step before getting into the code would be to write up an issue describing what metrics Linkerd should output and describing the overall approach to how we determine the source.

fantayeneh:
Great, raised https://github.com/linkerd/linkerd/issues/2137
Will close this.

PR #2134, opened 2018-10-01
Add support for socket options for Linkerd and Namerd server endpoints

### LINKED ISSUE #1449: Feature - support zero downtime config changes
Currently, changing the linkerd config yaml file does not take effect without restarting the linkerd process.  However, restarting the process will lose any connections that are initiated in the time between shutting down and starting up again.  This means that (unless we want to lose some traffic during config changes), we cannot push new linkerd config to existing machines.  

Both haproxy and nginx both have solutions to this - in effect you can specify new configs in both of these and either restart without connections lost (haproxy) or reload the config (nginx) without shutting down.  Linkerd has an `/admin/shutdown` endpoint, but to my understanding (and in my testing) this is a graceful shutdown, but does not prevent loss of connections in the time between shutdown and subsequent restart.

If neither the haproxy or the nginx approach is workable, another solution I have seen is to offer an `/admin/offline` endpoint that tells the healthcheck to fail, but does not affect anything else.  This causes any load balancer that uses that health check to eventually mark the node down and quit sending traffic.  At that point, it can be safely restarted and once the linkerd process comes up again, the health check will pass again and the load balancer will add it back.   This solution requires some additional scripting work to orchestrate everything, but it is workable.

dadjeibaah:
A recent update to Finagle allows `ListeningServer` sockets to enable `SO_REUSEPORT`. This feature allows two or more Linkerd or Namerd instance to bind to the same port enabling zero downtime configuration changes.

This PR adds  `SO_REUSEPORT` to Linkerd routers, Linkerd admin servers and all Namerd control plane interfaces. Note that this feature is only available to certain Linux distributions. There also may need to be additional work to understand how this feature can be used when Linkerd is deployed in a daemonset environment. 

fixes #1449

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
@@ -4,11 +4,11 @@ import com.twitter.finagle.Stack
 import com.twitter.finagle.transport.Transport
 
 case class SocketOptionsConfig(
-  disableTcpNoDelay: Boolean = true,
-  reuseAddrEnabled: Boolean = true,
-  reusePortEnabled: Boolean = false
+  noDelay: Boolean = true,
+  reuseAddr: Boolean = true,
+  reusePort: Boolean = false
 ) {
   def params = Stack.Params.empty +
-    Transport.Options(disableTcpNoDelay, reuseAddrEnabled, reusePortEnabled)
+    Transport.Options(noDelay, reuseAddr, reusePort)
 
 }
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
@@ -1,0 +1,14 @@
+package com.twitter.finagle.buoyant
+
+import com.twitter.finagle.Stack
+import com.twitter.finagle.transport.Transport
+
+case class SocketOptionsConfig(
+  disableTcpNoDelay: Boolean = true,
+  reuseAddrEnabled: Boolean = true,
+  reusePortEnabled: Boolean = false
+) {
+  def params = Stack.Params.empty +
+    Transport.Options(disableTcpNoDelay, reuseAddrEnabled, reusePortEnabled)
+
+}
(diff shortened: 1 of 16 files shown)


zackangelo:
Awesome! 👏

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
```diff
@@ -0,0 +1,14 @@
+package com.twitter.finagle.buoyant
+
+import com.twitter.finagle.Stack
+import com.twitter.finagle.transport.Transport
+
+case class SocketOptionsConfig(
+  disableTcpNoDelay: Boolean = true,
```
adleong: I think the name of this parameter is backwards.  It maps to the finagle argument `noDelay`.

More generally, I think we should remove the `enabled` or `disable` words and just make all of these positive:

`tcpNoDelay = true, resuseAddr = true, reusePort = false`

dadjeibaah: Yea, good point. I was running through a couple of different versions to see what names would describe the options well. I do agree that `disable-` makes the word more negative.

linkerd/docs/config.md
```diff
@@ -270,3 +271,15 @@ Key | Default Value | Description
...
+
+### Socket Options
+
+Linkerd supports configuring socket level options for any given interface.
+i.e. the admin and router interfaces. These configurations are only available 
+on Linux 3.9 distributions and newer.
+
+Key | Default Value | Description
+--- | ------------- | -----------
+noDelay | true | If set to true, disables the use of `TCP_NODELAY` on a socket interface
```
adleong: I think this is backwards.  If set to true it enables TCP_NODELAY

dadjeibaah: According the the docs in Finagle's Transport.scala it looks like if it is set to true it is disabled and enabled if it is false.

```
/**
   * $param the options (i.e., socket options) of a `Transport`.
   *
   * @param noDelay enables or disables `TCP_NODELAY` (Nagle's algorithm)
   *                option on a transport socket (`noDelay = true` means
   *                disabled). Default is `true` (disabled).
   *
   * @param reuseAddr enables or disables `SO_REUSEADDR` option on a
   *                  transport socket. Default is `true`.
   *
   * @param reusePort enables or disables `SO_REUSEPORT` option on a
   *                  transport socket (Linux 3.9+ only). This option is only
   *                  available when using finagle-netty4 and native epoll support
   *                  is enabled. Default is `false`.
   */
```

PR #2140, opened 2018-10-03
Add request timeout to consul client to avoid stale service discovery updates

### LINKED ISSUE #2131: Restart consul watches after a timeout
When Linkerd/Namerd establishes a watch on Consul, it issues a long polling request that is satisfied when the watch fires.  There have been reports of issues where Linkerd does not receive a response when the watch fires and then keeps the long polling request open forever, resulting in holding on to stale data.  (e.g. https://github.com/linkerd/linkerd/issues/2121)

To mitigate the impact of that scenario, we should time out these long polling requests after a certain period and re-issue them.  This guarantees that we will eventually get up to date, even if a watch fails to fire.

dadjeibaah:
There have been issues where Namerd with a consul namer fails to update its watch state in situations were a long-polling request to consul never returns with service discovery info.

This PR adds a request timeout configuration for the client that issues these requests. The timeout is set to 10 minutes, which is double the timeout that consul already provides for [blocking queries](https://www.consul.io/api/index.html#blocking-queries). When Linkerd's consul client reaches the 10 minute timeout, the client throws an `IndividualRequestTimeoutException` which triggers a new request to be issued in [SvcAddr](https://github.com/linkerd/linkerd/blob/master/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala#L84) to the consul agent.

Tests were performed manually with a Test server that responded to any consul client request after 10 minutes. 

fixes #2131 

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -80,6 +80,19 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
+          case Throw(e: IndividualRequestTimeoutException) =>
+            // update state with last known state if we receive a API request timeout
+            stats.errors.incr()
+            log.log(
+              failureLogLevel,
+              "consul datacenter '%s' service '%s' observation error %s." +
+                " Last known state is %s",
+              datacenter, key.name, e, currentValueToLog
+            )
+            state.update(currentValueToLog)
+            val backoff #:: nextBackoffs = backoffs
+            // subsequent errors are logged as DEBUG
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
 
           case Throw(e) =>
             // update state with Addr.Neg, log error and continue polling with backoff
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -81,7 +81,7 @@ private[consul] object SvcAddr {
             stopped = true
             Future.Unit
           case Throw(e: IndividualRequestTimeoutException) =>
-            // update state with last known state if we receive a API request timeout
+            // catch request timeout exceptions for Consul API. Use last known good state.
             stats.errors.incr()
             log.log(
               failureLogLevel,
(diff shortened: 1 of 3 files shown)


namer/consul/src/test/scala/io/buoyant/namer/consul/SvcAddrTest.scala
```diff
@@ -150,4 +150,57 @@ class SvcAddrTest extends FunSuite with Matchers with Awaits {
...
+    val api = apiStub { (_, _, _, _, _, _) =>
+      requestCounter.incrementAndGet() match {
+        case 1 =>
+          invoked.setDone()
+          val response = Indexed[Seq[ServiceNode]](Seq(initServiceUpdate), Some("1"))
+          Future.value(response)
+        case 2 =>
+          secondInvoke.setDone()
+          Future.exception(new IndividualRequestTimeoutException(1.minutes))
+      }
```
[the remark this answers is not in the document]
dadjeibaah: Ahh, the "Dr. Strange" package, is there an example file I can look at to see how this works?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -80,6 +80,18 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
+          case Throw(e: IndividualRequestTimeoutException) =>
+            // update state with last known state if we receive a API request timeout
```
adleong: I don't think this comment is accurate anymore

namer/consul/src/test/scala/io/buoyant/namer/consul/SvcAddrTest.scala
```diff
@@ -150,4 +150,84 @@ class SvcAddrTest extends FunSuite with Matchers with Awaits {
...
+      tc.advance(12.minutes)
+      timer.tick()
+      changes match {
+        case Addr.Bound(addrs, _) => addrs.head shouldBe Address
+          .Inet(firstAddr, Addr.Metadata("endpoint_addr_weight" -> 1.0))
+        case _ => fail("received unexpected Addr on timed out service discovery request")
+      }
+
+      // Advance timer to go past backoff sleep future in SvcAddr
+      tc.advance(5.minutes)
```
adleong: Do we need both of this `tc.advance` call and the next one?  Should they be combined?

PR #2144, opened 2018-10-08
Add a new consul backed interpreter to Linkerd 

### LINKED ISSUE #2116: Create a new Linkerd interpreter that reads dtabs directly from a Consul discovery service.
Namerd, sometimes falls into an out-of-sync state when reading dtabs from a consul dtab store. One to fix this situation is by introducing a new interpreter config for Linkerd that reads dtabs from a Consul SD. 
We need to investigate and see if we can reuse some of the implementation already available on Linkerd and layout what the configuration options would look like.

dadjeibaah:
There are certain circumstances were we would like to retrieve dtabs stored in a consul agent directly from Linkerd.

This PR adds support to connect and retrieve dtabs to be used as a new interpreter in Linkerd. This PR also includes an interpreter state endpoint to allow for diagnostics and interpreter monitoring.

fixes #2116


interpreter/consul/src/main/scala/io/buoyant/interpreter/consul/ConsulDtabCache.scala
```diff
@@ -0,0 +1,91 @@
...
+import com.google.common.cache.{CacheBuilder, CacheLoader}
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.finagle.{Dtab, Failure, Path}
+import com.twitter.io.Buf
+import com.twitter.logging.Logger
+import com.twitter.util._
+import io.buoyant.consul.v1._
+import io.buoyant.namer.{InstrumentedActivity, InstrumentedVar}
+
+class ConsulDtabCache(
```
adleong: there seems to be a large amount of overlap here with ConsulDtabStore.  Is there a way to refactor things to share code between the two of them?  Or even better, can the Consul interpreter just use a ConsulDtabStore under the hood?

adleong: A bonus of this approach is that it helps to ensure compatibility between the consul dtab store and the consul interpreter.

dadjeibaah: Yea, that is correct. My dilemma was that I wanted to just use the `observe` method present in `ConsulDtabStore`, but my options were either to use the entire class or create a shared class that could be used by the consul interpreter and consul dtab storage. I didn't want to use the `ConsulDtabStore` class since it had other  CRUD operations for dtabs which I didn't want to include. I also had issues creating a shared class in that it would require me to restructure the dependencies for the consul, namer and linkerd modules. That would have caused a huge diff. I was thinking about following up with a separate PR that does the actual restructuring but probably the best bet is to just reuse `ConsulDtabStore` in the consul interpreter module.

namerd/storage/consul/src/main/scala/io/buoyant/namerd/storage/consul/ConsulDtabStore.scala
```diff
@@ -23,6 +23,7 @@ class ConsulDtabStore(
   datacenter: Option[String] = None,
   readConsistency: Option[ConsistencyMode] = None,
   writeConsistency: Option[ConsistencyMode] = None,
+  overrideHandlerUrl: Option[String] = None,
```
adleong: Maybe just require the handler url to be explicitly passed in, rather than passing in an override.

dadjeibaah: Yea that could work, but I wanted the default cause to always to be the `storage/....` endpoint optionally set the URL if we wanted it changed. I guess since the `ConsulDtabStore` isn't used anywhere else it'd be fine to make it more explicit.

dadjeibaah:
Tests were done using a consul agent that contained a dtab in consul's kv store. Requests were sent periodically through Linkerd as dtabs in consul were updated. Requests were routed correctly on every dtab change.

Issue #2132, opened 2018-09-29
OutOfDirectMemoryError


zackangelo:
Issue Type: Bug

We appear to be experiencing a direct memory leak in linkerd. We were observing `OutOfDirectMemory` errors quite frequently (almost daily) for a while and then we tuned the `-XX:MaxDirectMemory` JVM flag to increase the limit because we suspected that the value was too low to support the amount of I/O linkerd was doing. However, it only appears to have made the problem less frequent. Restarting linkerd corrects the issue for a while.

```
io.netty.util.internal.OutOfDirectMemoryError: failed to allocate 1048576 byte(s) of direct memory (used: 535822630, max: 536870912)
 at io.netty.util.internal.PlatformDependent.incrementMemoryCounter(PlatformDependent.java:640)
 at io.netty.util.internal.PlatformDependent.allocateDirectNoCleaner(PlatformDependent.java:594)
 at io.netty.buffer.PoolArena$DirectArena.allocateDirect(PoolArena.java:764)
 at io.netty.buffer.PoolArena$DirectArena.newChunk(PoolArena.java:740)
 at io.netty.buffer.PoolArena.allocateNormal(PoolArena.java:244)
 at io.netty.buffer.PoolArena.allocate(PoolArena.java:226)
 at io.netty.buffer.PoolArena.allocate(PoolArena.java:146)
 at io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:324)
 at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:181)
 at io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:172)
 at io.netty.buffer.AbstractByteBufAllocator.ioBuffer(AbstractByteBufAllocator.java:133)
 at io.netty.channel.DefaultMaxMessagesRecvByteBufAllocator$MaxMessageHandle.allocate(DefaultMaxMessagesRecvByteBufAllocator.java:80)
 at io.netty.channel.epoll.EpollRecvByteAllocatorHandle.allocate(EpollRecvByteAllocatorHandle.java:71)
 at io.netty.channel.epoll.AbstractEpollStreamChannel$EpollStreamUnsafe.epollInReady(AbstractEpollStreamChannel.java:786)
 at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe$1.run(AbstractEpollChannel.java:412)
[...]

PR #2146, opened 2018-10-10
set maxConcurrentStreamsPerConnection default 1000

siggy:
`maxConcurrentStreamsPerConnection`'s default value was unlimited,
meaning Linkerd could run out of memory if a client created enough
streams, particularly if the client is leaking streams but not properly
closing them.

Change default `maxConcurrentStreamsPerConnection` value to 1000,
disallow unlimited via a Linkerd config file.

Related to #2132 and https://github.com/grpc/grpc-go/pull/2354

Signed-off-by: Andrew Seigner <siggy@buoyant.io>

### DIFF
--- a/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
+++ b/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
@@ -255,7 +255,7 @@ case class RetryBufferSize(
 
 class H2ServerConfig extends ServerConfig with H2EndpointConfig {
 
-  var maxConcurrentStreamsPerConnection: Option[Int] = None
+  var maxConcurrentStreamsPerConnection: Option[Long] = None
   var addForwardedHeader: Option[AddForwardedHeaderConfig] = None
 
   @JsonIgnore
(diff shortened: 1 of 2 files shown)


linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
```diff
@@ -255,7 +255,7 @@ case class RetryBufferSize(
 
 class H2ServerConfig extends ServerConfig with H2EndpointConfig {
 
-  var maxConcurrentStreamsPerConnection: Option[Int] = None
+  var maxConcurrentStreamsPerConnection: Option[Long] = None
```
dadjeibaah: I think we may need to include the `@JsonDeserialize` annotation since this is a long similar to `classificationTimeoutMs`. It would look something like this:

```scala
 @JsonDeserialize(contentAs = classOf[java.lang.Long])
  var maxConcurrentStreamsPerConnection: Option[Long] = None
```

siggy: Ah, nice catch. I'm going to revert the `Long` -> `Int` change, and handle it in `withEndpointParams`.

PR #2153, opened 2018-10-17
removes all load balancers from the balancer registry

### LINKED ISSUE #2152: Linkerd runs into OOM when receiving empty address sets in its namers.
Linkerd runs into an OOM issue due to a Finagle bug in the `BalancerRegistry` module. It looks like Finagle creates empty load balancers that are never cleared every time a namer gets an empty address set update. More details at https://github.com/twitter/finagle/issues/735

dadjeibaah:
Linkerd runs into an issue where it runs out of memory when it receives empty address sets after a while from any configured namer. This is due to Finagle's Balancer registry not managing balancer state correctly and inadvertently store balancer instances in its `BalancerRegistry`  that are never removed.

This PR is set up to effectively "disable" the `BalancerRegistry` module by immediately removing `Balancer` instances when they are created. Once the [issue](https://github.com/twitter/finagle/issues/735) is fixed we can remove this workaround.

Tests were performed with a locally running Linkerd instance backed by a fs namer. Removing and readding an ip address in the `cat` file caused the bug to appear. After this fix, the size of the `BalancerRegistery` does not grow. Tests were also done with all the various load balancer configurations with Linkerd.

fixes #2152

### DIFF
--- a/linkerd/core/src/test/scala/io/buoyant/linkerd/ClientTest.scala
+++ b/linkerd/core/src/test/scala/io/buoyant/linkerd/ClientTest.scala
@@ -2,6 +2,8 @@ package io.buoyant.linkerd
 
 import com.twitter.finagle.Path
 import com.twitter.finagle.liveness.{FailureAccrualFactory, FailureAccrualPolicy}
+import com.twitter.finagle.loadbalancer.LoadBalancerFactory._
+import com.twitter.finagle.loadbalancer.buoyant.DeregisterLoadBalancerFactory
 import com.twitter.finagle.loadbalancer.{FlagBalancerFactory, LoadBalancerFactory}
 import com.twitter.finagle.ssl.client.SslClientConfiguration
 import com.twitter.finagle.transport.Transport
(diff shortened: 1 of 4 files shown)


router/core/src/main/scala/com/twitter/finagle/loadbalancer/buoyant/DeregisterLoadBalancerFactory.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  override def newBalancer[Req, Rep](
+    endpoints: Activity[IndexedSeq[EndpointFactory[Req, Rep]]],
+    emptyException: NoBrokersAvailableException,
+    params: Stack.Params
+  ): ServiceFactory[Req, Rep] = {
+    val underlying = lbf.newBalancer(endpoints, emptyException, params)
+
+    // A balancer is wrapped in a ServiceFactoryProxy so we need to unwrap it so that we
+    // can actually remove it from the registry.
```
adleong: This is so gross 👍 👍 👍

router/core/src/main/scala/com/twitter/finagle/loadbalancer/buoyant/DeregisterLoadBalancerFactory.scala
```diff
@@ -0,0 +1,38 @@
+package com.twitter.finagle.loadbalancer
+package buoyant
+
+import com.twitter.finagle.{NoBrokersAvailableException, ServiceFactory, ServiceFactoryProxy, Stack}
+import com.twitter.util.Activity
+
+/**
+ * Wraps a load balancer factory implementation and unregisters it from the
+ * Finagle global balancer registry.effectively disabling the use of the global registry.
```
adleong: typo on this line

router/core/src/main/scala/io/buoyant/router/Router.scala
```diff
@@ -371,11 +373,12 @@ object StackRouter {
     stk.result
   }
 
-  val defaultParams: Stack.Params =
+  val defaultParams: Stack.Params = {
```
adleong: don't think these braces are necessary

linkerd/core/src/test/scala/io/buoyant/linkerd/ClientTest.scala
```diff
@@ -38,17 +48,31 @@ class ClientTest extends FunSuite {
...
+      case _ => fail("Unexpected load balancer configured")
+    }
+    assert(barBal.toString == "ApertureLeastLoaded")
 
     // bas, not configured, gets default values
     val basParams = client.clientParams.paramsFor(Path.read("/#/io.l5d.fs/bas"))
-    val LoadBalancerFactory.Param(basBalancer) = basParams[LoadBalancerFactory.Param]
-    assert(basBalancer == FlagBalancerFactory)
+    val Param(basBalancer) = basParams[LoadBalancerFactory.Param]
+    println(basBalancer)
```
adleong: rm

dadjeibaah: 🤦‍♂️ Thanks

PR #2163, opened 2018-10-31
Add Prometheus exception counters

### LINKED ISSUE #2118: Improve Prometheus time series names for exception counters
Prometheus metric names for exception counters have the exception in the name (e.g., `rt:server:exn:io_netty_handler_codec_DecoderException`). This makes it difficult to aggregate or display by exception because it requires you to know in advance all of the possible exceptions. 

Perhaps a better way would be to consolidate all exceptions into a single counter and put the exception name in a label (e.g. `rt:server:exceptions{exception="io_netty_handler_codec_DecoderException"}`). That way, the exception names could be grouped or regex matched and you can select all of them using the same time series.

kleimkuhler:
### Problem
Prometheus metric names for exception counters have the exception in the name (e.g., rt:server:exn:io_netty_handler_codec_DecoderException). This makes it difficult to aggregate or display by exception because it requires you to know in advance all of the possible exceptions.

### Solution
Consolidate all exceptions into a single counter and put the exception name in a label (e.g. rt:server:exn{exception="io_netty_handler_codec_DecoderException"}). That way, the exception names could be grouped or regex matched and you can select all of them using the same time series.

### Validation
Two tests were added for client and server side.

Fixes #2118

### DIFF
--- a/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
+++ b/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
@@ -81,8 +81,16 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
         (Seq("rt", "client"), labels0 :+ ("client" -> escapeLabelVal(id)))
       case Seq("rt", "client", "service", path) if !labelExists(labels0, "service") =>
         (Seq("rt", "client", "service"), labels0 :+ ("service" -> escapeLabelVal(path)))
+      case Seq("rt", "client", "failures", path) if !labelExists(labels0, "exception") =>
+        (Seq("rt", "client", "failures"), labels0 :+ ("exception" -> escapeLabelVal(path)))
+      case Seq("rt", "client", "exn", path) if !labelExists(labels0, "exception") =>
+        (Seq("rt", "client", "exn"), labels0 :+ ("exception" -> escapeLabelVal(path)))
       case Seq("rt", "server", srv) if !labelExists(labels0, "server") =>
         (Seq("rt", "server"), labels0 :+ ("server" -> escapeLabelVal(srv)))
+      case Seq("rt", "server", "failures", path) if !labelExists(labels0, "exception") =>
+        (Seq("rt", "server", "failures"), labels0 :+ ("exception" -> escapeLabelVal(path)))
+      case Seq("rt", "server", "exn", path) if !labelExists(labels0, "exception") =>
+        (Seq("rt", "server", "exn"), labels0 :+ ("exception" -> escapeLabelVal(path)))
       case _ => (prefix0, labels0)
     }
 
--- a/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
+++ b/telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
@@ -75,22 +75,21 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
     val (prefix1, labels1) = prefix0 match {
       case Seq("rt", router) if !labelExists(labels0, "rt") =>
         (Seq("rt"), labels0 :+ ("rt" -> escapeLabelVal(router)))
-      case Seq("rt", "service", path) if !labelExists(labels0, "service") =>
-        (Seq("rt", "service"), labels0 :+ ("service" -> escapeLabelVal(path)))
-      case Seq("rt", "client", id) if !labelExists(labels0, "client") =>
-        (Seq("rt", "client"), labels0 :+ ("client" -> escapeLabelVal(id)))
+
+      // Add label for peer { "service", "client", "server" }
+      case Seq("rt", peer, identifier) if !labelExists(labels0, peer) =>
+        (Seq("rt", peer), labels0 :+ (peer -> escapeLabelVal(identifier)))
+
+      // Handle client service case
       case Seq("rt", "client", "service", path) if !labelExists(labels0, "service") =>
         (Seq("rt", "client", "service"), labels0 :+ ("service" -> escapeLabelVal(path)))
-      case Seq("rt", "client", "failures", path) if !labelExists(labels0, "exception") =>
-        (Seq("rt", "client", "failures"), labels0 :+ ("exception" -> escapeLabelVal(path)))
-      case Seq("rt", "client", "exn", path) if !labelExists(labels0, "exception") =>
-        (Seq("rt", "client", "exn"), labels0 :+ ("exception" -> escapeLabelVal(path)))
-      case Seq("rt", "server", srv) if !labelExists(labels0, "server") =>
-        (Seq("rt", "server"), labels0 :+ ("server" -> escapeLabelVal(srv)))
-      case Seq("rt", "server", "failures", path) if !labelExists(labels0, "exception") =>
-        (Seq("rt", "server", "failures"), labels0 :+ ("exception" -> escapeLabelVal(path)))
-      case Seq("rt", "server", "exn", path) if !labelExists(labels0, "exception") =>
-        (Seq("rt", "server", "exn"), labels0 :+ ("exception" -> escapeLabelVal(path)))
+
+      // Add label for exception { "failures", "exn" }
+      case Seq("rt", peer, "failures", path) if !labelExists(labels0, "exception") =>
+        (Seq("rt", peer, "failures"), labels0 :+ ("exception" -> escapeLabelVal(path)))
+      case Seq("rt", peer, "exn", path) if !labelExists(labels0, "exception") =>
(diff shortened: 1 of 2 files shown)


telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -81,8 +81,16 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
         (Seq("rt", "client"), labels0 :+ ("client" -> escapeLabelVal(id)))
       case Seq("rt", "client", "service", path) if !labelExists(labels0, "service") =>
         (Seq("rt", "client", "service"), labels0 :+ ("service" -> escapeLabelVal(path)))
+      case Seq("rt", "client", "failures", path) if !labelExists(labels0, "exception") =>
+        (Seq("rt", "client", "failures"), labels0 :+ ("exception" -> escapeLabelVal(path)))
+      case Seq("rt", "client", "exn", path) if !labelExists(labels0, "exception") =>
```
dadjeibaah: What do you think about using the word `exceptions` instead of `exn`? To me, it seems better to have it spelled out so that its descriptive.

kleimkuhler:
I like the suggestion on adding these counters for services as well. I'll get that change up!

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
kleimkuhler: I agree it would be more descriptive. The reason I kept it as `exn` is because that's the naming convention of the metric tree. I'm not sure if that would introduce any confusion when trying to debug and realizing that the node `exn` is in fact labeled as `exceptions` in the output.

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -75,14 +75,21 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
     val (prefix1, labels1) = prefix0 match {
       case Seq("rt", router) if !labelExists(labels0, "rt") =>
         (Seq("rt"), labels0 :+ ("rt" -> escapeLabelVal(router)))
-      case Seq("rt", "service", path) if !labelExists(labels0, "service") =>
-        (Seq("rt", "service"), labels0 :+ ("service" -> escapeLabelVal(path)))
-      case Seq("rt", "client", id) if !labelExists(labels0, "client") =>
-        (Seq("rt", "client"), labels0 :+ ("client" -> escapeLabelVal(id)))
+
+      // Add label for peer { "service", "client", "server" }
```
adleong: I don't think peer is the right name for this.  I'd call these "stacks".  e.g. the service stack, the client stack, and the server stack.

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -75,14 +75,21 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
...
+      // Handle client service case
       case Seq("rt", "client", "service", path) if !labelExists(labels0, "service") =>
         (Seq("rt", "client", "service"), labels0 :+ ("service" -> escapeLabelVal(path)))
-      case Seq("rt", "server", srv) if !labelExists(labels0, "server") =>
-        (Seq("rt", "server"), labels0 :+ ("server" -> escapeLabelVal(srv)))
+
+      // Add label for exception { "failures", "exn" }
+      case Seq("rt", peer, "failures", path) if !labelExists(labels0, "exception") =>
+        (Seq("rt", peer, "failures"), labels0 :+ ("exception" -> escapeLabelVal(path)))
+      case Seq("rt", peer, "exn", path) if !labelExists(labels0, "exception") =>
```
adleong: Will the exception name ever be multiple segments?  e.g. in the case of nested exceptions?  do we need to capture metrics like:
`Seq("rt", "client", "exn", "com.twitter.finagle.SomeParentException", "java.net.SomeChildException")` ?

telemetry/prometheus/src/test/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeterTest.scala
```diff
@@ -144,4 +144,68 @@ class PrometheusTelemeterTest extends FunSuite {
     val rsp = await(handler(Request(prometheusPath))).contentString
     assert(rsp == "linkerd_rt:server:requests{rt=\"incoming\", server=\"127.0.0.1/4141\"} 1\n")
   }
+
+  test("exception stats are labelled for service") {
+    val (stats, handler) = statsAndHandler
+    // Replicate stack trace for some failure bar:baz
+    stats.scope("rt", "incoming", "service", "/#/foo", "failures").counter("bar").incr()
+    stats.scope("rt", "incoming", "service", "/#/foo", "failures").counter("bar:baz").incr()
```
adleong: Related to my question above, is this how these types of nested exceptions are actually recorded?  Or would it be more like `stats.scope("rt", "incoming", "service", "/#/foo", "failures", "bar").counter("baz").incr()` ?

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
kleimkuhler: Yes exceptions will be nested, but the stack trace prints out each layer of the nested structure. In the case of an exception that is nested `foo:bar:baz`, it will be printed across three lines like this: `foo`, `foo:bar`, `foo:bar:baz`. The counter will count each layer of the nested structure.

telemetry/prometheus/src/test/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeterTest.scala
kleimkuhler: Also related to my answer above, this is how I've seen them recorded. If you have seen differently then I would be interested in seeing that!

dadjeibaah: @kleimkuhler I think I understand what @adleong means when referring to how nested exceptions are recorded. I took a look at how the failure stats get generated for a Finagle HTTPServer. Check out this [code](https://github.com/twitter/finagle/blob/63d5858f0a7d5f926ec8107233bf95d42c6fcda7/finagle-base-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L47) in the Finagle code base. It looks like counter receives a list of exception class name strings and then is incremented. That means that your test would have something like:
```scala
stats.scope("rt", "incoming", "service", "/#/foo", "failures").counter("Exception1", "Exception2", "Exception3").incr()
```

kleimkuhler: @dadjeibaah Okay thanks for the link - I understand what you mean. I can work on adding a catch for multiple exceptions within one counter.

telemetry/prometheus/src/main/scala/io/buoyant/telemetry/prometheus/PrometheusTelemeter.scala
```diff
@@ -75,14 +82,21 @@ class PrometheusTelemeter(metrics: MetricsTree, private[prometheus] val handlerP
...
+
+      // Handle client service case
       case Seq("rt", "client", "service", path) if !labelExists(labels0, "service") =>
         (Seq("rt", "client", "service"), labels0 :+ ("service" -> escapeLabelVal(path)))
-      case Seq("rt", "server", srv) if !labelExists(labels0, "server") =>
-        (Seq("rt", "server"), labels0 :+ ("server" -> escapeLabelVal(srv)))
+
+      // Add label for exception { "failures", "exn" }
+      case Seq("rt", stack, "failures", exception) =>
+        (Seq("rt", stack, "failures"), addException(labels0, exception))
```
adleong: I think this could be more simply written as something like:

```
  case Seq("rt", stack, "failures", exception@_*) => (..., labels0 :+ ("exception" -> escapeLabelVal(exception.mkString(":")))))
```
```

kleimkuhler: That's what I was originally working with, but a nested exception is built incrementally - not all at once. In the case of `.counter("bar", "qux")`, the first pass through `writeMetrics` will only see `bar`, and the second pass will only see `qux`.

We want to append `qux` to the already existing exception `bar`, but we can't do that in one go.

kleimkuhler:
No problem! Yes I see what you mean. Thinking about it, I think a low-hanging fruit before looking at a complete refactor would be changing the type of `labels` to `Map`. It could avoid a lot of the member checking and string manipulation in `addException`. I could swing back around with some general changes for that after a merge.

PR #2164, opened 2018-11-03
stabilizeExistence's inner var should watch the underlying var

### LINKED ISSUE #2079: L5d fails with DynBoundTimeoutException after target service is updated

## Getting Help ##

## Filing a Linkerd issue ##

Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
When  k8s deployment is applied and new version is installed, L5d fails with "E 0720 15:10:25.148 UTC THREAD10: service failure: com.twitter.finagle.naming.buoyant.DynBoundTimeoutException: Exceeded 10.seconds binding timeout while connecting to /#/io.l5d.k8s/default/http/_my_service_name_ for name: /svc/_my_service_name_"

L5d should update new service endpoints and send traffic to new pods

k8s deployed l5d and target service
servers config for usedroute:
```
  servers:
  - port: XXXX
    ip: 0.0.0.0
    clearContext: true
  bindingCache:
    paths: 100
    trees: 100
    bounds: 100
    clients: 10
    idleTtlSecs: 5
```
rollingupdate details:
```
  replicas: 2
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 0
      maxSurge: 50%
  minReadySeconds: 20
```

**Environment**:
- linkerd/namerd version, config files: 1.4.2
- Platform, version, and config files (Kubernetes, DC/OS, etc): kubernetes in AWS (not EKS)

[same issue was in December](https://github.com/linkerd/linkerd/issues/1730)

adleong:
Fixes #2079 

The `stabilizeExistence` method returns a type that consists of an outer `Var[Activity.State[Option[Var[T]]]]` which contains an inner `Var[T]`.  The outer `Var.async` which is driven by updates from the underlying unstable Activity.  This mostly behaves as intended with changes to the underlying unstable Activity causing updates to the outer and inner Vars as appropriate.  However, the inner Var is not an `Var.async` and is only updated so long as the outer Var is observed.  This means that if the inner Var is observed and the outer Var is not, changes will not propagate from the underlying unstable Activity to the inner Var.

This can cause a load balancer to get get stuck with an address set which never updates, even when new data comes in from service discovery.  Consider this scenario:

A first call to the Kubernetes namer is made and the namer calls `stabilizeExistence` on the underlying endpoints Activity.  The Activity (which is essentially the outer Var) is observed and the inner `Name.Bound` (which contains the inner Var) is cached as part of a client stack and also observed.  Then the observation on the outer Var is closed.  The inner Var is still observed but will not receive any updates from the endpoints Activity that backs it.  Since the client stack keeps this Var[Addr] cached, it will hold on to this stale Var for the lifetime of the client stack.

We solve this by making both the inner and outer Vars to be `Var.async` with each of them observing the underlying Activity.  This way, each Var will get updates if and only if that Var is observed.

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
@@ -76,14 +76,15 @@ object ExistentialStability {
         }
       }
 
-      // This keeps track of the state of the outer Var:
-      // None means the state was not Activity.Ok
-      // Some(false) means the state was Activity.Ok(Some)
-      // Some(true) means the state was Activity.Ok(None)
-      //
+      // This keeps track of the state of the outer Var.
       // We need to track these three states because we only want to update the outer Var when we
       // transition between states.
-      var exists: Option[Boolean] = None
+      sealed trait State
+      object DoesExist extends State
+      object DoesNotExist extends State
+      object NotOk extends State
+      var exists: State = NotOk
+
       val mu = new {}
 
       val outer = Var.async[Activity.State[Option[Var[T]]]](Activity.Pending) { update =>
@@ -93,28 +94,28 @@ object ExistentialStability {
               case Activity.Ok(Some(t)) if inner == null =>
                 // T created.
                 inner = makeInner(t)
-                exists = Some(true)
+                exists = DoesExist
                 update() = Activity.Ok(Some(inner))
-              case Activity.Ok(Some(_)) if exists != Some(true) =>
+              case Activity.Ok(Some(_)) if exists != DoesExist =>
                 // T re-created.
-                exists = Some(true)
+                exists = DoesExist
                 update() = Activity.Ok(Some(inner))
-              case Activity.Ok(Some(_)) if exists == Some(true) =>
+              case Activity.Ok(Some(_)) if exists == DoesExist =>
               // T modified.
               // Existence has not changed so the outer Var should not be updated.
-              case Activity.Ok(None) if exists != Some(false) =>
+              case Activity.Ok(None) if exists != DoesNotExist =>
                 // T deleted.
-                exists = Some(false)
+                exists = DoesNotExist
                 update() = Activity.Ok(None)
-              case Activity.Ok(None) if exists == Some(false) =>
+              case Activity.Ok(None) if exists == DoesNotExist =>
               // Spurious update in None state.  We can just ignore it.
               case Activity.Pending =>
                 // Pending.
-                exists = None
+                exists = NotOk
                 update() = Activity.Pending
               case Activity.Failed(e) =>
                 // Failed.
-                exists = None
+                exists = NotOk
                 update() = Activity.Failed(e)
             }
           }
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
@@ -11,7 +11,7 @@ object ExistentialStability {
 
   type VarUp[T] = Var[T] with Updatable[T]
 
-  implicit class ExistentialVar[T <: AnyRef](val unstable: Var[Option[T]])
+  implicit class ExistentialVar[T](val unstable: Var[Option[T]])
     extends AnyVal {
     /**
      * We can stabilize this by changing the type to Var[Option[Var[T]]].
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
@@ -1,6 +1,7 @@
 package com.twitter.finagle.buoyant
 
-import com.twitter.util.{Activity, Updatable, Var}
+import com.twitter.logging.Logger
+import com.twitter.util._
 
 /**
  * Extensions to `Var`/`Activity` allowing a `Var` containing an `Option` to be split
@@ -11,7 +12,9 @@ object ExistentialStability {
 
   type VarUp[T] = Var[T] with Updatable[T]
 
-  implicit class ExistentialVar[T](val unstable: Var[Option[T]])
+  val logger = Logger()
+
+  implicit class ExistentialVar[T <: AnyRef](val unstable: Var[Option[T]])
     extends AnyVal {
     /**
      * We can stabilize this by changing the type to Var[Option[Var[T]]].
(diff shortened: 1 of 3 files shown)


adleong:
@ylopatin-earnin and @chrisgoffinet: if you'd like to help me validate this fix, you can pull this branch (alex/stab-stab-stab) or the docker image: buoyantio/linkerd:1.5.1-stab

adleong:
After leaving some tests running over the weekend, this is looking pretty solid.  @ylopatin-earnin, is there a way for you to verify this fix in your environment?

k8s/src/test/scala/io/buoyant/k8s/EndpointsNamerTest.scala
```diff
@@ -1165,27 +1165,31 @@ class EndpointsNamerTest extends FunSuite with Awaits {
...
     val namer = new MultiNsNamer(Path.read("/test"), None, api.withNamespace, Stream.continually(1.millis))(timer)
 
-    @volatile var state: Activity.State[NameTree[Name]] = Activity.Pending
+    @volatile var state: Set[Address] = Set.empty
 
     val activity = namer.lookup(Path.read("/srv/http/sessions"))
 
-    eventually {
-      val addrs = await(activity.toFuture.map {
-        case NameTree.Leaf(Name.Bound(vaddr)) => vaddr.sample() match {
```
adleong: The problem with this test is that we only sample the vaddr, never observe it long enough for it to reach its final value.  Previously this was fine because observing the outer Var drove updates to the vaddr.  With our change to stabilizeExistence, we need to observe the vaddr itself if we want it to receive updates.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -56,44 +69,54 @@ object ExistentialStability {
...
+                    }
+                  }
+                }
+                update() = Activity.Ok(Some(inner))
+              case Activity.Ok(None) if exists == Some(false) =>
+              // Do nothing.
+              case Activity.Ok(None) if exists != Some(false) =>
+                exists = Some(false)
+                update() = Activity.Ok(None)
+              case Activity.Pending =>
```
hawkw: Nit/TIOLI: maybe we should match `Pending` and `Failed` in the same match arm, like this:
```scala
case state =>
  exists = None
  update() = state
```
That makes the code a little less repetitive.

On the other hand, this does rely on the rest of the match exhaustively matching over all the possible `Activity.Ok(...)` patterns...which it _does_, but it might be possible to miss one in a future change... 🤷‍♀️

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -20,33 +20,46 @@ object ExistentialStability {
...
-            exists = false
-            update() = None
+          case Some(t) => update() = t
+          case _ =>
+        }
+      }
+
+      // Initial value for outer Var.
+      val init = unstable.sample().map { t =>
+        if (inner == null) {
```
hawkw: TIOLI: It looks like we do this null check every time we call `makeInner`...maybe we should factor that out so that it's part of `makeInner`?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -20,33 +20,46 @@ object ExistentialStability {
...
+        }
+        inner
+      }
+
+      @volatile var exists = init.isDefined
+      val mu = new {}
+
+      Var.async[Option[Var[T]]](init) { update =>
+        unstable.changes.respond { state =>
+          mu.synchronized {
```
hawkw: I notice there's a change in this branch from synchronizing each match arm separately to synchronizing the whole match block...is that also a necessary part of this change? If so, it would be great to get a mention of that in the commit message for future reference...

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -20,33 +20,46 @@ object ExistentialStability {
...
-            // T deleted
-            exists = false
-            update() = None
+          case Some(t) => update() = t
+          case _ =>
+        }
+      }
+
+      // Initial value for outer Var.
+      val init = unstable.sample().map { t =>
```
dadjeibaah: Can this `.sample()` run into a race condition with the `makeInner` definition? Is that something we should be concerned about?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -20,33 +20,46 @@ object ExistentialStability {
...
+        if (inner == null) {
+          inner = makeInner(t)
+        }
+        inner
+      }
+
+      @volatile var exists = init.isDefined
+      val mu = new {}
+
+      Var.async[Option[Var[T]]](init) { update =>
```
dadjeibaah: What happens if we let this `Var.asnyc`'s init value be `None`?

k8s/src/test/scala/io/buoyant/k8s/EndpointsNamerTest.scala
```diff
@@ -1165,27 +1165,31 @@ class EndpointsNamerTest extends FunSuite with Awaits {
     val timer = new MockTimer
     val namer = new MultiNsNamer(Path.read("/test"), None, api.withNamespace, Stream.continually(1.millis))(timer)
 
-    @volatile var state: Activity.State[NameTree[Name]] = Activity.Pending
+    @volatile var state: Set[Address] = Set.empty
```
dadjeibaah: Does this test exercise the code that once had the bug? Are we confident that if we reproduce the issue at the unit test level, we do not see the issue come up?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -20,33 +20,46 @@ object ExistentialStability {
...
+
+      // Initial value for outer Var.
+      val init = unstable.sample().map { t =>
+        if (inner == null) {
+          inner = makeInner(t)
+        }
+        inner
+      }
+
+      @volatile var exists = init.isDefined
```
olix0r: does this actually need to be @volatile? it appears that all access is synchronized around `mu`

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
olix0r: This sample will open and immediately close an unstable observation.  Then, below, we immediately open the observation up again. Is it possible to avoid redundant state transitions?

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
adleong: Good question.  With the synchronized blocks just within the arms there is a potential race condition where two threads could pick the same arm because the case matching happens outside of the synchronize block.  I never observed this race condition happening in practice but to be safe I expanded the synchronized block to make sure this can never happen.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
adleong: This actually doesn't work because state is of type `Activity.State[Option[T]]` and update expects type `Activity.State[Option[Var[T]]`.

I bet you might be able to pull this off with Scala3's union types because `Activity.Pending | Activity.Failed` is a subtype of `Activity.State[Nothing]` but I don't think this can be done in scala2.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
adleong: gonna LI because `makeInner` returns a Var.  Moving the null check into `makeInner` means we have to move the assignment into `makeInner` as well which seems less clear to me.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -56,44 +69,54 @@ object ExistentialStability {
...
+      //
+      // We need to track these three states because we only want to update the outer Var when we
+      // transition between states.
+      @volatile var exists: Option[Boolean] = None
+      val mu = new {}
+
+      val outer = Var.async[Activity.State[Option[Var[T]]]](Activity.Pending) { update =>
+        unstable.states.respond { state =>
+          mu.synchronized {
+            state match {
```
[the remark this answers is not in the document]
adleong: done

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
adleong: I was able to remove this by taking @dadjeibaah's suggestion below of using None as the initial value.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
adleong: Good idea.

k8s/src/test/scala/io/buoyant/k8s/EndpointsNamerTest.scala
adleong: Added a new unit test.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
adleong: volatile removed.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
```diff
@@ -56,44 +65,63 @@ object ExistentialStability {
...
+              case Activity.Ok(Some(t)) if inner == null =>
+                // T created.
+                inner = makeInner(t)
+                exists = Some(true)
+                update() = Activity.Ok(Some(inner))
+              case Activity.Ok(Some(_)) if exists != Some(true) =>
+                // T re-created.
+                exists = Some(true)
+                update() = Activity.Ok(Some(inner))
+              case Activity.Ok(Some(_)) if exists == Some(true) =>
```
olix0r: tioli: `exists.getOrElse(false)` and `!exists.getOrElse(true)` are a bit easier to read than e.g. `exists != Some(false)`

olix0r: ah but that's not the same thing, because the None type.  Consider using an dedicated marker `enum` with three stats rather than `Option[Boolean]`

adleong: I considered this, but though it might be more boilerplate than necessary.  But you're probably right, it will be more clear.

olix0r: It may be, but feel free to disregard if it's not

adleong: I opted for a sealed trait because scala enums give me the heebie jeebies

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/ExistentialStability.scala
hawkw: Turns out I totally forgot how pattern matching works in Scala. Man, languages with subtyping are _wild_...

PR #2169, opened 2018-11-13
move request authorizers to client config section

dadjeibaah:
RequestAuthorizers are modules that reside in the client stack of Linkerd. The configuration file needs to reflect that as well. This PR moves the loading of `RequestsAuthorizers` so that it happens in the during client config initialization.

Tests were done using the `IstioRequestAuthorizer` and a custom plugin. Tests were also done for `io.l5d.global` and `io.l5d.static` configurations.

fixes #2169 


linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/H2Config.scala
```diff
@@ -177,11 +166,22 @@ class H2PrefixConfig(prefix: PathMatcher) extends PrefixConfig(prefix) with H2Cl
...
+  var requestAuthorizers: Option[Seq[H2RequestAuthorizerConfig]] = None
 
   @JsonIgnore
   override def params(vars: Map[String, String]): Stack.Params =
     withEndpointParams(super.params(vars))
       .maybeWith(forwardClientCert.map(ForwardClientCertFilter.Enabled))
+      .maybeWith(loggerParam)
+
+  @JsonIgnore
+  private[this] def loggerParam = requestAuthorizers.map { configs =>
```
adleong: lets take this opportunity to rename this.  (they used to be called "loggers" before they were renamed to "requestAuthorizers")

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -121,12 +122,23 @@ class HttpStaticClient(val configs: Seq[HttpPrefixConfig]) extends HttpClient wi
...
 
   @JsonIgnore
   override def params(vars: Map[String, String]): Stack.Params = {
     super.params(vars)
       .maybeWith(forwardClientCert.map(ForwardClientCertFilter.Enabled))
+      .maybeWith(loggerParam)
+  }
+
+  @JsonIgnore
+  private[this] val loggerParam: Option[param.RequestAuthorizer] = requestAuthorizers.map { configs =>
```
adleong: ditto

linkerd/examples/istio.yaml
```diff
@@ -16,8 +16,9 @@ routers:
     experimental: true
     discoveryHost: localhost
     apiserverHost: localhost
-  requestAuthorizers:
-  - kind: io.l5d.k8s.istio
-    mixerHost: localhost
   servers:
   - port: 4140
+  client:
```
adleong: we should perhaps remove this example file, considering Istio support is deprecated.

dadjeibaah:
> This looks correct to me. Have you been able to test it? I recommend writing a simple request authorizer plugin the just logs something and making sure you can load it and that it runs on each request.

I was able to start Linkerd and observe the `/config.json` endpoint. It looked like it was able to load just fine. Writing a plugin and testing it is a good idea. I will try that out.

dadjeibaah:
Good point, I will not that down for the next release.

PR #2171, opened 2018-11-14
Fix HTTP 204 response logging

### LINKED ISSUE #2166: Linkerd logs incorrect message about Content-Length for HTTP 204 responses
If Linkerd proxies an HTTP response with status code 204, it logs the following message:

>Response with a status code of 204 must not have a Content-Length header field thus the field has been removed. Content-Length: 0

This message is logged, even if the response does not have a Content-Length header field.

This is due to an underlying issue in Finagle (https://github.com/twitter/finagle/issues/739) which is in turn due to an underlying issue in Netty (https://github.com/netty/netty/pull/6862).

This issue can be worked around at runtime by disabling a specific logger though the Linkerd admin server:

```
curl localhost:9990/logging.json --data 'level=OFF&logger=com.twitter.finagle.http.codec.ResponseConformanceFilter%24'
```

We could also add code when Linkerd starts up to disable this logger programmatically.

kleimkuhler:
# Problem
If Linkerd proxies an HTTP response with status code 204, it logs the following message:

> Response with a status code of 204 must not have a Content-Length header field thus the field has been removed. Content-Length: 0

This message is logged, even if the response does not have a Content-Length header field.

# Solution
We can disable the specific logger causing this issue when starting up Linkerd.

# Validation
When running the h1 server through SSS and responding to requests with
```
val rsp = Response()
rsp.contentString = ""
rsp.statusCode = 204
Future.value(rsp)
```
The log output is empty

This is different from the current behavior which logs the following
```
1114 13:09:37.434 PST THREAD3761: Response with a status code of 204 must not have a Content-Length header field thus the field has been removed. Content-Length: 0
```

Fixes #2166


linkerd/main/src/main/scala/io/buoyant/linkerd/Main.scala
```diff
@@ -33,6 +34,9 @@ object Main extends App {
     val build = Build.load("/io/buoyant/linkerd/build.properties")
     log.info("linkerd %s (rev=%s) built at %s", build.version, build.revision, build.name)
 
+    // #2166 -- Linkerd logs incorrect message about Content-Length for HTTP 204 responses
```
adleong: A link in this comment to the underlying Finagle and/or Netty issues that cause this would be good.

kleimkuhler: @adleong Hey I updated the issue to link back to finagle

PR #2173, opened 2018-11-14
Add KeepAlive options for sockets

jreichhold:
**Problem**
LinkerD doesn't currently support setting SO_KEEPALIVE parameter in Finagle/Netty which defaults to false. This leads to the l5d server hanging onto connections to dead processes that didn't close the socket properly. Without this setting we are seeing a slow growth in connections (file descriptors) on l5d servers from the l5d routing layer. These connections are slowing being left in ESTABLISHED state on the destination l5d even after the routing l5d was killed days and weeks before.

**Solution**
KeepAlive is settable on the server connections. Also support write/read timeouts on liveness from Finagle
https://twitter.github.io/finagle/docs/com/twitter/finagle/transport/Transport$$Liveness.html

**Validation**
First validated that created connections have keepalive enabled. This was accomplished by `netstat -tnope | grep java`

Running canary with auto scaling and verifying that sockets to dead servers are collected after the keep alive settings on the server have updated.
```tcp        0      0 xxx.xx.xx.171:4141      xxx.xx.xx.250:53836     ESTABLISHED 1000       297717      21582/java       keepalive (930.26/0/0)```

Watched keepalive from a single dead machine remove 63 connections from a host after failing the probe count as configured in OS in
 
net.ipv4.tcp_keepalive_time
net.ipv4.tcp_keepalive_intvl
net.ipv4.tcp_keepalive_probes

CPU and memory consumption are both also observed to be lower on the new host

Fixes https://github.com/linkerd/linkerd/issues/2170

### DIFF
--- a/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
+++ b/finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
@@ -14,12 +14,12 @@ case class SocketOptionsConfig(
 ) {
   def params: Stack.Params = {
     val writeTimeout: Duration = writeTimeoutMs match {
-      case Some(seconds) => Duration.fromMilliseconds(seconds)
+      case Some(milliseconds) => Duration.fromMilliseconds(milliseconds)
       case _ => Duration.Top
     }
 
     val readTimeout: Duration = readTimeoutMs match {
-      case Some(seconds) => Duration.fromMilliseconds(seconds)
+      case Some(milliseconds) => Duration.fromMilliseconds(milliseconds)
       case _ => Duration.Top
     }
 

### CODE REVIEW

File: linkerd/docs/config.md [hunk 1]
```diff
@@ -283,3 +283,6 @@ Key | Default Value | Description
 noDelay | true | If set to true, enables the use of `TCP_NODELAY` on a socket interface
 reuseAddr | true | If set to true, enables the `SO_REUSEADDR` option
 reusePort | false | If set to true, enables the `SO_REUSEPORT` option, which can be used to bind another Linkerd process to the same interface port.
+readTimeoutSeconds | Unbounded | Configures this client or server with given transport-level socket read timeout.
```
dadjeibaah: s/Unbounded/unbounded

dadjeibaah: I also think we should shorten the new options. For example, we could shorten `readTimeoutSeconds` to `readTImeoutSecs` to keep it consistent with other configuration field names.

namerd/iface/control-http/src/test/scala/io/buoyant/namerd/iface/HttpControlServiceConfigTest.scala
```diff
@@ -51,13 +51,15 @@ class HttpControlServiceConfigTest extends FunSuite {
...
-    val expectedOpts = SocketOptionsConfig(reusePort = true)
+    val expectedOpts = SocketOptionsConfig(reusePort = true, readTimeoutSeconds = Some(60), keepAlive = Some(true))
     val yaml = """
       |kind: io.l5d.httpController
       |socketOptions:
       |  noDelay: true
       |  reuseAddr: true
       |  reusePort: true
+      |  readTimeoutSeconds: 60
+      |  keepAlive: true
```
dadjeibaah: Thanks for modifying the test with the new change! For completeness, I think we should also include `writeTimeoutSeconds` as well.

dadjeibaah:
@jreichhold, might sound silly to ask you this for a simple PR but do you mind adding some more information on the PR that describes the problem you are saying and outlining how your change fixes that problem. This [CONTRIBUTING.md](https://github.com/linkerd/linkerd/blob/master/CONTRIBUTING.md) is the guideline we use for PRs.

finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
```diff
@@ -2,13 +2,30 @@ package com.twitter.finagle.buoyant
...
+  reusePort: Boolean = false,
+  writeTimeoutMs: Option[Long] = None,
+  readTimeoutMs: Option[Long] = None,
+  keepAlive: Option[Boolean] = None
 ) {
-  def params = Stack.Params.empty +
-    Transport.Options(noDelay, reuseAddr, reusePort)
+  def params: Stack.Params = {
+    val writeTimeout: Duration = writeTimeoutMs match {
+      case Some(seconds) => Duration.fromMilliseconds(seconds)
```
adleong: nit: probably call this variable `milliseconds` instead of `seconds`.

jreichhold: done

PR #2175, opened 2018-11-15
Fix 2174 issue with label keys not escaped and add test

jreichhold:
Signed-off-by: Jonathan Reichhold <jonathan.reichhold@offerupnow.com>


adleong:
Thanks for filing this PR, @jreichhold!  I originally took out this escaping because we explicitly control all label names and can ensure that they're properly escaped before they make it to this part.  A different bug allowed arbitrary label names in, which caused this problem.  Therefore, I believe that https://github.com/linkerd/linkerd/pull/2176 will fix this without needing to add another layer of escaping.

Does that sound right to you?

PR #2185, opened 2018-12-04
Update linkerd/namerd scripts to check Java version

### LINKED ISSUE #2165: Can't start linkerd in java9+
If I startup linkerd with a JAVA_HOME targetting java9+, I get this error:
```
# /io.buoyant/linkerd/linkerd
OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.
Unrecognized VM option 'PrintGCDateStamps'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

It seems to be related to this PR: 
https://github.com/linkerd/linkerd/pull/2119

Java 9+ has a whole new logging system for GC logging. Probably need to make this new logging optional or switch based on java version

kleimkuhler:
# Problem
If we startup Linkerd with a `JAVA_HOME` targeting Java 9, we get this error:

```
sudo env JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-9.0.4.jdk/Contents/Home linkerd/target/scala-2.12/linkerd-1.5.2-SNAPSHOT-exec linkerd/examples/fs.yaml
```

```
Java HotSpot(TM) 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.
Unrecognized VM option 'PrintGCDateStamps'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

# Solution
We check the Java version that will be used to startup Linkerd. If the Java version is 9.x, then we will display a warning that GC logging will not be used and skip setting `GC_LOG_OPTION`. This allows Linkerd to startup, but without GC logging. We suggest changing the Java version that `JAVA_HOME` points to, or upgrading to Java SE 10 (similar to [here](https://www.oracle.com/technetwork/java/javase/downloads/jdk9-downloads-3848520.html))

# Validation

```
sudo env JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk-9.0.4.jdk/Contents/Home linkerd/target/scala-2.12/linkerd-1.5.2-SNAPSHOT-exec linkerd/examples/fs.yaml
```

```
GC logging feature not available for Linkerd with Java SE 9. Consider setting JAVA_HOME environment variable to an upgraded SE, or upgrading to JAVA SE 10.
Java HotSpot(TM) 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.
[...]
I 1204 10:15:05.239 PST THREAD1: linkerd 1.5.2-SNAPSHOT (rev=67694a18ec328d3c3827ecc636468f991e497992) built at 20181204-101330
W 1204 10:15:05.503 PST THREAD1: failed to create Hotspot JVM interface, using NilJvm instead
I 1204 10:15:05.635 PST THREAD1: Finagle version 18.9.1 (rev=63182f815cb20549ced80c99465879609b05d94b) built at 20180926-144723
I 1204 10:15:07.802 PST THREAD1: Tracer: com.twitter.finagle.zipkin.thrift.ScribeZipkinTracer
I 1204 10:15:07.842 PST THREAD1: connecting to usageData proxy at Set(Inet(stats.buoyant.io/104.28.23.233:443,Map()))
I 1204 10:15:08.235 PST THREAD1: serving http admin on /127.0.0.1:9990
I 1204 10:15:08.257 PST THREAD1: serving http on localhost/127.0.0.1:4140
I 1204 10:15:08.305 PST THREAD1: initialized
```

Fixes #2165

### DIFF
--- a/project/LinkerdBuild.scala
+++ b/project/LinkerdBuild.scala
@@ -234,12 +234,12 @@ object LinkerdBuild extends Base {
   val ConfigFileRE = """^(.*)\.yaml$""".r
 
   val execScriptJvmOptions =
-    """|DEFAULT_JVM_OPTIONS="-Djava.net.preferIPv4Stack=true             \
+    """|DEFAULT_JVM_OPTIONS="$DEFAULT_JVM_OPTIONS                        \
+       |   -Djava.net.preferIPv4Stack=true                               \
        |   -Dsun.net.inetaddr.ttl=60                                     \
        |   -Xms${JVM_HEAP_MIN:-32M}                                      \
        |   -Xmx${JVM_HEAP_MAX:-1024M}                                    \
        |   -XX:+AggressiveOpts                                           \
-       |   -XX:+UseConcMarkSweepGC                                       \
        |   -XX:+CMSParallelRemarkEnabled                                 \
        |   -XX:+CMSClassUnloadingEnabled                                 \
        |   -XX:+ScavengeBeforeFullGC                                     \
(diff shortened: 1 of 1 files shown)


dadjeibaah:
Running a linkerd docker image with this change displays this error:
```bash
$ docker run -p 9990:9990 -v `pwd`/basic.config.yml:/config.yaml -v `pwd`/disco:/disco buoyantio/linkerd:1.5.2-SNAPSHOT /config.yaml
/io.buoyant/linkerd/1.5.2-SNAPSHOT/bundle-exec: 15: /io.buoyant/linkerd/1.5.2-SNAPSHOT/bundle-exec: [[: not found
```
It might the fact that you the docker image runs the script with `sh` and not `bash`.

A quick search on the interwebs shows that you can do logical regex shell expressions with [grep]( 
https://stackoverflow.com/questions/21115121/how-to-check-pattern-match-by-using-bin-sh-not-by-bin-bash) but I'm not sure if its the best way.

kleimkuhler:
> Running a linkerd docker image with this change displays this error:
> 
> ```shell
> $ docker run -p 9990:9990 -v `pwd`/basic.config.yml:/config.yaml -v `pwd`/disco:/disco buoyantio/linkerd:1.5.2-SNAPSHOT /config.yaml
> /io.buoyant/linkerd/1.5.2-SNAPSHOT/bundle-exec: 15: /io.buoyant/linkerd/1.5.2-SNAPSHOT/bundle-exec: [[: not found
> ```
> 
> It might the fact that you the docker image runs the script with `sh` and not `bash`.
> 
> A quick search on the interwebs shows that you can do logical regex shell expressions with [grep](https://stackoverflow.com/questions/21115121/how-to-check-pattern-match-by-using-bin-sh-not-by-bin-bash) but I'm not sure if its the best way.

It looks like `case` is the best performance wise for just a quick regex check. I made the change.

adleong:
Does this actually work with JDK 10?  My understanding was that these GC flags were removed in favor of a new system in JDK 9.  I would expect that these flags would only work on JDK 8 or earlier.

adleong:
If we detect that the java version uses this new gc logging system, we should set the flags for it:

https://dzone.com/articles/disruptive-changes-to-gc-logging-in-java-9

kleimkuhler:
@adleong @dadjeibaah Ok my last commit introduced some significant changes from the first go at this.

The script still checks the version, but it enables GC logging no matter what Java version is being used (much better - thanks Alex). How this happens is `GC_LOG_OPTIONS` is now set the way that the specific Java version requires. For Java 9+, this is with a single `-Xlog`. The selectors, file, and decorators are equivalent to the old tags we used. I translated them via the article linked above: https://dzone.com/articles/disruptive-changes-to-gc-logging-in-java-9

project/LinkerdBuild.scala
```diff
@@ -360,30 +360,38 @@ object LinkerdBuild extends Base {
          |    jars="$jars:$jar"
          |  done
          |fi
+         |
          |export MALLOC_ARENA_MAX=2
          |
          |# Configure GC logging directory
          |if [ -z "$GC_LOG" ]; then
-         |  GC_LOG="/var/log/namerd"
+         |  GC_LOG="/var/log/linkerd"
```
adleong: I think this should still be namerd?

project/LinkerdBuild.scala
```diff
@@ -447,30 +455,38 @@ object LinkerdBuild extends Base {
...
```
adleong: I think this should be namerd?

PR #2186, opened 2018-12-06
Use last known good state when consul responds with an 5xx error

### LINKED ISSUE #2145: Namerd returning empty dtab after consul failure

## Getting Help ##

## Filing a Linkerd issue ##

Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:
The consul cluster became unresponsive. Namerd noticed and emptied the dtab for every service. This then ended up on Linkerd, causing it to not be able to route requests.

I would expect that when Namerd identifies that Consul is not responding, it will retain the known-good state of services.

Block traffic between consul and namerd.

**Environment**:
Linkerd config:

    - protocol: thrift
      label: port-10000
      thriftProtocol: binary
      interpreter:
        kind: io.l5d.namerd
        dst: /$/inet/namerd.service.consul/4100
        namespace: 10000
      servers:
      ......
      client:
       .....

Namerd config:

        interfaces:
          port: 4100
          ip: 0.0.0.0
        - kind: io.l5d.httpController
          port: 4321

        namers:
        - kind: io.l5d.consul
          host: 127.0.0.1
          port: 8500
          includeTag: false
          useHealthCheck: true
          healthStatuses:
          - passing
          consistencyMode: stale

        storage:
          kind: io.l5d.inMemory
          namespaces:
              port-10000: /svc => /#/io.l5d.consul/dc/my_service

Linkerd / Namerd version: 1.4.6

Namerd log:

    W 1008 13:54:35.666 UTC THREAD23 TraceId:9220a1acd06f2ba0: consul datacenter 'dc' service 'my_service' observation error UnexpectedResponse(500: rpc error: EOF). Last known state is 
    Bound(Set(Inet(/x.x.x.x:yyyyy,Map(endpoint_addr_weight -> 1.0)), 
    Inet(/x.x.x.x:yyyyy,Map(endpoint_addr_weight -> 1.0)), 
    Inet(/x.x.x.x:yyyyy,Map(endpoint_addr_weight -> 1.0))),Map())

Linkerd log:

    I 1008 13:54:43.618 UTC THREAD86 TraceId:cd339c016207d896: #/io.l5d.consul/dc/my_service: name resolution is negative (local dtab: Dtab())

dadjeibaah:
After #1863, when the consul namer receives a 5xx response from consul, the namer evaluates to `Addr.Neg`. This behaviour is desired when a service resolution is performed with a non-existent DC. However, if consul sends a 5xx response that isn't caused by a request to a non-existent DC, we want to make sure that the last known good `Addr` is used instead. #1863 causes the namer to never use the last known good state and incorrectly resolves a logical name with `Addr.Neg`. This happens because it is difficult to reliably differentiate a 5xx response that requires an `Addr.Neg` or one that requires the use of the last known good state.

This PR adds a new check in `SvcAddr` that determines if an exception is one that is a result of resolving a service name to a non-existent DC. This new check is somewhat brittle since it reads the value of the response content string. This PR also changes the "catch-all" case that captures any exception from consul and lets `SvcAddr` use the previous resolution state.

A couple of unit tests have been changed to reflect this new behavior and I was able to test this out locally with reproduction steps similar to #2145 

fixes #2145

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -9,7 +9,6 @@ import io.buoyant.consul.v1
 import io.buoyant.consul.v1.UnexpectedResponse
 import io.buoyant.namer.{InstrumentedVar, Metadata}
 import java.net.{InetAddress, InetSocketAddress}
-import java.rmi.UnexpectedException
 import scala.util.control.NoStackTrace
 
 private[consul] case class SvcKey(name: String, tag: Option[String]) {
@@ -83,6 +82,9 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
+            // this exception case checks if we queried for a service in a datacenter that
+            // doesn't exist. We capture this case so that we can return Addr.Neg to prevent
+            // service name resolution from timing out.
           case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage =>
             log.log(
               failureLogLevel,
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -6,9 +6,10 @@ import com.twitter.finagle.util.DefaultTimer
 import com.twitter.logging.Level
 import com.twitter.util._
 import io.buoyant.consul.v1
-import io.buoyant.namer.{Metadata, InstrumentedVar}
+import io.buoyant.consul.v1.UnexpectedResponse
+import io.buoyant.namer.{InstrumentedVar, Metadata}
 import java.net.{InetAddress, InetSocketAddress}
-
+import java.rmi.UnexpectedException
 import scala.util.control.NoStackTrace
 
 private[consul] case class SvcKey(name: String, tag: Option[String]) {
@@ -22,6 +23,8 @@ private[consul] object SvcAddr {
 
   private[this] val ServiceRelease =
     new Exception("service observation released") with NoStackTrace
+
+  private[this] val DatacenterErrorMessage = "No path to datacenter"
 
   case class Stats(stats: StatsReceiver) {
     val opens = stats.counter("opens")
@@ -80,6 +83,16 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
+          case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage =>
+            log.log(
+              failureLogLevel,
+              s"consul datacenter $datacenter service ${key.name} " +
+                s"lookup request timed out due to invalid dc error $e."
+            )
+            state.update(Addr.Neg)
+            val backoff #:: nextBackoffs = backoffs
+            // subsequent errors are logged as DEBUG
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
           case Throw(e: IndividualRequestTimeoutException) =>
             // catch request timeout exceptions for Consul API. Use last known good state.
             stats.errors.incr()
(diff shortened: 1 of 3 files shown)


namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
```diff
@@ -573,72 +573,10 @@ class ConsulNamerTest extends FunSuite with Awaits {
     @volatile var state: Activity.State[NameTree[Name]] = Activity.Pending
     namer.lookup(Path.read("/dc1/servicename/residual")).states respond { state = _ }
 
-    assert(state == Activity.Ok(NameTree.Neg))
-  }
-
-  test("Namer state returns Neg then Bound then Neg when a datacenter becomes available and then becomes unavailable") {
```
dadjeibaah: We no longer need this test since this tests the behavior of having a datacenter to become unavailable. we don't expect this to happen at runtime.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -80,6 +82,19 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
+            // this exception case checks if we queried for a service in a datacenter that
+            // doesn't exist. We capture this case so that we can return Addr.Neg to prevent
+            // service name resolution from timing out.
+          case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage =>
+            log.log(
+              failureLogLevel,
+              s"consul datacenter $datacenter service ${key.name} " +
```
adleong: This isn't a timeout, right?  it's just the consul API reporting that the requested DC does not exist.  I think we can improve the error message here to reflect that.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -80,6 +82,19 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
+            // this exception case checks if we queried for a service in a datacenter that
+            // doesn't exist. We capture this case so that we can return Addr.Neg to prevent
+            // service name resolution from timing out.
+          case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage =>
```
adleong: maybe include a link to the consul issue you filed in a comment here, as justification for why we need to do this.

namer/consul/src/test/scala/io/buoyant/namer/consul/ConsulNamerTest.scala
adleong: Are you sure? I don't see any reason why a DC couldn't become unavailable.  I would expect in that case that the vaddr would should go from having a value to being neg.

dadjeibaah: Looking through the comments in #2145, the user mentioned that 
>  A DC doesn't usually cease to exist during runtime
Based on that, I felt that the test wasn't necessary since this issue only happens on the initial call to resolve the logical. But since this can be considered edge case, it might be helpful to have this test.

edio:
I'm not 100% certain, but I believe, consul responds with the same message, when DC exists, but network paritioning happens, so a server receiving an initial request for resolution can't communicate with a server from _existent_ but not accessible DC.

It would be great if namerd could survive such intermittent failures. I do see the necessity of returning `Neg` right away if DC does not exist, though...

Maybe we should consider previous state when making a decision whether to update to `Neg` or not? If an address was ever resolved to a non-neg, it is very unlikely, I believe, that `No path to datacenter` is a truthful message. If you still think, current behavior is more correct, maybe we could agree on some toggle in config to let multiple behaviors?

dadjeibaah:
@edio, thanks for shedding light on this. Interesting that Consul still sends this message even when there is seemingly nothing wrong with the server. We could modify this so that it only returns Neg when the previous state is Pending.

edio:
@dadjeibaah 
> modify this so that it only returns Neg when the previous state is Pending

I think this is the right solution. I will create an issue on GH.

PR #2192, opened 2018-12-11
Fix silent failure when watchable object receives an HTTP 404 from k8s apiserver

### LINKED ISSUE #2167: namerd stops watching dtabs in k8s if apiserver intermittently throws 404
## Filing a Linkerd issue ##

Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:
namerd stops watching dtabs if k8s apiserver throws 404 for dtabs custom resource

Bare minimum, namerd should log a WARNING that says it stopped watching dtab resource so that operator can diagnose the issue easily. Perhaps, namerd should continue to try to restart the watch

We are still trying to figure out why apiserver throws 404 when it shouldn't because resource exists. But for now, do the following
1) create k8s dtabs custom resource
2) start namerd
3) add new dtabs
4) delete the custom resource
5) namerd stops watching without logging any WARN/INFO/ERROR message

Most likely this [line of code](https://github.com/linkerd/linkerd/blob/c03116375eace9b3ec4acaf60e5f671083903314/k8s/src/main/scala/io/buoyant/k8s/resources.scala#L235) is the culprit

**Environment**:
- linkerd/namerd version, config files: 1.5.0
- Platform, version, and config files (Kubernetes, DC/OS, etc): k8s 1.10.5
- Cloud provider or hardware configuration: doesn't matter

Let me know if you need more information!

dadjeibaah:
The implemented `restartWatches` method  in `NsListResource` and `NsObjectResource` currently have a bug where HTTP 404 responses from a k8s apiserver could cause the method to fail silently. This has the effect of stopping watch streams whenever HTTP 404 response is encountered causing  k8s API resources to be removed from Linkerd and Namerd's watch context. 

This PR fixes the issue by pattern matching the response to a `get` request in the `restartWatches` method since it returns an `Option[O]` where O is the type of a k8s resource object. In the case where we receive a `None` the method needs to return a default empty value in order for the underlying `Watchable` object to restart the watch successfully.

fixes #2167

### DIFF
--- a/k8s/src/main/scala/io/buoyant/k8s/resources.scala
+++ b/k8s/src/main/scala/io/buoyant/k8s/resources.scala
@@ -152,15 +152,18 @@ private[k8s] class ListResource[O <: KubeObject: TypeReference, W <: Watch[O]: O
   override protected def restartWatches(
     labelSelector: Option[String],
     fieldSelector: Option[String]
-  ): Future[(Seq[W], Option[String])] =
+  ): Future[(Seq[W], Option[String])] = {
     get(
       labelSelector = labelSelector,
       fieldSelector = fieldSelector,
       retryIndefinitely = true
-    ).map { maybeList =>
-      val list = maybeList.get // list resources should always exist
-      (list.items.map(od.toWatch), list.metadata.flatMap(_.resourceVersion))
+    ).map {
+      case Some(list) => (list.items.map(od.toWatch), list.metadata.flatMap(_.resourceVersion))
+      case None =>
+        log.debug(s"Initial request for api path $watchPath resulted in empty response")
+        (Seq.empty, None)
     }
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


thedebugger:
@dadjeibaah hey so i think this won't fix our problem. Let me try to capture the state transitions
1) namerd makes a watch dtab storage call to the API server 
2) if the call succeeds watch is established, namerd listens for events. This PR logs the watch events if i'm reading the code correctly. Right?
3) if the first fails for any reason (this call fails with 404s), namerd tries to restart the watch by making a GET call (to get updated version) on the resource. We see 404s in this call too. So i think we should add a WARN for this case.

Let me know if this makes sense. Feel free to ping me on slack if you want to discuss more.

k8s/src/main/scala/io/buoyant/k8s/resources.scala
```diff
@@ -231,11 +234,15 @@ private[k8s] class NsObjectResource[O <: KubeObject: TypeReference, W <: Watch[O
...
+    ).map {
+      case Some(nsObjects) =>
+        val watch = Seq(nsObjects).flatMap { obj => Seq(od.toWatch(obj)) }
+        // object watches don't supply a resource version, since the version associated
+        // with the object is tied to when the object was last modified, and therefore
+        // it may be too old to successfully establish a watch
+        (watch, None)
+      case None =>
+        log.debug(s"Initial request for api path $watchPath resulted in empty response")
+        (Seq.empty, None)
```
adleong: Isn't this already the behavior?  `None.toSeq.flatMap{ ... }` will be `Seq.empty`.  If you just want to add the logging, you can hide it behind an `if (watch.isEmpty)`

dadjeibaah:
I am going to close this since I am unable to reproduce the issue this PR is intended to fix. Trying to create a unit test also did not recreate the scenario. The issue was that if namerd tried to watch a dtab and the first call receives a 404 and then, after a retry receives a 404 again, Namerd will stop watching dtabs. However, during my tests, when namerd received a 404 for the first time from, the watch was unable to be established and a log line was displayed through this line of code

https://github.com/linkerd/linkerd/blob/6702ab00c1baf025a7695dcab8208be6330d98f5/namerd/storage/k8s/src/main/scala/io/buoyant/namerd/storage/K8sDtabStore.scala#L67

Please add anymore context if there is something I am missing @thedebugger.

PR #2205, opened 2019-01-15
Retain last state on "No path to datacenter" error

### LINKED ISSUE #2200: Allow namerd to survive intermittent consul "no path to datacenter" error
## Filing a Linkerd issue ##

Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
https://github.com/linkerd/linkerd/pull/2186 allows _namerd_ to fallback to last known state of an observed service in _consul_, unless consul responds with `500: No path to datacenter`.

This allows _namerd_ to reply immediately with `Neg` if queried DC is not known to consul. This behavior is required for correct cross-dc fallbacks for example.

However, consul responds with the same error when network partition occurs. Hence, even if DC exists, but consul cluster is intermittently in a bad shape, namerd will update all services to `Neg`.

_Namerd_ won't update services to `Neg`, if they are already resolved and consul responds with `500: No path to datacenter`. Only `Pending → Neg` transition should be allowed on that response from _consul_.

More details are in the corresponding pull-request
https://github.com/linkerd/linkerd/pull/2186

edio:
Consul responds with "No path to datacenter" when it is queried about DC
that does not exist (case 1) or DC that can't be contacted due to
network partition (case 2). Currently consul namer responds with
Addr.Neg on such response.
This makes sense in (case 1), because we want to fallback to
explore alternatives in dtab asap when we try to resolve new name.
However in (case 2) we want to retain previously known state (if it
exists), so namerd could survive intermittent consul failures caused by
network partition.

This commit adjusts consul namer so it only returns Addr.Neg if current
state is Addr.Pending. Otherwise, previous address is retained.

Fixes #2200

### DIFF
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -96,7 +96,7 @@ private[consul] object SvcAddr {
             state.update(Addr.Neg)
             val backoff #:: nextBackoffs = backoffs
             // subsequent errors are logged as DEBUG
-            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentState))
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, Addr.Neg))
           case Throw(e: IndividualRequestTimeoutException) =>
             // catch request timeout exceptions for Consul API. Use last known good state.
             stats.errors.incr()
--- a/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
+++ b/namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
@@ -67,7 +67,7 @@ private[consul] object SvcAddr {
     InstrumentedVar[Addr](Addr.Pending) { state =>
       stats.opens.incr()
       @volatile var stopped: Boolean = false
-      def loop(blockingIndex: Option[String], backoffs: Stream[Duration], failureLogLevel: Level, currentValueToLog: Addr): Future[Unit] = {
+      def loop(blockingIndex: Option[String], backoffs: Stream[Duration], failureLogLevel: Level, currentState: Addr): Future[Unit] = {
 
         if (stopped) Future.Unit
         else getAddresses(blockingIndex).transform {
@@ -82,12 +82,12 @@ private[consul] object SvcAddr {
             )
             stopped = true
             Future.Unit
-            // this exception case checks if we queried for a service in a datacenter that
-            // doesn't exist. We capture this case so that we can return Addr.Neg to prevent
-            // service name resolution from timing out. Since Consul's HTTP 5xx API response is
-            // overloaded and represents different error states, its necessary for us to capture this
-            // exception and resolve to Addr.Neg see https://github.com/hashicorp/consul/issues/4901
-          case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage =>
+          // this exception case checks if we queried for a service in a datacenter that
+          // doesn't exist. We capture this case so that we can return Addr.Neg to prevent
+          // service name resolution from timing out. Since Consul's HTTP 5xx API response is
+          // overloaded and represents different error states, its necessary for us to capture this
+          // exception and resolve to Addr.Neg see https://github.com/hashicorp/consul/issues/4901
+          case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage && currentState == Addr.Pending =>
             log.log(
               failureLogLevel,
               s"consul datacenter $datacenter service ${key.name} " +
@@ -96,7 +96,7 @@ private[consul] object SvcAddr {
             state.update(Addr.Neg)
             val backoff #:: nextBackoffs = backoffs
             // subsequent errors are logged as DEBUG
-            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentState))
           case Throw(e: IndividualRequestTimeoutException) =>
             // catch request timeout exceptions for Consul API. Use last known good state.
             stats.errors.incr()
@@ -104,11 +104,11 @@ private[consul] object SvcAddr {
               failureLogLevel,
               "consul datacenter '%s' service '%s' lookup request timed out %s." +
                 " Last known state is %s",
-              datacenter, key.name, e, currentValueToLog
+              datacenter, key.name, e, currentState
             )
             val backoff #:: nextBackoffs = backoffs
             // subsequent errors are logged as DEBUG
-            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentState))
 
           case Throw(e) =>
             // do not update state, log error and continue polling with backoff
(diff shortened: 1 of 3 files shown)


namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -82,12 +82,12 @@ private[consul] object SvcAddr {
...
-            // service name resolution from timing out. Since Consul's HTTP 5xx API response is
-            // overloaded and represents different error states, its necessary for us to capture this
-            // exception and resolve to Addr.Neg see https://github.com/hashicorp/consul/issues/4901
-          case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage =>
+          // this exception case checks if we queried for a service in a datacenter that
+          // doesn't exist. We capture this case so that we can return Addr.Neg to prevent
+          // service name resolution from timing out. Since Consul's HTTP 5xx API response is
+          // overloaded and represents different error states, its necessary for us to capture this
+          // exception and resolve to Addr.Neg see https://github.com/hashicorp/consul/issues/4901
+          case Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage && currentState == Addr.Pending =>
```
edio: Now I think, do we really need to parse the error message? Don't we want to have the same behavior as we have on this line for all errors? Wouldn't it be correct to return `Neg`on any error if we haven't resolved anything yet instead of just stucking in `Pending`?

This would simplify code as well.

adleong: I think you're right.  The difference in behavior from today would be that if we get a 5XX which is not a "No path to DC" error while the Addr is pending, we would immediately update it to Neg (and then retry) instead of leaving it as Pending and retrying.  But I think that's okay.

It's been a while since I've read this code so I'm trying to remember how it all works.  I don't see anything in here for dealing with 404s.  Do you remember how SvcAddr deals with service deletion?  When a service is deleted, does that catalog call just return an empty address list?

edio: > The difference in behavior from today would be that if we get a 5XX which is not a "No path to DC" error while the Addr is pending, we would immediately update it to Neg (and then retry) instead of leaving it as Pending and retrying. But I think that's okay.

Pending is generally bad, because, I believe, it will make linkerd to timeout eventually without letting namerd exploring other branches of dtab.

Let's describe all possible kinds of errors we can expect from consul
1. Consul agent can't be contacted
already handled in code partially by `Throw(e: IndividualRequestTimeoutException)`, partially by `Throw(e)`
Namer should retain last known state :heavy_check_mark:
If it is an initial request, Namer should return `Addr.Neg` :question: Currently it is not the case, and I think it is not entirely correct. `Addr.Pending` will be returned right now and if it remains like that for more than 10 seconds, linkerd will return 502. If it was `Addr.Neg`, namerd could try other discovery systems in the meantime (for example, try dns, or try contacting consul server directly instead of local agent (given proper namerd configuration and dtab of course)). 
Basically it is a choice between "let requests hang for 10 seconds and then fail" and "let requests fail right away". But with the latter option we can try other discovery systems before failing (calls to other consul datacenters will likely fail too).

2. Consul agent can be contacted but its response is not a response we expect
already handled in code by `Throw(e)`
Basically same as case 1. In the code the only difference is the message logged.

3. Consul agent can be contacted but, but not all servers can be contacted (No path to datacenter error)
Handled in code by `Throw(e: UnexpectedResponse) if e.rsp.contentString == DatacenterErrorMessage`
Namerd should retain last known state :heavy_check_mark:  (fixed with this commit)
If it is an initial request, Namer should return `Addr.Neg`.
Here, again, it is a choice between "let requests hang for 10 seconds and then fail" and "let requests fail right away". However, with the second option we can try other datacenters as well.

Would you agree, that behavior in those 3 cases looks very similar, and in code it is actually the same, so 3 cases can be merged into one? Only logged messages should be different.

adleong: @edio This is a great write-up!  Yes, I agree, I think it makes sense to merge these 3 cases.

edio: @adleong I'll fix the code shortly. Thanks!

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -117,11 +117,11 @@ private[consul] object SvcAddr {
...
-              datacenter, key.name, e, currentValueToLog
+              datacenter, key.name, e, currentState
             )
             val backoff #:: nextBackoffs = backoffs
             // subsequent errors are logged as DEBUG
-            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentValueToLog))
+            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, currentState))
 
           case Return(v1.Indexed(_, None)) =>
             // If consul doesn't return an index, we're in bad shape.
```
edio: I just checked 2 months of our logs (14 namerd instances talking to 5 datacenters in a pretty flaky consul cluster) and this message hasn't happenned even once.

@adleong , do you happen to recall why was this branch added in the first place? As of today, it looks to me, as if this was effectively dead code.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
edio: @dadjeibaah , my point is that other cases should behave the same way "No path to datacenter" case behaves after this PR.

If a name was never resolved — return `Addr.Neg` and let fallback to other parts of dtab
If a name was ever resolved before — return result of previous resolution

The code that handles all cases will look like this (still working on tests, so can't push)

```
          case Throw(e) =>
            // Update state only if it is Pending to not let linkerd hang on name resolution
            // Otherwise retain last known state to allow namerd survive intermittent failures
            val effectiveState = if (currentState == Addr.Pending) {
              state.update(Addr.Neg)
              Addr.Neg
            } else {
              currentState
            }

            log.log(
              failureLogLevel,
              "consul datacenter %s service %s lookup request received invalid dc error %s. Current state is %s",
              datacenter, key.name, e, effectiveState
            )
            // subsequent errors are logged as DEBUG
            Future.sleep(backoff).before(loop(None, nextBackoffs, Level.DEBUG, effectiveState))
```

Let me know please if I'm still missing something with this chane.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
adleong: @edio I think you're right.  It looks like these two branches used to do different things (a 5XX would update the state to Neg whereas a timeout would not) but a recent change (https://github.com/linkerd/linkerd/commit/a17406b43a928a59c8d1bdd94ddcb0f0bc8588cb) effectively make these branches identical.  I think we can safely unify them.

edio: @adleong , I believe, though, you replied to a wrong thread ;)

adleong: whoops, you're right.  I don't think this was added for any specific use-case, but just as standard defensive programming.  See: https://github.com/linkerd/linkerd/pull/632/files#diff-ea0487e1d8c6b88af1a70cdbf4dc02f6R55

edio: I see. I personally think, this code, does more harm than good. I concluded from the code, that there's some case, where consul behaves like this. And, as I said, in the last 2 months we haven't seen this branch of code executed even once, so I'm almost certain, it is indeed never executed.

I don't have strong opinion on this, but I'd prefer to remove this branch of code. Would you mind if I did that?

adleong: I see where you're coming from.  This branch also is questionable because it halts the request loop instead of restarting the watch or something like that, which is probably not great.  But still, since it's mostly unrelated to this change, I'd prefer we don't make changes unless we have a clear motivation for doing so.

edio: I reverted my changes in that place in the code to keep this PR as relevant to the issue as possible.

PR #2215, opened 2019-01-30
Add support for setting the backlog queue size under socketOptions

### LINKED ISSUE #2214: Add support for backlog under socketOptions
Finagle supports the ability to set the backlog queue size against the socket. Add a new option under socketOptions to support this. This is useful if you want the OS to rate limit the rate of incoming connections by the queue size.

chrisgoffinet:
Fixes #2214

Signed-off-by: Chris Goffinet <chris@threecomma.io>

### DIFF
--- a/linkerd/docs/config.md
+++ b/linkerd/docs/config.md
@@ -275,7 +275,7 @@ enabled | true | If set to true, data is sent to Buoyant once per hour
 ### Socket Options
 
 Linkerd supports configuring socket level options for any given interface.
-i.e. the admin and router interfaces. These configurations are only available 
+i.e. the admin and router interfaces. These configurations are only available
 on Linux 3.9 distributions and newer.
 
 Key | Default Value | Description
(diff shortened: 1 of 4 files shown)


linkerd/docs/config.md
```diff
@@ -286,3 +286,4 @@ reusePort | false | If set to true, enables the `SO_REUSEPORT` option, which can
 readTimeoutMs | unbounded | Configures this client or server with given transport-level socket read timeout in milliseconds.
 writeTimeoutMs | unbounded | Configures this client or server with given transport-level socket write timeout in milliseconds.
 keepAlive | false | If set to true, enables the `SO_KEEPALIVE` option, which will enable keep alive on the socket.
+backlog | None | If set will adjust the backlog queue size of the socket, a default of None falls back to the OS defined value of SOMAXCONN
```
dadjeibaah: TIOLI: Reworded this a little bit. If set, this config will adjust the backlog queue size of the socket, a default of None falls back to the OS defined value of SOMAXCONN

adleong: Alternatively, just put `SOMAXCONN` in the Default value column instead of `None`.

chrisgoffinet: Unfortunately there is no default for SOMAXCONN, it depends on the kernel :), by setting it to None, it tells Finagle to not set it, and let the OS handle it

PR #2220, opened 2019-02-11
Fix typo in docs

JoeWrightss:
"explicity" to "explicitly" in line 62 and 94.

### DIFF
--- a/linkerd/docs/response_classifier.md
+++ b/linkerd/docs/response_classifier.md
@@ -91,7 +91,7 @@ parameter `successStatusCodes` that accepts a user-defined list of gRPC
 status codes that should always be classified as `Success`.
 
 <aside class="notice">
-When defining `successStatusCodes`, it is up to the user to explicity classify
+When defining `successStatusCodes`, it is up to the user to explicitly classify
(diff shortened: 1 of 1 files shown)


istio-proto/src/main/protobuf/proxy/v1/config/dest_policy.proto
```diff
@@ -59,7 +59,7 @@ message DestinationPolicy {
 // If tags are omitted, the policy applies for all versions of the
 // service. Policy CANNOT BE empty.
 // *Note:* Destination policies will be applied only if the corresponding
-// tagged instances are explicity routed to. In other words, for every
+// tagged instances are explicitly routed to. In other words, for every
```
adleong: This is a direct mirror of https://github.com/istio/api/blob/6e481630954efcad10c0ab43244e8991a5a36bfc/proxy/v1/config/dest_policy.proto#L62 so we shouldn't edit this directly, even to fix typos

JoeWrightss: Thank you for reminding me, @adleong.😸

PR #2221, opened 2019-02-12
add support for HTTP/2 ping frame and add failure detector for mesh interpreter

### LINKED ISSUE #2219: Support PING frame in H2
In the HTTP/2 spec (https://http2.github.io/http2-spec/#PING) there is support for handling PING frames. I suggest we support this for the mesh interpreter. The biggest reason would be for users of Linkerd who put Namerd in front of a load balancer. In cloud environments like GCP where they provide elastic load balancing, they have a fixed idle timeout of 600s that isn't configurable. I've also seen this problem happen in action, so it's not theoretical. We've had to fallback to the ThriftMux interpreter because Finagle implements its own ping like mechanism to keep the connection open.

I've considered the option of enabling `SO_KEEPALIVE` on the client socket, but the issue we face here is that currently Netty/Finagle does not expose the ability to set custom keepalive settings (intv, probes, etc) which means that _even_ if we were to support this option, it would require the user to adjust their TCP settings at the host level and for large infrastructure environments, this is a non-starter. I think overall it makes since that the H2 implementation support `PING` instead of relying on the transport.

dadjeibaah:
The current HTTP/2 implementation does not support ping frames between an H2 server and client. This makes it difficult to implement a failure detector in Linkerd or Namerd to ensure that connections are alive and in a good state.

To fix this issue, This PR, adds a "listener" to the existing implementation that notifies the transport when a PING frame is received on both the client and the server. This PR also implements a failure detector for HTTP/2 connections between Linkerd and Namerd which ensures that no stale connections between the two instance stay up longer than a given period of time.

I conducted two different tests.
- Linkerd <> HA Proxy <> Namerd
1. Start HA Proxy with a tcp timeout of 1000ms for the client and server facing interface of the proxy
2. Start up Linkerd and Namerd. Linkerd streams dtab updates through `io.l5d.mesh` through HA Proxy
3. Set up a failure threshold for Linkerd's `interpreter` with a `minPeriod` of 500ms and a closeTimeout` of 1000ms
4. Observe the connections between Linkerd and HA Proxy. They should never be closed by HA Proxy since we are sending Ping frames to Namerd. Before this PR. The connection would always be torn down by HA Proxy with this same configuration.

- Linkerd <> 2 x Namerd
1. Start two instances of Namerd 
2. Start Linkerd and point its `interpreter` to both Namerd. 
3. Send a request through Linkerd, this should make Linkerd stream dtabs from the first Namerd instance.
4. `SIGSTOP` the first Namerd process. This forces the HTTP/2 connection to trigger the failure detector which will close the connection. Linkerd will switch to using the second Namerd process for dtab resolution.
  

Fixes #2219 

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
@@ -37,5 +37,7 @@ object H2Transport {
     def goAway(err: GoAway, deadline: Time): Future[Unit]
 
     final def goAway(err: GoAway): Future[Unit] = goAway(err, Time.Top)
+
+    def ping(): Future[Unit]
   }
 }
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
@@ -38,6 +38,6 @@ object H2Transport {
 
     final def goAway(err: GoAway): Future[Unit] = goAway(err, Time.Top)
 
-    def ping(): Future[Unit]
+    def sendPing(): Future[Unit]
   }
 }
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
@@ -38,6 +38,11 @@ object H2Transport {
 
     final def goAway(err: GoAway): Future[Unit] = goAway(err, Time.Top)
 
+    /**
+     * Writes a Non-ACK PING frame on the transport. This method does not send
+     * ACK PING frames on a stream. ACK PING frames are automatically
+     * sent by the underlying Netty H2 implementation.
+     */
     def sendPing(): Future[Unit]
   }
 }
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
@@ -180,7 +180,9 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
           // client sent a ping and got acknowledgement. Satisfy promise to let failure detector
           // know the connection is still alive.
           val prevPingPromise = pingPromise.getAndSet(null)
-          prevPingPromise.setDone()
+          if (prevPingPromise != null) {
+            prevPingPromise.setDone()
+          }
         }
         transport.read().transform(loop)
 
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
@@ -11,7 +11,7 @@ import com.twitter.logging.Logger
 import com.twitter.util._
 import io.netty.handler.codec.http2._
 import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger, AtomicReference}
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 
@@ -21,12 +21,7 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
   protected[this] def prefix: String
   protected[this] def stats: StatsReceiver
   protected[this] def failureThreshold: Option[FailureDetector.Config] = None
-
   protected[this] def transport: Transport[Http2Frame, Http2Frame]
-  protected[this] def failureDetector: FailureDetector = failureThreshold match {
-    case None => FailureDetector(NullConfig, () => Future.never, stats)
-    case Some(cfg) => FailureDetector(cfg, (ping _), stats)
-  }
 
   protected[this] lazy val writer: H2Transport.Writer = Netty4H2Writer(transport)
   /**
@@ -50,32 +45,44 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
     streams.size()
   }
 
-  @volatile private[this] var pingFramePromise: Promise[Unit] = null
+  private[this] val pingPromise: AtomicReference[Promise[Unit]] = new AtomicReference(null)
 
   protected[this] val OutstandingPingEx = Failure("Outstanding ping on HTTP/2 connection")
   protected[this] val eventLoop = transport.context match {
-    case trans: HasExecutor => trans.executor
-    case _ => null
+    case trans: HasExecutor => Some(trans.executor)
+    case _ => None
   }
 
   protected[this] def ping(): Future[Unit] = {
     val done = new Promise[Unit]
-    eventLoop.execute(
-      new Runnable {
-        override def run(): Unit = {
-          if (pingFramePromise == null) {
-            writer.ping()
-            pingFramePromise = done
-          } else {
-            done.setException(OutstandingPingEx)
+    eventLoop match {
+      case None =>
+        // Immediately satisfy the promise to prevent the failure detector from marking the
+        // the connection as closed if we don't have a way to execute a ping request on another
+        // thread.
+        done.setDone()
+        done
+      case Some(executor) =>
+        executor.execute(
+          new Runnable {
+            override def run(): Unit = {
+              val prevPromise = pingPromise.getAndSet(done)
+              if (prevPromise == null || prevPromise.isDone) {
+                val _ = writer.sendPing()
+              } else {
+                done.setException(OutstandingPingEx)
+              }
+            }
           }
-        }
-      }
-    )
-    done
+        )
+        done
+    }
   }
 
-  private[this] val detector = failureDetector
+  protected[this] val failureDetector: FailureDetector = failureThreshold match {
+    case None => FailureDetector(NullConfig, null, null)
+    case Some(cfg) => FailureDetector(cfg, (ping _), stats)
+  }
 
   private[this] val closedId: AtomicInteger = new AtomicInteger(0)
   @tailrec private[this] def addClosedId(id: Int): Unit = {
@@ -139,11 +146,6 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
     transport.read().rescue(wrapRemoteEx)
 
   protected[this] def demux(): Future[Unit] = {
-    detector.onClose.ensure {
-      log.debug(s"[%s] did not receive HTTP/2 ping ack from remote [%s]", prefix, transport.context.remoteAddress)
-      goAway(GoAway.InternalError); ()
-    }
-
     lazy val loop: Try[Http2Frame] => Future[Unit] = {
       case Throw(e) if closed.get =>
         log.debug("[%s] dispatcher closed: %s", prefix, e)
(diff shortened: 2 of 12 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
```diff
@@ -37,5 +37,7 @@ object H2Transport {
     def goAway(err: GoAway, deadline: Time): Future[Unit]
 
     final def goAway(err: GoAway): Future[Unit] = goAway(err, Time.Top)
+
+    def ping(): Future[Unit]
```
adleong: The semantics of this are a bit unexpected, so this should definitely have a comment.  I think the returned future is satisfied when the ping is written, not when the ack is received as one might expect.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -17,10 +20,15 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
   protected[this] def log: Logger
   protected[this] def prefix: String
   protected[this] def stats: StatsReceiver
+  protected[this] def failureThreshold: Option[FailureDetector.Config] = None
 
   protected[this] def transport: Transport[Http2Frame, Http2Frame]
-  protected[this] lazy val writer: H2Transport.Writer = Netty4H2Writer(transport)
+  protected[this] def failureDetector: FailureDetector = failureThreshold match {
+    case None => FailureDetector(NullConfig, () => Future.never, stats)
```
adleong: the second two arguments are ignored when `NullConfig` is passed in.  Consider passing `null` for them so that their values aren't red herrings.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -42,6 +50,33 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
     streams.size()
   }
 
+  @volatile private[this] var pingFramePromise: Promise[Unit] = null
+
+  protected[this] val OutstandingPingEx = Failure("Outstanding ping on HTTP/2 connection")
+  protected[this] val eventLoop = transport.context match {
+    case trans: HasExecutor => trans.executor
+    case _ => null
```
adleong: Do we ever expect to hit this branch?  If so, consider making `eventLoop` an `Option`.  Otherwise, consider throwing a more explicit error here instead of letting this trigger an NPE somewhere else.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -42,6 +50,33 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
...
+  protected[this] def ping(): Future[Unit] = {
+    val done = new Promise[Unit]
+    eventLoop.execute(
+      new Runnable {
+        override def run(): Unit = {
+          if (pingFramePromise == null) {
+            writer.ping()
+            pingFramePromise = done
+          } else {
+            done.setException(OutstandingPingEx)
```
adleong: possible race condition where `done` could be set to null before this is called

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -42,6 +50,33 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
...
+    case _ => null
+  }
+
+  protected[this] def ping(): Future[Unit] = {
+    val done = new Promise[Unit]
+    eventLoop.execute(
+      new Runnable {
+        override def run(): Unit = {
+          if (pingFramePromise == null) {
+            writer.ping()
```
adleong: personally, I like writing `val _ = write.ping()` to signal that there is a fire-and-forget future being dropped on the floor here.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -42,6 +50,33 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
     streams.size()
   }
 
+  @volatile private[this] var pingFramePromise: Promise[Unit] = null
```
adleong: Consider making this an AtomicReference and using cas state transitions to avoid race conditions.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -104,6 +139,11 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
     transport.read().rescue(wrapRemoteEx)
 
   protected[this] def demux(): Future[Unit] = {
+    detector.onClose.ensure {
```
adleong: This will register an additional callback every time this dispatcher receives a frame.  I think this should probably be set up when the detector is created.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -104,6 +139,11 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
...
+      log.debug(s"[%s] did not receive HTTP/2 ping ack from remote [%s]", prefix, transport.context.remoteAddress)
+      goAway(GoAway.InternalError); ()
```
adleong: I'm looking at the mux implementation for comparison and it doesn't explicitly tear down the connection on close of the detector.  Instead, it just sets its status to the detector's status.

I guess some other layer of the stack closes the service if the status becomes closed, maybe?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
```diff
@@ -128,9 +168,21 @@ trait Netty4DispatcherBase[SendMsg <: Message, RecvMsg <: Message] {
...
 
-      case Return(f: Http2SettingsFrame) =>
+      case Return(_: Http2SettingsFrame) =>
         // We received a settings frame, keep reading from the transport
         transport.read().transform(loop)
+      case Return(f: Http2PingFrame) =>
+        if (f.ack() && pingFramePromise != null) {
+          // client sent a ping and got acknowledgement. Satisfy promise to let failure detector
+          // know the connection is still alive.
+          pingFramePromise.setDone()
```
adleong: possible race condition where this could become null

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -298,6 +298,22 @@ object H2FrameCodec {
         ctx.fireChannelRead(data)
         0 // bytes are marked as consumed via WindowUpdateFrame writes
       }
+
+      override def onPingRead(
```
adleong: do we need this? we never need to handle pings ourselves, right?

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
```diff
@@ -298,6 +298,22 @@ object H2FrameCodec {
...
+        ctx: ChannelHandlerContext,
+        data: Long
+      ): Unit = {
+        val ping = new DefaultHttp2PingFrame(data)
+        ctx.fireChannelRead(ping); ()
+      }
+
+      override def onPingAckRead(
```
adleong: perhaps add a comment that we need to propagate ping acks so that we can inform the failure detector that an ack has been received.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
olix0r: Consider renaming the method to `sendPing()`, which may help to clarify this.

adleong:
Can you talk a bit about how to test this?  Have you used wireshark to observe the ping frames?  Is there a way to simulate a ping timeout and observe the behavior?

finagle/h2/src/main/scala/io/netty/handler/codec/http2/H2FrameCodec.scala
dadjeibaah: That's right. I think this must have been left over code.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
dadjeibaah: Looking at this a little more, the `done` references the promise within the method and in this case, when we set the exception here, we are notifying the Failure detector that things went wrong.

adleong: ah, you're right

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4DispatcherBase.scala
dadjeibaah: Testing this out with the SIGSTOP scenario we discussed. Without this, the client side stops pinging but the connection is still `ESTABLISHED`. I am not really sure if there is some kind of cleanup layer in the stack that detects if a service is closed. But I think we may have to tear down the connection manually.

olix0r: My understanding is that failure detection is interacts with the load balancer & probation, so that "failed" nodes are not considered for new requests.

These pings _can_ cause the _operating system_ to notify the process that a closed connection was actually closed -- i.e. if writing to the socket fails. (Otherwise, we may not try to write to the socket and wouldn't have cause to be notified of the disconnect).

SIGSTOPing a process for just a short time probably wouldn't be enough to cause the OS to tell the process the connection has closed. It should be sufficient to see the load balancer put the endpoint in probation; though this would only impact the LB, really, if there are more than one endpoint to choose from.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2Transport.scala
```diff
@@ -37,5 +37,7 @@ object H2Transport {
     def goAway(err: GoAway, deadline: Time): Future[Unit]
 
     final def goAway(err: GoAway): Future[Unit] = goAway(err, Time.Top)
+
+    def sendPing(): Future[Unit]
```
adleong: :+1: to calling this `sendPing`.  I still think a comment on the semantics wouldn't hurt.

adleong:
It seems that Linkerd is still attempting to send a request to `127.0.0.1:4322` (the Namerd which is stopped) and I guess this triggers a local goaway.  I would have expected the load balancer to not even attempt to send requests to that endpoint.

dadjeibaah:
I looked into this some more and it looks like there might be some behavior in the`SingletonPool` module that forces the service associated with `127.0.0.1:4322` to always stay open because the pool ensures that there is a most one service that is available for that connection. I will investigate further to see why that is the case.

dadjeibaah:
@chrisgoffinet @thedebugger have you been able to test out this branch? What are your thoughts on the current implementation for failure detection between Linkerd and Namerd connections? RIght now, if the failure detector on the client side is triggered, the client will terminate the connection and establish a new one. WDYT?

thedebugger:
@dadjeibaah I'll try if i can go over this tonight. Thanks for the PR

thedebugger:
@dadjeibaah @adleong we are trying to test this today. Hopefully, I will have an update on it later today

PR #2222, opened 2019-02-15
Issue 2212: Fix Namerd memory leak

### LINKED ISSUE #2212: Namerd memory leak
Creating this issue as a follow-up regarding the discussions with @dadjeibaah in slack channel.

Issue Type:
- [ ] Bug report

**What happened**:
Namerd's memory usage keeps increasing and will not perform garbage collection until its usage hit 80%~100%. A possible consequence is that, if the GC cannot be finished properly, the n4d pods will unable to serve any requests unless it's restarted. A side fact is that, linkerd's memory usuage has ever gone up higher than 20%, which might mean Namerd is not releasing the objects/resources properly which caused a memory leak.

Namerd Memory Usage Flow
screen shot 2019-01-28 at 1 59 43 pm

Linkerd Memory Usage Flow
screen shot 2019-01-28 at 1 58 33 pm
 
Namerd's memory usage flow should be similar to what it's shown for Linkerd.

N/A

**Environment**:
Namerd:1.6.0

Namerd config file
```
admin:
  ip: 0.0.0.0

telemetry:
- kind: io.l5d.prometheus
  prefix: l5d_n4d_

storage:
  kind: io.l5d.zk
  pathPrefix: /dtabs
  zkAddrs:
  - host: hostname1
    port: 2181
  - host: hostname2
    port: 2181
  - host: hostname3
    port: 2181

namers:
- kind: io.l5d.k8s
  prefix: -
  host: 127.0.0.1
  port: 8001
  transformers:
  - kind: io.l5d.k8s.daemonset
    namespace: mesh
    k8sHost: 127.0.0.1
    k8sPort: 8001
    port: in-http
    service: l5d
- kind: io.l5d.k8s
  prefix: -
  host: 127.0.0.1
  port: 8001
  transformers:
  - kind: io.l5d.k8s.daemonset
    namespace: mesh
    k8sHost: 127.0.0.1
    k8sPort: 8001
    port: in-grpc
    service: l5d
- kind: io.l5d.k8s
  prefix: -
  host: 127.0.0.1
  port: 8001

interfaces:
- kind: io.l5d.mesh
  ip: 0.0.0.0
  port: 4321
- kind: io.l5d.httpController
  ip: 0.0.0.0
  port: 4180
```

Kubernetes

kleimkuhler:
# Problem

Namerd's memory usage keeps increasing and will not perform garbage collection
until its usage hit 80%~100%. A possible consequence is that, if the GC cannot
be finished properly, the n4d pods will unable to serve any requests unless it's
restarted. A side fact is that, linkerd's memory usuage has ever gone up higher
than 20%, which might mean Namerd is not releasing the objects/resources
properly which caused a memory leak.

# Solution

namerd keeps open watches for every `getDtab` request that is made. When the value of that request is fulfilled, the watch for that request is kept in the `ZKWatcherManager` as seen in [1] after 5k requests sent.

namerd should remove the watch for a dtab once a value has been fulfilled. In order to do that, `def ApacheZooKeeper.getDataWatch` should handle interrupts for `FutureCancelledExceptions`. When it handles this interrupt, it now removes the watch from `ZKWatcherManager`. Adding this functionality required forking our own version of `ApacheZooKeper.scala` and is the reason for the addition of this file.

Once we are able to remove watches, `def ZkSession.watchedOperation` no longer should fire and forget its retries. This required refactoring how to handle `WatchedState` changes to `loop()` only in certain cases.

The result of these changes is seen in [2] after 5k requests sent.

# Validation

[1] and [2] show the changes reflected by this change. `ZKWatchManager` now properly removes watches when values for dtabs have been fulfilled. It continues to work as expected for streams.

Fixes #2212

Co-authored-by: Alex Leong <alex@buoyant.io>

[1]
screen shot 2019-02-15 at 12 00 06 pm

[2]
screen shot 2019-02-15 at 11 58 50 am


namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/client/buoyant/ZkSession.scala
```diff
@@ -123,71 +124,83 @@ class ZkSession(
       @volatile var closed = false
 
       def loop(): Future[Unit] = {
```
adleong: This is so much more readable than what was here before!  :star:

adleong: This method really is the heart of the zk watch code and is very semantically dense.  Perhaps adding some comments will help future spelunkers understand it more easily.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/client/buoyant/ZkSession.scala
```diff
@@ -123,71 +124,83 @@ class ZkSession(
...
 
           case Return(Watched(value, state)) =>
             val ok = Activity.Ok(value)
             retryBackoff.reset()
             u() = ok
 
-            val _ = state.changes.respond {
-              case WatchState.Pending =>
-              // Ignore updates WatchState is Pending.
+            state.map { s =>
```
adleong: This could use a comment explaining that this map is just to perform side effects (logging and updating the activity)

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/client/buoyant/ZkSession.scala
```diff
@@ -123,71 +124,83 @@ class ZkSession(
...
+                    sessionState, sessionId
+                  )
+                  u() = Activity.Failed(new Exception("" + sessionState))
+
+                case WatchState.SessionState(sessionState) =>
+                  logger.error("Unexpected session state %s. Session: %s", sessionState, sessionId)
+                  u() = Activity.Failed(new Exception("" + sessionState))
+              }
+              s
+            }.changes.filter {
```
adleong: This filter could use a comment explaining that this, in combination with the `toFuture()` below, creates a future that is satisfied when the state becomes Determined, ConnectedReadOnly, SaslAuthenticated, SyncConnected, or an unknown state.

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/client/buoyant/ZkSession.scala
kleimkuhler: Comments have been added to each stage of the chain of operations!

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/client/buoyant/ZkSession.scala
kleimkuhler: I like that right now, the result of each operation is changed together with the addition of comments for each stage. I think that introducing an intermediate variable could possibly reduce readability as a reader may get hung up on the naming of that variable. I can separate the chain if you feel strongly about it, but otherwise I inclined to leave it. Let me know your thoughts!

namerd/storage/zk/src/main/scala/com/twitter/finagle/serverset2/client/apache/buoyant/ApacheZooKeeper.scala
```diff
@@ -0,0 +1,114 @@
...
+        protected val print = { m: String =>
+          logger.info(m)
+        }
+      }, statsWatcher)
+    } else
+      Watched(wrappedZk, statsWatcher)
+  }
+}
+
+private class ApacheFactory extends ClientFactory[ZooKeeperRW] {
```
adleong: I think this is unused

adleong:
@adw12382: are you able to test if this change fixes the Namerd memory issue for you?  Would it be helpful for us to provide a docker image for you to test?

adw12382:
> @adw12382: are you able to test if this change fixes the Namerd memory issue for you? Would it be helpful for us to provide a docker image for you to test?

@adleong : I just come back from vacation today, and thanks for the awesome work regarding the issue! I will test it today or tomorrow and I think I should be able to generate a docker image from the branch. I will post my updates here.

adw12382:
I just got the following Leak Suspects report from an image built from the issue-2212 branch. 
screen shot 2019-02-21 at 12 52 51 pm

It seems to me that the zookeeper memory leak is fixed according to the report, but not sure what's the issue, which seems not critical, showing in the report though. Also, the namerd pod has been up and running for 1.5 hours in environment where live traffic goes through, and the namerd pods' memory usage stays around 10% line, which is much lower than what we observed previously.

I will let it run in our environment for several days and see if it works as expected, and will post updates here if any.

adw12382:
@adleong Just curious that is there any garbage collection specific config can be set for n4d container?

dadjeibaah:
@adw12382 Namerd exposes JVM flags through a `LOCAL_JVM_OPTIONS` environment variable. You could add whatever GC configuration you are interested through this variable and Namerd should pick it up.

kleimkuhler:
@adw12382 Glad to hear you are seeing the expected fix from this change.

The `io.netty.buffer.PoolCheck` suspect is something you can keep an eye on as your environment continues to run. During my testing, it consistently showed up as a suspect, but the heap percentage did not scale similarly to the ZooKeeper `HashMap`. I do not think it is an issue, but if you observe differently, then we can address that in a separate issue.

I will merge this as it currently is so that the ZooKeeper issue is fixed!

adw12382:
Hi there, just a quick update, the image has been up for 4 days and the memory usage climbs much more slower than how it was previously. Looking forward to the new release :octocat:

kleimkuhler:
@adw12382 Glad to hear! We plan on releasing a new version soon.

adw12382:
image
Just an update, it looks like the memory usage keeps increasing over time. I feel a potential leak is still there. I am curious to see see what happens after the usage bumps up to 100% again

dadjeibaah:
Interesting @adw12382, are the sharp cliffs caused by GC or is that Namerd being restarted? We should do more memory analysis over time and see what might be the cause.

adw12382:
hI @dadjeibaah ,

The lines that dropped and then came up again in same color are caused by GC. The lines that suddenly disappeared are caused by pod restarting. Besides, the lines appeared around 2/22 are reflecting status of the image built from the bugfix branch. All the previous lines are reflecting the namerd-1.6.0.

Thanks

PR #2226, opened 2019-02-19
Add `socketOptions` to `ClientConfig`

### LINKED ISSUE #2218: Support socketOptions on client
At the moment we've added support for setting custom socketOptions on the admin and routers, we should expand this for clients, so they can enable things like keepAlive

kleimkuhler:
# Problem

At the moment we've added support for setting custom socketOptions on the admin and routers, we should expand this for clients, so they can enable things like keepAlive

# Solution

Add `socketOptions` to `ClientConfig` so that a user can set the expected properties.

Closes #2218

### DIFF
--- a/linkerd/examples/client-options.yaml
+++ b/linkerd/examples/client-options.yaml
@@ -27,6 +27,7 @@ routers:
       loadBalancer:
         kind: aperture
       socketOptions:
+        keepAlive: true
         reusePort: true
         readTimeoutMs: 61000
         writeTimeoutMs: 61000
--- a/linkerd/examples/client-options.yaml
+++ b/linkerd/examples/client-options.yaml
@@ -26,6 +26,10 @@ routers:
       # this client only
       loadBalancer:
         kind: aperture
+      socketOptions:
+        reusePort: true
+        readTimeoutMs: 61000
+        writeTimeoutMs: 61000
     # Wildcards and variable capture can be used in prefixes
     - prefix: "/#/*/{service}"
       tls:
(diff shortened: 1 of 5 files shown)


adleong:
Have you been able to test that the socket options get set correctly?

One suggestion for how to test this would be to build off this branch and run Linkerd in docker or on a Linux machine, establish a connection through Linkerd, and check the socket timer with `netstat -o`.  The socket timer should be controllable with the keepalive socket option.  See http://veithen.io/2013/12/19/inspecting-socket-options-on-linux.html for more info.

linkerd/examples/client-options.yaml
```diff
@@ -26,6 +26,10 @@ routers:
       # this client only
       loadBalancer:
         kind: aperture
+      socketOptions:
+        reusePort: true
+        readTimeoutMs: 61000
+        writeTimeoutMs: 61000
```
dadjeibaah: Let's also add the `keepAlive` option here. It doesn't hurt to have more documentation 😄

kleimkuhler:
@adleong Thanks for the question. I have confirmed that `socketOptions` is set correctly on clients.

After running Linkerd in docker and establishing a connection to `localhost:9990`, I can see the following difference in `netstat -to` when I specify the following `socketOptions` for clients:

```
client:
    socketOptions:
      noDelay: true
      reuseAddr: true
      reusePort: true
      readTimeoutMs: 61000
      writeTimeoutMs: 61000
      keepAlive: true
      backlog: 128
```

Without `socketOptions`:

```
$ netstat -to
$ tcp        0      0 localhost:34354         localhost:9990          ESTABLISHED off (0.00/0/0)
```

With `socketOptions`:

```
$ netstat -to
$ tcp        0      0 localhost:47580         localhost:9990          ESTABLISHED keepalive (7189.26/0/0)
```

PR #2229, opened 2019-02-19
Upgrade Finagle to 19.1.0

kleimkuhler:


adleong:
It looks like something changed related to what trace events we send: according to the failing tests there is now an extra wire receive.  We should figure out what caused that change and if it makes sense or not.

kleimkuhler:
Okay I found the change: https://github.com/twitter/finagle/commit/25e5551b0853395f0e742a43ef6b5d60f5521bfc

`WireTracingFilter` was extended to the server. We are asserting the annotation keys for the server. Up until Finagle 19.1.0 the server would not generate these traces, but now it does.

So these additional traces are expected and updating the tests is the correct fix.

PR #2231, opened 2019-02-20
Remove the duplicated words in code comments

longkb:
Although it is spelling mistakes, it might make an affects while reading.

Co-Authored-By: Nguyen Phuong An <AnNP@vn.fujitsu.com>
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>

### DIFF
--- a/interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
+++ b/interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
@@ -232,8 +232,7 @@ object Client {
    * gRPC stream, publishing updates into the Var's state.
    *
    * When the stream returns an error, the Var may be updated (as
-   * according to `toT`) and no further updates will will be
-   * published.
+   * according to `toT`) and no further updates will be published.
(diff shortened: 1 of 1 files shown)


interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -305,7 +304,7 @@ object Client {
    * gRPC stream, publishing updates into the Activity's state.
    *
    * When the stream returns an error, the state is updated to
-   * Activity.Failed and no further updates will will be published.
+   * Activity. Failed and no further updates will be published.
```
adleong: `Failed` is a member of `Activity` so I think this was correct without the space: `Activity.Failed`.

longkb: > `Failed` is a member of `Activity` so I think this was correct without the space: `Activity.Failed`.

Thanks you for your comment. I have push a change to revert it :)

PR #2239, opened 2019-02-26
Update docs links from HTTP to HTTPS for security and consistency

tuanvcw:
This PR replaces old HTTP links in docs by new, correct HTTPS links for security and consistency.

PR #2240, opened 2019-02-26
Update all links to HTTPS for security and consistency

tuanvcw:
This is the follow-up of #2239.
After checking docs and updating links from HTTP to HTTPS, I realized there are still many deprecated HTTP links in other places. So I decided to update all of them with this PR.


tuanvcw:
Waiting on code owner review from @olix0r and/or @wmorgan

PR #2244, opened 2019-02-27
Ignore Unix swap files `*.swp` in .gitignore

truongnh1992:
Signed-off-by: Nguyen Hai Truong <truongnh@vn.fujitsu.com>


dadjeibaah:
@truongnh1992 is the `.swp` file being inadvertently added to your commits? Is there anyway you can include this file in your global `.gitignore` file?

PR #2258, opened 2019-03-21
adds support for dynamically modifying log levels in Namerd

### LINKED ISSUE #2254: namerd doesn't provide an interface to change logging level for individual loggers
## Filing a Linkerd issue ##

Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:
 We would like to run namerd in trace mode to figure out the leak in re: [here](https://github.com/linkerd/linkerd/issues/2251). But there is a lot of noise in the logs re: [here](https://gist.github.com/thedebugger/85be50892e75b0128d537ab7021e90ae). AFAIT, all linkerd/namerd logs are being logged with same logger which is a result of using com.twitter.logging.logger? As a result, we can turn off/on logging for specific loggers. I also noticed that namerd doesn't expose logging.json on admin endpoint.

Namerd should expose an interface to customize loggers and format. 

Run namerd with trace logs in k8s

Nope. Let me know if you need more information

dadjeibaah:
Namerd's dashboard currently does not have the ability to dynamically change log levels for modules running within the Namerd process. This makes it difficult to debug issues that may arise in Namerd since you are left to rely on the current log levels set at startup time.

This PR adds a new "Logging" page which allows you to change log levels dynamically. This PR also includes a nav bar for Namerd to assist in switching between the dtab playground and the new logging page.

Fixes #2254 

### DIFF
--- a/admin/src/main/resources/io/buoyant/admin/js/main-namerd.js
+++ b/admin/src/main/resources/io/buoyant/admin/js/main-namerd.js
@@ -23,7 +23,12 @@ require.config({
 require([
   'jQuery',
   'bootstrap',
-  'src/dashboard_delegate'
-], function ($, bootstrap, namerdDtabPlayground) {
-  new namerdDtabPlayground();
+  'src/dashboard_delegate',
+  'src/logging'
+], function ($, bootstrap, namerdDtabPlayground, loggingConfig) {
+  if(window.location.pathname.indexOf("/dtab") >= 0){
+    new namerdDtabPlayground();
+  } else if(window.location.pathname.endsWith("/logging")){
+    new loggingConfig();
+  }
 });
--- a/namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
+++ b/namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
@@ -31,7 +31,6 @@ class DtabHandler(
 
   def renderDtab(name: String, dtab: Dtab): Future[Response] = {
     val body = s"""
-          <div class="container-fluid">
             <div class="row">
               <div class="col-lg-6">
                 <h2 class="router-label-title">Namespace "$name"</h2>
@@ -39,20 +38,18 @@ class DtabHandler(
             </div>
             <div class="delegator">
             </div>
-          </div>
 
           <script id="data" type="application/json">{"namespace": "$name", "dtab": ${
       DelegateApiHandler
         .Codec.writeStr(dtab)
     }}</script>
-          <script data-main="../files/js/main-namerd" src="../files/js/lib/require.js"></script>
+          <script data-main="/files/js/main-namerd" src="/files/js/lib/require.js"></script>
       """
     render(body)
   }
 
   def renderError(e: Throwable): Future[Response] = {
     val body = s"""
-          <div class="container-fluid">
             <div class="row">
               <div class="col-lg-6">
                 <h2 class="router-label-title">Dtab Error</h2>
@@ -60,31 +57,14 @@ class DtabHandler(
                 <pre>$e</pre>
               </div>
             </div>
-          </div>
     """
     render(body)
   }
 
   def render(body: String): Future[Response] = {
-    val content = s"""
-      <!doctype html>
-      <html>
-        <head>
-          <title>namerd admin</title>
-          <link rel="shortcut icon" href="../files/images/favicon.png" />
-          <link type="text/css" href="../files/css/lib/bootstrap.min.css" rel="stylesheet"/>
-          <link type="text/css" href="../files/css/fonts.css" rel="stylesheet"/>
-          <link type="text/css" href="../files/css/dashboard.css" rel="stylesheet"/>
-          <link type="text/css" href="../files/css/delegator.css" rel="stylesheet"/>
-        </head>
-        <body>
-        $body
-        </body>
-      </html>
-    """
     val response = Response()
-    response.contentType = MediaType.Html + ";charset=UTF-8"
-    response.contentString = content
+    response.contentType = MediaType.Html
+    response.contentString = body
     Future.value(response)
   }
 }
--- a/namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
+++ b/namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
@@ -43,7 +43,7 @@ class DtabHandler(
       DelegateApiHandler
         .Codec.writeStr(dtab)
     }}</script>
-          <script data-main="/files/js/main-namerd" src="/files/js/lib/require.js"></script>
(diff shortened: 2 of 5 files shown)


namerd/main/src/main/scala/io/buoyant/namerd/NamerdAdmin.scala
```diff
@@ -21,15 +23,116 @@ object NamerdAdmin {
...
+  def logging(adminHandler: NamerdAdmin): Seq[Handler] = {
+    val adminFilter = new NamerdFilter(adminHandler)
+    Seq(
+      Handler("/logging.json", new LoggingApiHandler()),
+      Handler("/logging", adminFilter.andThen(new LoggingHandler(adminHandler)))
+    )
+  }
+
+  def apply(nc: NamerdConfig, namerd: Namerd): Seq[Handler] = {
+    val handler = new NamerdAdmin(Seq(NavItem("dtabs", "dtabs"), NavItem("logging", "logging")))
```
adleong: The `dtabs` and `Namerd` nav items link to the same place and I think are redundant

namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
```diff
@@ -31,60 +31,40 @@ class DtabHandler(
...
             <div class="delegator">
             </div>
-          </div>
 
           <script id="data" type="application/json">{"namespace": "$name", "dtab": ${
       DelegateApiHandler
         .Codec.writeStr(dtab)
     }}</script>
-          <script data-main="../files/js/main-namerd" src="../files/js/lib/require.js"></script>
+          <script data-main="/files/js/main-namerd" src="/files/js/lib/require.js"></script>
```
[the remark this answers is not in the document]
dadjeibaah: They pretty much are the same thing. The former is a relative path so we go up a directory to get to the js file whereas the latter is an absolute path. I will revert this since "it ain't broke".

namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
```diff
@@ -31,60 +31,40 @@ class DtabHandler(
...
-          <link type="text/css" href="../files/css/dashboard.css" rel="stylesheet"/>
-          <link type="text/css" href="../files/css/delegator.css" rel="stylesheet"/>
-        </head>
-        <body>
-        $body
-        </body>
-      </html>
-    """
     val response = Response()
-    response.contentType = MediaType.Html + ";charset=UTF-8"
```
kleimkuhler: Why is this being removed?

namerd/main/src/main/scala/io/buoyant/namerd/DtabListHandler.scala
```diff
@@ -12,7 +12,7 @@ class DtabListHandler(
   override def apply(req: Request): Future[Response] =
     store.list().toFuture.map { list =>
       val response = Response()
-      response.contentType = MediaType.Html + ";charset=UTF-8"
```
kleimkuhler: Same question here about why this is being removed

namerd/main/src/main/scala/io/buoyant/namerd/NamerdAdmin.scala
```diff
@@ -1,10 +1,12 @@
...
-import com.twitter.finagle.{Namer, Path}
+import com.twitter.finagle.http.{MediaType, Request, Response}
+import com.twitter.finagle.{Namer, Path, Service, SimpleFilter}
 import com.twitter.server.handler.ResourceHandler
-import io.buoyant.admin.Admin.Handler
+import com.twitter.util.Future
+import io.buoyant.admin.Admin.{Handler, NavItem}
 import io.buoyant.admin.names.DelegateApiHandler
-import io.buoyant.admin.{Admin, ConfigHandler, StaticFilter}
+import io.buoyant.admin._
```
kleimkuhler: TIOLI: This is now importing a lot more than just the three classes we need

namerd/main/src/main/scala/io/buoyant/namerd/NamerdAdmin.scala
```diff
@@ -21,15 +23,116 @@ object NamerdAdmin {
...
-  def dtabs(dtabStore: DtabStore, namers: Map[Path, Namer]) = Seq(
-    Handler("/", new DtabListHandler(dtabStore)),
-    Handler("/dtab/delegator.json", new DelegateApiHandler(ns => ConfiguredNamersInterpreter(namers.toSeq))),
-    Handler("/dtab/", new DtabHandler(dtabStore))
-  )
+  def dtabs(dtabStore: DtabStore, namers: Map[Path, Namer], adminHandler: NamerdAdmin) = {
+    val adminFilter = new NamerdFilter(adminHandler)
+    Seq(
+      Handler("/", adminFilter.andThen(new DtabListHandler(dtabStore))),
+      Handler("/dtab/delegator.json", new DelegateApiHandler(ns => ConfiguredNamersInterpreter(namers.toSeq))),
```
kleimkuhler: TIOLI: `ns` is never used so you could just go with `_`

namerd/main/src/main/scala/io/buoyant/namerd/DtabHandler.scala
dadjeibaah: This is removed so that [this](https://github.com/linkerd/linkerd/pull/2258/files#diff-e1ab574e9f83b4a3327606f6d43bbfa9R132) line in NamerdFilter can evaluate to true. The `contains` checks for the exact match of a string within an `Option`. So I had to change it to satisfy the condition so that we could build the final rendered HTML correctly.

namerd/main/src/main/scala/io/buoyant/namerd/DtabListHandler.scala
dadjeibaah: See comment above.

admin/src/main/resources/io/buoyant/admin/js/main-namerd.js
```diff
@@ -23,7 +23,12 @@ require.config({
 require([
   'jQuery',
   'bootstrap',
-  'src/dashboard_delegate'
-], function ($, bootstrap, namerdDtabPlayground) {
-  new namerdDtabPlayground();
+  'src/dashboard_delegate',
+  'src/logging'
+], function ($, bootstrap, namerdDtabPlayground, loggingConfig) {
+  if(window.location.pathname.indexOf("/dtab") >= 0){
```
adleong: this is how you do "contains" in javascript???

dadjeibaah: sadly yes. In JS's defense, this was how it was done pre-ES6, now you can use `.contains` but that is not supported in all browser versions.

PR #2261, opened 2019-03-28
Remove delegate tree from rich exceptions

adleong:
RichConnectionFailedException and RichNoBrokersException make a call to the `delegete` when they are populated to provide the delegate tree in the text of the exception.  In cases where there is a high rate of these exceptions being created (for example, when there are many requeues) this results in a high rate of calls to `delegate`.  If Namerd is used for delegation, this results in a high rate of requests to Namerd which can have Denial of Service effects.

We remove the delegation tree information from the rich exceptions.  This ensures that Namerd will not be impacted when there is a high rate of these exceptions.  The rich exceptions will still contain the service name and client name, so the delegation tree can still be retrieved manually from the dtab UI in the admin server if it is desired.

### DIFF
--- a/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala
@@ -8,8 +8,7 @@ import io.buoyant.namer.{DelegateTree, Delegator, RichActivity}
 
 class RichNoBrokersAvailableException(
   path: Dst.Path,
-  dtab: Option[Dtab],
-  resolutions: Option[Seq[String]]
+  dtab: Option[Dtab]
 ) extends RequestException {
 
   private[this] def formatDtab(d: Dtab): String =
@@ -21,8 +20,6 @@ class RichNoBrokersAvailableException(
     s"""Unable to route request!
 
 service name: ${path.path.show}
-resolutions considered:
-${resolutions.getOrElse(Seq("unknown")).map("  " + _).mkString("\n")}
 dtab:
 ${formatDtab(dtab.getOrElse(Dtab.empty))}
 base dtab:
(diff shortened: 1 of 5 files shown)


router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala
```diff
@@ -34,43 +31,6 @@ ${formatDtab(path.localDtab)}
 
 object RichNoBrokersAvailableException {
 
-  def apply(path: Dst.Path, namer: NameInterpreter): Future[Nothing] = {
```
dadjeibaah: Minor import cleanup since we no longer use the Delegator
```diff
diff --git a/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala b/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala
index 26dcd1654..cdcfa4c8d 100644
--- a/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala
@@ -2,9 +2,7 @@ package com.twitter.finagle.naming.buoyant
 
 import com.twitter.finagle._
 import com.twitter.finagle.buoyant.Dst
-import com.twitter.finagle.naming.NameInterpreter
 import com.twitter.util.Future
-import io.buoyant.namer.{DelegateTree, Delegator, RichActivity}
 
 class RichNoBrokersAvailableException(
   path: Dst.Path,

```

router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichNoBrokersAvailableException.scala
```diff
@@ -8,8 +8,7 @@ import io.buoyant.namer.{DelegateTree, Delegator, RichActivity}
 
 class RichNoBrokersAvailableException(
   path: Dst.Path,
-  dtab: Option[Dtab],
```
dadjeibaah: Import cleanup is needed here as well.
```diff
diff --git a/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichConnectionFailedException.scala b/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichConnectionFailedException.scala
index e5bd44f17..b41287f75 100644
--- a/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichConnectionFailedException.scala
+++ b/router/core/src/main/scala/com/twitter/finagle/naming/buoyant/RichConnectionFailedException.scala
@@ -1,10 +1,8 @@
 package com.twitter.finagle.naming.buoyant
 
 import com.twitter.finagle.buoyant.Dst
-import com.twitter.finagle.naming.NameInterpreter
 import com.twitter.finagle._
 import com.twitter.util.Future
-import io.buoyant.namer.{DelegateTree, Delegator}
 import java.net.SocketAddress
 import scala.util.control.NoStackTrace
 ```

PR #2271, opened 2019-05-07
Guard thrift serializer from throwing BufferUnderflowException

### LINKED ISSUE #2259: interpreter state admin endpoint fails with BufferUnderflowException on few nodes
## Filing a Linkerd issue ##

Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:
Call to `http/interpreter_state/io.l5d.namerd/httproutes.json` fails with 500 - linkerd logs BufferUnderflowException. 

It should return the state of all watches

We are seeing it on a few nodes that have this problem. It fails all the time on the nodes that have the problem.

Exception stack trace https://gist.github.com/thedebugger/416ddb566d482d840ac1d94491553bf2

**Environment**:
- linkerd/namerd version, config files: both 1.5.2  
- Platform, version, and config files (Kubernetes, DC/OS, etc): 1.10.11
- Cloud provider or hardware configuration: shoudn't matter

dadjeibaah:
fixes #2259

### DIFF
--- a/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ByteBufferSerializers.scala
+++ b/namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ByteBufferSerializers.scala
@@ -12,12 +12,12 @@ object ByteBufferSerializers {
     path.map(utf8).mkString("/", "/", "")
 
   def stamp(bb: ByteBuffer): String = {
-    if (bb.array().length > 0) bb.duplicate().getLong.toString else ""
+    if (bb.remaining() >= 8) bb.duplicate().getLong.toString else ""
   }
 
   def ipv4(bb: ByteBuffer): String = {
     val dup = bb.duplicate()
-    if (dup.array().length > 0)
+    if (dup.remaining() >= 4)
       s"${dup.get()}.${dup.get()}.${dup.get()}.${dup.get()}"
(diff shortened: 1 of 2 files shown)


namerd/iface/interpreter-thrift/src/main/scala/io/buoyant/namerd/iface/ByteBufferSerializers.scala
```diff
@@ -11,11 +11,14 @@ object ByteBufferSerializers {
   def path(path: Seq[ByteBuffer]): String =
     path.map(utf8).mkString("/", "/", "")
 
-  def stamp(bb: ByteBuffer): String =
-    bb.duplicate().getLong.toString
+  def stamp(bb: ByteBuffer): String = {
+    if (bb.array().length > 0) bb.duplicate().getLong.toString else ""
```
adleong: probably better not to assume that the BB is array-backed.  we can use `bb.remaining()` to get number of available bytes in the buffer.

Also, longs are 8 bytes so we should ensure that there are at least 8 bytes remaining: https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#getLong()

dadjeibaah: TIL

PR #2277, opened 2019-05-23
Add additional logging to track mesh clients interacting with Namerd

### LINKED ISSUE #2245: support for trace logs for all outbound communication

## Getting Help ##

## Filing a Linkerd issue ##

Issue Type:

- [ ] Bug report
- [X] Feature request

**What happened**:
We are seeing numerous cases where 1) namerd state diverge from k8s apiserver - dtabs and endpoint 2) linkerd state diverge from namerd e.g. linkerd thinks only 4 pods are available, but namerd sees 10. We suspect events are getting dropped in namerd and linkerd. To figure out where things are going wrong, linkerd and namerd should trace all inbound and outbound watch (or relevant) events.

Namerd and linkerd should log all inbound and outbound events. E.g.
1. Namerd: received X event from apiserver
2. Namerd: making X API call to apiserver
3. Namerd: sending event X to linkerd Y along with success/failure
4. Namerd: received an open stream from Linkerd X
4. Linkerd: making X api call to namerd
6. Linkerd: received X api call to namerd

If we can also log on namerd that linkerd with "hostname" is connected to namerd, that would be really helpful too. IPs are NATed. So we lose the observability what linkerd is connected to what namerd.
 
N/A

Let me know if you know more information

**Environment**:
- linkerd/namerd version, config files: linkerd 1.5.2, namerd 1.5.2, http2 mesh, thriftmux
- Platform, version, and config files (Kubernetes, DC/OS, etc): 1.10.5
- Cloud provider or hardware configuration: not relevant

dadjeibaah:
There have been reports of Linkerd mesh clients not receiving updates from a Namerd server. There currently are watch state endpoints in Linkerd that tell you the last known value received from Namerd. This information is helpful in figuring out when Linkerd gets into this state. However, this information does not reveal the sequence of events that led up to that state.

This PR adds logging in the mesh client that logs every state change the Linkerd process is notified of i.e. Address set changes observed by Namerd. Given that this PR increases the number of log lines that will be available for diagnosis, each mesh client will now contain a unique identifier that will be injected when a watch request is initiated between Linkerd and Namerd for better analysis. This will help identify which client gets into a bad state when crawling through logs and will help in situations were remote IPs (NATed IPs) do not easily identify a Linkerd client in the Namerd logs.

fixes #2245

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2HeaderInjector.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2HeaderInjector.scala
@@ -1,0 +1,27 @@
+package com.twitter.finagle.buoyant.h2
+
+import com.twitter.finagle
+import com.twitter.finagle._
+import com.twitter.util.Future
+
+object H2HeaderInjector {
+  def module(headerMap: Map[String, String]): Stackable[ServiceFactory[Request, Response]] = {
+    new finagle.Stack.Module0[ServiceFactory[Request, Response]] {
+      override def make(next: ServiceFactory[Request, Response]): ServiceFactory[Request, Response] =
+        new HeaderInjector(headerMap).andThen(next)
+
+      override def role: Stack.Role = Stack.Role("H2HeaderInjector")
+      override def description: String = "Add arbitrary headers to H2 requests"
+    }
+  }
+}
+
+class HeaderInjector(injectHeaders: Map[String, String]) extends SimpleFilter[Request, Response] {
+  override def apply(request: Request, service: Service[Request, Response]): Future[Response] = {
+    val req = request.dup()
+    injectHeaders.foreach { kvPair =>
+      req.headers.add(kvPair._1, kvPair._2); ()
+    }
+    service(req)
+  }
+}
--- a/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
+++ b/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
@@ -1,11 +1,13 @@
 package io.buoyant.grpc.runtime
 
-import com.twitter.finagle.{Failure, Service => FinagleService}
+import com.twitter.finagle.{Service => FinagleService}
 import com.twitter.finagle.buoyant.h2
+import com.twitter.logging.Logger
 import com.twitter.util.{Future, Return, Throw, Try}
 
 object ServerDispatcher {
 
+  val log = Logger.get(this.getClass.getName)
   trait Service {
     def name: String
     def rpcs: Seq[Rpc]
@@ -16,7 +18,6 @@ object ServerDispatcher {
   }
 
   object Rpc {
-
     class UnaryToUnary[Req, Rsp](
       val name: String,
       serve: Req => Future[Rsp],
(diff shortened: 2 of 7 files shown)


grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
```diff
@@ -99,11 +103,12 @@ object ServerDispatcher {
         h2.Response(h2.Status.Ok, frames)
     }
 
-    private[this] def respondStreaming[Rsp](codec: Codec[Rsp], msgs: Stream[Rsp]): h2.Response = {
+    private[this] def respondStreaming[Rsp](codec: Codec[Rsp], msgs: Stream[Rsp], reqMeta: Seq[(String, String)]): h2.Response = {
       val frames = h2.Stream()
       def loop(): Future[Unit] =
         msgs.recv().transform {
           case Return(Stream.Releasable(s, release)) =>
+            log.trace(s"Streaming response metadata: ${reqMeta.mkString(":")}")
```
adleong: This says response metadata but it's printing the request headers, right?

adleong: What's the intent here?  Each time there's a response message, the request headers get printed?

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2HeaderInjector.scala
```diff
@@ -0,0 +1,27 @@
...
+
+      override def role: Stack.Role = Stack.Role("H2HeaderInjector")
+      override def description: String = "Add arbitrary headers to H2 requests"
+    }
+  }
+}
+
+class HeaderInjector(injectHeaders: Map[String, String]) extends SimpleFilter[Request, Response] {
+  override def apply(request: Request, service: Service[Request, Response]): Future[Response] = {
+    val req = request.dup()
```
adleong: is this necessary?  aren't headers mutable?

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
dadjeibaah: That's correct since the request headers may contain information useful for the particular request.

finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/H2HeaderInjector.scala
dadjeibaah: I remember from my last run-in with H2 request mangling that I had to `dup()` this for some reason, alas, I do not remember. I think we don't need this.

adleong:
Thanks for putting this together, @dadjeibaah.  I think a simpler solution here would be for Namerd to simply include the remote (client) address in its log messages rather than generating a UUID on the client.  The UUID approach has some problematic drawbacks:
* It's not immediately clear from the Namerd logs how to map a UUID to a Linkerd instance.  To do this, you must scan all Linkerd instance logs for a matching UUID
* Generating and propagating the UUID adds extra complexity to the code
* Adding extra metadata through request headers breaks from idiomatic gRPC and requires us to bake in special handling for this in our gRPC code.

I can see why a UUID solution would be desirable in specific situations where the client address is not available (such as when the request goes through a NAT) but I think that requirement is specialized enough that it doesn't merit putting this logic in master.

PR #2284, opened 2019-06-13
Upgrade finagle to 19.5.1

dadjeibaah:


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4ServerDispatcher.scala
```diff
@@ -59,14 +59,13 @@ class Netty4ServerDispatcher(
     val save = Local.save()
     try {
       Contexts.local.let(RemoteInfo.Upstream.AddressCtx, transport.context.remoteAddress) {
-        transport.context.peerCertificate match {
-          case None =>
+        val peerCerts = transport.context.sslSessionInfo.peerCertificates
+        if (peerCerts.isEmpty) {
+          service(req).rescue(wrapServiceEx)
+        } else {
+          Contexts.local.let(Transport.peerCertCtx, peerCerts.head) {
```
adleong: TIOLI: I tend to prefer the safe `headOption` instead of the unsafe `head`:

ie `peerCerts.headOption match { ...`

Issue #2201, opened 2019-01-08
Add ability to suppress Linkerd error response body and header


chrismikehogan:
Issue Type:

- [ ] Bug report
- [X] Feature request

# Problem
We are trying to use Linkerd's dtab feature to build very large, dynamic routing tables. The goal to use Linkerd (instead of, e.g. Consul) to control which versions of which applications get routed to when a a request for a particular service is received.

This requires us to dynamically build most of our dtab, and requires that the dtab have an explicit dentry for every service in our eco system. E.g.:

```
...
/svc/cool-thing => /tagged/ver-1.3/cool-thing;
/svc/lame-thing => 0.1 * /tagged/ver-2.0/lame-thing & 0.9 * /tagged/ver-1.8/lame-thing;
...
```

This means that our dtab will get *very* large! The example I'm working with at the moment is 1811 lines long, and 143kb.

When linkerd errors in some way (say when a requested service is not found), it currently 502s and sends back an error message that contains, among other things, the entire dtab in both the body and a response header.

Fortunately, Linkerd appears to truncate the response header field instead of blowing up. But I think it would be useful to be able to suppress this verbose response by default, exposing it only if the client provided a debug header.

# Possible solution

Provide a configuration option to suppress verbose error responses by default.
Add support for a debug header that results in the verbose error response.

dadjeibaah:
👋 @chrismikehogan, Thanks for filing this. FWIW, there is a config parameter called `clearContext` that can be set at the router level of the config. It's a server side configuration so you won't have the ability to dynamically turn it on or off per request.

chrismikehogan:
I saw that flag in the configuration docs, but it wasn't obvious to me that it would do what I wanted, in addition to what the docs said it does:
> if true, all headers that set Linkerd contexts are removed from inbound requests. Useful for servers exposed on untrusted networks.

We definitely don't wan't the behavior described in the docs, even if it got us the suppressed error responses.

dadjeibaah:
Ah, that's right, it also gets applied to inbound requests as well so that wouldn't work.

adleong:
Truncating the response sounds great to me

chrismikehogan:
Truncating would work perfectly for us as well, even more so if the level of truncation could be controlled.

We could take or leave the concept debug header adding additional info.

menya84:
[...]

Issue #2196, opened 2018-12-18
Revise maxRequestKB and maxResponseKB meaning after #2189


edio:
Issue Type:

- [ ] Bug report
- [x] Feature request

Do you think it makes sense to revise the necessity or behavior of `maxRequestKB`  and `maxResponseKB` after #2189 ?

I don't know the initial motivation behind those, but I guess that it may be to protect linkerd from OOMs.

To me it would make sense to have following set of properties:

* `streamAfterContentLengthKB` - current behavior. Effectively defines how much space in heap large unchunked request can take at most
* `maxRequestKB`, `maxResponseKB` - max request/response size, both chunked and unchunked. Having it only for unchunked requests as it is now does not have much sense, since we already have `streamAfterContentLengthKB` (unless of course the motivation behind those properties is something else than preventing OOM). Having these limits for all requests/responses on the other hand, makes total sense for me: we may want to protect services in our service mesh from unreasonably large requests without polluting services logic.
However in this case, I believe, we should make it unlimited by default (right now finagle allows 2 GiB max, but I think we can try discussing this with finagle team).

dadjeibaah:
Thanks for filing this @edio. My understanding is that `maxRequestKB` and `maxResponseKB` were initially added to tame large requests from being sent to the router according to this PR https://github.com/linkerd/linkerd/pull/557. Having these configurations with `streamAfterContentLengthKB` seems a bit redundant, however, I do agree with you that these configs could be used to protect services from large requests. I did some digging too and it looks like some if not all of these http configs in the past were discussed in some form. Issue #1327 further discusses the `maxResponseKB` and `maxRequestKB` params.

adleong:
I'm actually a bit confused about the behavior of `maxRequestKB` and `maxResponseKB`.  I think the next steps for this ticket are to verify the behavior of these properties in all different cases:
* chunked message
* non-chunked message
* non-chunked streaming message

dadjeibaah:
This investigation may result in the removal of `maxRequestKB` and `maxResponseKB` as config parameters and may be a breaking change in Linkerd just as a heads up.

zaharidichev:
will give you a bit more clarity on that one shortly

PR #2291, opened 2019-07-03
a possible fix for #2201

dtacalau:
Here https://api.linkerd.io/1.6.1/linkerd/index.html#http-1-1-protocol, I've found two http config options here which can be used to truncate large headers or response bodies as requested for #2201:
maxHeadersKB | 8 | The maximum size of all headers in an HTTP message.
maxResponseKB | 5120 | The maximum size of a non-chunked HTTP response payload.

maxHeadersKB was already implemented and supported for l5d-err headers, I've done some tests with large dtabs and the l5d-err headers are truncated to the size specified by maxHeadersKB.

maxResponseKB was only parsed by HttpConfig.scala but was not implemented/supported by the rest of the Linderd code. This pull request adds support for maxResponseKB limit when constructing responses containing l5d-err header, what it does it just truncates the error response body to the limit specified by the maxResponseKB. The option name is somehow misleading, as one might expected the truncation to be done for the http response packet size instead of the http response body size, more precisely for error response body size. 

With this fix someone setting the config maxResponseKB / hparam.MaxResponseSize to a value lower than the 5120 default, will obtain truncation of the err response body size.

IMO A better approach would be to have two new config options specific to err headers and err response bodies, called maxErrHeadersKB and maxErrResponseKB. In this way users would benefit by the standard limits for headers and reponses and also have a much finer control over err headers and err responses.

### DIFF
--- a/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
+++ b/linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
@@ -5,7 +5,8 @@ import com.twitter.finagle.buoyant.{Dst => BuoyantDst}
 import com.twitter.finagle.context.{Contexts, Deadline => FDeadline}
 import com.twitter.finagle.http.{param => hparam, _}
 import com.twitter.finagle.tracing._
-import com.twitter.util.{Future, Return, Throw, Time, Try}
+import com.twitter.util.{Future, Return, StorageUnit, Throw, Time, Try}
+import com.twitter.conversions.StorageUnitOps._
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets.ISO_8859_1
 import java.util.Base64
@@ -49,6 +50,19 @@ import scala.collection.breakOut
 object Headers {
   val Prefix = "l5d-"
 
+  /**
+   * A param specific to error responses, the maximum size of a HTTP error response payload.
+   * By default it has the same value as maxResponseKB. Example:
+   * - protocol: http
+   *   maxResponseKB: 5120
+   */
+  object param {
+    final case class MaxErrResponseSize(size: StorageUnit) extends AnyVal
+    implicit object MaxErrResponseSize extends Stack.Param[MaxErrResponseSize] {
+      val default = MaxErrResponseSize(5.megabytes)
+    }
+  }
+
   object Ctx {
 
     /**
@@ -65,15 +79,15 @@ object Headers {
      * HttpTraceInitializer.serverModule.
      */
     val serverModule: Stackable[ServiceFactory[Request, Response]] =
-      new Stack.Module1[hparam.MaxHeaderSize, ServiceFactory[Request, Response]] {
+      new Stack.Module2[hparam.MaxHeaderSize, Headers.param.MaxErrResponseSize, ServiceFactory[Request, Response]] {
         val role = Stack.Role("ServerContext")
         val description = "Extracts linkerd context from http headers"
 
         val deadline = new Deadline.ServerFilter
-        def dtab(maxHeaderSize: Int) = new Dtab.ServerFilter(maxHeaderSize)
+        def dtab(maxHeaderSize: Int, maxErrResponseSize: Int) = new Dtab.ServerFilter(maxHeaderSize, maxErrResponseSize)
 
-        def make(maxHeaderSize: hparam.MaxHeaderSize, next: ServiceFactory[Request, Response]) =
-          deadline.andThen(dtab(maxHeaderSize.size.bytes.toInt)).andThen(next)
+        def make(maxHeaderSize: hparam.MaxHeaderSize, maxErrResponseSize: Headers.param.MaxErrResponseSize, next: ServiceFactory[Request, Response]) =
+          deadline.andThen(dtab(maxHeaderSize.size.bytes.toInt, maxErrResponseSize.size.bytes.toInt)).andThen(next)
       }
 
     val clearServerModule: Stackable[ServiceFactory[Request, Response]] =
@@ -255,12 +269,12 @@ object Headers {
        *
        * @todo use DtabFilter.Injector once it is released.
        */
-      class ServerFilter(maxHeaderSize: Int) extends SimpleFilter[Request, Response] {
+      class ServerFilter(maxHeaderSize: Int, maxErrResponseSize: Int) extends SimpleFilter[Request, Response] {
 
         def apply(req: Request, service: Service[Request, Response]) =
           get(req.headerMap) match {
             case Throw(e) =>
-              Future.value(Err.respond(e.getMessage, Status.BadRequest, maxHeaderSize))
+              Future.value(Err.respond(e.getMessage, Status.BadRequest, maxHeaderSize, maxErrResponseSize))
             case Return(dtab) =>
               clear(req.headerMap)
               FDtab.local ++= dtab
@@ -473,7 +487,7 @@ object Headers {
   object Err {
     val Key = Prefix + "err"
 
-    def respond(msg: String, status: Status = Status.InternalServerError, maxHeaderSize: Int): Response = {
+    def respond(msg: String, status: Status = Status.InternalServerError, maxHeaderSize: Int, maxErrResponseSize: Int): Response = {
       val rsp = Response(status)
       val header = URLEncoder.encode(msg, ISO_8859_1.toString)
       rsp.headerMap(Key) = if (header.length > maxHeaderSize) {
@@ -482,7 +496,13 @@ object Headers {
         header
       }
       rsp.contentType = MediaType.Txt
-      rsp.contentString = msg
+
+      if (msg.length > maxErrResponseSize) {
+        rsp.contentString = msg.substring(0, maxErrResponseSize)
+      } else {
+        rsp.contentString = msg
+      }
+
       rsp
     }
   }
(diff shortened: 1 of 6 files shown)


adleong:
Thanks for working on this, @dtacalau!  I think that you may be right about `maxHeadersKB` being unused... see [this discussion](https://github.com/linkerd/linkerd/issues/2196) for more details.  

I also agree with you that adding a new `maxErrResponseKB` property is probably a less confusing way forward.  Would you be interested in implementing that?

dtacalau:
@adleong I'm interested in implementing the maxErrResponseKB property, but will need Adam's ok for that.

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -65,15 +81,17 @@ object Headers {
      * HttpTraceInitializer.serverModule.
      */
     val serverModule: Stackable[ServiceFactory[Request, Response]] =
-      new Stack.Module1[hparam.MaxHeaderSize, ServiceFactory[Request, Response]] {
+      new Stack.Module3[hparam.MaxHeaderSize, hparam.MaxResponseSize, Headers.param.MaxErrResponseSize, ServiceFactory[Request, Response]] {
```
adleong: I think we shouldn't bring `MaxResponseSize` into this at all.  It should be sufficient to just check the `MaxErrResponseSize`.

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -49,6 +52,19 @@ import scala.collection.breakOut
 object Headers {
   val Prefix = "l5d-"
 
+  /**
+   * A param specific to error responses, the maximum size of a non-chunked HTTP error response payload.
```
adleong: I don't think we ever chunk encode error responses.

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -473,7 +491,10 @@ object Headers {
   object Err {
     val Key = Prefix + "err"
 
-    def respond(msg: String, status: Status = Status.InternalServerError, maxHeaderSize: Int): Response = {
+    private[this] val log = Logger.get("FramingFilter.ServerFilter")
```
adleong: I don't think this is in FramingFilter

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -482,7 +503,24 @@ object Headers {
...
-      rsp.contentString = msg
+
+      /* determine the max response size as a minimum(maxResponseSize, maxErrResponseSize) */
+      val maxResponseSizeNow = if (maxResponseSize > maxErrResponseSize) {
+        maxErrResponseSize
+      } else {
+        maxResponseSize
+      }
+
+      /**
```
adleong: I think this comment should be removed.

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
dtacalau: done

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
dtacalau: removed words from comment

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
dtacalau: removed the log val, was needed only for debugging

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
dtacalau: done

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -5,12 +5,15 @@ import com.twitter.finagle.buoyant.{Dst => BuoyantDst}
...
 import com.twitter.finagle.tracing._
-import com.twitter.util.{Future, Return, Throw, Time, Try}
+import com.twitter.util.{Future, Return, StorageUnit, Throw, Time, Try}
+import com.twitter.conversions.StorageUnitOps._
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets.ISO_8859_1
 import java.util.Base64
 import scala.collection.breakOut
 
+import com.twitter.logging.Logger
```
adleong: I think this is unused

adleong:
You'll also need to add the DCO signoff to your commits before we can merge this: https://github.com/apps/dco

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/HttpInitializerTest.scala
```diff
@@ -139,12 +140,13 @@ class HttpInitializerTest extends FunSuite with Awaits with Eventually {
...
     val maxReqSize = hparam.MaxRequestSize(40.kilobytes)
     val maxRspSize = hparam.MaxResponseSize(50.kilobytes)
+    val maxErrRspSize = Headers.param.MaxErrResponseSize(30.kilobytes)
     val streaming = hparam.Streaming(false)
     val compression = hparam.CompressionLevel(3)
 
     val router = HttpInitializer.router
       .configured(maxHeaderSize).configured(maxInitLineSize)
-      .configured(maxReqSize).configured(maxRspSize)
+      .configured(maxReqSize).configured(maxRspSize).configured(maxErrRspSize)
```
zaharidichev: Maybe add a newline for continuity/readability.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/HttpInitializerTest.scala
```diff
@@ -154,6 +156,7 @@ class HttpInitializerTest extends FunSuite with Awaits with Eventually {
     assert(sparams[hparam.MaxInitialLineSize] == maxInitLineSize)
     assert(sparams[hparam.MaxRequestSize] == maxReqSize)
     assert(sparams[hparam.MaxResponseSize] == maxRspSize)
+    assert(sparams[Headers.param.MaxErrResponseSize] == maxErrRspSize)
```
zaharidichev: `Headers.param.MaxErrResponseSize -> param.MaxErrResponseSize` would probably look a bit better

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
```diff
@@ -49,6 +52,19 @@ import scala.collection.breakOut
...
   val Prefix = "l5d-"
 
+  /**
+   * A param specific to error responses, the maximum size of a HTTP error response payload.
+   * By default it has the same value as maxResponseKB. Example:
+   * - protocol: http
+   *   maxResponseKB: 5120
+   */
+  object param {
+    case class MaxErrResponseSize(size: StorageUnit)
```
zaharidichev: TIOLI: Its uasually good practice to have your case classes being `final`. Can read more about it [here](https://nrinaudo.github.io/scala-best-practices/tricky_behaviours/final_case_classes.html). Also you can have the case class extending `AnyVal` to avoid the extra allocation (not that it will matter much)

zaharidichev:
@dtacalau Seems like you have some minor conflicts you have to sort out but other than that seems like everyone agrees this is good to merge.

linkerd/protocol/http/src/main/scala/com/twitter/finagle/buoyant/linkerd/LinkerdHeaders.scala
dtacalau: done

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/HttpInitializerTest.scala
dtacalau: renamed to hErrParam

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/HttpInitializerTest.scala
dtacalau: done

dtacalau:
@adleong sorry about that, I removed the conflict markers.

PR #2295, opened 2019-07-13
Enable streaming in Consul HTTP client

### LINKED ISSUE #2288: Stream chunked responses from Consul API
Consul's Agent API automatically switches to chunked responses when service discovery information becomes significantly large. Based on #2286, the HTTP client in the Consul namer does not handle chunked responses properly because the JSON parser expects the JSON data to be the entire payload when it begins parsing. The namer should be able to handle this scenario.

zaharidichev:
This work is continuation of #2287. In essence the change enables `streaming` for the consul http client and as a result of that makes handling large consul responses easier (i.e. not limited by max size config). The PR exposes the `fixedLengthStreamedAfterKB` which as described in the docs

> configures `fixedLengthStreamedAfter` limit, which effectively turns on streaming (think `withStreaming(true)`). The `fixedLengthStreamedAfter`, however, disables streaming for sufficiently small messages of known fixed length.

In order to reproduce the initial issue and validate the solution you can use the method described by @dadjeibaah in #2287. Note that if you want to see the initial issue you need to first run against [this](https://github.com/linkerd/linkerd/commit/753fc177f969f136cd40ddaf8ae0128a49c12f6e) version of the code.  After this PR is applied you can run the example again and convince yourself that exception is not thrown despite the fact that the message size configuration is not tweaked.

Fixes #2288 

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
@@ -3,7 +3,7 @@ package io.buoyant.consul.v1
 import com.fasterxml.jackson.databind.{DeserializationFeature, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
-import com.twitter.finagle.http.Request
+import com.twitter.finagle.http.{Request, Response}
 import com.twitter.finagle.buoyant.RetryFilter
 import com.twitter.finagle.param.HighResTimer
 import com.twitter.finagle.service.{RetryBudget, RetryPolicy}
@@ -13,6 +13,7 @@ import com.twitter.finagle._
 import com.twitter.io.Buf
 import com.twitter.util._
 import io.buoyant.consul.log
+
 import scala.util.control.NonFatal
 
 // a thunked version of the api call such that we can peek at the request before making the call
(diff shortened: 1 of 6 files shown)


consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -85,18 +86,24 @@ trait BaseApi extends Closable {
...
   mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
 
-  def parseJson[T: Manifest](buf: Buf): Try[T] = {
-    val Buf.ByteArray.Owned(bytes, begin, end) = Buf.ByteArray.coerce(buf)
-    Try(mapper.readValue[T](bytes, begin, end - begin))
-  }
+  private def parse[T: Manifest](rsp: Response): Future[T] =
+    for {
+      data <- rsp match {
+        case chunked if rsp.isChunked => chunked.chunkReader.accumulate.map({ case (buf, _) => Buf.ByteArray.coerce(buf) })
```
adleong: We don't need access to the trailers (indeed, we discard them here) so we can use the slightly more ergonomic `.reader` API instead of `.chunkReader`.  

I think this could simply be `Reader.readAll(chunked.reader)`

consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
```diff
@@ -85,18 +86,24 @@ trait BaseApi extends Closable {
...
+        case nonChunked => Future.value(Buf.ByteArray.coerce(nonChunked.content))
+      }
+      parsed <- data match {
```
adleong: my scala destructuring is a bit rusty but I think you might be able to simplify this as:

```
for {
  Buf.ByteArray.Owned(bytes, begin, end) <- rsp match {
    ...
  }
} yield mapper.readValue(bytes, begin, end - begin)
```

or even as 

```
val content = if rsp.isChunked {
  ...
} else {
  ...
}
content.map {
  case Buf.ByteArray.Owned(bytes, begin, end) => mapper.readValue(...)
}
```

zaharidichev:
@adleong I already did manual testing and it works. I guess @dadjeibaah can also try it out manually to verify further.

adleong:
Great work, @zaharidichev!

PR #2297, opened 2019-07-15
Return node and service metadata from consul in namerd response

### LINKED ISSUE #2269: Return meta attached to consul service in consul namer response
Issue Type:

- [ ] Bug report
- [x] Feature request

**What happened**:
Service registered in consul with metadata attached
```
$ curl --request PUT --data '\
{\
"ID": "foo-1",\
"Name": "foo",\
"Address": "10.10.10.10",\
"Port": 10101,\
  "Meta": {\
    "foo": "bar"\
  },\
}' http://127.0.0.1:8500/v1/agent/service/register
```
When querying via namerd, metadata from consul isn't returned by the namer:
```
{
"type": "bound",
"addrs": [
  {
    "ip": "10.10.10.10",
    "port": 10101,
    "meta": {
      "endpoint_addr_weight": 1
    }
  }
],
"meta": {}
}
```

I'd expect `"foo": "bar"` meta attached to the consul service to be present in the namerd response.

Currently meta in consul namer response carries some finagle-specific data. We need to make sure, that metadata from consul can't clash with the meta-data namer already stores.

zaharidichev:
This allows namerd to return metadata that is received from Consul for both services and nodes. 

Fixes #2269

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
@@ -159,7 +159,7 @@ class HealthApi(
 case class Node(
   Node: Option[String],
   Address: Option[String],
-  Meta: Option[MetaData]
+  Meta: Option[Metadata]
 )
 
 case class Service_(
@@ -168,7 +168,7 @@ case class Service_(
   Address: Option[String],
   Tags: Option[Seq[String]],
   Port: Option[Int],
-  Meta: Option[MetaData]
+  Meta: Option[Metadata]
 )
 
 case class ServiceHealth(
@@ -190,6 +190,6 @@ case class ServiceNode(
   ServiceAddress: Option[String],
   ServicePort: Option[Int],
   Status: Option[HealthStatus.Value],
-  ServiceMeta: Option[MetaData],
-  NodeMeta: Option[MetaData]
+  ServiceMeta: Option[Metadata],
+  NodeMeta: Option[Metadata]
 )
--- a/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
@@ -141,7 +141,9 @@ class HealthApi(
           service.flatMap(_.Tags),
           service.flatMap(_.Address),
           service.flatMap(_.Port),
-          checks.reduceOption(HealthStatus.worstCase)
+          checks.reduceOption(HealthStatus.worstCase),
+          service.flatMap(_.Meta),
+          node.flatMap(_.Meta)
         )
       }
       val nodes = if (statuses == Set(HealthStatus.Passing)) {
@@ -156,7 +158,8 @@ class HealthApi(
 
 case class Node(
   Node: Option[String],
-  Address: Option[String]
+  Address: Option[String],
+  Meta: Option[MetaData]
 )
 
 case class Service_(
@@ -164,7 +167,8 @@ case class Service_(
   Service: Option[String],
   Address: Option[String],
   Tags: Option[Seq[String]],
-  Port: Option[Int]
+  Port: Option[Int],
+  Meta: Option[MetaData]
 )
 
 case class ServiceHealth(
@@ -185,5 +189,7 @@ case class ServiceNode(
   ServiceTags: Option[Seq[String]],
   ServiceAddress: Option[String],
   ServicePort: Option[Int],
-  Status: Option[HealthStatus.Value]
+  Status: Option[HealthStatus.Value],
+  ServiceMeta: Option[MetaData],
+  NodeMeta: Option[MetaData]
 )
--- a/consul/src/main/scala/io/buoyant/consul/v1/package.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/package.scala
@@ -8,7 +8,7 @@ package object v1 {
   type Client = Service[http.Request, http.Response]
   type IndexedServiceNodes = Indexed[Seq[ServiceNode]]
   type IndexedServiceMap = Indexed[Map[String, Seq[String]]]
-  type MetaData = Map[String, Any]
+  type Metadata = Map[String, String]
 
   trait ConsulApiError extends Throwable {
     def rsp: http.Response
--- a/consul/src/main/scala/io/buoyant/consul/v1/package.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/package.scala
@@ -8,6 +8,7 @@ package object v1 {
   type Client = Service[http.Request, http.Response]
   type IndexedServiceNodes = Indexed[Seq[ServiceNode]]
   type IndexedServiceMap = Indexed[Map[String, Seq[String]]]
+  type MetaData = Map[String, Any]
 
   trait ConsulApiError extends Throwable {
     def rsp: http.Response
--- a/interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
+++ b/interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
@@ -1,5 +1,8 @@
 package io.buoyant.interpreter.mesh
 
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
 import com.twitter.finagle._
 import com.twitter.finagle.buoyant.h2
 import com.twitter.finagle.buoyant.h2.Reset
@@ -10,7 +13,7 @@ import com.twitter.util._
 import io.buoyant.admin.Admin
 import io.buoyant.config.Parser
 import io.buoyant.grpc.runtime.Stream
-import io.buoyant.namer.{DelegateTree, Delegator, InstrumentedActivity, InstrumentedVar, Metadata}
+import io.buoyant.namer.{DelegateTree, Delegator, InstrumentedActivity, InstrumentedVar}
 import io.linkerd.mesh
 import io.linkerd.mesh.Converters._
 import java.net.{InetAddress, InetSocketAddress}
@@ -44,6 +47,9 @@ object Client {
   private[this] val _rescueUnit: PartialFunction[Throwable, Future[Unit]] = {
     case NonFatal(_) => Future.Unit
   }
+
+  private[this] val mapper = new ObjectMapper with ScalaObjectMapper
+  mapper.registerModule(DefaultScalaModule)
 
   private[this] class Impl(
     root: Path,
@@ -334,9 +340,11 @@ object Client {
     case mesh.Endpoint(Some(_), Some(ipBuf), Some(port), pmeta) =>
       val ipBytes = Buf.ByteArray.Owned.extract(ipBuf)
       val ip = InetAddress.getByAddress(ipBytes)
-      val meta = Seq.empty[(String, Any)] ++
-        pmeta.flatMap(_.nodeName).map(Metadata.nodeName -> _)
-      Address.Inet(new InetSocketAddress(ip, port), Addr.Metadata(meta: _*))
+      Address
+        .Inet(
+          new InetSocketAddress(ip, port),
+          pmeta.mapValues(v => Try(mapper.readValue[Map[String, Any]](v)).getOrElse(v))
+        )
   }
 
   private[this] val fromReplicas: mesh.Replicas => Addr = {
@@ -347,8 +355,12 @@ object Client {
       case mesh.Replicas.OneofResult.Failed(mesh.Replicas.Failed(msg)) =>
         Addr.Failed(msg.getOrElse("unknown"))
 
-      case mesh.Replicas.OneofResult.Bound(mesh.Replicas.Bound(paddrs)) =>
-        Addr.Bound(paddrs.collect(_collectFromEndpoint).toSet)
+      case mesh.Replicas.OneofResult.Bound(mesh.Replicas.Bound(paddrs, meta)) =>
+        Addr
+          .Bound(
+            paddrs.collect(_collectFromEndpoint).toSet,
+            meta.mapValues(v => Try(mapper.readValue[Map[String, Any]](v)).getOrElse(v))
+          )
     }
   }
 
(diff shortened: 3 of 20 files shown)


zaharidichev:
@adleong, thanks for taking a look!

I assume that is best verified through looking at `/delegator.json`. So when I have consul + namerd + linkerd (configured to use namerd as an interpreter via the experimental http interface): 

1. I add some service data to consul: 
```
$ curl --request PUT --data '\
{\
"ID": "foo-1",\
"Name": "foo",\
"Address": "10.10.10.10",\
"Port": 10101,\
  "Meta": {\
    "foo": "bar"\
  },\
}' http://127.0.0.1:8500/v1/agent/service/register
```

2. I do 
`curl -X GET 'http://127.0.0.1:9990/delegator.json?namespace=http&path=/svc/foo'`

3. I am looking at: 
```json
{
    "type": "delegate",
    "path": "/svc/foo",
    "delegate": {
        "type": "leaf",
        "path": "/#/io.l5d.consul/dc1/foo",
        "dentry": {
            "prefix": "/svc",
            "dst": "/#/io.l5d.consul/dc1"
        },
        "bound": {
            "addr": {
                "type": "bound",
                "addrs": [
                    {
                        "ip": "10.10.10.10",
                        "port": 10101,
                        "meta": {
                            "endpoint_addr_weight": 1,
                            "node_meta": {
                                "consul-network-segment": ""
                            }
                        }
                    }
                ],
                "meta": {
                    "service_meta": {
                        "foo": "bar"
                    }
                }
            },
            "id": "/#/io.l5d.consul/dc1/foo",
            "path": "/"
        }
    }
}
```

So if this is the correct way to check that (as I suppose), I assume the answer is yes. Another detail that I stumbled upon during my codebase exploration is the fact that there is a thrift interface as well that I have not checked. I assume this also needs to transmit the metadata correctly?

zaharidichev:
@adleong I added configuration option to opt-in to the metadata transfer from consul. Regarding the mesh interface. I tested it and nothing is broken. However if we want to cary the metadata over the mesh interface, we end up in a similar situation as with the thrift interface. We need quite a bit of logic to serdes `Map[String,Any]`. Not sure whether its worth it if there is no user case for it. Your call

adleong:
@zaharidichev Awesome.  Given that the mesh interface is the one we recommend, I think it's worth investigating how to propagate this data.  The simplest method is probably to expand the gRPC response object to include a Map<String, String> metadata field.  What do you think?

zaharidichev:
@adleong, yeah dumping the whole json metadata in the value of the map as a `String` will certainly work and thats what I considered as well. Its a bit of a hack but if the mesh interface is whats being promoted then its worth it. I will do that

adleong:
Does the consul metadata contain structured (nested) data, or is it just key-value pairs?  If it's just KV pairs, we should be able to encode it as a protobuf Map<String, String> rather than encoding it as json.

zaharidichev:
It can be structured. However we can still encode that as a string by simply putting the json string there (hence my comment that its a bit hacky)

zaharidichev:
@adleong This now carries the metadata over the mesh interface without a problem

consul/src/main/scala/io/buoyant/consul/v1/ConsulApi.scala
```diff
@@ -185,5 +189,7 @@ case class ServiceNode(
   ServiceTags: Option[Seq[String]],
   ServiceAddress: Option[String],
   ServicePort: Option[Int],
-  Status: Option[HealthStatus.Value]
+  Status: Option[HealthStatus.Value],
+  ServiceMeta: Option[MetaData],
```
adleong: I think metadata is usually written as one word which means it would be capitalized as `Metadata`.

mesh/core/src/main/protobuf/resolver.proto
```diff
@@ -39,13 +39,7 @@ message Endpoint {
   int32 port = 3;
 
   /** Endpoint-specific diagonstic metadata */
-  message Meta {
-    /**
-     * In scheduled environments, a diagnostic name for the parent node.
-     */
-    string nodeName = 1;
```
adleong: We should use field number 5 to preserve backwards compatibility.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -334,9 +340,11 @@ object Client {
     case mesh.Endpoint(Some(_), Some(ipBuf), Some(port), pmeta) =>
       val ipBytes = Buf.ByteArray.Owned.extract(ipBuf)
       val ip = InetAddress.getByAddress(ipBytes)
-      val meta = Seq.empty[(String, Any)] ++
-        pmeta.flatMap(_.nodeName).map(Metadata.nodeName -> _)
-      Address.Inet(new InetSocketAddress(ip, port), Addr.Metadata(meta: _*))
+      Address
+        .Inet(
+          new InetSocketAddress(ip, port),
+          pmeta.mapValues(v => Try(mapper.readValue[Map[String, Any]](v)).getOrElse(v))
```
adleong: If the protobuf filed is a `map<string, string>` then we don't need to use an objectMapper and deserialize JSON.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
```diff
@@ -347,8 +355,12 @@ object Client {
       case mesh.Replicas.OneofResult.Failed(mesh.Replicas.Failed(msg)) =>
         Addr.Failed(msg.getOrElse("unknown"))
 
-      case mesh.Replicas.OneofResult.Bound(mesh.Replicas.Bound(paddrs)) =>
-        Addr.Bound(paddrs.collect(_collectFromEndpoint).toSet)
+      case mesh.Replicas.OneofResult.Bound(mesh.Replicas.Bound(paddrs, meta)) =>
+        Addr
+          .Bound(
+            paddrs.collect(_collectFromEndpoint).toSet,
+            meta.mapValues(v => Try(mapper.readValue[Map[String, Any]](v)).getOrElse(v))
```
adleong: Same here

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -20,6 +21,8 @@ private[consul] case class SvcKey(name: String, tag: Option[String]) {
 
 private[consul] object SvcAddr {
 
+  type MetaWithAddresses = (v1.MetaData, Set[Address])
```
adleong: isn't this type basically the same as `Addr.Bound`?

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -46,20 +49,30 @@ private[consul] object SvcAddr {
     preferServiceAddress: Option[Boolean] = None,
     tagWeights: Map[String, Double] = Map.empty,
     stats: Stats,
-    stateWatch: v1.PollState[String, v1.IndexedServiceNodes]
+    stateWatch: v1.PollState[String, v1.IndexedServiceNodes],
+    transferMetaData: Option[Boolean] = None
```
adleong: Let's make this just a `Boolean` throughout the consul namer and library code.  The ConsulInitializer should be responsible for taking the `Option[Boolean]` from the config and applying a default.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -46,20 +49,30 @@ private[consul] object SvcAddr {
...
         blockingIndex = index,
-        consistency = consistency,
-        retry = false
+        consistency = consistency
       )
       v1.InstrumentedApiCall.execute(apiCall, stateWatch)
-        .map(indexedToAddresses(preferServiceAddress, tagWeights))
+        .map {
+          case v1.Indexed(nodes, idx) => {
+            val meta: v1.MetaData = nodes.headOption.flatMap(_.ServiceMeta) -> transferMetaData match {
```
adleong: The `headOption` confused me at first.  It might be good to add a comment explaining that all nodes are assumed to have the same ServiceMeta.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -113,11 +126,11 @@ private[consul] object SvcAddr {
...
             Future.exception(NoIndexException)
 
-          case Return(v1.Indexed(addrs, xConsulIndex)) =>
+          case Return(v1.Indexed((srvMeta, addresses), xConsulIndex)) =>
             stats.updates.incr()
-            val addr = addrs match {
+            val addr = addresses match {
               case addrs if addrs.isEmpty => Addr.Neg
-              case addrs => Addr.Bound(addrs, meta)
+              case addrs => Addr.Bound(addrs, meta ++ Map(Metadata.serviceMeta -> srvMeta))
```
adleong: I don't think we need to bury the service meta under a "serviceMeta" key.  We can just do `meta ++ srvMeta`.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -148,47 +161,69 @@ private[consul] object SvcAddr {
...
       case None => 1.0
       case Some(Nil) => 1.0
       case Some(ws) => ws.max
     }
-    val meta = Addr.Metadata((Metadata.endpointWeight, weight))
-    Try(InetAddress.getAllByName(ip)
-      .toTraversable
-      .map(singleIP => Address.Inet(new InetSocketAddress(singleIP, port), meta)))
+    val meta = Addr.Metadata((Metadata.endpointWeight, weight)) ++ Map(
+      Metadata.nodeMeta -> m
```
adleong: No need to bury this under a nodeMeta key.

namerd/iface/mesh/src/main/scala/io/buoyant/namerd/iface/mesh/ResolverService.scala
```diff
@@ -67,7 +73,7 @@ object ResolverService {
 
     case Addr.Bound(addrs, meta) =>
       val paddrs = addrs.collect(_collectToEndpoint).toSeq
-      mesh.Replicas.OneofResult.Bound(mesh.Replicas.Bound(paddrs))
+      mesh.Replicas.OneofResult.Bound(mesh.Replicas.Bound(paddrs, meta.mapValues(v => mapper.writeValueAsString(v))))
```
adleong: Since we don't plan on serializing json here, we should just be able to use `toString`.

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -46,20 +49,30 @@ private[consul] object SvcAddr {
...
+
+    def getAddresses(index: Option[String]): Future[v1.Indexed[MetaWithAddresses]] = {
       val apiCall = consulApi.serviceNodes(
         key.name,
         datacenter = Some(datacenter),
         tag = key.tag,
         blockingIndex = index,
-        consistency = consistency,
-        retry = false
+        consistency = consistency
```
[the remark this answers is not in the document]
zaharidichev: because `false` is the default value here for this parameter and therefore redundant (unless we want to be overly explicit)

consul/src/main/scala/io/buoyant/consul/v1/package.scala
```diff
@@ -8,6 +8,7 @@ package object v1 {
   type Client = Service[http.Request, http.Response]
   type IndexedServiceNodes = Indexed[Seq[ServiceNode]]
   type IndexedServiceMap = Indexed[Map[String, Seq[String]]]
+  type MetaData = Map[String, Any]
```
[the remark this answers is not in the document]
zaharidichev: Yes, indeed I missed on the fact that they limit the value type to be a `string` and therefore it cannot be an arbitrary object graph.

interpreter/mesh/src/main/scala/io/buoyant/interpreter/mesh/Client.scala
zaharidichev: My initial thought was that the proto field is a string that represents either a leaf json value or a json object, hence I was trying to deserialize it as a tree first, to avoid the effect of having a json object represented as a string. However, now that I see that the consul api enforces all values to be string in the metadata map, then this gymnastics is not needed at all.

mesh/core/src/main/protobuf/resolver.proto
zaharidichev: I guess we should also till be populating the field 4 as well.

namer/consul/src/main/scala/io/buoyant/namer/consul/ConsulInitializer.scala
```diff
@@ -65,7 +66,8 @@ case class ConsulConfig(
   preferServiceAddress: Option[Boolean] = None,
   weights: Option[Seq[TagWeight]] = None,
   fixedLengthStreamedAfterKB: Option[Int],
-  tls: Option[TlsClientConfig] = None
+  tls: Option[TlsClientConfig] = None,
+  transferMetaData: Option[Boolean] = None
```
adleong: I think this should be capitalized as `transferMetadata`

namer/consul/src/main/scala/io/buoyant/namer/consul/SvcAddr.scala
```diff
@@ -46,20 +47,33 @@ private[consul] object SvcAddr {
...
       )
       v1.InstrumentedApiCall.execute(apiCall, stateWatch)
-        .map(indexedToAddresses(preferServiceAddress, tagWeights))
+        .map {
+          case v1.Indexed(nodes, idx) => {
+            // we use the metadata from the first node, because all nodes from
+            // the same service are assumed to have identical metadata.
+            // For reference: https://www.consul.io/api/catalog.html
+            val meta: v1.Metadata = nodes.headOption.flatMap(_.ServiceMeta) -> transferMetaData match {
```
adleong: Take it or leave it (TIOLI) suggestion:

This might be clearer as:

```
val meta = nodes.headOption.flatMap(_.ServiceMeta) match {
  case Some(serviceMeta) if transferMetadata => serviceMeta
  case _ => Map.empty
}
```

PR #2300, opened 2019-07-24
Introduce maxHops param to terminate requests that have hopped too much

### LINKED ISSUE #1411: proxy loop issue in linkerd

zaharidichev:
This PR introduces `maxCallDepth` param that defaults to 1000 at the moment.  This parameter, combined with a `MaxCallDepthFilter` allows us to detect requests that have done too many hops and return a proper `400` status and message.

Fixes #1411 

### DIFF
--- a/linkerd/docs/protocol-h2.md
+++ b/linkerd/docs/protocol-h2.md
@@ -86,6 +86,7 @@ initialStreamWindowBytes | 64KB | Configures `SETTINGS_INITIAL_WINDOW_SIZE` on s
 maxConcurrentStreamsPerConnection | 1000 | Configures `SETTINGS_MAX_CONCURRENT_STREAMS` on new streams.
 maxFrameBytes | 16KB | Configures `SETTINGS_MAX_FRAME_SIZE` on new streams.
 maxHeaderListByts | none | Configures `SETTINGS_MAX_HEADER_LIST_SIZE` on new streams.
+maxCallDepth | 10 | If set, limits the number of maximum hops. The number of calls is derived by inspecting the Via header. This can be used to prevent proxy loops.
 
 ## HTTP/2 Service Parameters
 
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/Message.scala
@@ -43,6 +43,7 @@ object Headers {
   val Path = ":path"
   val Scheme = ":scheme"
   val Status = ":status"
+  val Via = "via"
 
   def apply(pairs: Seq[(String, String)]): Headers =
     new Impl(pairs)
--- a/linkerd/docs/protocol-http.md
+++ b/linkerd/docs/protocol-http.md
@@ -64,6 +64,7 @@ Key | Default Value | Description
 --- | ------------- | -----------
 addForwardedHeader | null | If set, a `Forwarded` header is added to all requests.  See [below](#http-1-1-forwarded).
 timestampHeader | null | If set, the specified header will be added to outbound requests with a timestamp. See [below](#http-1-1-timestamp).
+maxCallDepth | 1000 | If set, limits the number of maximum hops. The number of hops is derived by inspecting the Via header. This can be used to prevent proxy loops.
 
 <a name="http-1-1-timestamp"></a>
 ### Adding Timestamp Headers ###
--- a/linkerd/protocol/h2/src/e2e/scala/io/buoyant/linkerd/protocol/h2/H2EndToEndTest.scala
+++ b/linkerd/protocol/h2/src/e2e/scala/io/buoyant/linkerd/protocol/h2/H2EndToEndTest.scala
@@ -544,4 +544,38 @@ class H2EndToEndTest extends FunSuite {
       await(router.close())
     }
   }
+
+
+  test("returns 400 for requests that have more than allowed hops") {
+    val config =
+      s"""|routers:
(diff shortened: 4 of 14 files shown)


linkerd/docs/protocol-h2.md
```diff
@@ -86,6 +86,7 @@ initialStreamWindowBytes | 64KB | Configures `SETTINGS_INITIAL_WINDOW_SIZE` on s
 maxConcurrentStreamsPerConnection | 1000 | Configures `SETTINGS_MAX_CONCURRENT_STREAMS` on new streams.
 maxFrameBytes | 16KB | Configures `SETTINGS_MAX_FRAME_SIZE` on new streams.
 maxHeaderListByts | none | Configures `SETTINGS_MAX_HEADER_LIST_SIZE` on new streams.
+maxHops | 10 | If set, limits the number of maximum hops. The number of hops is derived by inspecting the Via header. This can be used to prevent proxy loops.
```
adleong: What do you think of calling this parameter "maxCallDepth"?  Do you think that would be clearer?

router/base-http/src/main/scala/io/buoyant/router/http/MaxHopsFilter.scala
```diff
@@ -0,0 +1,47 @@
...
+  final case class MaxHopsExceeded(numHops: Int)
+    extends Exception(s"Maximum number of hops ($numHops) has been exceeded. Please check for proxy loops.")
+    with NoStackTrace
+
+  final case class Param(value: Int) extends AnyVal {
+    def mk(): (Param, Stack.Param[Param]) = (this, Param.param)
+  }
+
+  object Param {
+    implicit val param = Stack.Param(Param(10))
```
adleong: With a pure Linkerd self loop, we currently get to 1226 hops before hitting the `431 Request Header Fields Too Large`.  Let's set this default value larger to more closely match the behavior today and not break existing apps which have a call depth greater than 10.  What about 1000?

router/base-http/src/test/scala/io/buoyant/router/http/MaxHopsFilterTest.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  implicit object HttpRequestLike extends RequestLike[Request, HeaderMap] {
+    override def headers(request: Request): HeaderMap = request.headerMap
+  }
+
+
+  def service(maxHops: Int) = new MaxHopsFilter[Request, HeaderMap, Response](
+    maxHops,
+    Via
+  ) andThen Service.mk[Request, Response](_ => Future.value(Response()))
```
adleong: In the Linkerd codebase we usually avoid infix style and prefer the explicit:

```
).andThen(Service.mk[...](...))
```

router/base-http/src/main/scala/io/buoyant/router/http/MaxHopsFilter.scala
```diff
@@ -0,0 +1,47 @@
...
+
+class MaxHopsFilter[Req, H: HeadersLike, Rep](maxHops: Int, headerKey: String)(implicit requestLike: RequestLike[Req, H]) extends SimpleFilter[Req, Rep] {
+
+  def numHops(viaValue: String) = viaValue.split(",").length
+
+  def apply(req: Req, svc: Service[Req, Rep]): Future[Rep] = {
+    val headersLike = implicitly[HeadersLike[H]]
+
+    headersLike.get(requestLike.headers(req), headerKey) match {
+      case Some(v) if numHops(v) > maxHops => throw MaxHopsExceeded(maxHops)
```
adleong: use `Future.exception(MaxHopsExceeded(maxHops)` instead of throwing

PR #2303, opened 2019-07-26
Introduce option to enable dtab compression for Consul

### LINKED ISSUE #2202: Allow for compressed dtabs in Consul
Issue Type:

- [ ] Bug report
- [X] Feature request

This is motivated by the same use case describedin #2201. I'll copy/paste that use case here:

## Use case
We are trying to use Linkerd's dtab feature to build very large, dynamic routing tables. The goal to use Linkerd (instead of, e.g. Consul) to control which versions of which applications get routed to when a a request for a particular service is received.

This requires us to dynamically build most of our dtab, and requires that the dtab have an explicit dentry for every service in our eco system. E.g.:

...
/svc/cool-thing => /tagged/ver-1.3/cool-thing;
/svc/lame-thing => 0.1 * /tagged/ver-2.0/lame-thing & 0.9 * /tagged/ver-1.8/lame-thing;
...

This means that our dtab will get very large! The example I'm working with at the moment is 1811 lines long, and 143kb.

## Problem
We're using Consul as the KV store. Consul enforces a 512kb limit on the base64-encoded values. While my example dtab at 143kb is close to what our current "worst case" would be, we're concerned that this will not scale with time.

## Proposed solution
We'd like to have the ability to read/write gzipped (or otherwise compressed) dtabs from Consul (or any other datastore). The dynamic dtabs compress extremely well -- my current example is only 16k. The compression gains are only slightly lost with base64 encoding that Consul demands -- my final encoded, gzipped dtab was 21k.

zaharidichev:
Allows for making use of Gzip compression when interacting with Consul KV Api. 

Fixes #2202 

### DIFF
--- a/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/BaseApi.scala
@@ -13,6 +13,7 @@ import com.twitter.finagle._
 import com.twitter.io.{Buf, Reader}
 import com.twitter.util._
 import io.buoyant.consul.log
+import scala.collection.immutable.Map
 import scala.util.control.NonFatal
 
 // a thunked version of the api call such that we can peek at the request before making the call
@@ -117,4 +118,6 @@ object Headers {
   val Index = "X-Consul-Index"
 }
 
-case class Indexed[T](value: T, index: Option[String])
+case class Indexed[T](value: T, index: Option[String]) {
+  def mapValue[R](f: T => R): Indexed[R] = copy(value = f(value))
+}
--- a/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
+++ b/consul/src/main/scala/io/buoyant/consul/v1/KvApi.scala
@@ -1,21 +1,19 @@
 package io.buoyant.consul.v1
 
 import java.util.Base64
-
-import com.twitter.conversions.DurationOps._
 import com.twitter.finagle.http
-import com.twitter.finagle.service.Backoff
 import com.twitter.finagle.stats.{DefaultStatsReceiver, StatsReceiver}
 import com.twitter.util._
 
 object KvApi {
-  def apply(c: Client, backoff: Stream[Duration]): KvApi = new KvApi(c, s"/$versionString", backoff)
+  def apply(c: Client, backoff: Stream[Duration], enableValueCompression: Boolean = false): KvApi = new KvApi(c, s"/$versionString", backoff, enableValueCompression)
 }
 
 class KvApi(
   val client: Client,
   val uriPrefix: String,
   val backoffs: Stream[Duration],
+  val enableValueCompression: Boolean,
   val stats: StatsReceiver = DefaultStatsReceiver
 ) extends BaseApi with Closable {
   val kvPrefix = s"$uriPrefix/kv"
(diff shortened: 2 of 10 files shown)


zaharidichev:
@adleong @dadjeibaah 

Thanks for the review. So yes, validating that manually was pretty straightforward. Here are the steps you can take: 

- Fire up consul: ./consul agent -dev -advertise 127.0.0.1
- Start Namerd: 
```
admin:
  ip: 0.0.0.0
  port: 9991
interfaces:
  - kind: io.l5d.httpController
storage:
  kind: io.l5d.consul
  compressValues: true
```
- Create a dtab entry via the API: 
```
curl -X POST \
  http://127.0.0.1:4180/api/1/dtabs/graduation \
  -H 'Content-Type: application/json' \
  -H 'Host: 127.0.0.1:4180' \
  -d '[{"prefix":"/yeezy","dst":"/kanye"}]'
```
- Do a `curl http://127.0.0.1:8500/v1/kv/namerd/dtabs/graduation` and you will get something like this: 
```
[
    {
        "LockIndex": 0,
        "Key": "namerd/dtabs/graduation",
        "Flags": 0,
        "Value": "SDRzSUFBQUFBQUFBQU5PdlRFMnRxclMxMDg5T3pLdE1CUUFBeWttbkRnQUFBQT09",
        "CreateIndex": 16,
        "ModifyIndex": 16
    }
]
```

This is straight from consul. After you base64 decode the value and decompress it with Gzip,
you can convince yourself that indeed Consul is stroing the gzipped representation of `/yeezy=>/kanye`
- Now do a curl -X GET http://127.0.0.1:4180/api/1/dtabs/graduation in order to verify
that the KVApi can indeed decompress the dtab: 

```
[
    {
        "prefix": "/yeezy",
        "dst": "/kanye"
    }
]
```

If there is a better way to verify that let me know.

zaharidichev:
Now its important to keep in mind that this is a rather primitive solution in a sense that when you set this compression to `true`, you assume that your consul instance will only contain compressed dtabs, so you cannot really go back. So if one wants to transition to this approach while keeping their old data in consul, its not possible.. Instead what we would have to do it maybe wrap the data we store in consul in some wrapper such as `case class Entry(data: String, compressed: Boolean)`
So then the logic (in order to be able to support values written to consul before this change) would be something like: 

```scala
match parseContentAs[Entry](rsp.content) {
   case Failure(_) => rsp.content
   case Success(Entry(data, true)) => decompress(data)
   case Success(Entry(data, false)) => data
}
```

This is if we really want to keep it backwards compatible wrt the data already living in a Consul instance. @chrismikehogan maybe it would be useful to share a bit more details around your usecase in order to determine whether this is actually necessary because this try-this-then-that pattern has some perf implications as well.

adleong:
@zaharidichev You raise some really good points.  I haven't heard back from anyone who had been requesting this feature.  I think we should put this aside for now (perhaps close the PR but keep the branch) until we get more clarity on the use-case.  This implementation is solid but I'd rather not merge a feature that isn't actually useful to anyone.  Does that make sense?

zaharidichev:
Yes, certainly makes sense. Lets avoid codebase bloating. If we get more clarity on the request, I will pick that up again.

adleong:
Closing for now due to lack of clarity on requirements.

PR #2309, opened 2019-08-16
Avoid leaking memory when receiving error response from remote

### LINKED ISSUE #2307: Leaking H2 frames when  H2 server goes down
During some of my memory leaking investigations, I came across a problem that leads to OOM. 

- LinkerD config:
```
admin:
  port: 9995

routers:
  - protocol: http
    dtab: |
      /svc => /$/inet/127.1/9995;
    servers:
      - port: 4145
  - protocol: h2
    h2AccessLog: logs/access.log
    h2AccessLogRollPolicy: daily
    h2AccessLogAppend: true
    h2AccessLogRotateCount: -1
    dtab: |
      /svc => /$/inet/127.1/9999;
    identifier:
      kind: io.l5d.header.token
      header: ":authority"
    servers:
      - port: 4150
        maxConcurrentStreamsPerConnection: 300
        addForwardedHeader:
          by: {kind: "ip:port"}
          for: {kind: "ip:port"}
        initialStreamWindowBytes: 1048576 # 1MB
    client:
      initialStreamWindowBytes: 1048576 # 1MB
```

Steps to reproduce: 
1. Fire up linkerd with VM option:
`-Xmx128M -XX:MaxDirectMemorySize=64M -Dio.netty.noUnsafe=true`
2. Start dealiner server:

        java -Dlogback.configurationFile=logback.xml \
                -Djava.util.logging.config.file=logging.properties \
                -classpath deadliner.jar \
                 com.bigcommerce.deadliner.DeadlinerServer

3. Start dealiner client:

       java -Dlogback.configurationFile=logback.xml \
              -classpath deadliner.jar \
               com.bigcommerce.deadliner.DeadlinerClient \
               127.0.0.1 \
               4150 \
               20 \
               1000000 \
               1
4. Wait until around 50000 requets have been issues by the client
5. Observe your heap memory and confirm its stable
6. Kill the  deadliner server.
7. Observe that your heap grows and GC activity does not seem to be fixing the problem...
8. Wait for OOM to happen

Looking at the memory footprint over the time of this experiment you can observe this: 
Screenshot 2019-08-13 at 16 03 30

Looking at the heap dumps, you can also see that after the server gets killed, there begins accumulation of `PoolChunk` objects on the heap, which indicates we are leaking frames somewhere (i.e. failing to call `.release()`): 

**At 50k requests(before server gets killed):**
Screenshot 2019-08-13 at 15 46 02

**Right before OOM:**
Screenshot 2019-08-13 at 15 50 57

Also, if we override the `.finalize()` method on `AsyncQueueReader` to detect queues being GCs while still empty, we will quickly see that this starts happening right after we kill the server. Obviously we are not draining our stream and releasing the frames at some point, which causes `ByteBufs` not being released and as consequence `PoolChunks` getting stuck on the heap.

zaharidichev:
Whenever we end up with a error Response from the remote, we need to make sure that we properly take care of the `rcvMsg` future in the `Netty4StreamTransport` and ensure that there are no pending frames in the queue. If that is the case and we dont drain them, they will hang around in memory for the lifecycle of the application. More details on what this ultimatelly causes can be found in #2307

Fixes #2307 

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
@@ -389,6 +389,17 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
    */
   def send(msg: SendMsg): Future[Future[Unit]] = {
 
+    // if we receive a response with an error status,
+    // it is important to cancel the stream of the rcvMessage
+    // since there might be frames there that will never make
+    // it to the remote and will cause a memory leak...
+    def onErrorResponseWritten(): Unit = msg match {
+      case Response(Status.ClientError(_) | Status.ServerError(_)) =>
+        onRecvMessage.onSuccess(_.stream.cancel(Reset.Refused))
+        ()
+      case _ => // do nothing if resp is ok
+    }
+
     val headersF = writeHeaders(msg).onFailure {
       // If writeHeaders fails, the state will be left in SendPending
       // so we must cancel the send stream ourselves.
@@ -397,7 +408,7 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message] 
     }
     val streamFF = headersF.map { _ =>
       if (msg.stream.isEmpty) Future.Unit
-      else localResetOnCancel(writeStream(msg.stream))
+      else localResetOnCancel(writeStream(msg.stream).onSuccess(_ => onErrorResponseWritten()))
     }
 
     val writeF = streamFF.flatten
--- a/finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
+++ b/finagle/h2/src/test/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransportTest.scala
@@ -3,14 +3,15 @@ package netty4
 
 import com.twitter.concurrent.AsyncQueue
 import com.twitter.finagle.Failure
+import com.twitter.finagle.buoyant.h2.netty4.Netty4StreamTransport.NullStatsReceiver
 import com.twitter.finagle.stats.InMemoryStatsReceiver
 import com.twitter.util._
 import io.buoyant.test.FunSuite
 import io.netty.buffer.Unpooled
-import io.netty.handler.codec.http2._
+import io.netty.handler.codec.http2.{DefaultHttp2DataFrame, _}
 import java.net.SocketAddress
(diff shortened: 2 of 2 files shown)


finagle/h2/src/main/scala/com/twitter/finagle/buoyant/h2/netty4/Netty4StreamTransport.scala
```diff
@@ -389,6 +389,17 @@ private[h2] trait Netty4StreamTransport[SendMsg <: Message, RecvMsg <: Message]
    */
   def send(msg: SendMsg): Future[Future[Unit]] = {
 
+    // if we receive a response with an error status,
+    // it is important to cancel the stream of the rcvMessage
+    // since there might be frames there that will never make
+    // it to the remote and will cause a memory leak...
+    def onErrorResponseWritten(): Unit = msg match {
```
adleong: I'm a bit confused about this.  `msg` is the message to be sent to the remote, not a message received from the remote.  Why do we match on the send message?

zaharidichev: So lets say we have a `client -> linkerd -> server` situation. The server is unresponsive, cannot be reached. Whats happening is that in this case the `ErrorReseter` will actually catch that `RichConnectionFailedExceptionWithPath` and translate it to `Future.value(LinkerdHeaders.Err.respond(e.exceptionMessage, Status.BadGateway)): Response` to be sent to the remote. In this case the remote is the client. Now, this is all fine because the client will get a proper description of what happened along with the appropriate status code. However this transport may still have frames in its recvQ that never made it to the server. We need to release these. This is why I match on the message that is being sent back to the client and if its indicating an error, I clear out the recvQ. At least this is my understanding. I might be completely wrong here. Let me know :)

adleong: Gotcha.  I think we can solve this in a more robust way elsewhere.  It seems that the problem is that when a connection cannot be established, the request that triggered that connection is dropped without draining the request's stream.  Take a look at RoutingFactory.scala.  It seems like perhaps if the call to `clientFactory` fails, the request should be properly disposed of, instead of just dropped.  

This is somewhat complicated by the fact that the request type is generic but the request stream discard behavior is protocol specific.  We deal with a similar problem elsewhere by using a ResponseDiscarder.  We can probably do something similar here and have a RequestDiscarder which is passed to this module as a stack param.

zaharidichev: Hmmm, yes I will take a closer look into that as I have the feeling that the  `Netty4StreamTransport` is probably not the best place to intercept that.

adleong:
Closing in favor of #2311

Issue #2304, opened 2019-08-05
Error during namerd startup with dcos-exec script


agaffney:
Issue Type:

- [X] Bug report
- [ ] Feature request

**What happened**:

When launching namerd from the `buoyantio/namerd:1.6.4-dcos` image, the following error is seen:

```
./dcos-exec: 72: read: 0: bad variable name
```

The error doesn't seem to actually hurt anything, but it appears that the `read` command here is malformed:

https://github.com/linkerd/linkerd/blob/master/project/LinkerdBuild.scala#L483

It's not clear what the intent of that code is, so I can't offer a suggestion to fix it.

**What you expected to happen**:

No error message

**How to reproduce it (as minimally and precisely as possible)**:

Run namerd from the `buoyantio/namerd:1.6.4-dcos` Docker image

**Anything else we need to know?**:

N/A

**Environment**:
- linkerd/namerd version, config files: 1.6.4
- Platform, version, and config files (Kubernetes, DC/OS, etc): DC/OS
- Cloud provider or hardware configuration: AWS

agaffney:
It looks like this was introduced in https://github.com/linkerd/linkerd/pull/2272 when the `-t` option was changed to `-r`, but the `0` argument to `-t` was not dropped.

dadjeibaah:
@agaffney thanks for filing this. We will look into this.

PR #2310, opened 2019-08-18
Remove parameter from -r conditional

cpretzer:

Removed the variable, per #2304 

I'll set up a DC/OS installation to test shortly


adleong:
@kleimkuhler would you mind reviewing this?

PR #2311, opened 2019-08-19
Release frames  in request stream on client factory failure

zaharidichev:
This PR is an alternative fix for #2307 based on the suggestion given by @adleong right [here](https://github.com/linkerd/linkerd/pull/2309#discussion_r314824524). This simply consumes and discards all frames from the request queue when an exception occurs and uses a `RequestDiscarder` to do so. 


router/core/src/main/scala/io/buoyant/router/RoutingFactory.scala
```diff
@@ -181,8 +191,17 @@ class RoutingFactory[Req, Rsp](
 
         // Service failures are recorded within the clientFactory's
         // path stack, too.
-        rsp <- service(req1).ensure {
+        rsp <- service(req1).transform {
```
adleong: This doesn't quite fit my mental model of request ownership.  I see this line as giving ownership of the request to `service` and therefore that service is now responsible for reading or discarding the request's stream.  I think the problem occurs when the above call to `clientFactory` fails and we don't give the request to the service at all.  In that case, it remains our responsibility to drain the stream.

Does it work to add a failure handler (`onFailure`) to the call to clientFactory?

zaharidichev: So I might be way off here so correct me if I am wrong.  The above call `clientFactory(dst, conn)` does not actualyl fail. Moreover, if it did fail, the initial fix that I proposed in #2309 would have not worked, because the request would have not reached the `Netty4StreamTransport`. So by the time we fail, the request is already in the recvQueue, waiting to be sent to the other side, which in my mind is a step further than the call to `clientFactory(dst, conn)`. So, in summary, we do give the request to the service and this is actually where the problem occurs, because we give it to it and forget about it :) Let me know if you have any other ideas.

adleong: Interesting!  So the call to `clientFactory` actually succeeds even though the connection can't be established?  Pretty weird.  Perhaps this is due to `FactoryToService` or something like that...

Yes, if we give away the request to the `service`, it also makes sense to give away control of the lifecycle.  We should figure out how far through the filter chain the request gets before encountering an error and make sure that whoever drops the request properly discards it.

adleong: Ah, maybe this is it:  In `class FactoryToService`, if the call to `factory()` fails, we drop the request without passing it along.  We could try creating our own `DiscardingFactoryToService` and replacing the standard FactoryToService module with our own.

zaharidichev: yes will try that out. The importat piece here is that although we need to drop the request, we certainly should not fail the stream alltogether, because if we do, the enriched error response will not reach the client back. So we should simply drain the queue (i.e. consume every frame and release it).

zaharidichev: Yeah, I have the feeling this will work well. Thanks for the useful tips @adleong

router/core/src/main/scala/io/buoyant/router/DiscardingFactoryToService.scala
```diff
@@ -0,0 +1,99 @@
...
+    }
+    def param[Req] = _param.asInstanceOf[Stack.Param[RequestDiscarder[Req]]]
+  }
+
+  /**
+   * Creates a [[com.twitter.finagle.Stackable]]
+   * [[DiscardingFactoryToService]]. This makes per-request service acquisition
+   * part of the stack so it can be wrapped by filters such as tracing.
+   */
+  def module[Req, Rep]: Stackable[ServiceFactory[Req, Rep]] = {
```
adleong: I don't think this module is ever used so we can omit defining it.  In `Router` it looks like we just use DiscardingFactoryToService directly.

router/core/src/main/scala/io/buoyant/router/DiscardingFactoryToService.scala
```diff
@@ -0,0 +1,99 @@
+package io.buoyant.router
+
+import com.twitter.finagle._
+import com.twitter.util.{Future, Time}
+import io.buoyant.router.DiscardingFactoryToService.RequestDiscarder
+
+/**
+ * Turns a [[com.twitter.finagle.ServiceFactory]] into a
```
adleong: We should add an attribution comment saying this code was copied+modified from finagle's FactoryToService.

PR #2315, opened 2019-08-20
Apply default socketoptions values when partial config is supplied.

fantayeneh:
Fixes #2313 


finagle/buoyant/src/main/scala/com/twitter/finagle/buoyant/SocketOptionsConfig.scala
```diff
@@ -27,7 +27,7 @@ case class SocketOptionsConfig(
     }
 
     Stack.Params.empty +
-      Transport.Options(noDelay, reuseAddr, reusePort) +
+      Transport.Options(noDelay.getOrElse(true), reuseAddr.getOrElse(true), reusePort.getOrElse(false)) +
```
zaharidichev: I think it would be good form to put these in some constant such as `DefaultNoDelay`, etc.

PR #2324, opened 2019-09-03
Fix content-type header in gRPC runtime

edio:
According to the spec https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md gRPC servers and clients must set content-type header to `application/grpc`.
Failing to do so makes valid gRPC applications (particularly ones written in go) to fail during request/response validation.

This PR updates buoyant gRPC runtime to set `content-type` header to `application/grpc+proto`.

This PR does NOT add `content-type` header validation on server though. This would be a breaking change which may cause denial of service for existing linkerd users.

### DIFF
--- a/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
+++ b/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
@@ -3,19 +3,21 @@ package io.buoyant.grpc.runtime
 import com.twitter.finagle.buoyant.h2
 
 private[runtime] object H2Headers {
-  private[this] val ContentType = "content-type"
-  private[this] val GrpcProtoContentType = "application/grpc+proto"
+  private[this] val AuthorityHeader = h2.Headers.Authority -> ""
+  private[this] val ContentTypeHeader = "content-type" -> "application/grpc+proto"
+  private[this] val MethodHeader = h2.Headers.Method -> h2.Method.Post.toString
+  private[this] val SchemeHeader = h2.Headers.Scheme -> "http"
 
   def responseHeaders(http2Status: h2.Status = h2.Status.Ok): h2.Headers = h2.Headers(
     h2.Headers.Status -> http2Status.toString,
-    ContentType -> GrpcProtoContentType
+    ContentTypeHeader
   )
 
   def requestHeaders(path: String): h2.Headers = h2.Headers(
-    h2.Headers.Scheme -> "http",
-    h2.Headers.Method -> h2.Method.Post.toString,
-    h2.Headers.Authority -> "",
     h2.Headers.Path -> path,
-    ContentType -> GrpcProtoContentType
+    AuthorityHeader,
+    ContentTypeHeader,
+    MethodHeader,
+    SchemeHeader
   )
 }
--- a/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
+++ b/grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
@@ -1,0 +1,21 @@
+package io.buoyant.grpc.runtime
+
+import com.twitter.finagle.buoyant.h2
+
+private[runtime] object H2Headers {
+  private[this] val ContentType = "content-type"
+  private[this] val GrpcProtoContentType = "application/grpc+proto"
+
+  def responseHeaders(http2Status: h2.Status = h2.Status.Ok): h2.Headers = h2.Headers(
+    h2.Headers.Status -> http2Status.toString,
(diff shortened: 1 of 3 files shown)


grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
```diff
@@ -148,7 +149,13 @@ class ServerDispatcher(services: Seq[ServerDispatcher.Service])
...
     }.toMap
 
-  override def apply(req: h2.Request): Future[h2.Response] =
+  override def apply(req: h2.Request): Future[h2.Response] = {
+    // According to gRPC spec, server must return 415 if content-type is not set to
+    // application/grpc to allow for graceful error handling when plain H2 client
+    // makes request to a gRPC endpoint.
+    // This, however, would be a breaking change on linkerd side.
+    // Postponing content-type validation for now.
+    // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
```
zaharidichev: Maybe add an issue as a reminder that this needs to be implemented for the next major release

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
```diff
@@ -0,0 +1,21 @@
...
+  private[this] val ContentType = "content-type"
+  private[this] val GrpcProtoContentType = "application/grpc+proto"
+
+  def responseHeaders(http2Status: h2.Status = h2.Status.Ok): h2.Headers = h2.Headers(
+    h2.Headers.Status -> http2Status.toString,
+    ContentType -> GrpcProtoContentType
+  )
+
+  def requestHeaders(path: String): h2.Headers = h2.Headers(
+    h2.Headers.Scheme -> "http",
```
zaharidichev: maybe we can put that in as val the same way we do for `GrpcProtoContentType`.

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
```diff
@@ -0,0 +1,21 @@
...
+    h2.Headers.Method -> h2.Method.Post.toString,
+    h2.Headers.Authority -> "",
```
zaharidichev: Same as above

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
```diff
@@ -0,0 +1,21 @@
+package io.buoyant.grpc.runtime
+
+import com.twitter.finagle.buoyant.h2
+
+private[runtime] object H2Headers {
+  private[this] val ContentType = "content-type"
```
adleong: we can reuse the constant defined at `com.twitter.finagle.http.Fields.ContentType`

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/ServerDispatcher.scala
adleong: :+1: to filing a github issue and linking to it in this comment

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
edio: I'd have to adjust dependencies for this to work. IMO, not worth it for a single string constant. But let me know please if you believe it's better that way, I'll update deps and code then

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
edio: I decided to change the code slightly. Although `Headers` are mutable, tuples are not.

So I decided declare entire tuples as fields on the object (and maybe, if we are unlucky with JIT, save few heap allocations in runtime). This way, I believe, it's more readable, and gives less reasons to extract string literals.

edio:
Is it possible to retrigger CI? I believe, the failures are not related

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
zaharidichev: Good thinking!

grpc/runtime/src/main/scala/io/buoyant/grpc/runtime/H2Headers.scala
adleong: fair enough!

PR #2327, opened 2019-09-10
Respect maxConcurrentStreams 

zaharidichev:
This commit adds the functionality to respect maxConcurrentStreams for server dispatchers. The reason for this is that H2 settings such as `maxConcurrentStreams` are merely advertisement based. We as a server need to protect against misbehaving clients that do not respect the settings we have advertised. Otherwise we are prone to memory leaks and all sorts of resource depletion problems. 

This  PR is a limited version of #2325 which was a more opinionated version of this work. 

### DIFF
--- a/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/H2.scala
+++ b/finagle/h2/src/main/scala/com/twitter/finagle/buoyant/H2.scala
@@ -2,6 +2,7 @@ package com.twitter.finagle.buoyant
 
 import com.twitter.finagle.buoyant.h2._
 import com.twitter.finagle.buoyant.h2.netty4._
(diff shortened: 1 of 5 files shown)


zaharidichev:
@adleong We can easily avoid iterration by maintaining counters or splitting things in multiple maps. I think that adds complexity to begin with. Also I think counting streams that are in StreamLocalReset is the correct thing to do. After all these streams are still occupying some sort of resources. I think it is correct to assume that in well behaved scenarios these streams will be evicted from the map once the other side closes as well, so its safe to simply do a `.size`.  Will add a comment about that one. 

On another note, had to google what a _spelunkers_ is... 

> A spelunker is an explorer of caves. If you hope to one day be a spelunker, you probably have a love of dark, damp spaces and headlamps.

Spot on! Probably the best word to describe anyone who delves into the Netty4BaseDispatcher code

Issue #2114, opened 2018-08-29
Support "b3" trace header


adriancole:
Let's support at least reading "b3" header from a single string, most commonly traceid-spanid-1
It would also be nice to support optionally writing this, especially in message providers or others with constrained environments.

### Expected behavior

As discussed on openzipkin/b3-propagation#21 and first implemented here: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3SingleFormat.java https://github.com/openzipkin/brave/blob/master/brave/src/test/java/brave/propagation/B3SingleFormatTest.java we should be able to parse "b3" and ideally also write it depending on config.

### Actual behavior

right now, `ZipkinTracePropagator` writes the historical "X-B3-" headers, which we should still support. However, if we see a header named "b3" we should read that instead.

### Steps to reproduce the behavior

Turn on tracing and see if `b3=80f198ee56343ba864fe8b2a57d3eff7-05e3ac9a4f6e3b90-1-e457b5a2e4d86bd1` results in the same context as:

```
X-B3-TraceId: 80f198ee56343ba864fe8b2a57d3eff7
X-B3-ParentSpanId: 05e3ac9a4f6e3b90
X-B3-SpanId: e457b5a2e4d86bd1
X-B3-Sampled: 1
```

adleong:
Thanks, @adriancole!  I think this would be a very simple change to https://github.com/linkerd/linkerd/blob/master/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala  Are you interested in working on this?

adriancole:
yep can do if no one else in a timeout :P

dtacalau:
@adleong  can you please assign this issue to me?

dtacalau:
Would like to have a discussion about the design around this enhancement:
- if we see a request containing a header named b3, read it, ignore any other the historical "X-B3-" headers, write b3 to propagated request ( just read "b3" => propagate "b3"). I don't think we need a config change here.
- if we see a request containing historical "X-B3-" headers and no "b3" header, read "X-B3-", write "X-B3-" to propagated request (current behavior still supported).

@adriancole @adleong not sure I understand what this request means: "It would also be nice to support optionally writing this"?  Is this covered by the changes described above or do you mean to always write single header "b3", if asked by config, no matter if input was "b3" or historical "X-B3-"?

Thanks!

adriancole:
[...]

PR #2329, opened 2019-09-24
Fix for #2114 Support b3 trace header 

dtacalau:
With this fix Linkerd will propagate b3 single header the same way it propagates x-b3- multi headers. If a request contains "b3" single header and "X-B3-" multi headers, the trace context from "b3" is preferred and "x-b3-" will be ignored.  "b3" is  handled the same way as finagle: extract the traceid from "b3" header, write back "X-B3-*" multi headers, and let the flow continue as if it has received "x-b3-". 

No config changes were needed, "b3" is not written on the way out, not sure if it's a good idea since finagle does not write "b3".

Also added support for 128bit traceId has been implemented with this fix because "b3" supports 128bit traceIds.

This has been tested in two ways:
- unit tests for "b3" single header parsing and handling, also this unit tests are good for testing the handling of "x-b3-" multi headers
- manual testing by sending a request with various "x-b3-" headers to a l5d router which has tracePropagator configured, and checked the behaviour hasn't changed after implementing "b3" support.

Signed-off-by: dst4096 <dst4096@gmail.com>


linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
```diff
@@ -61,28 +59,122 @@ object ZipkinTrace {
...
+          }
+          case _ => ()
+        }
+      def handleTraceAndSpanIds(headers: HeaderMap, a: String, b: String): Unit = {
+        headers.set(ZipkinTraceHeader, a)
+        headers.set(ZipkinSpanHeader, b)
+        ()
+      }
+
+      val _ = caseInsensitiveGet(headers, ZipkinB3SingleHeader).map(_.split("-").toList) match {
```
zaharidichev: You do not need to assign the result of this operation to _. You probably got the impression that this needs to happen because of the compile time failure indicating that you are ignoring a non null value. But here you are not

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
```diff
@@ -61,28 +59,122 @@ object ZipkinTrace {
...
+        case Some(a) =>
```
zaharidichev: Since this is a side - effecting operation, you do not need to match on None really, as nothing useful is happening at that point.  Also,  matching on the size in this way is not very idiomatic. You can use cons matches I think. So it would look a bit like: 

```
      caseInsensitiveGet(headers, ZipkinB3SingleHeader).map(_.split("-").toList).foreach {
        case  Nil => ///
        case  a :: Nil =>
        case  a :: b :: Nil =>
        case  a :: b :: c :: Nil =>
      }
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
```diff
@@ -61,28 +59,122 @@ object ZipkinTrace {
...
+            case 1 =>
+              // either debug flag or sampled
+              handleSampled(headers, a(0))
+            case 2 =>
+              // this is required to be traceId, spanId
+              handleTraceAndSpanIds(headers, a(0), a(1))
+            case 3 =>
+              handleTraceAndSpanIds(headers, a(0), a(1))
+              handleSampled(headers, a(2))
+            case 4 =>
```
zaharidichev: What happens if the size of the list is more than 4, say 5. I think in this case as this is a non exhaustive match, you will get an exception.  In such cases, just ensure that your match is exhaustive or simply add a default match case

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
zaharidichev: Not sure there is a need for `.toList` here

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
```diff
@@ -61,28 +59,122 @@ object ZipkinTrace {
...
     }
   }
 
   def set(headers: HeaderMap, id: TraceId): Unit = {
     val _ = headers.set(ZipkinSpanHeader, id.spanId.toString)
-    val __ = headers.set(ZipkinTraceHeader, id.traceId.toString)
+
+    // support setting a 128bit traceid
+    if (id.traceIdHigh.isEmpty) {
+      val __ = headers.set(ZipkinTraceHeader, id.traceId.toString)
```
zaharidichev: no need for assignign this to a val `___`

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,305 @@
+package io.buoyant.linkerd.protocol.http
+
+import io.buoyant.linkerd.protocol.ZipkinTrace
+import io.buoyant.linkerd.protocol.ZipkinTracePropagator
+
```
zaharidichev: Unneccessary empty line

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,305 @@
...
+import com.twitter.finagle.http.Request
+
```
zaharidichev: And here

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,305 @@
...
+    assert(trace.isEmpty)
+
+    //b3 has been removed
+    assert(req.headerMap.get("b3").isEmpty)
+    // check "x-b3-sampled" has been added to request
+    assert(req.headerMap.keys == Set("x-b3-sampled"))
+
+    // expect to get the right sampled value which is 0
+    var sampler = ZipkinTrace.getSampler(req.headerMap)
+    sampler match {
```
zaharidichev: Instead of doing explicit `assert(false)` whenever you have an empty option, I think you can simply combine the two with the contains combinator such as `assert(sampler.contains(0.0f))`. This will be a lot more concise. That stands true for all the other places, were you yous this sort of pattern

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,305 @@
...
+    assert(req.headerMap.get("b3").isEmpty)
+    // check "x-b3-traceid" and "x-b3-spanid" and "x-b3-sampled" and "x-b3-parentspanid" have been added to request
+    assert(req.headerMap.keys == Set("x-b3-traceid", "x-b3-spanid", "x-b3-sampled", "x-b3-parentspanid"))
+
+    trace match {
+      case Some(tid) => {
+        assert(tid.traceId.toString().equals("a3ce929d0e0e4736"))
+        assert(tid.spanId.toString().equals("00f067aa0ba902b7"))
+        assert(tid.parentId.toString().equals("5b4185666d50f68b"))
+        assert(tid.sampled.toString() == "Some(false)")
```
zaharidichev: Use .contains here. This is an option that should contain a bool that is false so `assert(tid.sampled.contains(false))` should do.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,305 @@
...
+      }
+      case None => {
```
zaharidichev: I think generally, you do not need brackets surrounding the right sides of case statements unless you want to be really explicit and emphasize something

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,305 @@
...
+        assert(false)
+      } // "traceId does not exist"
+    }
+
+    // expect to get the right sampled value which is 0
+    // this also checks "x-b3-sampled" has been added to request
+    var sampler = ZipkinTrace.getSampler(req.headerMap)
```
zaharidichev: Can't this be a `val` ?

adriancole:
one thing to spot check in the implementation is if "b3" is consumed or not. I know some proxies pass headers, which could result in a stale "b3" value sent out on the other side. Probably not an issue here, but worth asking

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: Done, removed assignment.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: Done, removed all assignments.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: Done, removed

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: Done, removed

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: Done, no need for var, used val

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: Done, removed all surrounding brackets, C++ habits :)

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: Done, used sampled.contains(false), much cleaner.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: Done, used assert(sampler.contains(0.0f)), much cleaner, thanks

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: Done, not needed, array is fine.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: No need to change to cons matches, since I've dropped the toList

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: Done, added default match, cases where size of list is not as expected will just be ignored.

dtacalau:
> Nice work @dtacalau! Left a few comments, mainly around code style, etc. What I am wondering however is whether this same logic + tests should be implemented for the [h2 version of the code](https://github.com/linkerd/linkerd/blob/dcf89d6a94b86e3412297b8e45974fdd8c24a67d/linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ZipkinTracePropagator.scala).

Done, added support for b3 in h2 implementation, also I've updated unit test for h2.

adriancole:
PS as timing is interesting sometimes, a user of a gateway project just found what appears to be a propagation bug relating to my earlier comment. Basically verify that this product won't end up passing through either header choice unaffected, and it will save users from lengthy troubleshooting later https://github.com/spring-cloud/spring-cloud-sleuth/issues/1452

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ZipkinTracePropagator.scala
```diff
@@ -59,28 +59,121 @@ object ZipkinTrace {
...
+
+    // support setting a 128bit traceid
+    if (id.traceIdHigh.isEmpty) {
+      headers.set(ZipkinTraceHeader, id.traceId.toString)
+    } else {
+      headers.set(ZipkinTraceHeader, id.traceIdHigh.get.toString + id.traceId.toString)
+    }
+
+    headers.set(ZipkinParentHeader, id.parentId.toString)
+    headers.set(ZipkinSampleHeader, (if ((id.sampled exists { _ == true })) 1 else 0).toString)
```
zaharidichev: can use contains here + there is unnecessary parens

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
```diff
@@ -61,28 +58,121 @@ object ZipkinTrace {
...
+
+    // support setting a 128bit traceid
+    if (id.traceIdHigh.isEmpty) {
+      headers.set(ZipkinTraceHeader, id.traceId.toString)
+    } else {
+      headers.set(ZipkinTraceHeader, id.traceIdHigh.get.toString + id.traceId.toString)
+    }
+
+    headers.set(ZipkinParentHeader, id.parentId.toString)
+    headers.set(ZipkinSampleHeader, (if ((id.sampled exists { _ == true })) 1 else 0).toString)
```
zaharidichev: same here

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,283 @@
...
+    //b3 has been removed
+    assert(req.headerMap.get("b3").isEmpty)
+    // check "x-b3-traceid" and "x-b3-spanid" have been added to request
+    assert(req.headerMap.keys == Set("x-b3-traceid", "x-b3-spanid"))
+
+    trace match {
+      case Some(tid) =>
+        assert(tid.traceId.toString().equals("64fe8b2a57d3eff7"))
+        assert(tid.spanId.toString().equals("05e3ac9a4f6e3b90"))
+        assert(tid.traceIdHigh.toString().equals("Some(80f198ee56343ba8)"))
```
zaharidichev: again, there is no need for Some(), just check with `contains` or any other suitable combinator

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,283 @@
...
+      case None =>
+        assert(false)  // "traceId does not exist"
```
zaharidichev: not a great pattern in my opinion. You can do option.exists(arbitrary predicate). If you want to have seprate asserts for the separate conditions, probably do an `assert(trace.NonEmpty)` + `trace.forEach(... your asserts here....)`

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
```diff
@@ -61,28 +58,121 @@ object ZipkinTrace {
...
+
+      def handleTraceAndSpanIds(headers: HeaderMap, a: String, b: String): Unit = {
+        headers.set(ZipkinTraceHeader, a)
+        headers.set(ZipkinSpanHeader, b)
+        ()
+      }
+
+      caseInsensitiveGet(headers, ZipkinB3SingleHeader).map(_.split("-")) match {
+        case Some(a) =>
+          a.size match {
```
zaharidichev: TIOLI: I know that I said that toList was not needed and it is not if you are simply matching on the size. But dont you think it would be a bit more idiomatic to do cons matches? Then you can have the extracted parts of the list be named a lot more explicitly instead of needing to index into the array?  And generally matching on size when we have all the bells and whistles to do it in more expressive ways is a bit of a code smell.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
```diff
@@ -0,0 +1,283 @@
...
+  test("b3 single headers preferred over x-b3- multi headers") {
+    val ztp = new ZipkinTracePropagator()
+    val req = Request()
+    req.headerMap.add("b3", "a3ce929d0e0e4736-00f067aa0ba902b7-1")
+    req.headerMap.add("x-b3-traceid", "0000000000000001")
+    req.headerMap.add("x-b3-spanid", "0000000000000002")
+    req.headerMap.add("x-b3-sampled", "0")
+
+    val trace = ztp.traceId(req)
+    trace match {
```
zaharidichev: Again, making use of the combinators would probably make sense here.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
```diff
@@ -61,28 +58,121 @@ object ZipkinTrace {
   val ZipkinSampleHeader = "x-b3-sampled"
   val ZipkinFlagsHeader = "x-b3-flags"
 
+  /**
```
adleong: Ideally we could share code between the http and h2 implementations.  There's a `router/base-http` project that is where code shared between `router/http` and `router/h2` lives.  We could do a similar thing here and create a `linkerd/protocol/base-http` project to hold code which can be shared by `linkerd/protocol/http` and `linkerd/protocol/h2`.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ZipkinTracePropagator.scala
dtacalau: I have redone the whole block, see my other PRs.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: I have redone the whole block, see my other PRs.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: Done

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/ZipkinTracePropagatorTest.scala
dtacalau: redone tests, see my other PRs.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: redone it using list, see my other PR.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ZipkinTracePropagator.scala
```diff
@@ -59,28 +59,121 @@ object ZipkinTrace {
...
+   * the portable choice is to always write down "X-B3-"  even if we read "b3". Finagle does a similar thing in
+   * finagle-http-base: read "b3", write back "X-B3-".
+   * @note The "b3" encodings specs, single or multi: https://github.com/openzipkin/b3-propagation#http-encodings
+   * @param headers
+   * @return Option[TraceId]
+   */
   def get(headers: Headers): Option[TraceId] = {
-    val trace = caseInsensitiveGet(headers, ZipkinTraceHeader).flatMap(SpanId.fromString)
+    //if "b3" present convert it to "x-b3-" multio and let the flow continue as if it has received "x-b3-"
+    convertB3SingleHeaderToMultiHeaders(headers)
```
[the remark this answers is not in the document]
dtacalau: Yup, feels the same to me, I've redone the implementation, see my other PR.

linkerd/protocol/h2/src/main/scala/io/buoyant/linkerd/protocol/h2/ZipkinTracePropagator.scala
```diff
@@ -59,28 +59,121 @@ object ZipkinTrace {
...
+    }
+
+    headers.set(ZipkinParentHeader, id.parentId.toString)
+    headers.set(ZipkinSampleHeader, (if ((id.sampled exists { _ == true })) 1 else 0).toString)
+    headers.set(ZipkinFlagsHeader, id.flags.toLong.toString)
+    ()
   }
 
+  /**
+   * This method will also work on requests containing "b3" single header, because after the header is read and
```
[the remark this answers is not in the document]
dtacalau: Fixed in my other PR.

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/ZipkinTracePropagator.scala
dtacalau: Done, moved common code to linkerd/protocol/base-http and redone the whole implementation around it. See my other PR.

dtacalau:
The code from this PR has been rewritten and I've created another PR

Issue #2331, opened 2019-09-25
Streaming is disabled if you do not set streamingEnabled in 1.7.0


taer:
With this config
```
routers:
- protocol: http
  label: cgp
  compressionLevel: 3
  maxHeadersKB: 16
  maxInitialLineKB: 18
  #streamingEnabled: true
  #streamAfterContentLengthKB: 5
```
in linkerd 1.7.0, the effective config is 
```
routers:
- protocol: http
  label: cgp
  compressionLevel: 3
  maxHeadersKB: 16
  maxInitialLineKB: 18
  streamingEnabled: false
  streamAfterContentLengthKB: 5120
```
The case statement that handled the enable steaming didn't honor the default if streamAfterContentLengthKB was not present. Also, the finangle default for streaming is 5M, not 5K as indicated in the docs. So I removed the dependence on finagle's default.

taer:
We ran into this when pushing to production without streamingEnabled or streamAfterContentLengthKB set, assuming the defaults of true and 5K would take effect.

j0sh3rs:
We also experienced this issue in production -- the default behavior appeared to manifest as linkerd responding with 502 codes to our applications.  Reenabling streaming resolved the issue for us, despite it being unclear as to why the response codes manifested as 502s.

edio:
Same here. Explicitly setting supposedly default parameter `streamAfterContentLengthKB` resolves the issue.

adleong:
Fixed by https://github.com/linkerd/linkerd/pull/2332

PR #2332, opened 2019-09-25
Fix case where streaming can become disabled

taer:
Fix for #2331

### DIFF
--- a/config/src/main/scala/io/buoyant/config/ConfigError.scala
+++ b/config/src/main/scala/io/buoyant/config/ConfigError.scala
@@ -17,6 +17,9 @@ case class ConflictingSubtypes(t0: NamedType, t1: NamedType) extends ConfigError
 case class ConflictingLabels(name: String) extends ConfigError {
   def message = s"Multiple routers with the label $name"
 }
+case class ConflictingStreamingOptions(name: String) extends ConfigError {
+  def message = s"Conflicting streaming options set. Can't disable streaming and set streamAfterContentLengthKB in $name"
+}
 
 case class ConflictingPorts(
   addr0: InetSocketAddress,
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -18,6 +18,7 @@ import com.twitter.finagle.stack.nilStack
 import com.twitter.finagle.tracing.TraceInitializerFilter
 import com.twitter.finagle.{ServiceFactory, Stack}
 import com.twitter.logging.Policy
+import io.buoyant.config.ConflictingStreamingOptions
 import io.buoyant.linkerd.protocol.HttpRequestAuthorizerConfig.param
 import io.buoyant.linkerd.protocol.http._
 import io.buoyant.router.{ClassifiedRetries, Http, RoutingFactory}
@@ -231,6 +232,7 @@ case class HttpConfig(
     case (Some(true), None) => hparam.Streaming(5.kilobytes)
     case (None, None) => hparam.Streaming(5.kilobytes)
     case (Some(false), None) => hparam.Streaming(false)
+    case (Some(false), Some(_)) => throw ConflictingStreamingOptions(_label.getOrElse("unknown"))  //tried label property, but it NPE's in test
     case (_, Some(streamAfter)) => hparam.Streaming(streamAfter.kilobytes)
     case _ => hparam.Streaming(false)
   }
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -228,7 +228,9 @@ case class HttpConfig(
   var service: Option[HttpSvc] = None
 
   private val streaming = streamingEnabled -> streamAfterContentLengthKB match {
-    case (Some(true), None) => hparam.Streaming(true)
+    case (Some(true), None) => hparam.Streaming(5.kilobytes)
+    case (None, None) => hparam.Streaming(5.kilobytes)
+    case (Some(false), None) => hparam.Streaming(false)
     case (_, Some(streamAfter)) => hparam.Streaming(streamAfter.kilobytes)
     case _ => hparam.Streaming(false)
   }
--- a/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
+++ b/linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
@@ -232,7 +232,7 @@ case class HttpConfig(
     case (Some(true), None) => hparam.Streaming(5.kilobytes)
     case (None, None) => hparam.Streaming(5.kilobytes)
     case (Some(false), None) => hparam.Streaming(false)
-    case (Some(false), Some(_)) => throw ConflictingStreamingOptions(_label.getOrElse("unknown"))  //tried label property, but it NPE's in test
+    case (Some(false), Some(_)) => throw ConflictingStreamingOptions(label)
     case (_, Some(streamAfter)) => hparam.Streaming(streamAfter.kilobytes)
     case _ => hparam.Streaming(false)
   }
--- a/linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
+++ b/linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
@@ -1,6 +1,8 @@
 package io.buoyant.linkerd.protocol.http
 
+import com.twitter.conversions.StorageUnitOps._
 import com.twitter.finagle.buoyant.Dst
+import com.twitter.finagle.http.param.{FixedLengthStreamedAfter, Streaming}
 import com.twitter.finagle.http.{Method, Request}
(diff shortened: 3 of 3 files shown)


linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
```diff
@@ -46,6 +49,76 @@ class HttpConfigTest extends FunSuite with Awaits {
...
+                  |maxInitialLineKB: 4
+                  |streamAfterContentLengthKB: 5
+                  |streamingEnabled: true
+                  |servers:
+                  |- port: 5000
+      """.stripMargin
+    val config = parse(yaml)
+
+    val streaming = config.routerParams(Stack.Params.empty)[Streaming]
+    assert(streaming.disabled)
```
zaharidichev: I do not think this will be disabled

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
```diff
@@ -46,6 +49,76 @@ class HttpConfigTest extends FunSuite with Awaits {
...
+                  |  kind: io.l5d.methodAndHost
+                  |maxInitialLineKB: 4
+                  |streamingEnabled: false
+                  |servers:
+                  |- port: 5000
+      """.stripMargin
+    val config = parse(yaml)
+
+    val streaming = config.routerParams(Stack.Params.empty)[Streaming]
+    assert(streaming.enabled)
```
zaharidichev: Same here, according to your logic, this will not disable streaming as it will hit the `(_, Some(streamAfter)) ` case

taer:
I can't imagine anyone would want to disable streaming ever. But from what I can tell, disabling streaming forces a total buffer and a Content-length. The streamAfter I think seems to tell linkerd to convert Content-Length into chunked? I honestly don't 100% understand what that does. But if someone were to *want* to Content-Length everything, then you'd have to keep that streamEnable option. And you don't want to prevent someone who relied on that from upgrading. 

They are dependent on each other. 

You could have contentLengthEncodeAll:  true    -- but that's just the inverse.
You can't really convert chunkToContentLength based on a size, so no parameter makes sense there.
I assume streamAfterContentLengthKB means if CL is > X, then change the entire thing to chunked encoding response.

I think you're stuck w/ a boolean and the now optional steamAfter setting. We could fail startup for the illegal cases

Did I get the streamAfterContentLengthKB concept correct?

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
taer: Yup.. I had it right. Then formatted and changed the conditions to clean it up.. And just got lost in the code. :)  

Updated PR

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
taer: stupid brain fart commit. Fixed.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
```diff
@@ -46,6 +49,76 @@ class HttpConfigTest extends FunSuite with Awaits {
...
+                  |identifier:
+                  |  kind: io.l5d.methodAndHost
+                  |maxInitialLineKB: 4
+                  |servers:
+                  |- port: 5000
+      """.stripMargin
+    val config = parse(yaml)
+
+    val streaming = config.routerParams(Stack.Params.empty)[Streaming]
+    assert(streaming.enabled)
```
taer: @zaharidichev Any ideas on how to test the payload here?

zaharidichev:
@taer 
To get some more additonal context on things. I made an attempt at explaining some of the differences around chunked, streaming, etc right [here](https://github.com/linkerd/linkerd/issues/2196#issuecomment-512270795). Hope this is useful. 

> I assume streamAfterContentLengthKB means if CL is > X, then change the entire thing to chunked encoding response.

I beleive you are right here. 

The real question now is whether we should actually check for the illigal case. Can we really have streaming disabled and streamAfterContentLengthKB. And by the looks of it, the answer is no, because if you look at the  finagle [params](https://github.com/twitter/finagle/blob/develop/finagle-base-http/src/main/scala/com/twitter/finagle/http/param/params.scala#L84) you cna quickly see that when streaming is disabled, we are setting it to disabled and having `streamAfterContentLengthKB` does not play any role. So I would say, if you go ahead and add a check to fail the config initialization when this logical conflict arises, that will be great.

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
zaharidichev: Unfortunatelly finagle has that private, so I dont see a way of accessing it. Quite a pity indeed.

taer:
Should we just throw an exception with the message in it? Or does linkerd have a more fanciful way to handle this?

zaharidichev:
@taer 
I think you can resort to using the config package and do some validation there. You can also throw
a `ConfigError`.  You an take a look at how [these](https://github.com/linkerd/linkerd/blob/c03116375eace9b3ec4acaf60e5f671083903314/config/src/main/scala/io/buoyant/config/ConfigError.scala) are used around the codebase.

zaharidichev:
@taer Are you still keen on moving that forward, or I should just pick it up from here?

taer:
@zaharidichev Heya. Sorry about the delay. Had "normal" job things to work on. I'll do the simple throwing of the configException real quick and push that up

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
```diff
@@ -46,6 +50,96 @@ class HttpConfigTest extends FunSuite with Awaits {
...
+                  |servers:
+                  |- port: 5000
+      """.stripMargin
+    try {
+      parse(yaml)
+      fail()
+    } catch {
+      case _: InvalidDefinitionException => //Expected
+      case e: Throwable => fail(e)
+    }
```
zaharidichev: I think you can rewrite this in a more compressed manner: 

```scala
    intercept[InvalidDefinitionException] {
      parse(yaml)
    }
```

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
```diff
@@ -228,7 +229,10 @@ case class HttpConfig(
   var service: Option[HttpSvc] = None
 
   private val streaming = streamingEnabled -> streamAfterContentLengthKB match {
-    case (Some(true), None) => hparam.Streaming(true)
+    case (Some(true), None) => hparam.Streaming(5.kilobytes)
+    case (None, None) => hparam.Streaming(5.kilobytes)
+    case (Some(false), None) => hparam.Streaming(false)
+    case (Some(false), Some(_)) => throw ConflictingStreamingOptions(_label.getOrElse("unknown"))  //tried label property, but it NPE's in test
```
zaharidichev: If you are talking about `HttpConfigTest`, it does not NPE for me if use the `def label`

linkerd/protocol/http/src/test/scala/io/buoyant/linkerd/protocol/http/HttpConfigTest.scala
```diff
@@ -1,10 +1,13 @@
...
 import com.twitter.finagle.buoyant.Dst
+import com.twitter.finagle.http.param.{FixedLengthStreamedAfter, Streaming}
 import com.twitter.finagle.http.{Method, Request}
 import com.twitter.finagle.{Dtab, Path, Stack}
 import com.twitter.util.Future
-import io.buoyant.config.Parser
+import io.buoyant.config.{ConflictingStreamingOptions, Parser}
 import io.buoyant.linkerd.{IdentifierInitializer, RouterConfig}
 import io.buoyant.linkerd.protocol.{HttpConfig, HttpIdentifierConfig, HttpInitializer}
 import io.buoyant.router.Http
```
zaharidichev: The `com.twitter.conversions.StorageUnitOps._`, `FixedLengthStreamedAfter` and `ConflictingStreamingOptions` seem to be unused, can we just remove them ?

linkerd/protocol/http/src/main/scala/io/buoyant/linkerd/protocol/HttpConfig.scala
taer: Strange. It did once. Now it's happy.

config/src/main/scala/io/buoyant/config/ConfigError.scala
```diff
@@ -17,6 +17,9 @@ case class ConflictingSubtypes(t0: NamedType, t1: NamedType) extends ConfigError
 case class ConflictingLabels(name: String) extends ConfigError {
   def message = s"Multiple routers with the label $name"
 }
+case class ConflictingStreamingOptions(name: String) extends ConfigError {
+  def message = s"Conflicting streaming options set. Can't disable streaming and set streamAfterContentLengthKB in $name"
```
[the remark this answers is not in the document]
taer: I just cloned `ConflictingLabels` error.

Issue #2333, opened 2019-09-26
enabling zipkin b3 propagation leads to possible incorrect spans


dtacalau:
This could be a bug, not sure, was found while implementing and testing #2114.

**What happened**:
Problem is linkerd generates a new spanid and a new parentspanid for each router which has configured property "tracePropagator: kind: io.l5d.zipkin", but the parentspanid is different from the previous spanid.

**What you expected to happen**:
 I'd expect parentspanid of a new span to be the same as spanid of the previous span. We need to make sure the tracing spans are generated correctly w/ and w/o propagation enabled.

**How to reproduce it (as minimally and precisely as possible)**:
Here is my setup:
1. Start linkerd using a config similar with the one from https://linkerd.io/2016/05/17/distributed-tracing-for-polyglot-microservices . Linkerd config defines three routers listening on three different ports. The router on port 9000 forwards to the router on port 9001, which forwards to the router on port 9002, which forwards to the Linkerd admin service on port 9990.
2. Start zipkin service with SCRIBE transport enabled: export SCRIBE_ENABLED=true; java -jar zipkin.jar
3. Start wireshark capture on all interfaces
3. Use a http client from finale to send a request to linkerd 9000 port: git/finagle/doc/src/sphinx/code/quickstart$ ./sbt 'runMain Client', which by default will contain a traceid and a spanid using "x-b3-" headers

Scenario 1 (router on port 9000 has tracePropagator io.l5d.zipkin enabled, the other routers do not have tracePropagator) (linkerd config: 1-linkerd-one-router-with-propagation.yaml, capture: 1-capture-one-router-with-propagation.pcapng)
Packet 44
    - Client send a HTTP request to l5d 9000 port, containing (X-B3-SpanId: 74d95541b82adc24; X-B3-TraceId: 74d95541b82adc24)
    - x-b3- headers present, first span, spanid=traceid
    
Packet 49
    - l5d forwards the request to l5d 9001 port, (x-b3-spanid: 9f387655f61ca970; x-b3-parentspanid: 135fe09bbbd2206e; x-b3-traceid: 74d95541b82adc24)
    - new spanid different from packet 44
    - parentspanid != spanid of packet 44????
    - keeps the traceid    
Packet 54 - l5d forwars the request to l5d 9002 port, (x-b3-spanid: 9f387655f61ca970; x-b3-parentspanid: 135fe09bbbd2206e; x-b3-traceid: 74d95541b82adc24)
[...]

PR #2340, opened 2019-10-07
Moved http and h2 ZipkinTracePropagator to common http code

dtacalau:
Signed-off-by: dst4096 <dst4096@gmail.com>

### DIFF
--- a/linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
+++ b/linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
@@ -70,18 +70,11 @@ object ZipkinTrace {
   def clear[H: HeadersLike](headers: H): Unit = {
     val headersLike = implicitly[HeadersLike[H]]
 
-    headersLike.remove(headers, ZipkinSpanHeader)
-    headersLike.remove(headers, ZipkinTraceHeader)
-    headersLike.remove(headers, ZipkinParentHeader)
-    headersLike.remove(headers, ZipkinSampleHeader)
-    headersLike.remove(headers, ZipkinFlagsHeader)
-
-    headersLike.remove(headers, ZipkinSpanHeader.toUpperCase())
-    headersLike.remove(headers, ZipkinTraceHeader.toUpperCase())
-    headersLike.remove(headers, ZipkinParentHeader.toUpperCase())
-    headersLike.remove(headers, ZipkinSampleHeader.toUpperCase())
-    headersLike.remove(headers, ZipkinFlagsHeader.toUpperCase())
-
+    caseInsensitiveGetKey(headers, ZipkinSpanHeader).map(headersLike.remove(headers, _))
+    caseInsensitiveGetKey(headers, ZipkinTraceHeader).map(headersLike.remove(headers, _))
+    caseInsensitiveGetKey(headers, ZipkinParentHeader).map(headersLike.remove(headers, _))
+    caseInsensitiveGetKey(headers, ZipkinSampleHeader).map(headersLike.remove(headers, _))
+    caseInsensitiveGetKey(headers, ZipkinFlagsHeader).map(headersLike.remove(headers, _))
     ()
   }
 
@@ -94,6 +87,11 @@ object ZipkinTrace {
     headersLike.set(headers, ZipkinSampleHeader, (if ((id.sampled exists { _ == true })) 1 else 0).toString)
     headersLike.set(headers, ZipkinFlagsHeader, id.flags.toLong.toString)
     ()
+  }
+
+  private def caseInsensitiveGetKey[H: HeadersLike](headers: H, key: String): Option[String] = {
+    val headersLike = implicitly[HeadersLike[H]]
+    headersLike.toSeq(headers).iterator.collectFirst { case (k, v) if key.equalsIgnoreCase(k) => k }
   }
 
   private def caseInsensitiveGet[H: HeadersLike](headers: H, key: String): Option[String] = {
--- a/linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
+++ b/linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
@@ -1,0 +1,103 @@
+package io.buoyant.linkerd.protocol.http
+
+import com.twitter.finagle.{SimpleFilter, Stack}
+import com.twitter.finagle.buoyant.Sampler
+import com.twitter.finagle.http.util.StringUtil
+import com.twitter.finagle.tracing.{Flags, SpanId, Trace, TraceId, Tracer}
+import com.twitter.util.Try
+import io.buoyant.linkerd.{TracePropagator, TracePropagatorInitializer}
+import io.buoyant.router.http.{HeadersLike, RequestLike}
+
+class ZipkinTracePropagator[Req, H: HeadersLike](implicit requestLike: RequestLike[Req, H]) extends TracePropagator[Req] {
+  /**
+   * Read the trace id from the request, if it has one.
+   */
+  override def traceId(req: Req): Option[TraceId] = {
+    var traceId = ZipkinTrace.get(requestLike.headers(req))
+    ZipkinTrace.clear(requestLike.headers(req))
+    traceId
+  }
+
+  /**
+   * Return a sampler which decides if the given request should be sampled, based on properties
+   * of the request (zipkin or linkerd if zipkin not present).  If None is returned, the decision of whether to sample the request is deferred
+   * to the tracer.
+   */
+  override def sampler(req: Req): Option[Sampler] = {
+    if (Trace.hasId) {
+      Trace.id.sampled.map {
+        case true => Sampler(1.0f)
+        case false => Sampler(0.0f)
+      }
+    } else {
+      None
+    }
+  }
+
+  /**
+   * Write the trace id onto a request.
+   */
+  override def setContext(
+    req: Req,
+    traceId: TraceId
+  ): Unit = {
+    ZipkinTrace.set(requestLike.headers(req), traceId)
+  }
+}
+
+object ZipkinTrace {
+
+  val ZipkinSpanHeader = "x-b3-spanid"
+  val ZipkinParentHeader = "x-b3-parentspanid"
+  val ZipkinTraceHeader = "x-b3-traceid"
+  val ZipkinSampleHeader = "x-b3-sampled"
+  val ZipkinFlagsHeader = "x-b3-flags"
+
+  def get[H: HeadersLike](headers: H): Option[TraceId] = {
+    val trace = caseInsensitiveGet(headers, ZipkinTraceHeader).flatMap(SpanId.fromString)
+    val parent = caseInsensitiveGet(headers, ZipkinParentHeader).flatMap(SpanId.fromString)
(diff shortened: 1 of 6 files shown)


linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
```diff
@@ -0,0 +1,94 @@
...
+import io.buoyant.linkerd.{TracePropagator, TracePropagatorInitializer}
+import io.buoyant.router.http.{HeadersLike, RequestLike}
+
+class ZipkinTracePropagator[Req, H: HeadersLike](implicit requestLike: RequestLike[Req, H]) extends TracePropagator[Req] {
+  /**
+   * Read the trace id from the request, if it has one.
+   */
+  override def traceId(req: Req): Option[TraceId] = {
+    var traceId = ZipkinTrace.get(requestLike.headers(req))
+
```
adleong: Shouldn't we be clearing the trace headers here also?  As in `LinkerdHeaders.Ctx.Trace.clear(req.headers)`?

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  /**
+   * Return a sampler which decides if the given request should be sampled, based on properties
+   * of the request (zipkin or linkerd if zipkin not present).  If None is returned, the decision of whether to sample the request is deferred
+   * to the tracer.
+   */
+  override def sampler(req: Req): Option[Sampler] = {
+    var sampler = ZipkinTrace.getSampler(requestLike.headers(req)).map(Sampler(_))
+
```
adleong: Similarly, shouldn't we be clearing the sample headers?

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
```diff
@@ -0,0 +1,94 @@
...
+object ZipkinTrace {
+
+  val ZipkinSpanHeader = "x-b3-spanid"
+  val ZipkinParentHeader = "x-b3-parentspanid"
+  val ZipkinTraceHeader = "x-b3-traceid"
+  val ZipkinSampleHeader = "x-b3-sampled"
+  val ZipkinFlagsHeader = "x-b3-flags"
+
+  def get[H: HeadersLike](headers: H): Option[TraceId] = {
+    val headersLike = implicitly[HeadersLike[H]]
```
adleong: is this unused?

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
```diff
@@ -0,0 +1,94 @@
...
+        case v if v < 0 => 0.0f
+        case v if v > 1 => 1.0f
+        case v => v
+      }
+    }
+  }
+
+  private def caseInsensitiveGet[H: HeadersLike](headers: H, key: String): Option[String] = {
+    val headersLike = implicitly[HeadersLike[H]]
+    headersLike.iterator(headers).collectFirst { case (k, v) if key.equalsIgnoreCase(k) => v }
```
adleong: I think you can use `headersLike.toSeq(headers).iterator.collectFirst` without needing to add new methods to `HeadersLike`

router/base-http/src/main/scala/io/buoyant/router/http/HeadersLike.scala
```diff
@@ -12,4 +12,5 @@ trait HeadersLike[H] {
   def add(headers: H, k: String, v: String): Unit
   def set(headers: H, k: String, v: String): Unit
   def remove(headers: H, key: String): Seq[String]
+  def iterator(headers: H): Iterator[(String, String)]
```
adleong: I don't think we need to add this.

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
dtacalau: I removed that line of code, it didn't make sense to me to clear l5d-ctx header in a Zipkin propagator plugin. Am I missing something?

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
dtacalau: Not sure, same question as above: why would you clear l5d-sample for a Zipkin propagator plugin which deals with x-b3- headers?

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
dtacalau: yes, will remove it.

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
dtacalau: you're right, headersLike.toSeq(headers).iterator.collectFirst works fine.

router/base-http/src/main/scala/io/buoyant/router/http/HeadersLike.scala
dtacalau: that's right, will remove it.

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
adleong: The way that trace propagation works in Linkerd is that the trace context is read from the incoming request (in a server filter), a child span is created, and the whole context is stored in a thread local.  When the request is ready to be sent out (in a client filter), the trace context is read from the thread local and written into the request headers.

This means that when we read the trace context from a request, we should clear those headers.

dtacalau: Here's my idea of fixing this: clear the old tracing context just before setting the new tracing context. At least for Zipkin propagation, we cannot clear tracing context during get traceid because we need it to be available for get sampler call (which will use x-b3-sampled and x-b3-flags headers). Also, we cannot clear tracing context during get sampler call because we need it to be available for the get traceid call (same headers). Get traceid and get sampler do not depend on each other, so the context must be preserved for both calls. Note that this is only specific for Zipkin propagation. I've pushed the latest changes implementing this idea. Please let me know if it's OK!

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
dtacalau: Old trace context is now fully cleared before setting new trace context.

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
adleong: After looking at this code some more and reading https://github.com/openzipkin/b3-propagation again, I think this can actually be simpler.  The `X-B3-Sampled` only supports values of `0` and `1` which means that the value is fully recoverable from the `traceId` in the `traceId.sampled` field.

Therefore my suggestion is to have `get` read the traceId and clear all the headers.  Then, the `sampler` method can simply ignore the `req` entirely and use `Trace.id.sampled` to create a sampler.  

This relies on the fact that `traceId` will be called before `sampler` and will set the traceId into the thread local, but looking at `TraceInitializer`, I think this is a fair assumption.  (sorry if that contradicts some of my earlier suggestions)  We should also add some documentation to the TracePropagator trait to make these assumptions explicit.

dtacalau: Makes sense, here's how I'm thinking to implement your suggestion:
- have a default implementation of 
object TraceInitializer {
  abstract class ServerFilter {
    def sampler(req: Req): Option[Sampler] = {
      - if Trace.id set return Sampler(Trace.id.sampled) else None
- ZipkinTracePropagator will not overwrite the default sampler implementation above.
- LinkerdTracePropagator will have to overwrite it, no actual change in its source code.

Should be an easy change but I want to do testing before posting it. Let me know if we're on the same page with this. Thanks.

adleong: I don't think we need to make that the default implementation of `sampler`.  Safer to just explicitly make it the implementation for ZipkinTracePropagator.

dtacalau:
Latest changes after Alex's review:  
- X-b3 multi headers are now removed from the request after the traceId call.
- Sampler is now computed using the traceid from thread local context.
- headers are removed using both lower and upper case.

linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
```diff
@@ -0,0 +1,103 @@
...
+  def clear[H: HeadersLike](headers: H): Unit = {
+    val headersLike = implicitly[HeadersLike[H]]
+
+    headersLike.remove(headers, ZipkinSpanHeader)
+    headersLike.remove(headers, ZipkinTraceHeader)
+    headersLike.remove(headers, ZipkinParentHeader)
+    headersLike.remove(headers, ZipkinSampleHeader)
+    headersLike.remove(headers, ZipkinFlagsHeader)
+
+    headersLike.remove(headers, ZipkinSpanHeader.toUpperCase())
```
adleong: This isn't a true case insensitive remove since we only remove the header if it matches the case exactly or if it is all upper case.  To be truly case insensitive, we'd need to do something like in `caseInsensitiveGet` where we iterate over the header keys to find the matching ones, and then remove those from the header map.

dtacalau: yeah got it, fixed it to support mixed case too.

dtacalau:
> Thanks for making those updates, @dtacalau!
> 
> How were able to test this? I looked at our [hello world](https://github.com/linkerd/linkerd-examples/tree/master/docker/helloworld) application that we often use for testing this kind of thing and is used by our [distributed tracing blogpost](https://linkerd.io/2017/03/14/a-service-mesh-for-kubernetes-part-vii-distributed-tracing-made-easy/) but it looks like it only propagates the `l5d-ctx` headers and therefore wouldn't work with zipkin propagation.
> 
> It'd be great if you could put together steps for manually testing this that I could run through.
Just in case someone else needs this, steps for manual testing Zipkin w/ and w/o propagation can be found here (steps to reproduce the issue):  https://github.com/linkerd/linkerd/issues/2333. Zipkin x-b3 multi headers can be analyzed using wireshark to capture http traffic, Zipkin traces can be analyzed using local Zipkin service running on http://localhost:9411.

adleong:

PR #2342, opened 2019-10-07
fixes for x b3 sampled and flags get set

dtacalau:


adleong:

PR #2343, opened 2019-10-07
Fix for #2114 Support b3 trace header 

dtacalau:
Linkerd Zipkin propagator will now read b3 single header the same way it reads x-b3- multi headers. If a request contains the "b3" single header and multiple "X-B3-" headers, the trace context from "b3" is preferred and "x-b3-" headers will be ignored. 

B3 is also preferred and read for the sampling decision, if not present, sampling decision falls back to X-B3. 

On the way out "X-B3-" multi headers are still preferred over "B3". There's no way to know what the downstream is capable of, and it is more likely it supports "X-B3-*" vs not, so, the portable choice is to always write down "X-B3-"  even if we read "b3".

No config changes were needed for this solution.

### DIFF
--- a/linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
+++ b/linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
@@ -1,0 +1,94 @@
+package io.buoyant.linkerd.protocol.http
+
+import com.twitter.finagle.{SimpleFilter, Stack}
+import com.twitter.finagle.buoyant.Sampler
+import com.twitter.finagle.http.util.StringUtil
+import com.twitter.finagle.tracing.{Flags, SpanId, TraceId}
+import com.twitter.util.Try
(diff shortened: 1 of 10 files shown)


linkerd/protocol/base-http/src/main/scala/io/buoyant/linkerd/http/ZipkinTracePropagator.scala
```diff
@@ -0,0 +1,248 @@
...
+          case Flags(1) => samplerTrue // flags debug
+          case _ => { // flags invalid or not present, look for sampled field
+            sampled match {
+              case Some(true) => samplerTrue
+              case Some(false) => samplerFalse
+              case _ => samplerNone
+            }
+          }
+        }
+      }
```
zaharidichev: If you want to flatten/simplify your match as we spoke, I think you can have some kind an ADT that encodes all the variants that can be returned from `getFromB3SingleHeader()` . Then you can have something like 

getFromB3SingleHeader(v) match {
   case Variant1(x,y) => ....
   case Variant2(z) => ......
   etc, etc.... 
}

adleong:

PR #2353, opened 2019-10-30
[#2352] Add configuration server.serverSession to allow expiring server conne…

### LINKED ISSUE #2352: Impossible to time out server connections
Issue Type:

- [ ] Bug report
- [X] Feature request

**What happened**:
Clients do not correctly close connections, or network components silently drop connections without closing or resetting them.
After several days this can result in Linkerd running out of file handles.

It should be possible to expire server connections.

Start Linkerd with a very low limit of file handles.
Then open as many connections as possible to Linkerd, can be a plain socket, until Linkerd fails to accept the connections due to lack of file handles.

**Environment**:
- Platform, version, and config files (Kubernetes, DC/OS, etc): No Kubernetes or such, just plain VMs

robertpanzer:
…ctions

This PR fixes #2352.

There might be clients opening connections to Linkerd not closing them properly.
Also sometimes network components might silently drop connections to Linkerd.
This results in connections that are never closed and after a longer time this might accumulate to so many open connections until the Linkerd process runs out of file handles.
Right now I could not find a configuration parameter to control expiring of inbound connections like it is supported for client connections.

Therefore this PR proposes to add a configuration parameter `serverSession` to a server that behaves similar to the `clientSession`.
With the following configuration server connections will be closed after 2 seconds of idling or after a full life time of 30 seconds:
```yaml
servers:
- port: 4140
  serverSession:
    idleTimeMs: 2000
    lifeTimeMs: 30000
```

### DIFF
--- a/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
+++ b/linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
@@ -842,6 +842,107 @@ class HttpEndToEndTest
     }
   }
 
+  test("serverSession idleTimeMs should close server connections") {
+    val config =
+      s"""|routers:
+          |- protocol: http
+          |  experimental: true
+          |  dtab: |
+          |    /p/fox => /$$/inet/127.1/{fox.port} ;
+          |    /svc/century => /p/fox ;
+          |  servers:
+          |  - port: 0
+          |    serverSession:
+          |      idleTimeMs: 1500
+          |      lifeTimeMs: 3000
+          |""".stripMargin
+
+    serverIdleTimeMsBaseTest(config){ (router:Router.Initialized, stats:InMemoryStatsReceiver) =>
+      // Assert
+      def connectsCount = () => stats.counters(Seq("rt", "http", "server", "127.0.0.1/0", "connects"))
+      def activeConnectionsCount = stats.gauges(Seq("rt", "http", "server", "127.0.0.1/0", "connections"))
+      def idleConnectionsCount = () => stats.counters(Seq("rt", "http", "server", "127.0.0.1/0", "idle"))
+      def lifetimeConnectionsCount = () => stats.counters(Seq("rt", "http", "server", "127.0.0.1/0", "lifetime"))
+
+      // An incoming request through the Http.Router will establish an active connection; We expect to see it here
+      assert(connectsCount() == 1.0)
+      assert(activeConnectionsCount() == 1.0)
+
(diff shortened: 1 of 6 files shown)


zaharidichev:
@robertpanzer 

The reasoning behind this is solid. However, I do wonder whether this could be achieved with setting socket options on the server. You can read a bit more about it [right here](https://api.linkerd.io/1.7.0/linkerd/index.html#socket-options). Let me know if that helps?

robertpanzer:
The only option that I see that could help in such a scenario would be enabling SO_KEEPALIVE.
But as far as I can tell configuring this properly also requires changing the system settings as explained here: https://www.tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/
And this would be where it gets tricky, in large heterogenous networks it's not really feasible to configure this on that level.

linkerd/protocol/http/src/e2e/scala/io/buoyant/linkerd/protocol/HttpEndToEndTest.scala
```diff
@@ -842,6 +842,107 @@ class HttpEndToEndTest
...
+          |  servers:
+          |  - port: 0
+          |    serverSession:
+          |      idleTimeMs: 1500
+          |      lifeTimeMs: 3000
+          |""".stripMargin
+
+    serverIdleTimeMsBaseTest(config){ (router:Router.Initialized, stats:InMemoryStatsReceiver) =>
+      // Assert
+      def connectsCount = () => stats.counters(Seq("rt", "http", "server", "127.0.0.1/0", "connects"))
```
adleong: I don't think this needs to be a closure, I think it can simply be a def, similarly to how metrics are checked elsewhere in this file:

```
def connectsCount = stats.counters(Seq("rt", "http", "server", "127.0.0.1/0", "connects"))
```

robertpanzer: Thanks @adleong!
I updated the PR.
CircleCI seems to have failed on downloading a zookeeper dependency, could you please trigger a rebuild?

robertpanzer:
Thanks!
Oh yes, just realized I forgot to sign the second commit.
I just updated that commit with the sign off message and force-pushed.

adleong:
Sorry that it's taking us so long to get this merged, @robertpanzer!  The CI failure here is also happening on master.  I am looking into it.

robertpanzer:
Sure, thanks for letting me know!

adleong:
The build issue on master has been resolved and this is good to go!  Thanks for your patience, @robertpanzer!

robertpanzer:
\o/ Awesome! Thanks for merging!

PR #2361, opened 2019-12-03
Remove ZK candidate dep

### LINKED ISSUE #2360: Remove dependency on zookeeper-candidate
The artifact for `com.twitter.common.zookeeper" % "candidate" % "0.0.84"` on maven.twttr.com has been removed and the Linkerd build is broken as a result.  Our zk-leader-namer depends on this library to determine the leader of a zookeeper group.

We can remove this dependency and restore our build by copying the files and code we need from this package directly into the Linkerd code base.  The relevant files we depend on are https://github.com/twitter/commons/blob/master/src/java/com/twitter/common/zookeeper/Candidate.java and related files.  

Note that Finagle has done a similar thing to remove their dependency on zookeeper-candidate.  Some (but not all) of the files that we need from zookeeper-candidate have been copied to https://github.com/twitter/finagle/tree/develop/finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper

zaharidichev:
The artifact for com.twitter.common.zookeeper" % "candidate" % "0.0.84" on maven.twttr.com is not present anymore. Therefore we need to cpy the bits that we use from it and have it in our sourcecode.

Fixes #2360


namer/zk-leader/src/main/java/com/twitter/common/zookeeper/Group.java
```diff
@@ -0,0 +1,728 @@
+// =================================================================================================
```
adleong: Can you just use https://github.com/twitter/finagle/blob/develop/finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper/Group.java instead?

namer/zk-leader/src/main/java/com/twitter/common/zookeeper/ZooKeeperClient.java
```diff
@@ -0,0 +1,518 @@
+// =================================================================================================
```
adleong: can you use https://github.com/twitter/finagle/blob/develop/finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper/ZooKeeperClient.java?

namer/zk-leader/src/main/java/com/twitter/common/zookeeper/ZooKeeperUtils.java
```diff
@@ -0,0 +1,168 @@
+// =================================================================================================
```
adleong: can you use https://github.com/twitter/finagle/blob/develop/finagle-serversets/src/main/java/com/twitter/finagle/common/zookeeper/ZooKeeperUtils.java?

project/Deps.scala
```diff
@@ -13,6 +13,11 @@ object Deps {
 
   def twitterUtil(mod: String) =
     "com.twitter" %% s"util-$mod" % "19.5.1"
+
+  def twitterCommonUtil = "com.twitter.common" % "util" % "0.0.121"
```
adleong: I think the goal here is to stop depending on com.twitter.common altogether.  I think this means we also need to copy things like `com.twitter.common.base.Command` into our repo if they haven't already been copied into finagle-serversets

project/Deps.scala
```diff
@@ -22,9 +27,12 @@ object Deps {
 
   val boringssl = "io.netty" % "netty-tcnative-boringssl-static" % "2.0.19.Final"
 
-  def zkCandidate =
-    ("com.twitter.common.zookeeper" % "candidate" % "0.0.84")
-      .exclude("com.twitter.common", "util")
+  def apacheZookeeper = ("org.apache.zookeeper" % "zookeeper" % "3.5.0-alpha")
```
adleong: I think if we depend on finalge-serversets, we should automatically get this as a transitive dependency.

Issue #2363, opened 2019-12-17
Consul Namer doesn't update Mesh Interpreter when the Consul payload is significantly large


cpretzer:
### Issue Type:

- [x] Bug report
- [ ] Feature request

### **What happened**:
Version 1.7.0 of Linkerd and Namerd lose connectivity on an update when the Consul namer process a large number of replicas and sends a request with a large body.

In this scenario, Namerd is configured to use the Consul namer and Linkerd is configured to use the io.l5d.mesh interface.

When the Consul store changes, Namerd will trigger an update to Linkerd through the mesh interpreter. Namerd will send a few data payloads, which Linkerd receives, and then the traffic just stops altogether.

### **What you expected to happen**:
The update from Namerd should complete successfully and Linkerd should have an exact copy of the list of replicas that are stored in Namerd and Consul

### **How to reproduce it (as minimally and precisely as possible)**:
Reproducing this is a multi-step process which requires setting up Consul, Linkerd, and Namerd.

**Requires Java 8 or 9**

0. **Create a working directory**
   1. `mkdir consul-namer-repro`, for example
1. **Download Consul**
   1. Get the version for your OS: https://www.consul.io/downloads.html
   2. Make sure that you can run it from the working directory
2. **Download Linkerd and Namerd** 
   1. https://github.com/linkerd/linkerd/releases/download/1.7.0/namerd-1.7.0-exec
   2. https://github.com/linkerd/linkerd/releases/download/1.7.0/linkerd-1.7.0-exec
   3. Copy the files to the working directory
   4. Make the files executable: `chmod 755 *1.7.0-exec`
4. **Download configuration files**
   1. https://gist.github.com/cpretzer/36682f0485abb106c8ca3ae028f1580c
   2. click "Download Zip" and the tgz file will be compressed in another zip file
5. **Start the services**
Open a terminal window for each of the commands below (_You can background the processes, but it'll be more difficult to read the logs_)
```./start_consul.sh```: The consul logs should report that five services have been registered
```./start_namerd.sh```
```./start_linkerd.sh```
```docker run --name --rm nginx_one -p 8080:8080 nginx```
```docker run --name --rm nginx_two -p 8081:8080 nginx```
6. **Check the current replica counts**
This shows the number of replicas currently stored by *Namerd* and *Linkerd*, and they should all read `0`
```./admin-counts.sh```
7. Curl the Linkerd service to get the replicas
```curl http://localhost:4140/cat```
8. **Check the updated replica counts**
[...]

PR #2366, opened 2019-12-19
Add initialStreamWindowSizeKB

cpretzer:
This PR addresses #2363 and corrects #2364 by removing changes to `Build.scala` and `Changes.md`


interpreter/mesh/src/main/scala/io/buoyant/interpreter/MeshInterpreterInitializer.scala
```diff
@@ -32,6 +35,7 @@ object MeshInterpreterConfig {
   val DefaultRoot = Path.Utf8("default")
 
   val defaultRetry = Retry(1, 10.minutes.inSeconds)
+  val defaultWindowSizeKB = 1024
```
zaharidichev: I think bupingthat to 10 Mb is not unresonable and will give us greater guarantee that we will not see this problem show up oftenly

adleong: agreed.  10 Mb default gives us more wiggle room

PR #2387, opened 2020-04-21
Update to use scheduled workflows for nightly builds

grampelberg:


grampelberg:
@adleong @siggy @cpretzer I have no idea how to test this, but it should (tm) work.

.circleci/config.yml
```diff
@@ -137,3 +137,33 @@ jobs:
...
+        context: nightly
+
+  twitter_develop:
+    triggers:
+    - schedule:
+        cron: "0 0 * * *"
+        filters:
+          branches:
+            only:
+            - master
```
siggy, line 166: Recommend setting `twitter-develop` to `10 12 * * *` (5:10am PT) and `nightly` to `30 12 * * *` (5:30am PT). The intent is that when `twitter-develop` is broken, CI is only red for about 20 minutes in the early morning before `master` makes it go green.

grampelberg: TIL! Thanks @siggy, got it updated now.

grampelberg:
@adleong mind taking a peek again? I really want to get rid of that nightly cron trigger =)

PR #2389, opened 2020-05-05
Bump the node dependencies

grampelberg:


grampelberg:
@adleong yeah, it gets lint to pass. No idea how it was passing before (though may be related to the eslint version bump).

adleong:
circleci seems stuck and/or confused on this one.  @grampelberg maybe try pushing an empty commit to see if that unsticks it?

grampelberg:
@adleong I have no idea what is up with Circle status, but can you rerun the job? Pretty sure that error is transient.

Issue #2390, opened 2020-05-15
Serverset namer can't resolve shard fragment


JustinVenus:
Issue Type:

- [ X] Bug report
- [ X] Feature request

**What happened**:
I was not able to resolve a name in the form of `/service/prod/read:http#0`

**What you expected to happen**:
I expected to get shard member `0`  for service label `http` of the `Path(/service/prod/read)`.

**How to reproduce it (as minimally and precisely as possible)**:
1. configure linkerd to use serversets.
2. try to resolve a dtab in the `dtab playground on the admin interface` with a shard like `#0`

**Anything else we need to know?**:
It seems like the functionality could be added [here](https://github.com/linkerd/linkerd/blob/f65e6335fa40589810eabef41e5f345e95331401/namer/serversets/src/main/scala/io/buoyant/namer/serversets/ServersetNamer.scala#L50) and would allow the namer to behave inline with upstream [finagle](https://github.com/twitter/finagle/blob/b8523a4541953d6bbc6a64b9409d75ff162afcd9/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/naming/ServersetNamer.scala#L37).

**Environment**:
- linkerd/namerd version, config files:  ANY configured for serversets
- Platform, version, and config files (Kubernetes, DC/OS, etc): Mesos/Aurora
- Cloud provider or hardware configuration: AWS

JustinVenus:
If this functionality is useful I'd be happy to code it up and submit a PR.

cpretzer:
@JustinVenus a contribution would be awesome!

I'm not familiar with shards, is this for zookeeper?

JustinVenus:
@cpretzer serversets have a json payload on the znode in zookeeper.  The format is roughly like the following.

```json
{"status": "ALIVE", "additionalEndpoints": {"http": {"host": "host1", "port": 31238}}, "serviceEndpoint": {"host": "host1", "port": 31238}, "shard": 0}
```
Most serverset implementations (that I am familiar with) are able to advertise the shard (aka instanceId).  The default/upstream finagle implementation of the ServersetNamer allows one to craft a dtab to specify a given shard/instanceId.

My use case is to prefer a primary instanceId with the ability to failover to a standby like the following example.

```
/srv/foo => /zk/prod/read:http#0 | /zk/prod/read:http#1 | !   
```

cpretzer:
Got it! Thanks for teaching me something new 😀

cpretzer:
[...]

PR #2391, opened 2020-05-18
Update ServersetNamer to support dtabs with shard fragment

JustinVenus:
Implements the feature from https://github.com/linkerd/linkerd/issues/2390 that I requested to allow shards to be considered in DTABs.  This is exposing the same functionality that Finagle already supports.

This allows the following DTAB snipped to be considered valid.

```
/srv/foo => /zk/prod/read:http#0 | /zk/prod/read:http#1 | !
```

### DIFF
--- a/namer/serversets/src/main/scala/com/twitter/finagle/serversets2/BouyantZkResolver.scala
+++ b/namer/serversets/src/main/scala/com/twitter/finagle/serversets2/BouyantZkResolver.scala
@@ -1,0 +1,33 @@
+package com.twitter.finagle.serverset2
+
+import com.twitter.finagle.{Addr, Path, Resolver}
+import com.twitter.finagle.serverset2.naming.ServersetPath
+import com.twitter.util.Var
+
+trait BouyantZkResolver {
+  protected[this] def addrOf(
+    zkHosts: String,
+    zkPath: String,
+    endpoint: Option[String],
+    shardId: Option[Int]
+  ): Var[Addr]
+
+  def resolve(path: Path): Var[Addr] = ServersetPath.of(path) match {
+    case Some(ServersetPath(zkHosts, zkPath, endpoint, shardId)) =>
+      addrOf(zkHosts, zkPath.show, endpoint, shardId)
+    case _ =>
+      Var.value(Addr.Neg)
+  }
+}
+
+class BouyantZkResolverImpl(zk2: Zk2Resolver) extends BouyantZkResolver {
+  def this() = this(Resolver.get(classOf[Zk2Resolver]).get)
+
+  protected[this] def addrOf(
+    zkHosts: String,
+    zkPath: String,
+    endpoint: Option[String],
+    shardId: Option[Int]
+  ): Var[Addr] =
+    zk2.addrOf(zkHosts, zkPath, endpoint, shardId)
+}
--- a/namer/serversets/src/main/scala/io/buoyant/namer/serversets/ServersetNamer.scala
+++ b/namer/serversets/src/main/scala/io/buoyant/namer/serversets/ServersetNamer.scala
@@ -3,6 +3,7 @@ package io.buoyant.namer.serversets
 
 import com.twitter.finagle._
 import com.twitter.util.{Activity, Var}
+import com.twitter.finagle.serverset2.{BouyantZkResolver, BouyantZkResolverImpl}
 
 /**
  * The serverset namer takes Paths of the form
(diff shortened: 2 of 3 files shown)


project/Deps.scala
```diff
@@ -46,6 +47,8 @@ object Deps {
     "org.joda"  % "joda-convert" % "1.7"
   )
 
+  val mockito = "org.mockito" % "mockito-all" % "1.9.5"
```
cpretzer, line 50: Can the test be run without introducing this dependency?

It's probably not a big deal, just curious.

JustinVenus: It seemed a little less ugly (my opinion) to do it with a mockito.  If the reviewers/owners want the ServersetNamer refactored to just use `zk2.addrOf` I think it'll make a lot of sense to use mockito.

cpretzer: Sounds good to me. @adleong any thoughts here?

namer/serversets/src/main/scala/com/twitter/finagle/serversets2/BouyantZkResolver.scala
```diff
@@ -0,0 +1,15 @@
+package com.twitter.finagle.serverset2
+
+import com.twitter.finagle.{Addr, Resolver}
+import com.twitter.util.Var
+
+class BouyantZkResolver(zk2: Zk2Resolver) {
```
adleong, line 6: What is the purpose of this class?  Can't we just pass a Zk2Resolver to the ServersetNamer directly?

namer/serversets/src/main/scala/io/buoyant/namer/serversets/ServersetNamer.scala
```diff
@@ -49,7 +54,12 @@ class ServersetNamer(zkHost: String, idPrefix: Path) extends Namer {
         val addr = if (segments.nonEmpty && (segments.last contains ":")) {
           val Array(name, endpoint) = segments.last.split(":", 2)
           val zkPath = (segments.init :+ name).mkString("/", "/", "")
-          resolveServerset(zkHost, zkPath, endpoint)
+          if (!(endpoint contains "#")) {
```
adleong, line 57: By putting this inside the `if` on line 54, we're only able to use shard ids when also specifying an endpoint.  According to https://github.com/twitter/finagle/blob/develop/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/naming/ServersetPath.scala#L23 it looks like they should be able to be used independently.

Ideally here we would just use `ServersetPath`'s unapply to ensure that we parse this correctly.  Unfortunately, `ServersetPath` is marked as package private to `com.twitter`.  We get around that restriction elsewhere in Linkerd by moving classes to packages like `com.twitter.buoyant.namer.serversets`.  If you moved this class into that package, we could use `ServersetPath` and wouldn't need to worry about maintaining our own parsing code.

adleong, line 57: This is a moot point if we use `ServersetPath` as I recommend above, but I'll just point it out anyway: Linkerd style is to avoid using infix notation.  So this should be `endpoint.contains("#")`.  Furthermore, I'd flip the `if` and `else` blocks to remove the negation in the condition.

namer/serversets/src/test/scala/io/buoyant/namer/serversets/ServersetNamerTest.scala
```diff
@@ -1,16 +1,21 @@
...
 import com.twitter.util.{Activity, Var}
 import io.buoyant.namer.NamerTestUtil
 import java.net.{InetAddress, InetSocketAddress}
+import org.mockito.Matchers.{anyString, any}
+import org.mockito.Mockito.{verify, when, never}
 import org.scalatest.FunSuite
+import org.scalatestplus.mockito.MockitoSugar
 
-class ServersetNamerTest extends FunSuite with NamerTestUtil {
+class ServersetNamerTest extends FunSuite with NamerTestUtil with MockitoSugar {
```
adleong, line 13: Mockito is pretty cool but I would prefer not to take on an additional testing framework at this time for the sake of consistency with other tests in the project.

You should be able to accomplish the same thing just as easily by overriding the `resolveServerset` similarly to how we already override the `resolve` method in this test.

namer/serversets/src/main/scala/com/twitter/finagle/serversets2/BouyantZkResolver.scala
JustinVenus: I needed access to `private[twitter] Zk2Resolver.addrOf` [to resolve the shard component](https://github.com/twitter/finagle/blob/198d0eb3037b9fc7bfbd671b4aa3c1fd8ab1580a/finagle-serversets/src/main/scala/com/twitter/finagle/serverset2/Zk2Resolver.scala#L264) so I declared `BouyantZkResolver` in the com.twitter.finagle.serverset2 package for access.

namer/serversets/src/main/scala/io/buoyant/namer/serversets/ServersetNamer.scala
JustinVenus: Good find, I was wondering if that worked w/o the labeled part of the endpoint.

I'd be happy to refactor this part to use the `ServersetPath`'s unapply.

namer/serversets/src/test/scala/io/buoyant/namer/serversets/ServersetNamerTest.scala
JustinVenus: Thank you for the feedback @adleong.  I will make the changes you've requested.

JustinVenus:
The [failed test](https://circleci.com/gh/linkerd/linkerd/9890?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) doesn't appear to be related to anything I worked on.  Do I need to perform further action?

cpretzer:
@JustinVenus this is a transient failure. I'm rerunning the job now

PR #2394, opened 2020-05-27
Set MaxInitialLineSize from maxInitialSizeKB config

### LINKED ISSUE #2393: MaxInitialLineSize parameter is not set from maxInitialLineKB config value
Issue Type:

- [x] Bug report
- [ ] Feature request

**What happened**:
`maxInitialLineKB` config value is ignored by linkerd, resulting in 414 for request with long URLs

`maxInitialLineKB` value is used to set `MaxInitialLineSize` parameter for the router

set `maxInitialLineKB` to a 64KiB and submit a request with request line > 4KiB (default value for the `maxInitialLineKB`

This seems to be a regression in 1.7.x

**Environment**:
- linkerd/namerd version, config files: 1.7.x

edio:
MaxInitialLineSize parameter is not set from the config. This appears to
be a regression in 1.7.x

This commit adds test case to demonstrate the issue and a fix

Fixes #2393


edio:
@zaharidichev , yes we have tested the change on our dev environments: linkerd is able to process requests with >4KiB initial line with this change as expected.

PR #2404, opened 2020-07-06
Pin protobuf version to v0.1.1

cpretzer:
## What
Namerd uses protobuf files in the `linkerd2-proxy-api` repository which are retrieved with curl from the HEAD of the repository. This PR pins the version to [v0.1.1](https://github.com/linkerd/linkerd2-proxy-api/tree/v0.1.1/proto) of those files.

## Why
Ensures stability and a reliable version

## How
Added the LATEST_API_RELEASE variable to the pull-destination-proto.sh script which specifies the version of the linkerd2-proxy-api to use. This follows the design pattern of the `.proxy-version` configuration in the `linkerd2` repository

Also did a bit of refactoring to loop over files to download, rather than explicitly naming the files and their URLs.


cpretzer:
@adleong I updated the script to pass `shellcheck`

PR #2448, opened 2022-03-11
Add support for building without zookeeper

### LINKED ISSUE #2440: Request upgrade of all log4j 1.x to at least log4j 2.17.1, or patch vulnerabilities

## Getting Help ##

Full details at [CONTRIBUTING.md](https://github.com/linkerd/linkerd/blob/main/CONTRIBUTING.md).

## Filing a Linkerd issue ##

Issue Type:

- [ ] Bug report
- [X] Feature request

**What happened**:

Linkerd 1 currently uses log4j 1.x through its Netty and Zookeeper dependencies according to https://github.com/linkerd/linkerd/issues/2438#issuecomment-994911423

```
Hello, sharing what I've found regarding log4j vuln in Linkerd. I cloned the linkerd/linkerd repo and checked out the 1.6.3 tag which is the version we're using, then I added a dependency tree plugin to sbt plugins file and after executing it I only see log4j 1.2.16 and 1.2.17, I'll do the same with tag 1.7.4
I also started looking into netty and if I'm not mistaken log4j was updated to v2 on version 4.1.65.final netty/netty#11264
on linkerd 1.6.3 I see netty 4.1.31.final
on linkerd 1.7.4 I see netty 4.1.47.final and log4j 1.2.16 and 1.2.17
from these dependencies it seems like linkerd 1 is safe, but log4j 1 has other vulnerabilities, although not as critical as log4shell
```

Even though log4j 1.x is not vulnerable to CVE-2021-44228, it is still end of life and has a number of other vulnerabilities which our security scanner is picking up.  

We request that Linkerd 1 be upgraded so that either only log4j 2.17.1 (or higher) is used, or the vulnerabilities in log4j 1.x are patched, in order to fulfill our security requirements.

We are a paying customer of Bouyant.

**Environment**:

We are using Linkerd 1.7.3.

cpretzer:
* Add `nozk` config to namerd and linkerd projects
* Update Apache Curator version to 5.2.0

This change adds support for building the linkerd and namerd binaries without ZooKeeper to remove the dependency on log4j

Closes #2440


project/LinkerdBuild.scala
```diff
@@ -522,6 +542,29 @@ object LinkerdBuild extends Base {
...
+      .configDependsOn(NoZk)(BundleProjectsNoZk: _*)
+      .settings(inConfig(NoZk)(NoZkSettings))
+      .settings(
+        assembly := (assembly in Bundle).value,
+        docker := (docker in Bundle).value,
+        dockerBuildAndPush := (dockerBuildAndPush in Bundle).value,
+        dockerPush := (dockerPush in Bundle).value
+      )
+
+    val noZk = aggregateDir("namerd",
```
adleong, line 554: should this project just use the NoZk config only?

project/LinkerdBuild.scala
```diff
@@ -741,6 +804,28 @@ object LinkerdBuild extends Base {
...
+      .configDependsOn(NoZk)(BundleProjectsNoZk: _*)
+      .settings(inConfig(NoZk)(NoZkSettings))
+      .settings(
+        assembly := (assembly in Bundle).value,
+        docker := (docker in Bundle).value,
+        dockerBuildAndPush := (dockerBuildAndPush in Bundle).value,
+        dockerPush := (dockerPush in Bundle).value
+      )
+
+    val noZk = aggregateDir("linkerd",
```
adleong, line 816: ditto

project/LinkerdBuild.scala
```diff
@@ -818,6 +904,7 @@ object LinkerdBuild extends Base {
   val namerZkLeader = Namer.zkLeader
   val namerRancher = Namer.rancher
 
+
```
adleong, line 907: extra newline

project/Deps.scala
```diff
@@ -2,9 +2,9 @@ import sbt._
 
 object Deps {
 
-  val curatorFramework = "org.apache.curator" % "curator-framework" % "4.1.0"
-  val curatorClient = "org.apache.curator" % "curator-client" % "4.1.0"
-  val curatorDiscovery = "org.apache.curator" % "curator-x-discovery" % "4.1.0"
+  val curatorFramework = "org.apache.curator" % "curator-framework" % "5.2.0"
+  val curatorClient = "org.apache.curator" % "curator-client" % "5.2.0"
+  val curatorDiscovery = "org.apache.curator" % "curator-x-discovery" % "5.2.0"
```
adleong, line 7: do we need to bump deps as part of this change?  I thought we were just adding a nozk configuration.

project/LinkerdBuild.scala
```diff
@@ -890,4 +978,30 @@ object LinkerdBuild extends Base {
       Telemetry.all,
       Mesh.all
     )
+
+  val noZk = project("linkerd-no-zk", file("no-zk"))
```
adleong, line 982: do we need this?

project/plugins.sbt
```diff
@@ -28,3 +28,6 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
 
 // our grpc building extends the wrapped sbt-protobuf
 addSbtPlugin("com.github.gseitz" % "sbt-protobuf" % "0.5.3")
+
+// generate dependency graphs
+// addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
adleong, line 33: missing newline

adleong (commented):
Can you include instructions in the description for how to build the nozk configuration?  Have you been able to confirm the nozk artifacts don't contain the zk jars?

project/LinkerdBuild.scala
```diff
@@ -522,6 +542,29 @@ object LinkerdBuild extends Base {
       .settings(inConfig(OpenJ9)(OpenJ9Settings))
       .configDependsOn(Dcos)(dcosBootstrap)
       .settings(inConfig(Dcos)(DcosSettings))
+      .configDependsOn(NoZk)(BundleProjectsNoZk: _*)
```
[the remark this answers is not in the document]
cpretzer: Good catch, I left this by oversight

project/LinkerdBuild.scala
cpretzer: Updated the PR with this change

project/Deps.scala
cpretzer: I bumped this dep in order to get a version of Curator that doesn't rely on Zookeeper as a downstream dependency. It's easy enough to roll this back, and I left it because all the tests passed.

project/LinkerdBuild.scala
cpretzer: Removed 😄

project/Deps.scala
adleong: what about the netty and boringssl bumps below?  is bumping those versions necessary?  I'd prefer to change as little as possible.

project/LinkerdBuild.scala
```diff
@@ -540,6 +579,12 @@ object LinkerdBuild extends Base {
       .configDependsOn(Test)(BundleProjects: _*)
       .settings(publishArtifact := false)
       .withTests()
+
+    // val examplesNoZk = projectDir("namerd/examples")
```
adleong, line 583: should this be removed?

adleong (commented):
This looks pretty good

project/Deps.scala
cpretzer: @adleong I've rolled back the dependency versions, so please have a look and let me know if there are additional changes required

adleong (approved):
Thanks, @cpretzer, looks great!

DrCapt:
@adleong will there be a release of zookeeperless linkerd in the near future?

